PR #55, opened 2018-02-25
Version 1.0.0.

billhollings:


ePirat:
It would be great if v1.0.0 could get a git tag, else I am not sure how one would easily find Version 1.0.0 again in the future, in the git history.

billhollings:
Tag/Release v1.0.0 added.

PR #58, opened 2018-02-26
Minor markdown syntax fixes in docs

ePirat:

PR #59, opened 2018-02-26
Update User Guide Formatting

tobine:
Add space between "###" and following text so headings show up correctly.
Remove one instance of "###" where it was used along with underline.

### DIFF
--- a/Docs/MoltenVK_Runtime_UserGuide.md
+++ b/Docs/MoltenVK_Runtime_UserGuide.md
@@ -358,8 +358,7 @@ line with no arguments.
 
 
 <a name="spv_vs_msl"></a>
-Troubleshooting Shader Conversion
----------------------------------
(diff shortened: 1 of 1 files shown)


ePirat:
Somewhat duplicate of #57 and #58

Docs/MoltenVK_Runtime_UserGuide.md
```diff
@@ -358,7 +358,7 @@ line with no arguments.
 
 
 <a name="spv_vs_msl"></a>
-###Troubleshooting Shader Conversion
+Troubleshooting Shader Conversion
```
ePirat: Judging from the Table of Contents at the beginning, this should likely have the `###` and not the underline-heading, as it is a sub-section of the “Metal Shading Language Shaders” section.

PR #92, opened 2018-03-09
Wrap try/catch blocks for if exceptions disabled

davidhubbard:
SPIRV-Cross supports the macro `SPIRV_CROSS_EXCEPTIONS_TO_ASSERTIONS` for builds with exceptions disabled. MoltenVK support is just a couple #if statements. When exceptions are disabled, the program will just abort, not throw.

### DIFF
--- a/MoltenVKShaderConverter/MoltenVKSPIRVToMSLConverter/SPIRVToMSLConverter.cpp
+++ b/MoltenVKShaderConverter/MoltenVKSPIRVToMSLConverter/SPIRVToMSLConverter.cpp
@@ -192,9 +192,12 @@ MVK_PUBLIC_SYMBOL bool SPIRVToMSLConverter::convert(SPIRVToMSLConverterContext& 
     scOpts.vertex.flip_vert_y = context.options.shouldFlipVertexY;
     mslCompiler.CompilerGLSL::set_options(scOpts);
 
+#ifndef SPIRV_CROSS_EXCEPTIONS_TO_ASSERTIONS
 	try {
+#endif
 		_msl = mslCompiler.compile(&vtxAttrs, &resBindings);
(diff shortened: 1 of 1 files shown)


billhollings:
Is this change suggested because you **_want_** the app to abort/crash on shader compilation...and the try/catch interferes with the `abort()` in some way?

volcanoauthors:
Hi Bill,

Yes, that's what I'm going for: abort/crash the app if the shader compiler throws an error.

billhollings:
Is it accurate that what you need is to be able to compile MoltenVK without any exception handling at all?

If so...that is a potentially larger requirement...because although SPIR-V conversion is currently the only exception handling...it puts a requirement on any future addition of more generic exception handling within MoltenVK.

What is the source of the requirement to be able to compile MoltenVK without exception handling?

davidhubbard:
> Is it accurate that what you need is to be able to compile MoltenVK without any exception handling at all?

Yes, that's what I'm doing.

> it puts a requirement on any future addition of more generic exception handling within MoltenVK.

Well, I don't actually expect you to remember me or this request. I might send another pull request, but I don't expect to point to this one and say "you promised." This isn't going to keep me up at night.

> What is the source of the requirement to be able to compile MoltenVK without exception handling?

I'm writing Vulkan examples and documentation, capturing best practices and perhaps making it easier to learn Vulkan. https://github.com/ndsol/VolcanoSamples

In order to use only the subset of features that are the most broadly supported, the code does not use any Vulkan features not in the Vulkan Compatibility Initiative. By the same token, [exception handling is disabled by default on Android](https://developer.android.com/ndk/guides/cpp-support.html#c_exceptions). While devs are learning, I prefer they just not use exceptions - KISS principle.

PR #93, opened 2018-03-09
Fix vkGetInstanceProcAddr to work with 1.1 loader

mikes-lunarg:
In order to determine the instance version, the 1.1 loader takes advantage of the fact that a 1.0 ICD will return NULL for vkGetInstanceProcAddr(NULL, "vkEnumerateInstanceVersion")


billhollings:
@mikes-lunarg 

Thanks!

PR #99, opened 2018-03-13
Fix vkCmdCopyBuffer when copying unaligned regions.

### LINKED ISSUE #62: vkCmdCopyBuffer requires copy sizes to be a multiple of 4 on macOS
When calling [vkCmdCopyBuffer](https://www.khronos.org/registry/vulkan/specs/1.0/man/html/vkCmdCopyBuffer.html) under MoltenVK, each of the regions given needs to have a copy size that is a multiple of 4. Now, I understand that this is [a limitation of Metal on macOS](https://developer.apple.com/documentation/metal/mtlblitcommandencoder/1400767-copyfrombuffer?language=objc), but the Vulkan spec does not indicate there is any such limitation. Additionally, MoltenVK doesn't detect anything wrong and attempts to call the blit command encoder with unchanged parameters and triggers validation errors:

```
-[MTLDebugBlitCommandEncoder copyFromBuffer:sourceOffset:toBuffer:destinationOffset:size:]:898: failed assertion `size(30) must be a multiple of 4 bytes.'
```

I have a couple of questions:

* Is this considered "a bug" from MoltenVK's side? I understand that there may be things in the Vulkan API that cannot be emulated with full fidelity, although this seems to be one that can be worked around (even if it would be very roundabout and complicated).
* If not, is there a list of such discrepancies or instances where behavior is different from the spec?

mellinoe:
This fixes #62.

* Vulkan allows any valid copy regions to be used in this function. However, Metal's blit command encoder only allows copy offsets and sizes which are multiples of 4 bytes on macOS.
* Before this change, attempting to copy such a region would cause Metal validation errors to trigger.
* This adds logic to detect such copy regions and uses a special compute shader to perform a byte-by-byte copy from the source buffer to the destination buffer with the appropriate offsets.
* This new code path is only needed on macOS, because iOS does not share the same copy region restrictions.

@billhollings This is my first submission, so please let me know if there's something I have missed, or something I should be doing differently. I've tried to match the patterns and style that are nearby.

### DIFF
--- a/MoltenVK/MoltenVK/Commands/MVKCmdTransfer.mm
+++ b/MoltenVK/MoltenVK/Commands/MVKCmdTransfer.mm
@@ -555,9 +555,6 @@ void MVKCmdCopyBuffer::setContent(VkBuffer srcBuffer,
 }
 
 void MVKCmdCopyBuffer::encode(MVKCommandEncoder* cmdEncoder) {
-
-    id<MTLBlitCommandEncoder> mtlBlitEnc = cmdEncoder->getMTLBlitEncoder(kMVKCommandUseCopyBuffer);
-
     id<MTLBuffer> srcMTLBuff = _srcBuffer->getMTLBuffer();
     NSUInteger srcMTLBuffOffset = _srcBuffer->getMTLBufferOffset();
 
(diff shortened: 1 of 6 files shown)


billhollings:
@mellinoe 

Good work! Thanks very much for taking the time to do this!

And thanks for following the established factory design pattern to keep it compatible with the other similar components.

billhollings:
@mellinoe 

Can you test again under the code added in PR #104, please? See notes there. Thanks.

PR #104, opened 2018-03-14
 Derive vkCmdCopyBuffer() alignment requirement at runtime.

billhollings:


billhollings:
@mellinoe 

This is an update to your code from PR #99. I have updated your code to derive the 4-byte alignment limit at runtime, so that the code can adapt easier to possible changes to this limit in future versions of macOS Metal.

Can you test out the update in this PR against your original use cases that drove the need for PR #99, to make sure I haven't broken your original fix, please?

mellinoe:
@billhollings Sure thing, I'll test this out later today, but the change looks fine at a glance.

mellinoe:
@billhollings All of my tests related to this are passing, so things look good on my end. Thanks for letting me know about the change.

PR #110, opened 2018-03-21
Add CMake support

andystanton:
This is a proof of concept to test if MoltenVK can be built with CMake. The rationale is that it makes building Vulkan cross-platform projects easier.

Running `make` from the project root pulls or refreshes the external dependencies, builds them, then builds the converter libs & tool, and then builds MoltenVK as a static library.

Here is an example of how it can be used to create a standalone macOS app pulling in GLFW and MoltenVK as git submodules: https://github.com/andystanton/moltenvk-cmake-test

Other things that could be done:
* Build dynamic library
* Build static framework
* Build iOS versions
* Build demos
* Add install step to put library and headers in /usr/local
* Use cmake subdirectories for external dependencies instead of invoking cmake via a script
* Test xcode project generation from cmake

Can you provide feedback?


karl-lunarg:
One can always build the existing Xcode project in the repo with something like

```
xcodebuild -project MoltenVKPackaging.xcodeproj \
    GCC_PREPROCESSOR_DEFINITIONS='$GCC_PREPROCESSOR_DEFINITIONS MVK_LOGGING_ENABLED=0' \
    -scheme "MoltenVK (Release)" build
```
This can easily be done from another CMake-based project that pulls in MVK.

One challenge with this approach is that as long as there is a (non-generated) Xcode project and a CMake project, there will be dual maintenance.  If the CMake project evolved to the point where it generates an equivalent Xcode project using CMake's Xcode generator, then the CMake-generated Xcode project _could_ replace the current Xcode project.

Since MVK is fundamentally for macOS/iOS, I'm not sure it makes sense to convert it to CMake since it _doesn't have_ to be portable in the CMake sense.

Also:

* CMake support for Xcode project generation is good, but may not be complete.  There may be things in the current Xcode project that can't be created with the CMake Xcode generator,

* There was a discussion about this in the Vulkan ecosystem GitHub or someplace similar, and it was agreed that it probably wasn't worth doing this sort of work.  Adding functionality is probably more important.

* I think that the primary developer of this repo is more comfortable with Xcode than determining the correct CMake incantations to generate what he wants to end up in an Xcode project.

andystanton:
That reasoning makes sense - I'll close the PR. Thanks for the feedback @karl-lunarg!

PR #111, opened 2018-03-22
Externals as submodules

eile:
Fetching dependencies now becomes ```git submodule update --init``` (or ```git clone --recursive```). Since this is standard git, I don't think the fetch script is still needed.


karl-lunarg:
👎 

Deleting the fetch script removes the step of building glslang (and the SPIRV-Tools it fetches).  I'm not sure how MVK would build without this step.  @eile Did you test this?

I'm not opposed to moving back to submodules.  (This repo used submodules up until recently.)

But there would have to be a script or some other mechanism that builds the submodules that need building.  And I think that is missing from this PR.

There is other functionality that is in the deleted fetch script: the -sdk option and obtaining the glslang revision.  These can be handled by other projects by just replacing glslang with their own.  For example, the Vulkan-LoaderAndValidationLayers project gets its own specific version of glslang and builds it.  It could then just delete the glslang submodule in this repo and symlink to its own during its build process.  

I think that moving back to submodules is ok, but there still needs to be a script or something to build the submodules that need building.

eile:
> Deleting the fetch script removes the step of building glslang (and the SPIRV-Tools it fetches). I'm not sure how MVK would build without this step. @eile Did you test this?

Ah, I tested using the open CMake PR. I'll readd this part of the script and add a git submodule line.

billhollings:
As @karl-lunarg indicates...we moved away from submodules for two reasons:

- A build step is required on most of the submodules...so a script is needed anyway. This used to be called `makeAll` in the 1.0.0 version of MoltenVK.

- The versions of the `Vulkan-LoaderAndValidationLayers`, `glslang`, `SPIRV-Tools` and `SPIRV-Headers`dependencies need to be kept in sync with each other. These interactions are handled through Git revision hashes stored in a file. MoltenVK is now uses this as a consistent approach for handling all dependencies.

PR #133, opened 2018-04-16
Update V-LVL revision

karl-lunarg:
- pick up 1.1.73 headers
- cope with interface change in cube demo


billhollings:
@karl-lunarg Merged now.

Looks like I will need to submit a corresponding fix to the iOS `DemoViewController.mm`. I'll handle that.

PR #143, opened 2018-04-25
Separate categories from MVKOSExtensions

hanton:
Separate categories from MVKOSExtensions:

MTLTextureDescriptor+MoltenVK
MTLSamplerDescriptor+MoltenVK
CAMetalLayer+MoltenVK


billhollings:
Thanks!

PR #151, opened 2018-05-09
Change baked in shaders to precompiled shaders for performance and maintenance

hanton:
Change baked in shaders to precompiled shaders for performance and maintenance.

[Metal Programming Guide](https://developer.apple.com/library/content/documentation/Miscellaneous/Conceptual/MetalProgrammingGuide/Prog-Func/Prog-Func.html#//apple_ref/doc/uid/TP40014221-CH5-SW1)
> For the best performance, compile your Metal shading language source code into a library file during your app's build process in Xcode, which avoids the costs of compiling function source during the runtime of your app.

### DIFF
--- a/Docs/MoltenVK_Runtime_UserGuide.md
+++ b/Docs/MoltenVK_Runtime_UserGuide.md
@@ -52,7 +52,7 @@ distribution package, see the main [`README.md`](../README.md) document in the `
 About **MoltenVK**
 ------------------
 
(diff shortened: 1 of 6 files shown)


hanton:
hi @billhollings 

Thanks for providing the analysis data. It looks like the baked-in shaders solution doesn't impact performance that much. 

If the design of MoltenVK is preferred to keep the installation simple, the current drag-and-drop solution is good :)

PR #161, opened 2018-05-20
Fix an issue where the depth format in MVKCmdClearImage was not getting set correctly.

mellinoe:
Without this, clearing a depth stencil image causes a runtime crash with this message:

```
-[MTLRenderPipelineDescriptorInternal validateWithDevice:]:2400: failed assertion `No valid pixelFormats set.'
```

The fix is to simply keep track of the pixel format here.


billhollings:
Thanks. Good catch.

Issue #160, opened 2018-05-20
vkGetPhysicalDeviceImageFormatProperties returns "Success" for ETC2 formats when they are not supported


mellinoe:
To decide between BC3 and ETC2 textures for one of my projects, I first check if ETC2 is supported. I do this by calling vkGetPhysicalDeviceImageFormatProperties. On MoltenVK, this returns "Success", and fills in the structure with regular info (e.g. maxExtent=16384x16384x1). Notably, the VkPhysicalDeviceFeatures for this physical device has "false" for "textureCompressionETC2" -- so on some level it's known to be unsupported.

For completeness, I'm passing this to vkGetPhysicalDeviceImageFormatProperties:

format: VK_FORMAT_ETC2_R8G8B8_UNORM_BLOCK
type: VK_IMAGE_TYPE_2D
tiling: VK_IMAGE_TILING_OPTIMAL
usage: VK_IMAGE_USAGE_TRANSFER_SRC_BIT | VK_IMAGE_USAGE_TRANSFER_DST_BIT | VK_IMAGE_USAGE_TRANSFER_SAMPLED_BIT
flags: 0

It looks like [the implementation](https://github.com/KhronosGroup/MoltenVK/blob/d5272484acc4c5c5a70d71ce45768bc29d704167/MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm#L100) isn't doing any support checks, and will always return VK_SUCCESS.

billhollings:
I'll fix this per the further discussion in PR #162.

mellinoe:
I pulled the change down and ran a quick test. Everything looks correct to me now. Thanks for the quick fix!

PR #162, opened 2018-05-20
Correctly handle unsupported compressed formats in MVKPhysicalDevice::getImageFormatProperties

mellinoe:
This can be viewed as a partial fix for #160 . I am aware of some other format limitations in Metal, but I'm not sure what the best way to filter them out here is. For example, I am aware that the `R16_UNorm` format can only be used as a depth-stencil format on `macOS_GPUFamily1_v2` and `macOS_GPUFamily1_v3`. That kind of filtering seems more complicated, whereas this filtering is very straightforward, and the information is readily available.

### DIFF
--- a/MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
@@ -105,6 +105,16 @@ VkResult MVKPhysicalDevice::getImageFormatProperties(VkFormat format,
                                                      VkImageFormatProperties* pImageFormatProperties) {
     if (!pImageFormatProperties) { return VK_SUCCESS; }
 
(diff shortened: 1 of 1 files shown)


billhollings:
Thanks for the submission.

billhollings:
Thanks for the submission.

Format properties are all handled in the collection of `MVKFormatDesc` structures within `mvk_datatypes.mm`. The structure includes an `isSupported()` function which takes into consideration the OS versions...which you refer to.

It looks like we can simply add an `mvkFormatIsSupported()` function...and call it from `MVKPhysicalDevice::getImageFormatProperties()` and return `VK_ERROR_FORMAT_NOT_SUPPORTED` if not.

This should be fairly simple. I'll take care of it.

mellinoe:
@billhollings Thanks, much appreciated!

PR #171, opened 2018-06-01
Fix vsync not being enabled in IMMEDIATE present mode. MVK_MACOS was …

danginsburg:
…not being defined.


billhollings:
@danginsburg Thanks!

PR #194, opened 2018-07-09
Removes references to IOSurface framework on IOS

natbro:
and marks the IOSurface features as unavailable. A better solution would be to not support IOSurface features prior to iOS 11.0 when the framework became partially public, but this is a good initial solution.


mlfarrell:
Will this affect the ability to use MoltenVK with ARKit?  That is one of my major long-term goals for my project.

natbro:
It shouldn't impact your use of ARKit. iOS is shifting towards the use of shared metal textures (SMT) over global IOSurfaces. If you bump into an issue interacting with CVPixelBuffers from ARKit or the camera/video AVFoundation APIs let us know and we'll try to sort it out.

PR #201, opened 2018-07-24
CreatePipelines should not return SUCCESS when MSL translation fails.

prideout:
When vertex shader translation fails, we were already calling setConfigurationResult, but we were not doing this for fragment shaders.

### DIFF
--- a/MoltenVK/MoltenVK/GPUObjects/MVKPipeline.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKPipeline.mm
@@ -287,7 +287,6 @@ MTLRenderPipelineDescriptor* MVKGraphicsPipeline::getMTLRenderPipelineDescriptor
 			id<MTLFunction> mtlFunction = mvkShdrMod->getMTLFunction(&shaderContext, pSS->pSpecializationInfo, _pipelineCache).mtlFunction;
 			if ( !mtlFunction ) {
 				setConfigurationResult(mvkNotifyErrorWithText(VK_ERROR_INITIALIZATION_FAILED, "Fragment shader function could not be compiled into pipeline. See previous error."));
-				return nil;
 			}
 			plDesc.fragmentFunction = mtlFunction;
         }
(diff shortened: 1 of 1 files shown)


MoltenVK/MoltenVK/GPUObjects/MVKPipeline.mm
```diff
@@ -284,7 +284,12 @@
         // Fragment shader
         if (mvkAreFlagsEnabled(pSS->stage, VK_SHADER_STAGE_FRAGMENT_BIT)) {
 			shaderContext.options.entryPointStage = spv::ExecutionModelFragment;
-			plDesc.fragmentFunction = mvkShdrMod->getMTLFunction(&shaderContext, pSS->pSpecializationInfo, _pipelineCache).mtlFunction;
+			id<MTLFunction> mtlFunction = mvkShdrMod->getMTLFunction(&shaderContext, pSS->pSpecializationInfo, _pipelineCache).mtlFunction;
+			if ( !mtlFunction ) {
+				setConfigurationResult(mvkNotifyErrorWithText(VK_ERROR_INITIALIZATION_FAILED, "Fragment shader function could not be compiled into pipeline. See previous error."));
+				return nil;
```
billhollings: Thanks for catching this.

A missing fragment function is not fatal to the pipeline (although it won't render anything).

I agree that the configuration result should be set and logged. But returning `nil` will kill the pipeline itself. Not sure if this is the appropriate.

Thoughts?

prideout: That makes sense although returning nil is consistent with the existing behavior for vertex shaders. Would you like me to change the vertex shader behavior as well?

On a side note, the spec doesn't say that we can return VK_ERROR_INITIALIZATION_FAILED here, but I think it's good to do so. Adreno drivers seem to return VK_INCOMPLETE when their shader compilation step fails. Personally, I don't care too much about the actual result code as long as it is not VK_SUCCESS.

billhollings: I hear you...but in this case...consistency is secondary. Metal requires a render pipeline to have a non-nil vertex function...but it does NOT require it to have a non-nil fragment function.

So...yes...please set the configuration result as you are doing...but set the `plDesc.fragmentFunction` to the `mtlFunction` even if it is nil...and then continue on instead of returning nil.

That way a lazy app that does not check the return code will get a functional pipeline that does not have a fragment function.

Thanks.

Issue #202, opened 2018-07-26
Support for VK_FORMAT_A2B10G10R10_SNORM_PACK32 as vertex format


aerofly:
Hello,

Is there a reason why VK_FORMAT_A2B10G10R10_SNORM_PACK32 is not yet supported as a vertex format by MoltenVK? It can easily be fixed/enabled by setting the MTLVertexFormat to MTLVertexFormatInt1010102Normalized in the file mvk_datatypes.mm.

**`MVK_MAKE_FMT_STRUCT( VK_FORMAT_A2B10G10R10_SNORM_PACK32, MTLPixelFormatInvalid, MTLPixelFormatInvalid, kMTLFmtNA, kMTLFmtNA, 1, 1, 4, MTLVertexFormatInt1010102Normalized, kMVKFormatColorFloat, MVK_FMT_NO_FEATS, MVK_FMT_NO_FEATS ),
`**

We see no issue on our side with this change.

VK_FORMAT_A2B10G10R10_SNORM_PACK32 can be used to reduce GPU memory usage. The 10 bit precision is usually enough when using vertex normals, so one would safe 8 bytes per vertex when using this format.

By the way: Great job so far, for bringing Vulkan to Mac OS and iOS.

billhollings:
@aerofly 

It's likely an oversight. Mapping all the Vulkan and Metal formats...pixel & vertex...is an ongoing community effort.

As an open source projects...we encourage all users to submit Pull Requests when you find something amiss and have figured out a fix.

Could you submit your fix as a PR, please? Thanks!

aerofly:
@billhollings: Ok, I had a go at Pull Requests. Let me know if this works for you.

billhollings:
Fixed in PR #206. Closing.

PR #206, opened 2018-07-27
new vertex format VK_FORMAT_A2B10G10R10_SNORM_PACK32

aerofly:
VK_FORMAT_A2B10G10R10_SNORM_PACK32 as a vertex attribute has the MoltenVK equivalent of MTLVertexFormatInt1010102Normalized. This pull request adds this missing format.

Note: We have tested this on MacOS only, its not clear if this works on iOS as well.


aerofly:
VK_FORMAT_A2B10G10R10_SNORM_PACK32 and VK_FORMAT_A2R10G10B10_SNORM_PACK32 can be used as vertex formats, but Apple makes no distinction, so should MTLVertexFormatInt1010102Normalized be added for both Vulkan formats?

billhollings:
Fixes issue #202.

PR #230, opened 2018-08-27
Support the VK_KHR_maintenance1 extension.

cdavis5e:
Much of this was already supported, simply by being supported by Metal.
Of course, this support is incomplete: Metal doesn't yet allow you to
create a 2D texture view from a 3D texture.

### DIFF
--- a/MoltenVK/MoltenVK/GPUObjects/MVKDescriptorSet.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKDescriptorSet.mm
@@ -611,7 +611,7 @@ VkResult MVKDescriptorPool::allocateDescriptorSets(uint32_t count,
 												   const VkDescriptorSetLayout* pSetLayouts,
 												   VkDescriptorSet* pDescriptorSets) {
 	if (_allocatedSetCount + count > _maxSets) {
-		return mvkNotifyErrorWithText(VK_ERROR_INITIALIZATION_FAILED, "The maximum number of descriptor sets that can be allocated by this descriptor pool is %d.", _maxSets);
+		return mvkNotifyErrorWithText(VK_ERROR_OUT_OF_POOL_MEMORY_KHR, "The maximum number of descriptor sets that can be allocated by this descriptor pool is %d.", _maxSets);
 	}
 
 	for (uint32_t dsIdx = 0; dsIdx < count; dsIdx++) {
--- a/MoltenVK/MoltenVK/Commands/MVKCommandPool.h
+++ b/MoltenVK/MoltenVK/Commands/MVKCommandPool.h
@@ -142,6 +142,9 @@ public:
 	void freeCommandBuffers(uint32_t commandBufferCount,
 							const VkCommandBuffer* pCommandBuffers);
 
+	/** Release any held but unused memory back to the system. */
+	void trimCommandPool();
+
 
 #pragma mark Construction
 
--- a/MoltenVK/MoltenVK/Commands/MVKCommandPool.mm
+++ b/MoltenVK/MoltenVK/Commands/MVKCommandPool.mm
@@ -63,6 +63,10 @@ void MVKCommandPool::freeCommandBuffers(uint32_t commandBufferCount,
 	}
 }
 
+void MVKCommandPool::trimCommandPool() {
+	// TODO: Implement.
+}
+
 void MVKCommandPool::addCommandBuffer(MVKCommandBuffer* cmdBuffer) {
 	_commandBuffers.insert(cmdBuffer);
 }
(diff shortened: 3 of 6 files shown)


MoltenVK/MoltenVK/GPUObjects/MVKDescriptorSet.mm
```diff
@@ -611,7 +611,7 @@
 												   const VkDescriptorSetLayout* pSetLayouts,
 												   VkDescriptorSet* pDescriptorSets) {
 	if (_allocatedSetCount + count > _maxSets) {
-		return mvkNotifyErrorWithText(VK_ERROR_INITIALIZATION_FAILED, "The maximum number of descriptor sets that can be allocated by this descriptor pool is %d.", _maxSets);
+		return mvkNotifyErrorWithText(VK_ERROR_OUT_OF_POOL_MEMORY_KHR, "The maximum number of descriptor sets that can be allocated by this descriptor pool is %d.", _maxSets);
```
danginsburg: I think the behavior here needs to be dependent on whether the app enabled VK_KHR_maintenance1 in vkCreateDevice.  The old behavior is correct when its disabled.  In VK_KHR_maintenance1, I don't think it should do mvkNotifyErrorWithText even, just return VK_ERROR_OUT_OF_POOL_MEMORY_KHR.  It's normal for applications to expect a failure here.

billhollings: @danginsburg 

The logging of the `VK_ERROR_OUT_OF_POOL_MEMORY_KHR` result is consistent with that `VkResult` being labelled as an error (`VK_ERROR_...`). Are you recommending that for this one error result, we consider it to be a non-error `VkResult` (like say `VK_INCOMPLETE`)...and not log it?

billhollings:
@cdavis5e Thanks very much for submitting this PR!

Is the lack of 3D->2D copying and image views a concern? What happens if an app detects that this extension is supported and then tries to do one of those commands?

At the very least, we should probably detect that and log an error...otherwise Metal will likely raise an assertion. For more completeness...those commands could probably invoke compute shaders that handle those image transfers.

@danginsburg Regarding the definition of this extension, can you provide some input on the following, please?

- How do `VK_FORMAT_FEATURE_TRANSFER_SRC_BIT_KHR` & `VK_FORMAT_FEATURE_TRANSFER_DST_BIT_KHR` differ from `VK_FORMAT_FEATURE_TRANSFER_SRC_BIT` & `VK_FORMAT_FEATURE_TRANSFER_DST_BIT`?

- What were the "inconsistencies" between the return codes of `vkCreateGraphicsPipelines`, `vkCreateComputePipelines`, `vkAllocateDescriptorSets` and `vkAllocateCommandBuffers`...and how were they "fixed" by this extension?

MoltenVK/MoltenVK/GPUObjects/MVKDescriptorSet.mm
danginsburg: I would prefer that, yes.  Dota 2 relies on this failure and checks for the error.  Logging it produces spam which I don't think needs to be there.

danginsburg:
I would assume the difference between the _KHR and non-KHR variant is just from when it got promoted to core from the extension?  They must have the same value, right?

RE: vkAllocateCommandBuffers:

ifdef::VK_KHR_maintenance1[]
fname:vkAllocateCommandBuffers can: be used to create multiple command
buffers.
If the creation of any of those command buffers fails, the implementation
must: destroy all successfully created command buffer objects from this
command, set all entries of the pname:pCommandBuffers array to
dlink:VK_NULL_HANDLE and return the error.
endif::VK_KHR_maintenance1[]

vkAllocateDescriptorSets:
ifdef::VK_KHR_maintenance1[]
If the allocation fails due to no more space in the descriptor pool, and not
because of system or device memory exhaustion, then
ename:VK_ERROR_OUT_OF_POOL_MEMORY_KHR must: be returned.

fname:vkAllocateDescriptorSets can: be used to create multiple descriptor
sets.
If the creation of any of those descriptor sets fails, then the
implementation must: destroy all successfully created descriptor set objects
from this command, set all entries of the pname:pDescriptorSets array to
dlink:VK_NULL_HANDLE and return the error.
endif::VK_KHR_maintenance1[]

I couldn't find the specifics in the extension for vkCreate*Pipelines - it's probably to do also with partial failures but I'm not sure.

danginsburg:
> Are you okay with adding this extensions knowing that it can't support the 3D texture requirements?

Yes.

billhollings:
The error handling and the test for whether the extension is enabled described above are added in PR #237.

PR #232, opened 2018-08-30
Support the VK_KHR_shader_draw_parameters extension.

cdavis5e:
Update SPIRV-Cross to support shaders which use the builtins provided by
this extension.


billhollings:
@cdavis5e  Thanks for contributing this...and the associated SPIRV-Cross changes.

PR #236, opened 2018-08-31
Support the VK_KHR_push_descriptor extension.

cdavis5e:


billhollings:
@cdavis5e Nice work!

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
```diff
@@ -77,6 +77,20 @@
     if (properties) {
         properties->sType = VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_PROPERTIES_2_KHR;
         properties->properties = _properties;
+        auto* next = (VkStructureType*)properties->pNext;
+        while (next) {
+            switch (*next) {
```
billhollings: @cdavis5e This line is causing an `EXC_BAD_ACCESS` error when running `vulkaninfo` from the 1.0.82 Vulkan SDK.

cdavis5e: I think I know what's wrong. I forgot to account for the padding between the `sType` and `pNext` members.

cdavis5e: #245

Issue #233, opened 2018-08-31
Wrong comparison for count variable in extension enumeration


baldurk:
The check here is backwards - should be `extCnt <= *pCount`. For example if *pCount is set to something high like 100 it will fail the comparison and return VK_INCOMPLETE even though there's enough room in the output array for it.

https://github.com/KhronosGroup/MoltenVK/blob/259292c4d14621d25e8e2f23c760aac65d9ed1aa/MoltenVK/MoltenVK/Loader/MVKLayers.mm#L40-L45

billhollings:
Fixed in PR #237. Please test and close.

PR #237, opened 2018-09-01
 Vulkan extension enhancements.

billhollings:
Add Vulkan extensions as trackable first-class objects that can be marked as supported and enabled.

Adds error handling described in #230.
Fixes issue #233.


billhollings:
@cdavis5e 

You've added several Vulkan extensions recently. Thanks for taking the initiative!

This PR promotes Vulkan extensions to first class objects that can be tracked as supported or enabled. As a result, the mechanism for adding new extensions to the internal list is now different (and simpler)...so please take note of that if you add new extensions in the future.

In the process...I have included all the Vulkan extensions you have recently added in the new mechanism...but please feel free to test after this PR goes live to make sure everything is as good as you left it.

PR #244, opened 2018-09-05
MVKImageView: Create 3D MTLTextureViews for 2D image views of 3D textures.

cdavis5e:
This won't help using these image views from shaders, but it will help
attaching them to framebuffers. (Perhaps we should warn or raise an
error in the former case.)

I've left the warning in `validateImageViewConfig()` in place.

### DIFF
--- a/MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
@@ -640,12 +640,19 @@ MVKImageView::MVKImageView(MVKDevice* device, const VkImageViewCreateInfo* pCrea
 
 // Validate whether the image view configuration can be supported
 void MVKImageView::validateImageViewConfig(const VkImageViewCreateInfo* pCreateInfo) {
-	VkImageType imgType = ((MVKImage*)pCreateInfo->image)->getImageType();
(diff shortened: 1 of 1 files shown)


cdavis5e:
Don't merge this yet. It actually doesn't do what it was intended to do. (Yeah, I jumped the gun on this.) I'll have a fix hopefully shortly.

cdavis5e:
Well, this doesn't cover everything I had hoped to, but it will make some programs work.

MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
```diff
@@ -763,10 +775,12 @@
 	_useMTLTextureView = _image->_canSupportMTLTextureView;
 
 	// If the view is identical to underlying image, don't bother using a Metal view
+	bool is3D = _image->_mtlTextureType == MTLTextureType3D;
 	if (_mtlPixelFormat == _image->_mtlPixelFormat &&
-		_mtlTextureType == _image->_mtlTextureType &&
+		(_mtlTextureType == _image->_mtlTextureType ||
+		 (_mtlTextureType == MTLTextureType2D || _mtlTextureType == MTLTextureType2DArray) && is3D) &&
```
billhollings: Can you fix the syntax warning that arises here regarding the order of && and ||, please?

Issue #243, opened 2018-09-05
MoltenVK doesn't handle stencil views of depth-stencil image


baldurk:
Vulkan lets you create a stencil view of a depth-stencil image to sample the stencil. You do so by specifying `VkImageViewCreateInfo::subresourceRange.aspectMask == VK_IMAGE_ASPECT_STENCIL_BIT` on the image view, then when sampled as a unsigned integer texture the view will return stencil in red instead of depth in red.

This seems like it can be mapped to metal, albeit with similar sort of swizzling restrictions to other formats. Something like the following proof-of-concept hack allows me to read from red in my stencil view successfully:

```patch
diff --git a/MoltenVK/MoltenVK/GPUObjects/MVKImage.h b/MoltenVK/MoltenVK/GPUObjects/MVKImage.h
index 48e0e53..8b1cbef 100644
--- a/MoltenVK/MoltenVK/GPUObjects/MVKImage.h
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKImage.h
@@ -277,7 +277,7 @@ public:
 protected:
 	id<MTLTexture> newMTLTexture();
 	void initMTLTextureViewSupport();
-    MTLPixelFormat getSwizzledMTLPixelFormat(VkFormat format, VkComponentMapping components);
+    MTLPixelFormat getSwizzledMTLPixelFormat(VkFormat format, VkImageAspectFlags aspect, VkComponentMapping components);
     bool matchesSwizzle(VkComponentMapping components, VkComponentMapping pattern);
     const char* getSwizzleName(VkComponentSwizzle swizzle);
     void setSwizzleFormatError(VkFormat format, VkComponentMapping components);
diff --git a/MoltenVK/MoltenVK/GPUObjects/MVKImage.mm b/MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
index 6b84a1b..64b446b 100644
--- a/MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
@@ -464,7 +464,7 @@ MVKImage::MVKImage(MVKDevice* device, const VkImageCreateInfo* pCreateInfo) : MV
 
     _isDepthStencilAttachment = (mvkAreFlagsEnabled(pCreateInfo->usage, VK_IMAGE_USAGE_DEPTH_STENCIL_ATTACHMENT_BIT) ||
[...]

PR #252, opened 2018-09-10
Support views of the stencil aspect of depth/stencil images.

cdavis5e:
This requires Metal 1.2 (for the `MTLPixelFormatX24_Stencil8` and
`MTLPixelFormatX32_Stencil8` formats). Since Vulkan doesn't actually
have a corresponding format for that, add a new Metal feature for this.
Bump extension spec version.

Based on a change by @baldurk.


MoltenVK/MoltenVK/Vulkan/mvk_datatypes.mm
```diff
@@ -759,7 +759,7 @@ MVK_PUBLIC_SYMBOL MTLTextureUsage mvkMTLTextureUsageFromVkImageUsageFlags(VkImag
     }
     if ( mvkAreFlagsEnabled(vkImageUsageFlags, VK_IMAGE_USAGE_DEPTH_STENCIL_ATTACHMENT_BIT) ) {
         mvkEnableFlag(mtlUsage, MTLTextureUsageRenderTarget);
-        mvkDisableFlag(mtlUsage, MTLTextureUsagePixelFormatView);        // Clears bit. Do this last.
+        mvkEnableFlag(mtlUsage, MTLTextureUsagePixelFormatView);
```
billhollings: Should this always be enabled...or only for the stencil component of a D/S format...and only when the new `_device->_pMetalFeaturesstencilViews` flag is enabled?

cdavis5e: I debated whether or not to do that. I really wanted to do that, but I thought in the interests of simplicity that it might be better not to. It would require us to pass the pixel format and the device, respectively. I agree it would be more correct.

billhollings: Yeah...it's a PITA. But...as you say...more correct.  :^(

See for instance how format translation is handled. There is a basic static version as `mvkMTLPixelFormatFromVkFormat()` in `mvk_datatypes.h`...but all calls go through `MVKDevice::mtlPixelFormatFromVkFormat()`...all for the single `MTLPixelFormatDepth24Unorm_Stencil8` format.

At least `mvkMTLTextureUsageFromVkImageUsageFlags()` is only called from one place.

billhollings: BTW...you can't pass the `MVKDevice` to `mvkMTLTextureUsageFromVkImageUsageFlags()` because `mvk_datatypes.h` is a public Vulkan-like runtime C API...and knows nothing of C++ objects.

Better to delegate to the device...which can use the static method and then add any customization.

billhollings: Thinking about this further...I'm not so sure you'd need to change `mvkMTLTextureUsageFromVkImageUsageFlags()` to pass the format. You should be able to leave `mvkMTLTextureUsageFromVkImageUsageFlags()` where it was...and move your test for the `stencilViews` flag to a new `MVKDevice` method.

MoltenVK/MoltenVK/API/vk_mvk_moltenvk.h
```diff
@@ -54,7 +54,7 @@ extern "C" {
 #define MVK_VERSION     MVK_MAKE_VERSION(MVK_VERSION_MAJOR, MVK_VERSION_MINOR, MVK_VERSION_PATCH)
 
 
-#define VK_MVK_MOLTENVK_SPEC_VERSION            7
+#define VK_MVK_MOLTENVK_SPEC_VERSION            8
```
billhollings: Spec version is already 8 (I beat you to a change yesterday!).

Bump this to 9, please.

billhollings:
Fixes issue #243.

PR #255, opened 2018-09-13
Advertise some features we already support

cdavis5e:
Advertise a couple of features—namely `fullDrawIndexUint32`, and `drawIndirectFirstInstance`—that are already supported.

Also advertise `robustBufferAccess`; we don't really support it directly, but the spec says we have to. I haven't observed any issues with out-of-bounds accesses anyway, and I suspect that Apple wouldn't allow shaders to do crazy destabilizing things to the system, so I think we're OK here. (The feature itself does nothing in MoltenVK.)


billhollings:
@cdavis5e Thanks! Good catches!

PR #256, opened 2018-09-14
Fix memory leak not destroying spirv_cross objects.  When creating a …

danginsburg:
…lot of pipelines this was causing a large memory leak.


billhollings:
@danginsburg Thanks! Nice catch!

PR #259, opened 2018-09-18
Support the sampleRateShading feature.

cdavis5e:
This is for fragment shaders that run per-sample instead of
per-fragment. It implies that the `SampleId`, `SampleMask`, and
`SamplePosition` builtins, as well as per-sample interpolation, are all
available--using any of these causes the frag shader to run once per
sample. (In Metal, reading the color buffers may also cause a frag
shader to run per-sample.)

Update SPIRV-Cross to pull in a change that properly translates the
`SamplePosition` builtin to a form that Metal can understand.


billhollings:
@cdavis5e  Thanks!

PR #262, opened 2018-09-19
Support the multiViewport feature.

cdavis5e:
Clients can now set multiple viewports and scissor rects, but without a
geometry shader, it won't be much use. For now, however, we can support
`VK_EXT_shader_viewport_index_layer`, which will make this feature more
useful despite us not having geometry shaders.

### DIFF
--- a/MoltenVK/MoltenVK/Commands/MVKCommandEncoderState.h
+++ b/MoltenVK/MoltenVK/Commands/MVKCommandEncoderState.h
@@ -144,7 +144,7 @@ protected:
     void encodeImpl() override;
     void resetImpl() override;
 
-    MTLViewport _mtlViewport =  { 0, 0, 0, 0, 0, 0 };
+    std::vector<MTLViewport> _mtlViewports;
 };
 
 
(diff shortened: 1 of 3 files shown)


billhollings:
@cdavis5e 

Thanks. Just a couple of code convention adjustments needed.

Do you want to include `VK_EXT_shader_viewport_index_layer` in the list of supported extensions?

Do you need to pull in your latest SPIRV-Cross (viewport index) commit?

Also...while you are tweaking this...can you bump the value of `MVK_VERSION_PATCH`, please?

Every time the SPIRV-Cross pull commit in `SPIRV-Cross_repo_revision` changes...we should bump the the value of `MVK_VERSION_PATCH`. The reason is that the MoltenVK version number is a component of the `pipelineCacheUUID`...which is used to determine whether an external pipeline cache is still valid. In MoltenVK, the pipeline cache holds MSL, and any update to SPIRV-Cross might modify how the MSL is converted...and so it is safest to convert again and not to reload the MSL from cache.

Sorry...I should have caught that on your previous PR which did update `SPIRV-Cross_repo_revision`.

cdavis5e:
I intended to add `VK_EXT_shader_viewport_index_layer` in a later commit. I have another change that cleans up the extension list first. Do you want me to add that now?

(I also fixed the one you missed in `GPUObjects/MVKDevice.mm`. :)

PR #267, opened 2018-09-20
Support the VK_KHR_dedicated_allocation extension.

cdavis5e:
This is used to declare that a `VkDeviceMemory` object is for the sole
use of a particular buffer or image.

In fact, until Metal 1.2 on iOS (2.0 on Mac), Metal didn't really
support anything *but* dedicated allocations. MoltenVK hacks around this
limitation by manually syncing memory between buffers and images that
used the `VkDeviceMemory` object.

Maybe someday we can use the new `MTLHeap` object to support some sort
of sub-allocation scheme. But at this moment, it won't let you specify
the exact offset within the heap, nor can it be used for managed or (on
macOS) shared memory, nor will it let you directly map the heap memory
(something that could be worked around with an aliasable `MTLBuffer`
taking up the entire heap space), all of which makes it unsuitable for
implementing `VkDeviceMemory`. So, for now, we always prefer dedicated
allocation.


MoltenVK/MoltenVK/GPUObjects/MVKDeviceMemory.mm
```diff
@@ -105,30 +105,45 @@
 VkResult MVKDeviceMemory::addBuffer(MVKBuffer* mvkBuff) {
 	lock_guard<mutex> lock(_rezLock);
 
+	// If a dedicated alloc, ensure this buffer is the one and only buffer
+	// I am dedicated to.
+	if (_isDedicated && (_buffers.empty() || _buffers[0] != mvkBuff) )
+		return VK_ERROR_OUT_OF_DEVICE_MEMORY;
```
billhollings: When returning errors, MoltenVK convention is to log error text where appropriate using `mvkNotifyErrorWithText()` (eg. see 3 lines below). Is there a reason why you are not logging the error text here?

MoltenVK/MoltenVK/GPUObjects/MVKDeviceMemory.mm
```diff
@@ -105,30 +105,45 @@
...
 }
 
 VkResult MVKDeviceMemory::addImage(MVKImage* mvkImg) {
 	lock_guard<mutex> lock(_rezLock);
 
-	_images.push_back(mvkImg);
+	// If a dedicated alloc, ensure this image is the one and only image
+	// I am dedicated to.
+	if (_isDedicated && (_images.empty() || _images[0] != mvkImg) )
+		return VK_ERROR_OUT_OF_DEVICE_MEMORY;
```
billhollings: Ditto.

billhollings:
@cdavis5e Thanks!

PR #275, opened 2018-09-24
Better support subviews of 3D textures.

cdavis5e:
Now when we set up the `MTLRenderPassAttachmentDescriptor` for the 3D
texture, we set the `depthPlane` property instead of the `slice`
property. Combined with an appropriate setting for
`renderTargetArrayLength` in the base descriptor, this allows rendering
to a subset of a 3D texture.


billhollings:
@cdavis5e 

Thanks for all the submissions!

LunarG is looking to cut a Vulkan SDK release this week. I'm going to let them know tomorrow (Tuesday) a commit of MoltenVK for them to include in the SDK.

I'll pull these PR's in once the CI finishes.

And...if you'd like to see something additional and small make it into the next SDK...let me know...and get it submitted as soon as you can on Tuesday.

For safety...it might be prudent to leave something as large as the swizzling that you are working on for after this SDK release.

Sorry for the late notice. With you contributing so regularly...in future I'll keep you abreast of scheduling items like this.

Thanks...Bill

PR #276, opened 2018-09-24
MVKFenceSitter: Put a hash table read behind the lock.

cdavis5e:
To avoid a data race. Caught this with `ThreadSanitizer`.


billhollings:
@cdavis5e Nice catch!

PR #291, opened 2018-10-02
MVKPipeline: Fix getting the vertex attribute divisor struct.

cdavis5e:
Bah. I read the spec wrong. It's hung off the vertex input state, not
the top-level create info.


billhollings:
@cdavis5e Thanks for the fix.

PR #292, opened 2018-10-02
MVKDeviceMemory: Allow dedicated allocations to forget their resources.

cdavis5e:
Otherwise, if the resource is destroyed first (which it almost always
will be), then when the `VkDeviceMemory` is freed, it will access a
freed resource, at which point it will then proceed to fandango on core.


billhollings:
@cdavis5e Thanks.

PR #303, opened 2018-10-16
Support arbitrary swizzles of image data.

cdavis5e:
The swizzles are passed to shaders that need them using an "auxiliary
buffer", for which there must be room in the function argument table.
Code is inserted into the shaders to swizzle reads from sampled
textures--per the Vulkan spec, component mappings must only be applied
to sampled images.

Update SPIRV-Cross to pull in some fixes for handling swizzled image
data. Bump MoltenVK version.

Won't quite fix #302, but should fix a lot of the errors OP was seeing.

PR #305, opened 2018-10-17
Update shader caching for compatibility with texture swizzling.

billhollings:
MVKPipelineLayout set auxiliary buffer MSL index relative to already-calculated
push constant buffer MSL index.
MVKPipeline destructor release auxiliary buffer MTLBuffer instance.
Nudge MVK_VERSION_PATCH.

### DIFF
--- a/MoltenVK/MoltenVK/API/vk_mvk_moltenvk.h
+++ b/MoltenVK/MoltenVK/API/vk_mvk_moltenvk.h
@@ -48,7 +48,7 @@ extern "C" {
  */
 #define MVK_VERSION_MAJOR   1
 #define MVK_VERSION_MINOR   0
-#define MVK_VERSION_PATCH   24
+#define MVK_VERSION_PATCH   25
 
 #define MVK_MAKE_VERSION(major, minor, patch)    (((major) * 10000) + ((minor) * 100) + (patch))
 #define MVK_VERSION     MVK_MAKE_VERSION(MVK_VERSION_MAJOR, MVK_VERSION_MINOR, MVK_VERSION_PATCH)
@@ -631,7 +631,7 @@ typedef uint32_t MVKMSLSPIRVHeader;
 
 #endif // VK_NO_PROTOTYPES
 
-	
+
 #ifdef __cplusplus
 }
 #endif	//  __cplusplus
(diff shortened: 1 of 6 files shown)


billhollings:
@cdavis5e 

Unfortunately the swizzle enhancements in PR #303 broke Dota2 and the [Cinder Fish Tornado](https://github.com/brenwill/Cinder) demo.

Repairs in this PR to MoltenVK's shader caching and look-ups with the new shader parameters that you introduced has fixed the Data2 problems.

Before I merge, can you check out this PR and ensure that the changes here do not harm the apps you were testing against with your original changes, please?

With the swizzle enhancement, the Fish Tornado demo is now triggering a `Cannot yet map non-aliased parameter to Metal resource!` error from within `SPIRV-Cross`. I'm going to pull out the offending shaders and work on them in `SPIRV-Cross`. In the meantime, if you have any insights about what might cause that error, let me know.

Thanks.

cdavis5e:
> With the swizzle enhancement, the Fish Tornado demo is now triggering a `Cannot yet map non-aliased parameter to Metal resource!` error from within `SPIRV-Cross`. I'm going to pull out the offending shaders and work on them in `SPIRV-Cross`. In the meantime, if you have any insights about what might cause that error, let me know.
This error occurs when a sampled image is passed as a parameter to a function instead of being used as a global. ("Non-aliased" is because globals are lowered to function parameters in MSL--they "alias" the original globals. Parameters like this one that were present in the original SPIR-V don't.) Because any of the images the shader uses could conceivably be passed here, SPIRV-Cross can't figure out what its index in the texture argument table is, so it doesn't know which swizzle constant it needs to use here. One solution might be to pass the argument index along with the image; then in the function, use the parameter to look up the swizzle constant. Another might be to pass the swizzle constant itself.

billhollings:
@cdavis5e 

Thanks for the info about texture handling. Do you mean when the texture is passed as a function parameter within SPIR-V...as opposed to passed as a function parameter within MSL...which happens all the time in SPIRV-Cross because MSL has no globals?

Let me know once you've tested this PR against your test apps...since this PR changes some of the behaviour of your original design...in particular how the texture count and aux buffer MSL index is determined (I used a reference to the push constant indices that were already being calculated...and removed your map-construction code). I want to make sure I haven't broken anything on your side in the process before merging this PR.

Also...I have a few questions about the operation of the swizzling data and it would be good to have a short design discussion so I can understand some of the design requirements that went into the swizzling enhancement. Rather than address it here...I'll open a separate issue for that discussion.

billhollings:
Fixes issue #308.

Issue #321, opened 2018-10-31
MoltenVK was built without full bitcode Error


DiegoAce:
I'm finding it unusually difficult to build MoltenVK with bitcode for iOS. I noticed that the target MoltenVK-iOS has "Enable Bitcode" set to "Yes", but I still get "MoltenVK was built without full bitcode Error" when archiving my app with MoltenVK.framework. I have to add the user defined setting "BITCODE_GENERATION_MODE" with value "bitcode" on the MoltenVk and MoltenVKShaderConverter projects to build a framework with bitcode. Is this expected or am I missing something? I'm usually building with the "MoltenVK Package (Release) (iOS only)" target.

Also, in order to build the dylib with bitcode I had to add "-fembed-bitcode" to the create_dylib_ios.sh script.

billhollings:
Thanks for reporting that. It's the first we've heard of it.

Are you able to submit a pull request to add the changes that you needed to make?

If you do work on a PR...the organization of the build scripts just changed with PR #325...so pull the latest from the repo and merge locally first.

DiegoAce:
I submitted a pull request. I'm not 100% sure if this is what you want to do though. From my limited experience most libraries that I use I Archive them to create the bitcode, but I don't believe doing the Archive action on MoltenVKPackaging targets produced the libs. I'm not sure if I tried archiving the more specific MoltenVK-iOS target.

DiegoAce:
Ok, so I forgot how I used to get libs built in release with bitcode. I believe because Vulkan's dynamic layers are new to me and because MoltenVK includes a target that builds for release. On Xcode, building for release and Archiving have some differences, like the bitcode difference that I mentioned in my pull request. 
For other libraries I commonly use, like SDL, I would use the Archive action in SDL Xcode project to build for release with bitcode. I would then go get that lib or framework from the derived data. I'm not sure if that's how I should have been getting release/bitcode builds all along. Regardless, This same method does actually work with MoltenVK without setting "BITCODE_GENERATION_MODE=bitcode." Although, I still don't see anyway to build the dylib with bitcode without adding the "-fembed-bitcode" flag to the build script, since it's built by the script and not Xcode.
So I'm not sure what's the best practice for getting a release build with bitcode.

billhollings:
Thanks for the PR submission. Closing this now.

PR #326, opened 2018-10-31
Force Bitcode When Building

DiegoAce:
Enabling bitcode in Xcode only adds the -fembed-bitcode-marker flag, which doesn't actually create bitcode. The Archive action transforms that flag to -fembed-bitcode which actually creates the bitcode. Adding the User-Defined Setting BITCODE_GENERATION_MODE=bitcode forces bitcode to be created without archiving.


DiegoAce:
I added some pertinent comments about this in issue #321

PR #329, opened 2018-11-01
Remove destroyed resources from descriptor sets.

cdavis5e:
According to the Vulkan spec:

> "Entries that are not used by a pipeline can have uninitialized
>  descriptors or _descriptors of resources that have been destroyed_,
>  and executing a draw or dispatch with such a descriptor set bound
>  does not cause undefined behavior." [emphasis added]

But, in the current implementation, if a resource is destroyed, and a
descriptor set to which it was bound is subsequently bound to a
pipeline, then the now-freed object for the destroyed resource will be
passed to Metal, which almost always results in a segfault. To avoid
this, resources now keep track of descriptors to which they are bound,
and tell them to forget the resources when they are destroyed.
Conversely, if a descriptor set is destroyed before its bound resources,
the resources must now not notify the destroyed set. In this case, the
descriptor binding now tells the resource to forget the descriptor.

There is a race condition present if two threads attempt to destroy a
descriptor set and a resource at the same time. I've tried to mitigate
this with locking, but it won't help in the case where the resource
destruction happens first--the descriptor set will subsequently attempt
to modify a freed object. I don't know how to fix this.


cdavis5e:
I'm afraid putting this in `MVKBaseDeviceObject` isn't viable. It makes these objects non-copyable. Normally, this isn't a problem, since these objects are allocated off the heap, but `MVKPipelineLayout` objects apparently make copies of their associated descriptor set layout objects.

PR #333, opened 2018-11-01
Fix memoryTypes order on macOS to match spec

### LINKED ISSUE #314: Memory types out of order on macOS
[According to the specification](https://www.khronos.org/registry/vulkan/specs/1.1/html/vkspec.html#memory-device), the `memoryTypes` reported by the implementation in a `VkPhysicalDeviceMemoryProperties` must be sorted by `propertyFlags` sets:

> For each pair of elements `X` and `Y` returned in `memoryTypes`, `X` must be placed at a lower index position than `Y` if [...] the set of bit flags returned in the `propertyFlags` member of `X` is a strict subset of the set of bit flags returned in the `propertyFlags` member of `Y`.

On my MacBook Pro running macOS 10.14, running `vulkaninfo` from the 1.1.85.0 release shows:

```text
VkPhysicalDeviceMemoryProperties:
=================================
        memoryHeapCount       = 1
        memoryHeaps[0] :
                size          = 2147483648 (0x80000000) (2.00 GiB)
                flags:
                        VK_MEMORY_HEAP_DEVICE_LOCAL_BIT
        memoryTypeCount       = 3
        memoryTypes[0] :
                heapIndex     = 0
                propertyFlags = 0x1:
                        VK_MEMORY_PROPERTY_DEVICE_LOCAL_BIT
        memoryTypes[1] :
                heapIndex     = 0
                propertyFlags = 0xf:
                        VK_MEMORY_PROPERTY_DEVICE_LOCAL_BIT
                        VK_MEMORY_PROPERTY_HOST_VISIBLE_BIT
                        VK_MEMORY_PROPERTY_HOST_COHERENT_BIT
                        VK_MEMORY_PROPERTY_HOST_CACHED_BIT
        memoryTypes[2] :
                heapIndex     = 0
                propertyFlags = 0xb:
                        VK_MEMORY_PROPERTY_DEVICE_LOCAL_BIT
                        VK_MEMORY_PROPERTY_HOST_VISIBLE_BIT
                        VK_MEMORY_PROPERTY_HOST_CACHED_BIT
```

Based on my interpretation of the spec, I expect those last two entries to be swapped.  Applications using the suggested search loop to select a memory type will currently be using `COHERENT` memory with MoltenVK whether they requested it or not.

I had a quick look and this doesn't seem hard to fix; I would be happy to craft a PR, if that would be helpful?

spadix0:
Fixes #314

### DIFF
--- a/MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
@@ -1062,15 +1062,32 @@ void MVKPhysicalDevice::initMemoryProperties() {
 	//		- cmdbuff waitUntilCompleted (or completion handler)
 	//		- buffer/texture getBytes:
 
-	_memoryProperties.memoryHeapCount = 1;
-	_memoryProperties.memoryHeaps[0].flags = (VK_MEMORY_HEAP_DEVICE_LOCAL_BIT);
-	_memoryProperties.memoryHeaps[0].size = (VkDeviceSize)mvkRecommendedMaxWorkingSetSize(_mtlDevice);
-	_memoryProperties.memoryTypes[0].heapIndex = 0;
-	_memoryProperties.memoryTypes[0].propertyFlags = MVK_VK_MEMORY_TYPE_METAL_PRIVATE;	// Private storage
-	_memoryProperties.memoryTypes[1].heapIndex = 0;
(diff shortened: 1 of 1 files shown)


spadix0:
I thought this approach was easiest to read and least WET, although it may push the bounds of consistency...  easy enough to change if that's an issue.

Note that the index masks did *not* need to change; the swapped entries are always grouped together.

Tested on macOS w/cube and vulkaninfo.

billhollings:
@spadix0  Thanks!

Issue #241, opened 2018-09-03
Move scripts out of Xcode project files


ryandesign:
Diffs of Xcode project files are hard to read, especially when the diffs modify embedded scripts, since Xcode stores such scripts in a single line of the project file. A strategy to make diffs of such scripts easier to read, and to make editing such scripts easier, is to move them to separate files and have Xcode run them, using a method such as the one described in [this blog post](http://www.mokacoding.com/blog/better-build-phase-scripts/). Would you be willing to do this for your Xcode projects' scripts, or would you accept a pull request that does so?

billhollings:
@ryandesign 

This makes sense. I'll take care of it.

billhollings:
Added in PR #249.

PR #335, opened 2018-11-05
create_dylib.sh cleanup

ryandesign:
Hello, here are some changes to create_dylib.sh, which I had mentioned wanting to submit in #241.

You've already implemented one of the suggestions I wanted to make, which is combining the iOS and macOS scripts into one.

You've introduced a new variable `MVK_ARCH` to tell it what architecture to build for. Xcode already has a variable for that—`ARCHS`—so this PR switches to using that. It also accommodates `ARCHS` containing more than one value, should a universal build ever become relevant.

Since it's a library of C++ code being linked here, do so using clang++ instead of clang; this makes it unnecessary to manually add `-lc++`.

As far as I know it's always unnecessary to add `-lSystem` so remove that too.

And specify which C++ library to use, using the `-stdlib` flag.

I've been using these changes in the MacPorts MoltenVK port. We only build the macOS part, but I believe these changes should be correct for the iOS part too.

I think it should be possible to remove all the other variables that were added to the Xcode project before invoking the script, and to instead detect which values to use from within the script based on environment variables Xcode already sets. But `MVK_ARCH` is the only one I've done this for so far.

### DIFF
--- a/MoltenVK/MoltenVK.xcodeproj/project.pbxproj
+++ b/MoltenVK/MoltenVK.xcodeproj/project.pbxproj
@@ -875,7 +875,7 @@
 			);
 			runOnlyForDeploymentPostprocessing = 0;
 			shellPath = /bin/sh;
-			shellScript = "export MVK_OS=\"macosx\"\nexport MVK_ARCH=\"x86_64\"\nexport MVK_UX_FWK=\"AppKit\"\nexport MVK_MIN_OS_VERSION=${MACOSX_DEPLOYMENT_TARGET}\nexport MVK_IOSURFACE_FWK=\"-framework IOSurface\"\n\n\"${SRCROOT}/scripts/create_dylib.sh\"\n";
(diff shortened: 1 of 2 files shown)


billhollings:
@ryandesign 

Thanks for the update. I appreciate your `clang` option knowledge.

I'm curious...why did you need to invoke `printf` to make use of `${ARCHS}`?

ryandesign:
`ARCHS` contains the space-separated list of architectures e.g. `x86_64 i386` and I wanted to turn that into `-arch x86_64 -arch i386`; `printf` seemed like the easiest way to prepend a string to each member of the list.

In fact, at present, `ARCHS` will only contain a single value. I was trying to make MoltenVK build universal, and this was one of the fixes I needed to make. Once I got MoltenVK building universal, I realized that it's not useful because Metal is 64-bit only. But this change will make you ready for a possible future in which Macs or iPhones are once again available in multiple different architectures.

PR #352, opened 2018-11-20
Fix building for iOS < 12.0.  inputPrimitiveTopology requires iOS 12.…

danginsburg:
…0 and otherwise there are compile errors building MoltenVK with the older SDK.


MoltenVK/MoltenVK/Utility/MVKEnvironment.h
```diff
@@ -96,8 +96,13 @@
...
 #	define MVK_SUPPORT_IOSURFACE_BOOL    1
+#	define MVK_SUPPORT_INPUT_PRIMITIVE_TOPOLOGY_BOOL	1
 #endif
 
 #if MVK_IOS
 #	define MVK_SUPPORT_IOSURFACE_BOOL	(__IPHONE_OS_VERSION_MIN_REQUIRED >= __IPHONE_11_0)
+#	if !defined(__IPHONE_12_0)
+#		define __IPHONE_12_0 120000
+# 	endif
+#	define MVK_SUPPORT_INPUT_PRIMITIVE_TOPOLOGY_BOOL	(__IPHONE_OS_VERSION_MIN_REQUIRED >= __IPHONE_12_0)
```
cdavis5e: I actually think this check should use `MAX_ALLOWED` instead. The `MTLPrimitiveTopologyClass` enumeration should be available in that case. Also, the whole point of the `inputPrimitiveTopologyMVK` property is that we can use it regardless of whether the host supports it, and it will automatically set the real property if it's available.

danginsburg: Agreed, fixed.  I'm pretty agnostic as to how this gets fixed, I just wanted Bill to be aware that it wasn't building on Xcode 9.2 and this PR does fix that in case he wants to maintain building on older versions of Xcode/older SDKs.

billhollings:
@danginsburg 

Hmmm...so far...the focus has been on _**runtime**_ backwards compatibility...and as @cdavis5e mentions...that is why we have things like `inputPrimitiveTopologyMVK`...to cope with older OS runtimes that don't support newer calls.

We haven't focused on _**development**_ backwards compatibility yet because the cost-benefits are very different. Fewer developers are likely to be lagging...and for shorter periods of time. And there is also an extra level of challenge when building against SDK's that don't include the most recent function and constant references.

So...needless to say...I'd only want to start down that road if it was absolutely necessary...and I'd like to get some more info from you about it.

Up until now...we have only been moving to formal Xcode/SDK support once it is out of beta...and therefore all developers should be able to update.

What was the reason that you needed to build against an older Xcode/SDK version? Was it just a one-off...or should we expect that various development shops might be locked into an older Xcode/SDK release that they can't easily upgrade Xcode in time...say because a significant project is nearing completion?

If we do need to move in this direction...I'd probably want to try to consolidate all the changes into files like `MTLRenderPipelineDescriptor+MoltenVK.h/mm` by defining macros to fill in for constants and enums needed to support compiling against earlier SDK's. Centralizing like that should make it much simpler to maintain.

danginsburg:
> What was the reason that you needed to build against an older Xcode/SDK version?

I hadn't upgraded to Mojave and had a stable development environment and just was continuing to use it.  However, it's fine with me to require newer versions of the SDK, I just wanted to make sure you knew it wasn't building on older versions and that it was intentional.  I'll close this.

PR #367, opened 2018-12-04
Support the VK_KHR_16bit_storage extension.

cdavis5e:
This extension allows shaders to use 16-bit types in buffers, push
constants, and shader inputs and outputs. One step closer to Vulkan 1.1.


karl-lunarg:
When extensions are added, can they be added to the list in the MoltenVK_Runtime_UserGuilde.md?

billhollings:
@karl-lunarg Yes...you are correct. Thanks for noticing!

@cdavis5e Can you add the extensions you've just added to the list in `MoltenVK_Runtime_UserGuide.md`, please? Thanks.

Issue #306, opened 2018-10-19
Texture swizzling improvements


billhollings:
@cdavis5e 

Thanks very much for contributing the texture swizzling enhancement. That was a huge endeavour and a significant enhancement. Your contribution was most appreciated! Nice work!

In reviewing the improvements...and making them compatible with caching...I had a couple of design questions I thought I'd raise:

1. The auxiliary `MTLBuffer` is held by the `MVKPipeline`. But that one pipeline might be used in several renderings using different textures. This might set up a race condition where the GPU has not consumed the swizzle data for one texture before the CPU overwrites it with swizzle data for a different texture used in a later draw call. 

     I'm wondering if we could use `MVKCommandEncoder::setVertexBytes()`, etc. to dynamically set the swizzle content instead...similar to how `MVKPushConstantsCommandEncoderState` does. The `SPIRV-Cross` interaction would stay the same.

2. Texture swizzle slots between vertex and fragment shaders seem to overlap in the auxiliary buffer. Does this work if both the vertex and fragment shaders use textures with different swizzles?

3. Currently `swizzle_texture_samples` is always on for any shader that samples a texture. Shall we add a setting to `MVKConfiguration` to control this...to allow apps that know they don't need swizzling to turn it off to save on shader complexity and improve shader performance?

If these suggestions make sense to your overall design for texture swizzling...I can implement them.

cdavis5e:
> 1. The auxiliary `MTLBuffer` is held by the `MVKPipeline`. But that one pipeline might be used in several renderings using different textures. This might set up a race condition where the GPU has not consumed the swizzle data for one texture before the CPU overwrites it with swizzle data for a different texture used in a later draw call.

I see. Since this is state that depends on the bound descriptors (and later, the current draw parameters), it's not really viable to keep it in an `MTLBuffer` owned by the pipeline and shared between multiple draws.

>    I'm wondering if we could use `MVKCommandEncoder::setVertexBytes()`, etc. to dynamically set the swizzle content instead...similar to how `MVKPushConstantsCommandEncoderState` does. The `SPIRV-Cross` interaction would stay the same.

[...]

PR #376, opened 2018-12-09
MVKVector and small bug fix

aerofly:
This pull request replaces a few occurrences of std::vector with the custom MVKVector class. The MVKVector class features a special allocator that supports a configurable number of elements on the stack thus avoiding heap allocations if the vector size stays small.

I have not replaced every single occurrence of std::vector yet as I wanted to check first if this change is accepted.

This change greatly reduces the dynamic memory allocations during rendering, but there are still some objects that are allocated/freed from the Heap during rendering.

This pull requests also fixes a small issue if a shader with just a vertex stage is being used. In this case _needsFragmentAuxBuffer was true, even though no fragment shader was active:

`    if (_needsVertexAuxBuffer || _needsFragmentAuxBuffer) {
        _auxBuffer = [_device->getMTLDevice() newBufferWithLength: auxBufferSize options: MTLResourceStorageModeShared];	// retained
    }`


billhollings:
@aerofly Thanks for all your work on this...and thanks for dealing with all the conflicts while this has been sitting. Sorry I haven't had a chance to look at it yet...I've been traveling. I'll have a look tomorrow and pull it in if it looks good.

billhollings:
@aerofly Sounds good. Thanks for the review of testing.

billhollings:
@aerofly Since the point to this is to improve performance (by removing `mallocs`)...have you evaluated the performance of `MVKVector` vs `std::vector`?

@cdavis5e This is a fairly intrusive...but ideally invisible...change. It would be helpful to get more eyes on this. Feel free to evaluate design, behaviour and performance and provide feedback.

Thanks.

aerofly:
@billhollings: As performance varies wildly on different iOS devices in our case my measurements showed roughly 3-7% FPS performance increase and around 3-5% less CPU usage.

MoltenVK/MoltenVK/Utility/MVKVector.h
```diff
@@ -0,0 +1,496 @@
...
+    iterator( const size_t _index, const MVKVector &_vector ) : vector{ &_vector }, index{ _index } { }
+
+    iterator &operator=( const iterator &it )
+    {
+      vector = it.vector;
+      index = it.index;
+      return *this;
+    }
+
+    Type *operator->() const
```
cdavis5e: This iterator class is missing an `operator*`. I think that's why you were unable to use `std::copy_n()` or the `assign(begin, end)` idiom.

MoltenVK/MoltenVK/Utility/MVKVectorAllocator.h
```diff
@@ -0,0 +1,527 @@
...
+};
+
+
+//////////////////////////////////////////////////////////////////////////////////////////
+//
+// mvk_vector_allocator_default -> malloc based allocator for MVKVector
+//
+//////////////////////////////////////////////////////////////////////////////////////////
+template <typename T>
+class mvk_vector_allocator_default final
```
cdavis5e: This class is unused. I get that it's there in case we ever need it, but I think there's a case to be made that we could just add it when that time comes.

MoltenVK/MoltenVK/Utility/MVKVectorAllocator.h
```diff
@@ -0,0 +1,527 @@
...
+};
+
+
+//////////////////////////////////////////////////////////////////////////////////////////
+//
+// mvk_vector_allocator_with_stack -> malloc based MVKVector allocator with stack storage
+//
+//////////////////////////////////////////////////////////////////////////////////////////
+template <typename T, int N>
+class mvk_vector_allocator_with_stack
```
cdavis5e: This appears to be more or less compatible with the `std::allocator` template. In that case, you might be able (for our purposes, anyway) to get away with a templated alias:

```c++
template<typename T>
using MVKVector = std::vector<T, mvk_vector_allocator_with_stack<T, 8>>;
```

You could even add a non-type template parameter for the amount of inline space to reserve:

```c++
template<typename T, size_t N>
using MVKVector = std::vector<T, mvk_vector_allocator_with_stack<T, N>>;
```

Then, for example, the vector of `MTLViewport`s could be made to use a fixed size of 16, so it would never need to reallocate--well, at least as long as the maximum number of viewports remains 16.

cdavis5e:
Overall, I like this patch and the idea behind it. This reminds me of LLVM's `SmallVector` type, which has much the same reasoning behind it.

aerofly:
Thank you cdavis5e and billhollings for looking at my suggested changes. I made some minor adjustments to MVKVector to fix the issues cdavis5e mentioned. Again MVKVector is just a renamed version of our own std::vector like class we have been using for years in production code. The reason to use our own class over std::vector was to get better statistics on where and how a vector was used in our project.

I left in mvk_vector_allocator_default in case one wants to use MVKVector with no stack space at all.

As for using MVKVector, I recommend to create a better name for cases where it's clear by the Vulkan specification, that a certain number of elements is not exceed to use something like

```
template<typename T>
using MVKVector16 = MVKVector<Type, mvk_vector_allocator_with_stack<Type, 16>>;

MVKVector16<what_ever_type> variable;
```

MoltenVKShaderConverter/MoltenVKSPIRVToMSLConverter/SPIRVToMSLConverter.cpp
```diff
@@ -237,7 +237,7 @@ MVK_PUBLIC_SYMBOL bool SPIRVToMSLConverter::convert(SPIRVToMSLConverterContext&
 		mslOpts.aux_buffer_index = context.options.auxBufferIndex;
 		mslOpts.enable_point_size_builtin = context.options.isRenderingPoints;
 		mslOpts.disable_rasterization = context.options.isRasterizationDisabled;
-		mslOpts.swizzle_texture_samples = true;
+		mslOpts.swizzle_texture_samples = false;
```
billhollings: As per the [discussion](https://github.com/KhronosGroup/MoltenVK/issues/306#issuecomment-442178227) in #306...we'll probably make `mslOpts.swizzle_texture_samples = false;` the default in the end...but I would like to do so in a separate PR that contains flexibility for easily changing it. So can you remove this from this PR, please?

MoltenVK/MoltenVK/Utility/MVKVectorAllocator.h
billhollings: Did you evaluate the suggestion by @cdavis5e that the allocator might be all you need?

MoltenVKShaderConverter/MoltenVKSPIRVToMSLConverter/SPIRVToMSLConverter.cpp
aerofly: Yes, I will revert back to true again. Keep in mind this has a huge impact on performance in our case, e.g. we get 30 fps instead of 25 fps.

MoltenVK/MoltenVK/API/mvk_datatypes.h
```diff
@@ -295,6 +295,14 @@ VkExtent3D mvkMipmapBaseSizeFromLevelSize3D(VkExtent3D levelSize, uint32_t level
  */
 MTLSamplerAddressMode mvkMTLSamplerAddressModeFromVkSamplerAddressMode(VkSamplerAddressMode vkMode);
 
+#if MVK_MACOS
+/**
+ * Returns the Metal MTLSamplerBorderColor corresponding to the specified Vulkan VkBorderColor,
+ * or returns MTLSamplerBorderColorTransparentBlack if no corresponding MTLSamplerBorderColor exists.
+ */
+MTLSamplerBorderColor mvkMTLSamplerBorderColorFromVkBorderColor(VkBorderColor vkColor);
```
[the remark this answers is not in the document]
aerofly: Sorry, like I said, due to the changes that came in after I made my pull request I was probably messing up something. I will fix this again, give me a few days.

billhollings:
@aerofly Thanks for making changes.

No worries. This PR did not cause confusion. But it is a substantive change that influences a significant chunk of MoltenVK...so it's good to vet it well.

aerofly:
@billhollings Thank you for merging this pull request into MoltenVK, I really appreciate it. Thanks also for your time and feedback in getting my changes into MoltenVK properly.

One last question remains on my side: In MVKPipeline.mm I made a change on line 347:

`_needsFragmentAuxBuffer = false`

The line fixed an issue I got during rendering into a depthmap using a shader with no fragment shader. When no fragment shader was active I got a Metal validation error a few lines lower at

`    if (_needsVertexAuxBuffer || _needsFragmentAuxBuffer) {
        _auxBuffer = [_device->getMTLDevice() newBufferWithLength: auxBufferSize options: MTLResourceStorageModeShared];	// retained
    }`

As _needsFragmentAuxBuffer was true but auxBufferSize was zero.

PR #383, opened 2018-12-10
MVKDeviceMemory: Try creating an MTLBuffer before allocating heap memory.

### LINKED ISSUE #377: Binding memory to an object after mapping results in an invalid mapped pointer
I began experiencing strange memory corruption and crashes when dealing with host-visible mapped memory after updating MoltenVK on code that has no validation errors and no issues with Nvidia and Intel on Linux and Windows.  After building MoltenVK with the address sanitizer enabled I discovered the cause:

My application maps host-visible memory immediately after allocation but before it's bound to a buffer, the underlying behavior of which was changed in commit 958f56fb853cf8a9cdaec9f329d1efdc691f810d inside MVKDeviceMemory.mm. Calling vkMapMemory() on an allocated chunk of host visible memory that has not yet been bound to a buffer results in an internal call to ensureHostMemory(), which allocates memory with posix_memalign() and returns the pointer.  When this memory is later bound to a buffer object via a call to vkBindBufferMemory(), MVK internally calls ensureMTLBuffer(), which frees the previously allocated memory and leaves the calling application that mapped the memory with a pointer that points to freed memory.

Basic repro steps:

1. Create a buffer object //vkCreateBuffer()
2. Get the buffer memory requirements //vkGetBufferMemoryRequirements()
3. Allocate host visible memory for the buffer //vkAllocateMemory()
4. Map and get a pointer to the host visible memory //vkMapMemory()
5. Bind the memory to the buffer //vkBindBufferMemory()
6. The pointer to the memory mapped in step 4 is now invalid and writing to it causes undefined behavior.

cdavis5e:
If the device memory is subsequently bound to a `VkBuffer` after it has
been mapped, the heap memory will be freed, leaving dangling pointers in
the client. Instead, create an `MTLBuffer` right away--it is *device*
memory after all--and use its memory.

Another option might be to use `-[MTLDevice
newBufferWithBytesNoCopy:length:options:deallocator:]`, but that
requires that the entire allocation come from a single contiguous
virtual memory mapping, a requirement that is not necessarily satisfied
by `malloc(3)` memory.

Fixes #377.

### DIFF
--- a/MoltenVK/MoltenVK/GPUObjects/MVKDeviceMemory.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKDeviceMemory.mm
@@ -38,7 +38,7 @@ VkResult MVKDeviceMemory::map(VkDeviceSize offset, VkDeviceSize size, VkMemoryMa
 		return mvkNotifyErrorWithText(VK_ERROR_MEMORY_MAP_FAILED, "Memory is already mapped. Call vkUnmapMemory() first.");
 	}
 
-	if ( !ensureHostMemory() ) {
+	if ( !ensureMTLBuffer() && !ensureHostMemory() ) {
 		return mvkNotifyErrorWithText(VK_ERROR_OUT_OF_HOST_MEMORY, "Could not allocate %llu bytes of host-accessible device memory.", _allocationSize);
 	}
 


MoltenVK/MoltenVK/GPUObjects/MVKDeviceMemory.mm
```diff
@@ -38,7 +38,7 @@
 		return mvkNotifyErrorWithText(VK_ERROR_MEMORY_MAP_FAILED, "Memory is already mapped. Call vkUnmapMemory() first.");
 	}
 
-	if ( !ensureHostMemory() ) {
+	if ( !ensureMTLBuffer() && !ensureHostMemory() ) {
```
billhollings: An elegant solution!

I was concerned that large images can require more memory than a `MTLBuffer` can provide...that was a main reason why `MVKDeviceMemory` had to introduce host memory separate from `MTLBuffer` memory...but I see that `ensureMTLBuffer()` already guards against that in this situation...so this should cover both use cases.

PR #390, opened 2018-12-11
MVKImage: Round up byte alignment to the nearest power of 2.

cdavis5e:


billhollings:
Thanks. I just noticed this issue with some testing I was doing today...and was going to fix it...but I'm happy to have you beat me to it!

PR #391, opened 2018-12-11
Add new formats and fix existing ones

cdavis5e:
Add some more vertex formats. Add fallbacks in case they're not supported (this is the primary cause of most of the changed lines).

Compressed formats can only be sampled, and cannot be used in buffer views since they cannot be used to make linear textures. Nor can depth/stencil formats be used for buffer views for the same reason. On Mac, similar to all compressed formats, `VK_FORMAT_E5B9G9R9_UFLOAT_PACK32` can only be sampled. (We should really zero `linearTilingFeatures` for compressed and depth/stencil formats, since this isn't supported.)

Support `VK_FORMAT_FEATURE_BLIT_SRC_BIT` for depth/stencil formats. *Don't* support `VK_FORMAT_FEATURE_BLIT_DST_BIT` for linear textures on Mac, since they can't be used as render targets there.

Support `VK_FORMAT_A2R10G10B10_UNORM_PACK32`.


billhollings:
Thanks for all the detailed effort on this one!

PR #403, opened 2018-12-14
vkCmdPushDescriptorSet: Fix mapping of binding numbers to descriptor layouts.

cdavis5e:
In the past, this function just used the passed-in binding number to
index into the array of bindings. When the binding numbers and the
indices of the array matched up, this worked really well.

When they didn't, it broke spectacularly.

So now we keep track of which binding numbers map to which elements in
the array. That way, we don't have to assume that the array indices and
the binding numbers will match up.


cdavis5e:
(And yes, I'm aware this is my fault. 😬)

PR #406, opened 2018-12-15
Add missing include MVKEnvironment.h to MVKImage.mm.

### LINKED ISSUE #404: MVK_SUPPORT_IOSURFACE_BOOL not properly set in MVKImage
This line in `useIOSurface` seems to be detected incorrectly on macOS.
```
#if MVK_SUPPORT_IOSURFACE_BOOL
...
#endif
```
https://github.com/KhronosGroup/MoltenVK/blob/0ee58d2d42cfad22b43e0d1a12d209b22a57a8e4/MoltenVK/MoltenVK/GPUObjects/MVKImage.mm#L332

Even though the `metalFeatures.ioSurfaces` Metal-specific feature is detected correctly, and should be supported on this GPU on macOS 10.15.

I removed that `#if` check and compiled a static library, and everything seems to be working a lot better (not perfect, yet, but global IOSurface sharing is working at least).

I can't figure out if some combination of manual defines will work around it, but I think it should be detected automatically anyway, and that doesn't seem to be happening. Also `vkUseIOSurfaceMVK` currently fails silently with `VK_SUCCESS` when the detection fails, which was unexpected.

billhollings:
Fixes #404.

### DIFF
--- a/MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
@@ -21,6 +21,7 @@
 #include "MVKCommandBuffer.h"
 #include "mvk_datatypes.h"
 #include "MVKFoundation.h"
+#include "MVKEnvironment.h"
 #include "MVKLogging.h"
 #import "MTLTextureDescriptor+MoltenVK.h"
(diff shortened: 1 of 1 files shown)


paulreimer:
Thanks, that was really fast!

PR #411, opened 2018-12-17
vkCmdCopyImage: Cast source image to the destination format.

cdavis5e:
In which I am proved wrong again. While Metal indeed won't assert if the
pixel formats don't match, the results of attempting such a copy are...
mixed, to say the least. Sometimes, nothing is copied at all. Other
times, pixels are copied in weird ways that don't seem to conform at all
to expectation. (This is most apparent in the output of the CTS's image
copying tests.) It's best, then, to ensure that the source and
destination have the same format.

Unfortunately, in the process of writing this, my fears regarding not
being able to make uncompressed views of compressed images were
confirmed. So we can't do this for compressed images.


cdavis5e:
I've wondered about that. I keep thinking I really need to add a particular value for a compressed format to that enum.

PR #420, opened 2018-12-19
If user requests system default device curing vkCreateDevice, call …

mlfarrell:
Note.  I ensured the only call to MVKDevice constructor is through vkCreateDevice.  I wanted to do this earlier, but we cannot call the trigger method MTLCreateSystemDefaultDevice() if the user didn't want a high power device.  So deferring it until the vkCreateDevice call is the best method.

### DIFF
--- a/MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
@@ -1757,6 +1757,20 @@ uint32_t MVKDevice::expandVisibilityResultMTLBuffer(uint32_t queryCount) {
 MVKDevice::MVKDevice(MVKPhysicalDevice* physicalDevice, const VkDeviceCreateInfo* pCreateInfo) {
 
 	initPerformanceTracking();
+    
+#if MVK_MACOS
+    //on mac OS, the wrong GPU will drive the screen (graphics switching will not occur)
+    //unless we call this specific MTLCreateSystemDefaultDevice method to create the metal device
+    id<MTLDevice> device = physicalDevice->getMTLDevice();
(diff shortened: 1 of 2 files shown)


mlfarrell:
I don’t know what’s going on above. I signed the agreement multiple times.

cdavis5e:
It looks like GitHub wasn't able to match up the name and email of your commit with your GitHub account.

mlfarrell:
Looks like its taken care of.

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
```diff
@@ -1757,6 +1757,20 @@
...
+#if MVK_MACOS
+    //on mac OS, the wrong GPU will drive the screen (graphics switching will not occur)
+    //unless we call this specific MTLCreateSystemDefaultDevice method to create the metal device
+    id<MTLDevice> device = physicalDevice->getMTLDevice();
+    if (!device.headless && !device.lowPower) {
+        id<MTLDevice> sysDefaultDevice = MTLCreateSystemDefaultDevice();
+        
+        //lets be 100% sure this is the device the user asked for
+        if (sysDefaultDevice.registryID == device.registryID) {
+            physicalDevice->replaceMTLDevice(sysDefaultDevice);
```
billhollings: Is the requirement for this swap because Metal returns a different `MTLDevice` instance from `MTLCreateSystemDefaultDevice` for the same GPU that it returned in `MTLCopyAllDevices`?

mlfarrell: Sorry, accidentally resolved.
Yes that’s correct, see the conversation in the attached issue.

PR #422, opened 2018-12-19
Use an inline constant buffer for the auxiliary buffer.

cdavis5e:
As discussed and agreed upon in #306.

Since a pipeline might be shared between multiple draws, but those draws
could use different textures, it isn't tenable to have a single aux
buffer shared between those draws. Instead, we'll set up constant
buffers using the `setXxxBytes()` method on the command encoder. That
way, separate draws using the same pipeline won't interfere with each
other.

This change also uses separate buffers for the vertex and fragment
shaders, since they may have different sets of textures.


billhollings:
Thanks! This was on my todo list from #306. I'm happy you beat me to it! 👍

PR #423, opened 2018-12-19
Use a specific format type for compressed formats.

cdavis5e:
Instead of assuming that `kMVKFormatNone` means the format is
compressed. This makes the code doing these checks more obvious and
self-documenting.

Since I changed the public MoltenVK-specific API, I also bumped the
`VK_MVK_moltenvk` spec revision.


billhollings:
Thanks!

PR #425, opened 2018-12-21
Further reductions of heap allocations during rendering

aerofly:
As MVKVector is now accepted in MoltenVK, I did some minor adjustments to further reduce MoltenVK heap allocations during rendering. In our use case we are now down to negligible allocations < 10.

For push constants I introduced MVKVector128 which has 128 bytes fixed space on the stack which is what the spec has as a minimum requirement. E.g. its sufficient for 2 4x4 transform matrices as an example.


MoltenVK/MoltenVK/Utility/MVKVector.h
```diff
@@ -524,5 +524,10 @@ template<class Type, class Allocator = mvk_vector_allocator_with_stack<Type, 8>>
   }
 };
 
+
+template<typename Type>
+using MVKVector128 = MVKVector<Type, mvk_vector_allocator_with_stack<Type, 128>>;
```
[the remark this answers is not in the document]
aerofly: I apologize for this confusion and for not properly explaining what my changes do or intend to do. As for the merge conflicts, sorry also, we do not work with Git, so I am probably not using it properly. I will add my changes again and hopefully properly and try to apply a new pull request after Christmas in case you are interested in merging my changes.

But some more explanations:

As for MVKVector128: Again it's basically a std::vector with the exception, that it has 128 elements in its own class, which it can use, before a heap allocation takes place. That's all. If you need more than 128 elements, MVKVector will reserve this space from the heap, so it can of course grow. It's not a maximum number of elements. Really if you want another name for this, I have no issues with that.

The reason I was talking about 'stack' is when you use this MVKVector in a function:

`void function()
{
  MVKVector<int> test;
  test.push_back( 13 );
}`

In this case, no heap is used, instead the pre allocated elements of MVKVector ( which I defaulted to 8 ) are taken from the stack. So you are right, the pre-allocation takes place in the MVKVector class itself.

The main motivation for using a class like this in MoltenVK is, that in some places std::vector was used dynamically and therefor you do have heap allocations per frame. Again my measurements showed thousands of allocations per frame in our use case. My class will be obsolete in cases where you use the object pool mechanism you mentioned in a previous posting. From what I can see since my initial pull request, you already did make changes to better reuse commands that have std::vector data members. So in this case you could revert back to std::vector.

But there is another reason for using this class:

In case only the pre allocated number of elements are used MVKVector might give you a better cache usage, as accessing the elements avoids a cache miss in some situations.

Facebook's folly for example has a similar vector. If you read the first few sentences you might get a better idea:

https://github.com/facebook/folly/blob/master/folly/docs/small_vector.md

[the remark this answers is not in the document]
billhollings: Don't worry...it's not confusing. I think it the overall idea is great...and quite useful. Thanks for the background explanation...and for the reference to Folly's `small_vector`.

Does `MVKVector` also meet the requirement of contiguous memory after a spill-over?

I'm fine with all of this...except for the use of dedicated class names...per my explanation above.

Is it possible to move the dedicated size to a constructor parameter...so the size only appears on the line that creates the vector...instead of everywhere the vector is passed?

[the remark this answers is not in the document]
cdavis5e: > Don't worry...it's not confusing. I think it the overall idea is great...and quite useful. Thanks for the background explanation...and for the reference to Folly's `small_vector`.

See also [`llvm::SmallVector`](/llvm-mirror/llvm/blob/69ba13fd38262b31264a9447aa2d7d7af95daabd/include/llvm/ADT/SmallVector.h#L847). I mentioned that on the other PR #376.

> 
> I'm fine with all of this...except for the use of dedicated class names...per my explanation above.
> 
> I see Folly uses the dedicated size as a template parameter.

As does LLVM.

> 
> Is it possible to move the dedicated size to a constructor parameter...so the size only appears on the line that creates the vector...instead of everywhere the vector is passed?

In LLVM, this problem is solved by having `SmallVector` derive from a base, without the inline size NTTP, that has the guts of the vector implementation. Then references to the base class are passed around, and only the final `SmallVector` template itself needs the inline size as an NTTP. Strange that Folly doesn't do this too.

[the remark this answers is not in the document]
aerofly: @cdavis5e Thank you for the idea with the base class. I will investigate this in a few days. It's exactly what's missing to make this a cleaner interface.

@billhollings Yes, MVKVector meets the requirement of contiguous memory after a spill-over. Again all it is, is a simplified version of std::vector.

Anyway, give me a few more days after Christmas and I will try to implement c5davis5e recommendations to avoid using those special class names.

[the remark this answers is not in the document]
billhollings: Thanks for continuing to look at this. Looking forward to the updates!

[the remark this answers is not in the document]
billhollings: @aerofly Just wondering if are still planning to rework this.

If you're stuck for time...perhaps we could see if someone else can look at it to finish off this initial fixed size implementation.

It would be nice to be able to start to make use of the fixed allocations you've made possible!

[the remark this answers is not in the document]
aerofly: @billhollings I apologize for the delay. I have pretty much finished making the suggested changes. I will try to modify my pull request tomorrow. So in general usage is now as follows:

1) `MVKVector<uint32_t> dynamicOffsets`
  Same as using `std::vector<uint32_t> dynamicOffsets`

2) `MVKVector<uint32_t, 4> dynamicOffsets`
  Reserves 4 elements of uint32_t within the class itself.
  If size of vector stays <= 4 no heap allocations take place.
  If dynamicOffsets grows to > 4 elements, all elements are taken from heap.
  All elements are always in contiguous memory.
  
3) When passing around a MVKVector to a function, one declares the function as follows
  `void function( MVKVectorBase<uint32_t> &vector)`
  `{`
  `  // do something with vector...`
  `  vector.push_back( 13 );`
  `}`
  `  `
  `MVKVector<uint32_t, 4> dynamicOffsets;`
  `function( dynamicOffsets );`

So the only quirk is the fact that you have to use the name 'MVKVectorBase' when passing around MVKVector's to functions. This avoids your previous comment of having to know the number of pre allocated elements when calling a function, but you have to use this special name now.

I hope that's ok.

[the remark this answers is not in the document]
billhollings: Great. This is looking a bit more generic now. Thanks.

My only suggestion at this point is to wonder if we could flip the names? Make `MVKVector` the name of the base class...and something like `MVKVectorInline` the name of the pre-allocation subclass? 

That would result in:

```
void function( MVKVector<uint32_t> &vector)
{
// do something with vector...
vector.push_back( 13 );
}

MVKVectorInline<uint32_t, 4> dynamicOffsets;
function( dynamicOffsets );
```

That makes the generic  name `MVKVector` the thing that appears everywhere as it passed around to functions...and the specialized name `MVKVectorInline` only appears once...and only where the inline-allocation parameter that defines it appears.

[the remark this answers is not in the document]
aerofly: Sounds good. Let me see if I can implement this.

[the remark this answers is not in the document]
aerofly: Ok, I uploaded the new version of MVKVector alongside some code changes. Please review my changes.

You might better know what values to use for MVKVectorInline in certain places.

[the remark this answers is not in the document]
cdavis5e: I, uh, think you need to merge with `HEAD` again.

This seems to have previous changes that were merged to `HEAD` between last time and now included in it. Maybe you need to rebase?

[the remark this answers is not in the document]
billhollings: @aerofly 

Yeah...there is a fair bit of black art to git.

To keep your repo in sync with this repo...first...if you haven't already done it...add the Khronos repo as a remote to your local repo:

    git remote add upstream https://github.com/KhronosGroup/MoltenVK.git

You only have to do this once...and you can verify it via `git remote -v`...which will list all your remotes. This list will now include `origin` (yours on Github) and `upstream` (Khronos on Github).

Then...once you have committed to your local repo...you can periodically do:

    git pull upstream master

...to merge the current state of `upstream` into your local repo.

Depending on the overall order of commits...this will likely trigger a sync commit...which is fine. And you may or may not have to resolve conflicts if needed.

Then push to your GitHub repo as normal via `git push`.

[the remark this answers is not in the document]
aerofly: @billhollings Thank you very much, I am slowly understanding how its supposed to work. Quite a bit different workflow than I am used to.
So my pull request should be ok now, at least it seems to work on my side.
If you want me to perform any other changes, please let me know.

MoltenVK/MoltenVK/Commands/MVKCommandEncoderState.mm
```diff
@@ -402,7 +402,8 @@
 #pragma mark MVKResourcesCommandEncoderState
 
 // Updates the swizzle for an image in the given vector.
-static void updateSwizzle(MVKVector<uint32_t> &constants, uint32_t index, uint32_t swizzle) {
+template<typename T>
+static void updateSwizzle(T &constants, uint32_t index, uint32_t swizzle) {
```
cdavis5e: This should no longer need to be a template. In fact, you shouldn't need to change this at all anymore.

MoltenVK/MoltenVK/Commands/MVKCmdRenderPass.h
```diff
@@ -115,7 +115,7 @@ class MVKCmdSetViewport : public MVKCommand {
 
 private:
 	uint32_t _firstViewport;
-	MVKVector<MTLViewport> _mtlViewports;
+	MVKVectorInline<MTLViewport, 8> _mtlViewports;
```
cdavis5e: It should be possible to make this 16 now.

MoltenVK/MoltenVK/Commands/MVKCmdRenderPass.h
```diff
@@ -134,7 +134,7 @@ class MVKCmdSetScissor : public MVKCommand {
 
 private:
 	uint32_t _firstScissor;
-	MVKVector<MTLScissorRect> _mtlScissors;
+	MVKVectorInline<MTLScissorRect, 8> _mtlScissors;
```
cdavis5e: Ditto.

MoltenVK/MoltenVK/Commands/MVKCommandEncoderState.h
```diff
@@ -147,7 +147,7 @@ class MVKViewportCommandEncoderState : public MVKCommandEncoderState {
     void encodeImpl() override;
     void resetImpl() override;
 
-    MVKVector<MTLViewport> _mtlViewports;
+    MVKVectorInline<MTLViewport, 8> _mtlViewports;
```
cdavis5e: Another one that could be made 16.

MoltenVK/MoltenVK/Commands/MVKCommandEncoderState.h
```diff
@@ -176,7 +176,7 @@ class MVKScissorCommandEncoderState : public MVKCommandEncoderState {
     void encodeImpl() override;
     void resetImpl() override;
 
-    MVKVector<MTLScissorRect> _mtlScissors;
+    MVKVectorInline<MTLScissorRect, 8> _mtlScissors;
```
cdavis5e: Ditto.

MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
```diff
@@ -1027,7 +1027,8 @@
 		// If this image is not yet available, extract and signal the first semaphore and fence.
 
 		signaler = _availabilitySignalers.front();
-		_availabilitySignalers.pop_front();
+    _availabilitySignalers.erase( _availabilitySignalers.begin() );
```
cdavis5e: I'm concerned about the performance implications of this. I think there's a reason this code originally used `std::list`. Vectors generally do not do well with insertion/removal from the front.

MoltenVK/MoltenVK/GPUObjects/MVKPipeline.h
```diff
@@ -143,8 +142,8 @@ class MVKGraphicsPipeline : public MVKPipeline {
 	VkPipelineRasterizationStateCreateInfo _rasterInfo;
 	VkPipelineDepthStencilStateCreateInfo _depthStencilInfo;
 
-  MVKVector<MTLViewport> _mtlViewports;
-  MVKVector<MTLScissorRect> _mtlScissors;
+  MVKVectorInline<MTLViewport, 8> _mtlViewports;
```
cdavis5e: Two more to make 16.

billhollings:
Okay...looks good to go!

@aerofly Thanks very much for all your work...and rework...on this!

@cdavis5e Thanks for your thorough reviews of this!

I'd like to get this in the SDK release going out today...so I'm pulling this in as-is...and I will create an additional commit to cover the latest round of changes that @cdavis5e recommends...since they are basically usage changes. I hope that is okay with everyone.

MoltenVK/MoltenVK/Commands/MVKCommandEncoderState.mm
billhollings: @cdavis5e What do you mean by "shouldn't need to change this at all"?

billhollings: Hmmm...attempting to change this back from a template causes compile errors. Wasn't obvious to me why...so I'm leaving it for now.

cdavis5e: That is, it should've been possible to just use `MVKVector<uint32_t>&` as the type of `constants`.

MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
aerofly: @cdavis5e It really depends on the size of _availabilitySignalers. If this is smaller lets say than 20 elements, the new version should still be a bit faster as it avoids the heap allocation/deallocation which involves a system call. My initial thought was, that size of _availabilitySIgnalers will never exceed more than 4 elements as this list deals with swapchain images? If I am wrong, please revert back to the original version here.

MoltenVK/MoltenVK/GPUObjects/MVKPipeline.h
aerofly: Thank you for adjusting the sizes of those vectors. Optimizing this for common usage cases is exactly the right method.

MoltenVK/MoltenVK/Commands/MVKCommandEncoderState.mm
aerofly: Sorry for not commenting on my change here. If you use MVKVector as the function parameter, this means you are accessing the vector using virtual functions, e.g. element access or insertion is a virtual function call. When using the template function as above, this virtual function call is not necessary, instead the actual member functions of MVKVectorInline or MVKVectorDefault are called. The performance implications are absolutely negligible, so its fine to use the original version instead of my template version.

So feel free to revert it back to your original version, if you prefer that.

aerofly:
Thank you for approving this pull request, it makes it easier for us now to add newer versions of MoltenVK into our apps. The changes I did are already live in two games for Mac OS and iOS and so far we have not received any reports of crashes.

MoltenVK/MoltenVK/Commands/MVKCommandEncoderState.mm
billhollings: As mentioned...I tried reverting it back to `MVKVector<uint32_t>&` as part of this...but that causes strange compile errors in the `updateSwizzle()` function about `resize()` not being recognized...and the assignment operator not being valid.

aerofly: Yes sorry, there was a reason for adding the template. The MVKVector class does not offer all member functions. I wanted to avoid all those virtual function calls ( again no real performance implication ). If desired, I will add all those functions to this class to make the usage of MVKVector simpler.

billhollings: Hmmm...yeah...if this is going to be the integral part of MoltenVK that is now is...I think you'll need to add/adjust functions to make it compatible with `std::vector`.

Not a huge priority...since it's functional now...but when you have spare cycles...maybe add bits and pieces as you go (start with `resize()` and assignment to get that `updateSwizzle()` working). 

Otherwise it looks like we'll end up littering the MoltenVK codebase with unnecessary template functions.

Issue #429, opened 2018-12-28
mvkWaitForFences returns VK_SUCCESS with timeout 0


freesig:
`mvkWaitForFences` will always return `VK_SUCCESS` with a timeout of 0 even if the fence isn't ready.
```c++
VkResult mvkWaitForFences(uint32_t fenceCount,
						  const VkFence* pFences,
						  VkBool32 waitAll,
						  uint64_t timeout) {

	VkResult rslt = VK_SUCCESS;
	MVKFenceSitter fenceSitter(waitAll);

	for (uint32_t i = 0; i < fenceCount; i++) {
		((MVKFence*)pFences[i])->addSitter(&fenceSitter);
	}

	if ( !fenceSitter.wait(timeout) && timeout > 0 ) {
		rslt = mvkNotifyErrorWithText(VK_TIMEOUT, "Vulkan fence timeout after %llu nanoseconds.", timeout);
	}

	for (uint32_t i = 0; i < fenceCount; i++) {
		((MVKFence*)pFences[i])->removeSitter(&fenceSitter);
	}

	return rslt;
}
```
`&& timeout > 0 ` means that rslt can never be set to VK_TIMEOUT when timeout is 0.
The vulkan spec states:

> If timeout is zero, then vkWaitForFences does not wait, but simply returns the current state of the fences. VK_TIMEOUT will be returned in this case if the condition is not satisfied, even though no actual wait was performed.

billhollings:
Thanks for pointing this out.

This resulted from a typo during recent streamlining of fence handling.

Fixed now in PR #433.

`VK_TIMEOUT` is returned even on zero wait. And the error is only logged if the timeout is greater than zero.

Please retest with latest code and close this issue if fixed.

freesig:
yep this works now

Issue #430, opened 2018-12-28
Bizarrely high value returned from vkGetSwapchainImagesKHR() on macOS 10.14


tokyovigilante:
I'm working on what will (eventually) be a Vulkan renderer in Swift, and starting by porting the VkCube demo. Using 1.1.92 SDK with macOS Mojave on a 2018 13" RMBP (Intel Iris 655), but am having issues creating a swapchain. 

Firstly, getting the surface's capabilities with `vkGetPhysicalDeviceSurfaceCapabilitiesKHR` (previously created with `vkCreateMacOSSurfaceMVK` and backed by a CAMetalLayer-based view) returns a `VkSurfaceCapabilitiesKHR` struct which quotes a `maxImageCount` of 2, which is not good for triple-buffering. 

Secondly, I then create the swapchain as below, but get a presumably spurious result of 32766 images back from `vkGetSwapchainImagesKHR`, when I would expect 2 as passed in with the `VkSwapchainCreateInfoKHR` struct.

Abridged code below:

```Swift
        var createInfo = VkMacOSSurfaceCreateInfoMVK(
            sType: VK_STRUCTURE_TYPE_DISPLAY_SURFACE_CREATE_INFO_KHR,
            pNext: nil,
            flags: 0,
            pView: userData)
        result = vkCreateMacOSSurfaceMVK(instance._vkInstance, &createInfo, nil, &_vkSurface)
....
        var desiredNumberOfSwapchainImages: UInt32 = 3
        if desiredNumberOfSwapchainImages < surfaceCapabilities.minImageCount {
            desiredNumberOfSwapchainImages = surfaceCapabilities.minImageCount
        }
        if surfaceCapabilities.maxImageCount > 0 && desiredNumberOfSwapchainImages > surfaceCapabilities.maxImageCount {
            // Application must settle for fewer images than desired:
            desiredNumberOfSwapchainImages = surfaceCapabilities.maxImageCount
        }
        var swapchainCreateInfo = VkSwapchainCreateInfoKHR(
            sType: VK_STRUCTURE_TYPE_SWAPCHAIN_CREATE_INFO_KHR,
            pNext: nil,
            flags: 0,
            surface: surface._vkSurface,
            minImageCount: desiredNumberOfSwapchainImages,
            imageFormat: format,
            imageColorSpace: colorspace,
            imageExtent: size.vkExtent,
            imageArrayLayers: 1,
[...]

PR #433, opened 2018-12-30
Fix issues #429 & #430.

billhollings:
- Increase max swapchain images to 3.
- Return VK_TIMEOUT even on zero wait if fences not signalled.

Fixes issues #429 & #430.

### DIFF
--- a/MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
@@ -341,10 +341,9 @@ VkResult MVKPhysicalDevice::getSurfaceCapabilities(MVKSurface* surface,
 
     VkExtent2D surfExtnt = mvkVkExtent2DFromCGSize(mtlLayer.naturalDrawableSizeMVK);
 
-	// Metal supports 3 concurrent drawables, but if the swapchain is destroyed and
-	// rebuilt as part of resizing, one will be held by the current display image.
-	pSurfaceCapabilities->minImageCount = 2;
-	pSurfaceCapabilities->maxImageCount = 2;
+	// Metal supports 3 concurrent drawables.
+	pSurfaceCapabilities->minImageCount = 3;
+	pSurfaceCapabilities->maxImageCount = 3;
 
 	pSurfaceCapabilities->currentExtent = surfExtnt;
 	pSurfaceCapabilities->minImageExtent = surfExtnt;
--- a/MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
@@ -341,9 +341,8 @@ VkResult MVKPhysicalDevice::getSurfaceCapabilities(MVKSurface* surface,
 
     VkExtent2D surfExtnt = mvkVkExtent2DFromCGSize(mtlLayer.naturalDrawableSizeMVK);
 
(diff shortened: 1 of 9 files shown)


MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
```diff
@@ -341,10 +341,9 @@
 
     VkExtent2D surfExtnt = mvkVkExtent2DFromCGSize(mtlLayer.naturalDrawableSizeMVK);
 
-	// Metal supports 3 concurrent drawables, but if the swapchain is destroyed and
-	// rebuilt as part of resizing, one will be held by the current display image.
-	pSurfaceCapabilities->minImageCount = 2;
-	pSurfaceCapabilities->maxImageCount = 2;
+	// Metal supports 3 concurrent drawables.
+	pSurfaceCapabilities->minImageCount = 3;
```
cdavis5e: This would indicate to apps that they can't request double buffering.

billhollings: @cdavis5e Thanks for the sanity check.

Hmmm...yeah...I pondered that when I changed it.

But then I wondered how an app _would_ request double-buffering. From how I read the spec, `VkSwapchainCreateInfoKHR ::minImageCount` is just that...a _minimum_...which the implementation may be free to exceed if it works better for the implementation:

> `minImageCount` is the minimum number of presentable images that the application needs. The implementation will either create the swapchain with at least that many images, or it will fail to create the swapchain.

If the app passed 2 in for `minImageCount`...it seems to be the app saying to the implementation: "I won't take less than 2...but if you give me more...I'm cool with that"...rather than a more precise "just give me 2"...which would imply a more exact number...possibly named something like `imageCount`.  I assumed it was written that way because some implementations might be optimized for some larger number.

With that in mind...since the prominent use-case for Metal is triple-buffering...I took Metal to be one of those implementations that might just want to say "I know you can get by with 2...but 3 works better for me...so you're getting 3".

I'm happy to be convinced otherwise...and to drop `minImageCount` back to 2 if my assumptions are off-base.

KenThomases:
I know there are folks who explicitly want only double-buffering, not triple-buffering, in order to minimize latency. I think you will want some way for them to get that.

billhollings:
And per my thoughts above...in Vulkan...is it appropriate for that to be accomplished via `VkSwapchainCreateInfoKHR ::minImageCount`?

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
tokyovigilante: The [CAMetalLayer docs](https://developer.apple.com/documentation/quartzcore/cametallayer/2938720-maximumdrawablecount) suggest only 2 or 3 is possible for Metal. I guess if the user requests 2, then set `maximumDrawableCount` to 2, and to 3 for 3.  

`pSurfaceCapabilities->minImageCount` should definitely stay at 2 as that is a true representation of the underlying Metal capability. There are definitely circumstances (latency-critical apps) where 2 (ie double-buffering) would be preferred. 

Technically according to the spec, maximumDrawableCount could be 3 if the user requests mimimum 2 as you suggest, but that's unlikely to be what a user would expect.

mlfarrell: Does this change forget to also set maxImageCount?  Would that be left undefined?

Issue #431, opened 2018-12-28
Add build support for arm64e devices


DiligentGraphics:
When linking the project against libMoltenVK.a in iOS build, none of Vulkan symbols get resolved. I am using pre-built binaries from 1.1.92.1 SDK. I tried all static, dynamic and framework options, but result is the same. For macOS, everything works fine. I am building for arm64 in XCode 9.4.1.

DiligentGraphics:
I looked at MoltenVKPackaging XCode project, and it looks like the package is only built for i386 and x86_64 architectures, which would explain unresolved symbols. The question is then, what's the point in supplying pre-built binary that does not contain arm64 architecture?

billhollings:
Within MoltenVK...the `MoltenVKPackaging` _Xcode_ project does indeed build the iOS `arm64` versions too. They end up located in `Package/Latest/MoltenVK/iOS`.

Within the LunarG SDK...these same libraries are located in the `MoltenVK/iOS` directory.

You can verify the iOS functionality by running the iOS demos in MoltenVK...which use those built libraries (static, dynamic, and framework). I can also run the same MoltenVK demos by linking to the libraries in `MoltenVK/iOS` directory in the SDK, instead.

I'm not clear from your description what the context of your issues is...but perhaps their source is that the LunarG SDK itself (and the Vulkan layers it provides) does not officially support iOS development yet.

If so...for the time being...you may need to forego the LunarG SDK layers and link directly to the MoltenVK "driver" libraries in the `MoltenVK/iOS` directory in the SDK itself...or the `Package/Latest/MoltenVK/iOS` directory in MoltenVK if you build directly from MoltenVK using the `MoltenVKPackaging` _Xcode_ project.

DiligentGraphics:
My issue is not related to LunarG SDK layers because in fact I do not use them. I am trying to link my source code against pre-built MotlenVK libraries that I took from LunarG SDK. It works perfectly fine in MacOS build, but when I try to build the same source for iOS, none of Vulkan functions get resolved. 

I presumed that pre-built libraries are built with the MoltenVKPackaging xcode project, so I looked at the configuration for MoltenVK-iOS project and under Architectures found something that looks very confusing to me:

* Architectures: Standard Architectures (64-bit Intel)
* Base SDK: no SDK (macOS)
* Supported platforms: macOS
* Valid architectures: i386 x86_64

[...]

PR #437, opened 2018-12-31
Support iOS builds for arm64e architecture.

billhollings:
Fixes issue #431.

### DIFF
--- a/Docs/Whats_New.md
+++ b/Docs/Whats_New.md
@@ -28,6 +28,7 @@ Released 2018/12/31
 - MVKPipeline: Forbid vertex attribute offsets >= stride.
 - Fix handling of case where vertex bindings and binding indices don't match up.
 - Return VK_TIMEOUT even on zero wait if fences not signalled.
+- Support iOS builds for arm64e architecture.
 - Improvements to building external libraries.
 - Print Vulkan semantics when logging converted GLSL.
 
--- a/Docs/Whats_New.md
+++ b/Docs/Whats_New.md
@@ -20,6 +20,7 @@ Released 2018/12/31
 
 - Allow 2 or 3 swapchain images to support both double and triple buffering.
 - Force display to switch to GPU selected by vkCreateDevice() to avoid system view compositor having to copy from that GPU to display GPU.
+- Add support for uploading 3D S3TC-compressed images.
 - Use inline buffer for pipeline auxiliary buffer.
 - vkCmdCopyImage: Cast source image to the destination format.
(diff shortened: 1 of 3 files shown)


DiligentGraphics:
@billhollings Hmm... I still see the same link errors. It builds fine if I only have arm64 in Architectures. However when I replace it with arm64e, it still fails to link...

billhollings:
@DiligentGraphics 

Okay. Thanks for the feedback.

> I also checked the binary size: it changed from 10.3 Mb to 10.4 Mb (not sure if it is related to this change). Does that sound reasonable?

Yeah...I noticed almost no size change myself. I was expecting a more typical fat binary...something along the lines of twice the original size. I'm not sure if that is because both `arm64` architectures can share a lot of compiled code...or because it's actually not including `arm64e` for some reason.

Unfortunately I am not able to test `arm64e` on a device directly myself yet. As it sits now...the MoltenVK projects **_should_** be building for both architectures. Let's see what additional results arise from further testing as these devices and apps built for them become more common.

DiligentGraphics:
I tried to remove arm64 architecture completely and build arm64e only by editing xcode project files (I still do not know where these settings are hidden in the IDE). So when I removed arm64, the project failed to build.
There was also this warning:

```
None of the architectures in ARCHS (arm64) are valid for this target. Consider setting ARCHS to $(ARCHS_STANDARD) or updating it to include at least one value from VALID_ARCHS (arm64e). (in target 'MoltenVK-iOS')
```

Shouldn't you also update Architectures? As far as I understand, the intersection of Architectures and Valid Architectures get built.

sanvital:
> 
> > I also checked the binary size: it changed from 10.3 Mb to 10.4 Mb (not sure if it is related to this change). Does that sound reasonable?
> 
> Yeah...I noticed almost no size change myself. I was expecting a more typical fat binary...something along the lines of twice the original size. I'm not sure if that is because both `arm64` architectures can share a lot of compiled code...or because it's actually not including `arm64e` for some reason.
> 

This is an old thread (and it's basically dead given the merge), but I felt inspired to share my theory for the suspiciously small change on the size of the binary. I believe this observation is due to the fact that the arm64e slice refuses to support bitcode. If you `lipo -detailed_info` a universal executable you will note that the mac slices (i.e. simulator) are similarly small, as they also eschew bitcode.

I found this thread on the Apple developer forums, and added my two cents: [How to create arm64e slice with bitcode enabled?](https://forums.developer.apple.com/thread/125160). The answer, as far as I know, you cannot.

PR #442, opened 2019-01-04
Fix for multiple image copy regions + Remove arm64e support for now.

billhollings:
1. As described in Issue #431...support for `arm64e` architecture in Xcode is only Preview technology...and causes conflicts with the use of Bitcode. This PR removes support for `arm64e` until it is better handled by Xcode.

2. Fix Metal buffer offset with multiple copy regions in MVKCmdBufferImageCopy::encode().

@cdavis5e This second fix is to correct an issue where the Metal buffer offset was not being determined correctly when iterating multiple copy regions. The issue was introduced in PR #435...and so I needed to modify your code from that PR. I have no way of testing 3D compressed textures. Can you review and retest to ensure I have not now broken any of your changes from PR #435, please?

### DIFF
--- a/ExternalDependencies.xcodeproj/project.pbxproj
+++ b/ExternalDependencies.xcodeproj/project.pbxproj
@@ -3171,6 +3171,7 @@
 		A90FD89D21CC4EAB00B92BB2 /* Debug */ = {
 			isa = XCBuildConfiguration;
 			buildSettings = {
+				ARCHS = arm64;
 				BITCODE_GENERATION_MODE = bitcode;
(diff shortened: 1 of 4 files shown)


cdavis5e:
> @cdavis5e This second fix is to correct an issue where the Metal buffer offset was not being determined correctly when iterating multiple copy regions. The issue was introduced in PR #435...and so I needed to modify your code from that PR. I have no way of testing 3D compressed textures. Can you review and retest to ensure I have not now broken any of your changes from PR #435, please?

Looks good on my end. And I should really contribute the test I wrote for this back to the CTS.

PR #452, opened 2019-01-15
VkInstance: For idTech 6, always enable full image view swizzles.

cdavis5e:
id Software helped promulgate the original `GL_ARB_texture_swizzle`
extension in the first place. It's a safe bet anything they make is
going to need it.

### DIFF
--- a/Docs/Whats_New.md
+++ b/Docs/Whats_New.md
@@ -20,6 +20,7 @@ Released 2018/01/15
 
 - Support runtime config via runtime environment variables
 - Add full ImageView swizzling to config, and disable it by default.
+- For idTech 6, always enable full image view swizzling.
 - Add GPU switching to config, and enable it by default.
 - Add queue family specialization to config, and disable it by default.
 - Enable synchronous queue submits as config default.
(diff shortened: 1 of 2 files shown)


billhollings:
Hmmm...while I might be able to see the practical aspect of this...I'm not convinced it is appropriate to break documented behaviour in a hidden manner like this.

With the latest current update...anyone needing to swizzle will see a bunch of blatant "you need to swizzle this" errors in the logs. They can then turn it on via build settings, env vars, or an API call.

But someone using an `idTech` engine who _doesn't_ need to swizzle will just see their app run slower than it should for no obvious or documented reason.

danginsburg:
Just chiming in from my perspective: I'm guessing one reason this is necessary is that unlike something that runs through dxvk idTech has a native Vulkan renderer so there isn't a place to intercede?  I think it's fair game to use the appInfo for stuff like this, even if ugly.  

If you want to make it feel less one-off maybe you could do something like AppProfiles in the AMD Vulkan driver: https://github.com/GPUOpen-Drivers/xgl/blob/0d7c5a69ba314bfabe2d5dbe3e5e4d1ea3228845/icd/settings/settings.cpp#L98

billhollings:
Thanks @danginsburg 

Interesting to see this being done in native drivers.

One of the reasons that I introduced environment variable support for MoltenVK config was to solve this same issue...to remove the need for any special MoltenVK-specific API code.

The idea is that a game launch script on macOS can configure MoltenVK by setting env vars like `MVK_CONFIG_FULL_IMAGE_VIEW_SWIZZLE`...while running a pure, unchanged, Vulkan game and engine from another platform.

There are always several optional solutions to a problem, I guess.

Given env var config support...do we still need to special-case config info under-the-covers in this way? Are there situations where env vars would not be sufficient? Is there a reason that other Vulkan drivers take the under-the-covers approach?

cdavis5e:
I wanted this so that running, e.g. DOOM (2016) or Wolfenstein II would work out of the box. There's no Mac version at this time, so you'd have to run them via Wine. In that case, we could do this in Wine, e.g. have winevulkan.dll set this if the app info indicates an idTech engine. But that would leave any native version of idTech 6 games in the lurch. But I guess then those could have a script which sets the environment variable, or come with a version of MoltenVK which has this enabled by default. Also, I suppose it isn't that hard to make a script which runs Wine with all the correct settings. This is already common anyway, e.g. to run in a different WINEPREFIX. I'll withdraw this PR.

PR #460, opened 2019-01-18
Missing functions for MVKVector

aerofly:
As requested, I added a few missing functions to MVKVector, so that almost all std::vector members functions are now available for MVKVector as well. So in theory you can replace the templated function 

`template<typename T>
static void updateSwizzle(T &constants, uint32_t index, uint32_t swizzle) {`

with

`static void updateSwizzle(MVKVector &constants, uint32_t index, uint32_t swizzle) {`

Just to repeat: Performance of the templated function is higher since accessing MVKVector member functions involves a virtual function call. But in this case, performance implications are basically negligible.

@billhollings Could you please also look at the changes in MVKImage.mm. If the size of _availabilitySignalers is small, e.g. less than 20 elements, then there is no point in using std::list. MVKVector (e.g.std::vector) is still faster that std::list when removing the top element, again keeping in mind _availabilitySignalers does not contain a lot of elements.

I thought _availabilitySignalers never exceeds 4 elements as we are dealing with swap chain images?

### DIFF
--- a/MoltenVK/MoltenVK/GPUObjects/MVKImage.h
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKImage.h
@@ -20,8 +20,8 @@
 
 #include "MVKResource.h"
 #include "MVKSync.h"
+#include "MVKVector.h"
 #include <mutex>
-#include <list>
 
 #import <IOSurface/IOSurfaceRef.h>
 
@@ -396,7 +396,7 @@ protected:
 	uint32_t _swapchainIndex;
 	id<CAMetalDrawable> _mtlDrawable;
 	std::mutex _availabilityLock;
-	std::list<MVKSwapchainSignaler> _availabilitySignalers;
+	MVKVectorInline<MVKSwapchainSignaler, 4> _availabilitySignalers;
 	MVKSwapchainSignaler _preSignaled;
 	MVKSwapchainImageAvailability _availability;
 };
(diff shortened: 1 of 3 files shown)


billhollings:
@aerofly 

Thanks for the updates. Good stuff.

I understand what you are saying about `_availabilitySignalers`...and we can leave it as you have laid it out if it offers better performance. I guess this comes from not having to allocate an internal list element-holder to the heap?

I've cleaned up a couple of the left-overs...like removing the template from `updateSwizzle()`...and removing your commented-out `_availabilitySignalers.pop_front()`.

Thanks again.

aerofly:
@billhollings Thank you for approving this pull request. Yes, you are correct, avoiding the pop_front indeed saves the extra heap allocation. Also the _availabilitySignalers.push_back(signaler) does the same, so we are saving those 2 heap allocations together.

People sometimes forget that its actually faster to remove an element in std::vector in the front using erase, compared to a std::list pop_front heap deallocation which might involve a mutex lock. I have done some measurements and on a modern Windows PC with the usage case:
  container.pop_front();
  container.push_back( element );
Here, std::vector<int> with 1000 elements is roughly as fast as std::list<int>.

Of course, you can argue that this heap alllocation/deallocation is negligible, but it's the sum of all of them that makes the difference, that's what I tried to reduce with MVKVector.

Anyway, MoltenVK now has only very few heap allocations left per frame ( in our usage case < 10 ) which is very good now. If I have time I might look at those remaining in the future.

billhollings:
@aerofly

Amazing that a vector can shift up to 1000 elements as quickly as an element-holder-based linked list can! That is definitely an inherently counter-intuitive result...and I respect the fact that you took the time to investigate it.

Nice research! And thanks for contributing to making MoltenVK better!

PR #464, opened 2019-01-22
MVKRenderPass: Use combined store-resolve ops when requested and supported.

cdavis5e:
There is nothing in the Vulkan spec to suggest that
`VK_ATTACHMENT_STORE_OP_STORE` and a resolve attachment cannot both be
specified, nor that one or the other has no effect. So, if the app
supplies a color attachment with both `VK_ATTACHMENT_STORE_OP_STORE` and
a resolve attachment, that means it wants the output to be both stored
to the color attachment and resolved to the resolve attachment. In that
case, we should respect the application's wishes and do both. We can
only do that if the device supports this, though. All devices on Mac
starting with macOS 10.12 support this. On iOS, only the A9 and above
support combined store-resolve actions.

When this isn't supported, and the application asks for both, we should
probably schedule a second render pass which resolves the attachment,
as though `vkCmdResolveImage()` were called.


cdavis5e:
And I can see now that once #463 is merged, I'm going to have to bump the extension version *again*.

billhollings:
> And I can see now that once #463 is merged, I'm going to have to bump the extension version _again_.

Don't bother bumping the extension version so quickly. I'll only cut a release once all of these are in.

From a policy perspective...I think the HEAD can be considered WIP wrt version numbers.

billhollings:
And I'm going to try to resolve the conflict in `MVKPhysicalDeviceMetalFeatures` online here.

PR #478, opened 2019-01-28
Log to stderr instead of stdout.

cdavis5e:


Common/MVKLogging.h
```diff
@@ -183,7 +183,7 @@ static inline void MVKLogImplV(bool logToPrintf, bool logToASL, int aslLvl, cons
 	char lvlFmt[strlen(lvlStr) + strlen(format) + 5];
 	sprintf(lvlFmt, "[%s] %s\n", lvlStr, format);
 
-	if (logToPrintf) { vprintf(lvlFmt, args); }
+	if (logToPrintf) { vfprintf(stderr, lvlFmt, args); }
 //	if (logToASL) { asl_vlog(NULL, NULL, aslLvl, lvlFmt, args); }       // Multi-threaded ASL support requires a separate ASL client to be opened per thread!
```
billhollings: What's motivating the change? Most of the log messages are info level messages.

cdavis5e: When CrossOver (our product for running Windows programs on Mac and Linux) captures logs, it only captures `stderr`, on the assumption that anything of interest was written there. Wine writes all log messages, even informational ones (`TRACE`/`WARN`), to `stderr`. So does DXVK.

This is also more consistent with Unix tradition: the actual *output* of the program is written to `stdout`, while messages for the user, even informational ones, are usually written to `stderr`.

Issue #476, opened 2019-01-28
vkSetMTLTextureMVK doesn't call `retain` on the MTLTexture


mark-grimes:
N.B. this could be intended behaviour.

**Short version:**
When calling `vkSetMTLTextureMVK`, the `MVKImage::setMTLTexture` method releases the old `MTLTexture` but never retains the new one.

**Long version:**
I have a use case where I continually change the `MTLTexture` for a `VkImage` - basically for processing video frames retrieved from AVFoundation and feeding into a Vulkan compute shader.  To do this I:
1. Get a `CVImageBufferRef` for the frame from AVFoundation
2. Create a `CVMetalTextureRef` from this with `CVMetalTextureCacheCreateTextureFromImage`
3. Get a `id<MTLTexture>` from this with `CVMetalTextureGetTexture`
4. Apply the `MTLTexture` to the `VkImage` with `vkSetMTLTextureMVK`

Everything works except for unrestrained memory growth - I need to `CVBufferRelease` the `CVMetalTextureRef` (*big* memory growth, since this is per video frame).  If I do this then it crashes because the `CVMetalTextureRef` in turn releases the `MTLTexture` while Vulkan/MoltenVK is using it.  I've worked around this by calling `retain` on the `MTLTexture` before calling `vkSetMTLTextureMVK` - it all works and memory use is as expected (note no `release` on the next frame).  This is a bit of hassle because it's an ARC project, but it's possible, and I don't see this documented anywhere. So my code is now (per video frame):
1. Get a `CVImageBufferRef` for the frame from AVFoundation
2. Create a `CVMetalTextureRef` from this with `CVMetalTextureCacheCreateTextureFromImage`
3. Get a `id<MTLTexture>` from this with `CVMetalTextureGetTexture`
4. **Call `retain` on the `MTLTexture` through some work arounds for an ARC project**
5. Apply the `MTLTexture` to the `VkImage` with `vkSetMTLTextureMVK`
6. **Call `CVBufferRelease` with the `CVMetalTextureRef`**

Following through `vkSetMTLTextureMVK ` I find it releases the old texture as I would expect ([MVKImage.mm#L317](https://github.com/KhronosGroup/MoltenVK/blob/e5a890775bf6136d54b46c5b30670321b263c938/MoltenVK/MoltenVK/GPUObjects/MVKImage.mm#L317)), but it never calls retain on the new texture.

[...]

PR #480, opened 2019-01-28
 vkSetMTLTextureMVK() function retains texture object.

billhollings:
Fixes issue #476.


maddanio:
do you have any idea when this will land in the official lunarg release? because i just built agains 1.1.101.0 and it seems to still have this problem...

billhollings:
This should definitely be in the 1.1.101 SDK version.

What symptoms are you seeing? Does it behave differently if you use MoltenVK without the SDK?

PR #484, opened 2019-01-31
Support the VK_EXT_memory_budget extension.

cdavis5e:
This requires macOS 10.13 or iOS 11, for the `currentAllocatedSize`
property of `MTLDevice`. Ideally, we'd check for that method instead of
keying on the version.


billhollings:
Nice. I like the idea of turning off entire extensions based on OS version compatibility.

But it's highlighting that the negative macro logic is starting to look backwards. Next person that modifies `mvkIsSupportedOnPlatform()`...let's swap the negative macro logic to positive logic (eg. change `!MVK_IOS` to `MVK_MACOS` and vice-versa).

PR #489, opened 2019-02-05
Only warn on non-default swizzles when fullImageViewSwizzle is off.

cdavis5e:
Otherwise, the log will be spammed with hundreds of useless messages
that the `IDENTITY` swizzle is not supported.

### DIFF
--- a/MoltenVK/MoltenVK/Commands/MVKCommandEncoderState.mm
+++ b/MoltenVK/MoltenVK/Commands/MVKCommandEncoderState.mm
@@ -466,6 +466,11 @@ void MVKGraphicsResourcesCommandEncoderState::markDirty() {
 
 void MVKGraphicsResourcesCommandEncoderState::encodeImpl() {
 
+    bool fullImageViewSwizzle = false;
+    MVKPipeline* pipeline = _cmdEncoder->_graphicsPipelineState.getPipeline();
+    if (pipeline)
+        fullImageViewSwizzle = pipeline->fullImageViewSwizzle();
+
     encodeBinding<MVKMTLBufferBinding>(_vertexBufferBindings, _areVertexBufferBindingsDirty,
                                        [](MVKCommandEncoder* cmdEncoder, MVKMTLBufferBinding& b)->void {
                                            [cmdEncoder->_mtlRenderEncoder setVertexBuffer: b.mtlBuffer
@@ -492,7 +497,7 @@ void MVKGraphicsResourcesCommandEncoderState::encodeImpl() {
                                     _vertexAuxBufferBinding.index);
 
 	} else {
-		assertMissingSwizzles(_needsVertexSwizzle, "vertex", _vertexTextureBindings);
+		assertMissingSwizzles(_needsVertexSwizzle && !fullImageViewSwizzle, "vertex", _vertexTextureBindings);
 	}
 
     if (_fragmentAuxBufferBinding.isDirty) {
@@ -507,7 +512,7 @@ void MVKGraphicsResourcesCommandEncoderState::encodeImpl() {
                                       _fragmentAuxBufferBinding.index);
 
 	} else {
-		assertMissingSwizzles(_needsFragmentSwizzle, "fragment", _fragmentTextureBindings);
+		assertMissingSwizzles(_needsFragmentSwizzle && !fullImageViewSwizzle, "fragment", _fragmentTextureBindings);
     }
 
     encodeBinding<MVKMTLTextureBinding>(_vertexTextureBindings, _areVertexTextureBindingsDirty,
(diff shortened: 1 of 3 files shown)


cdavis5e:
And that's in release builds. In Debug builds, the assertion will trip.

billhollings:
@cdavis5e 

Under what conditions does this spamming happen? The default is to have `fullImageSwizzle` off now...and I am not seeing any spamming on any apps I test with.

Isn't the existing test in `MVKImageView::getSwizzledMTLPixelFormat()` sufficient to guard for the RGBA identity swizzle?

cdavis5e:
No, because it prints the message if the aux buffer binding is not marked dirty, but any image view has a non-default swizzle. It won't be marked dirty if the pipeline doesn't ask for it.

And for some reason, we're seeing this with `fullImageViewSwizzle`... ***on***?! That *definitely* shouldn't be happening. This is both with `MVK_CONFIG_FULL_IMAGE_VIEW_SWIZZLE` defined to 1 at build time, and at run time through the environment variable. Notably, we're not seeing the error that `MVKImageView` prints if you use a non-default swizzle

MoltenVK/MoltenVK/GPUObjects/MVKPipeline.h
```diff
@@ -99,13 +99,18 @@ class MVKPipeline : public MVKBaseDeviceObject {
 	/** Returns the current auxiliary buffer bindings. */
 	const MVKShaderAuxBufferBinding& getAuxBufferIndex() { return _auxBufferIndex; }
 
+	/** Returns whether or not full image view swizzling is enabled for this pipeline. */
+	bool fullImageViewSwizzle() const { return _fullImageViewSwizzle; }
```
billhollings: From what I can see...the value of this iVar and property never changes...and exists only to remember the state of `_device->_pMVKConfig->fullImageViewSwizzle` when the pipeline was created. Is that correct?

MoltenVK/MoltenVK/Commands/MVKCommandEncoderState.mm
```diff
@@ -464,6 +466,11 @@ static void assertMissingSwizzles(bool needsSwizzle, const char* stageName, MVKV
 
 void MVKGraphicsResourcesCommandEncoderState::encodeImpl() {
 
+    bool fullImageViewSwizzle = false;
+    MVKPipeline* pipeline = _cmdEncoder->_graphicsPipelineState.getPipeline();
+    if (pipeline)
+        fullImageViewSwizzle = pipeline->fullImageViewSwizzle();
```
billhollings: Following on my question in `MVKPipeline.h` below...is the purpose of this call and `fullImageViewSwizzle` local var to measure the state of `_device->_pMVKConfig->fullImageViewSwizzle` when the pipeline was created...and not when the draw/dispatch command is invoked? Is this because the status of that config value changes dynamically on some games you are testing?

MoltenVK/MoltenVK/GPUObjects/MVKPipeline.h
cdavis5e: That's correct.

MoltenVK/MoltenVK/Commands/MVKCommandEncoderState.mm
cdavis5e: No.

What I think is happening is:

1) A game creates some image views with non-default swizzles. These don't warn, because `fullImageViewSwizzle` is on.
1) The game creates a pipeline with `fullImageViewSwizzle` on, with a shader which uses storage image views, but no sampled image views.
1) The game sets these image views into a descriptor set, then binds the descriptor set with the pipeline.
1) When it's time to draw/dispatch, MoltenVK sees that some of the image views have non-default swizzles, but the pipeline didn't request an aux buffer, because the shader doesn't use sampled images. So it warns about *all* the image views, even the ones with default swizzles, even though the shader never actually samples from any of the image views.

billhollings: Ah! Okay. Thanks for the explanation.

The Vulkan spec includes:

> ...VkComponentMapping...This remapping must be identity for storage image descriptors,...

Do you interpret that to mean that only sampled images are to be swizzled? Or would we also need to apply swizzles in SPIRV-Cross to texture reads, etc?

If only sampled image views are to be swizzled...isn't the app barred from setting a non-identity swizzle on an image view that will be used without a sampler?

cdavis5e: I've heard back. The log messages were all eliminated. (Unfortunately, it didn't help with the fence timeout...) I can see that the converted shaders that do sampling have swizzle constants in them, so I know that `fullImageViewSwizzle` is definitely on.

I'm pretty sure that DXVK (for one) is being lazy. I know that it creates image views right off the bat for images it creates. It also has a tendency to create dummy images for unbound resources, and give them (0, 0, 0, 0) swizzles.

billhollings: Thanks for the update.

Issue #501, opened 2019-02-18
Is there any plan for Tessellation support in MoltenVK?


spencerkohan:
It is understandable that geometry shaders are not supported in MoltenVK as this feature is not implemented in Metal2, but Tessellation is available.  Is this on the roadmap for future versions?

cdavis5e:
I'm working on it right now. A change to support tessellation evaluation shaders was just accepted into [SPIRV-Cross](/KhronosGroup/SPIRV-Cross). Hopefully, I'll be able to get this in by next week's SDK release.

billhollings:
Duplicate of #56.

PR #508, opened 2019-02-21
Add support for tessellation.

### LINKED ISSUE #56: Request for tesselation support
just noticed MoltenVK has gone finally open source!
so just posting here request..
related to:
https://github.com/KhronosGroup/SPIRV-Cross/issues/120

cdavis5e:
At long last, tessellation comes to MoltenVK! With this change, clients
will now be able to specify tessellation shaders when creating
pipelines, and then draw tessellated patches with them.

Unfortunately, there seem to be a few gotchas with tessellation in
Metal. For one thing, tessellation pipelines in Metal are structured
very differently from Vulkan. There is no tessellation control or even
vertex stage. Instead, the tessellation evaluation shader takes the
place of the vertex function as a "post-tessellation vertex function."
The tessellation levels are supplied in a buffer to the tessellator,
which you are expected to populate. The most common way to do this is by
running a compute shader. MoltenVK thus runs the vertex shader and
tessellation control shader by themselves; a single `VkPipeline` object
then requires at least *three* `MTLPipelineState` objects.

But wait, there's more! The tessellation-control-as-compute stage uses
Metal's support for vertex-style stage input to a compute shader. But,
this support requires one to declare indexing *ahead of time*, when the
pipeline state is created. So a single `VkPipeline` object could have as
many as *five* `MTLPipelineState` objects.

Further, if there are more output than input control points for the
tessellation control stage, then later invocations may end up fetching
the wrong attributes! To get around this, this change uses index buffers
to ensure that all tessellation control shaders see the correct input.
Unfortunately, in the indexed draw case, this means that the incoming
index buffer needs to be munged.

Instancing is another pain point here. In Vulkan, as in OpenGL and
Direct3D, instancing is done in the vertex shader; but in Metal, it is
done at the tessellation evaluation stage. For this reason, only the
vertex stage of a tessellated draw supports instancing. Additional
memory is required to hold data for the extra vertices generated by
instancing. This also requires still more munging of index buffers for
indexed draws.

Indirect draws are even more painful. Because the number of vertices and
instances is unknown, storage for the maximum possible number of
vertices must be allocated. This change imposes a totally arbitrary
limit of 131072 vertices from a single draw, including all vertices
generated by instancing. On a Mac, this requires about 194-256 MB of
VRAM for all the temporary buffers.

There are some possible optimizations here. If we could prove that the
vertex shader's output doesn't depend on the instance ID, either
directly or through a per-instance attribute, then we could avoid
running the vertex and tess. control stages per instance, and take
advantage of Metal's support for tess. eval instancing. If we could
also prove that the vertex shader simply passes instance attributes
through (similarly with the tess. control shader), we could do this for
many more instanced draws as well. It should also be possible to cache
the output from the tess. control stage; if the draw comes up again, we
can then skip the vertex and tess. control stages entirely!

Fixes #56 and #501.


cdavis5e:
This change is incomplete. A bunch of CTS tests fail because shader I/O doesn't work right, and I haven't figured out why this is yet. Also, at lease one of the tests crashes.

billhollings:
@cdavis5e 

Thanks for taking this on. Congratulations! Nice work! 

Did you intend to leave this PR marked as _Draft_? It can't be merged into `master` while it is marked that way.

Are you wanting folks to evaluate, test, and provide feedback? Or are you waiting until more CTS tests pass?

cdavis5e:
> @cdavis5e
> 
> Thanks for taking this on. Congratulations! Nice work!
> 
> Did you intend to leave this PR marked as _Draft_? It can't be merged into `master` while it is marked that way.

Yes. Like I said, it's incomplete.

> 
> Are you wanting folks to evaluate, test, and provide feedback? Or are you waiting until more CTS tests pass?

I'd appreciate feedback on this. And maybe with more eyes, it'll be easier to figure out why the CTS tests fail.

billhollings:
For the Vulkan portability initiative...based on your descriptions above...or from the CTS results...it would be helpful if we could identify the Tessellation features that are inherently unsupportable in Metal...and expose them in the Vulkan portability extension [`VkPhysicalDevicePortabilitySubsetFeaturesEXTX`](https://github.com/KhronosGroup/Vulkan-Portability/blob/master/include/vulkan/vk_extx_portability_subset.h#L46) struct.

cdavis5e:
> For the Vulkan portability initiative...based on your descriptions above...or from the CTS results...it would be helpful if we could identify the Tessellation features that are inherently unsupportable in Metal...and expose them in the Vulkan portability extension [`VkPhysicalDevicePortabilitySubsetFeaturesEXTX`](https://github.com/KhronosGroup/Vulkan-Portability/blob/master/include/vulkan/vk_extx_portability_subset.h#L46) struct.

About the only one I can think of is isoline tessellation. It's not directly supported by Metal, and it's difficult to emulate. `PointsMode` might actually be another. I don't know yet if Metal will see that I set the `primitiveTopology` in the pipeline state to `Point` and render accordingly. Probably not.

ExternalRevisions/SPIRV-Cross_repo_revision
```diff
@@ -1 +1 @@
-d9ed3dcc7a7e62e5f95fd8f24e3d35e7e402ae92
+e37fbc6f72495e90597f6caddb4c9e309029df38
```
billhollings: `fetchDependencies` build generating numerous new warnings from SPIRV-Cross of this type:

```
External/SPIRV-Cross/spirv_msl.cpp:1898:43: warning:
      implicit conversion loses integer precision: 'std::__1::vector<unsigned int,
      std::__1::allocator<unsigned int> >::size_type' (aka 'unsigned long') to 'uint32_t'
      (aka 'unsigned int') [-Wshorten-64-to-32]
                uint32_t mbr_idx = ib_type.member_types.size();
                         ~~~~~~~   ~~~~~~~~~~~~~~~~~~~~~^~~~~~
```

cdavis5e: Just need to wait for KhronosGroup/SPIRV-Cross#870 to be merged...

billhollings:
> I'd appreciate feedback on this. And maybe with more eyes, it'll be easier to figure out why the CTS tests fail.

As a start...I blindly extracted all the `*.tessellation.*` CTS 1.1.4 tests and tried running them. 

Some passes. A number of fails (eg. `dEQP-VK.draw.shader_layer.tessellation_shader_1`) reporting:

    Vertex attribute gl_Position(1) is missing from the vertex descriptor.

And `dEQP-VK.tessellation.tesscoord.triangles_equal_spacing` crashes on `EXC_BAD_ACCESS`.

Do you want to lay down some direction as to which CTS testing is appropriate for this enhancement PR?

cdavis5e:
> > I'd appreciate feedback on this. And maybe with more eyes, it'll be easier to figure out why the CTS tests fail.
> 
> As a start...I blindly extracted all the `*.tessellation.*` CTS 1.1.4 tests and tried running them.
> 
> Some passes. A number of fails (eg. `dEQP-VK.draw.shader_layer.tessellation_shader_1`) reporting:
> 
> ```
> Vertex attribute gl_Position(1) is missing from the vertex descriptor.
> ```

Haven't tried running those yet.

> And `dEQP-VK.tessellation.tesscoord.triangles_equal_spacing` crashes on `EXC_BAD_ACCESS`.

On one of those testcases, Metal winds up generating far more vertices--about 1000 more--than the test was expecting. Try increasing the slop space to 2048 extra vertices on line 505 in vktTessellationCoordinatesTests.cpp.

> 
> Do you want to lay down some direction as to which CTS testing is appropriate for this enhancement PR?

So far, I've been running the `dEQP-VK.tessellation.*` tests. For me, the tests stop now when I get to  `dEQP-VK.tessellation.invariance.tess_coord_component_range.triangles_equal_spacing_ccw_point_mode` with:

```
Assertion 'numVertices <= maxNumVerticesInDrawCall' failed at ../external/vulkancts/modules/vulkan/tessellation/vktTessellationInvarianceTests.cpp:2104
```

billhollings:
Pulled in your latest...plus latest SPIRV-Cross.

Seeing a number of the following:

```
Test case 'dEQP-VK.tessellation.tesscoord.triangles_equal_spacing'..
2019-02-22 13:49:31.823096-0500 deqp-vk[13437:9836360] renderPipelineState inputPrimitiveTopology is MTLPrimitiveTopologyClassPoint but primitiveType is not MTLPrimitiveTypePoint
```

Is that related to the `PointMode` you mention above?

Also...a number of:

```
Test case 'dEQP-VK.tessellation.winding.default_domain.glsl_triangles_ccw'..
2019-02-22 13:49:32.418498-0500 deqp-vk[13437:9836360] renderPipelineState inputPrimitiveTopology is MTLPrimitiveTopologyClassTriangle but primitiveType is neither MTLPrimitiveTypeTriangle nor MTLPrimitiveTypeTriangleStrip
```

Then `dEQP-VK.tessellation.shader_input_output.patch_vertices_5_in_10_out` crashes with `SIGABRT`.

Do you want me to aid it debugging and fixing any of these...or is it best to leave you to it for now? You're probably running the same CTS tests that I am.

cdavis5e:
I'm not seeing any of this.

`SIGABRT` usually means that some `assert(3)` failed. Do you have any "Assertion failed" log lines you could show me that would shed light on this?

billhollings:
Strange that we are getting different CTS results. 

I'm running the 1.1.4 CTS suite on the `portability-dev` branch retrieved from Khronos Gerrit. I wouldn't expect that any of that should matter...as portability doesn't touch tessellation right now. But who knows?

What are your results from the 3 tests I mentioned above?

If you let me know which CTS suite and version you are running...I'll give it a go to confirm.

The `SIGABRT`is originating deep with Metal during the `kMVKGraphicsStageVertex` stage draw:

```
Test case 'dEQP-VK.tessellation.shader_input_output.patch_vertices_5_in_10_out'..
2019-02-23 13:35:58.667026-0500 deqp-vk[17922:10359860] renderPipelineState inputPrimitiveTopology is MTLPrimitiveTopologyClassTriangle but primitiveType is neither MTLPrimitiveTypeTriangle nor MTLPrimitiveTypeTriangleStrip
2019-02-23 13:35:58.667118-0500 deqp-vk[17922:10359860] Vertex Function(main0): Bytes are being bound at index 2 to a shader argument with write access enabled.
2019-02-23 13:35:58.678679-0500 deqp-vk[17922:10359860] *** Terminating app due to uncaught exception 'NSInvalidArgumentException', reason: '*** -[__NSSetM addObject:]: object cannot be nil'
*** First throw call stack:
(
	0   CoreFoundation                      0x00007fff42cd1ded __exceptionPreprocess + 256
	1   libobjc.A.dylib                     0x00007fff6ed9d720 objc_exception_throw + 48
	2   CoreFoundation                      0x00007fff42d113c0 -[CFPrefsConfigurationFileSource initWithConfigurationPropertyList:containingPreferences:] + 0
	3   CoreFoundation                      0x00007fff42c291ca -[__NSSetM addObject:] + 729
	4   MetalTools                          0x00007fff6217c0c4 +[MTLDebugResourceAccessTracker resourceTrackingRecordAccessesToArguments:bufferFuncArgsPtr:textureFuncArgsPtr:buffersBoundForWrite:texturesBoundForWrite:] + 775
	5   MetalTools                          0x00007fff621a85f9 -[MTLDebugCommandBuffer resourceTrackingRecordAccessesToArguments:bufferFuncArgsPtr:textureFuncArgsPtr:] + 101
	6   MetalTools                          0x00007fff621d2ff0 -[MTLDebugRenderCommandEncoder _resourceTrackingRecordBoundResourceAccesses] + 110
	7   MetalTools                          0x00007fff621d0c0d -[MTLDebugRenderCommandEncoder drawPrimitives:vertexStart:vertexCount:instanceCount:baseInstance:] + 271
	8   libMoltenVK.dylib                   0x0000000109885e7e _ZN10MVKCmdDraw6encodeEP17MVKCommandEncoder + 1246
```

billhollings:
BTW...just thinking about the upcoming SDK release...

In addition to CTS...are you running this PR against any real-world games that triggered this PR?

If so...and those tessellated games are running fine...and assuming this PR doesn't break anything existing...I'm good with including it in the upcoming release...and continuing to work on completing CTS conformance in the mainline evolution.

cdavis5e:
Hmm... It looks like you're running with validation enabled, and it doesn't like that I'm calling `setVertexBytes:...` on a writable buffer. I was afraid of this.

I'm running public `HEAD`.

billhollings:
Ah! Indeed. Validation on but set to log instead of assert on errors.

With validation off...`dEQP-VK.tessellation.*` runs to completion:

```
Test run totals:
  Passed:        61/432 (14.1%)
  Failed:        233/432 (53.9%)
  Not supported: 138/432 (31.9%)
  Warnings:      0/432 (0.0%)
```

cdavis5e:
I have a few more changes to push out, but they're dependent on some SPIRV-Cross changes that haven't been merged yet. With these, I have:

```
Test run totals:
  Passed:        148/432 (34.3%)
  Failed:        146/432 (33.8%)
  Not supported: 138/432 (31.9%)
  Warnings:      0/432 (0.0%)
```

The remaining failures:

* `*isoline*`: Obviously.
* `dEQP-VK.tessellation.shader_input_output.cross_invocation_per_vertex_mat4x3`: This happens because SPIRV-Cross doesn't translate a shader properly. The temp generated for the matrix input has the wrong name--or it's later using the wrong name to refer to it.
* `dEQP-VK.tessellation.shader_input_output.cross_invocation_per_patch_mat4x3`: This uses an array of matrices, which SPIRV-Cross doesn't support 😠.
* `dEQP-VK.tessellation.common_edge.*`: No idea.
* `dEQP-VK.tessellation.fractional_spacing.*`: Needs isolines.
* `dEQP-VK.tessellation.primitive_discard.*`: No idea.
* `dEQP-VK.tessellation.user_defined_io.*`: Mysterious failures to link the shader program over a missing `user(locn0)` varying, even though the tessellation evaluation shader writes to it... 😠Bug in Metal?
  * Some of these don't even make it to that point, since they use a custom block output, and SPIRV-Cross doesn't translate that correctly, either. I suspect this is related to the `cross_invocation_per_vertex_mat4x3` failure.

billhollings:
That's good news!

Once the SPIRV-Cross merge is done...and you add them and your other changes to this PR...are you ready to treat this PR as good-to-go for the upcoming SDK release? Or do you want to leave it in _Draft_ mode for a while longer until you can address the leftovers?

Regarding `*isoline*` & `dEQP-VK.tessellation.fractional_spacing.*`...looks like there are 90 of those. If this is something that is just not currently supported by Metal...there is the option of adding an `isolineTessellation` member to the [`VkPhysicalDevicePortabilitySubsetFeaturesEXTX`](https://github.com/KhronosGroup/Vulkan-Portability/blob/master/include/vulkan/vk_extx_portability_subset.h#L46) struct. With that in place I can modify the CTS `portability-dev` branch to treat those 90 tests as `Not supported` instead of `Failed`.

cdavis5e:
> * `dEQP-VK.tessellation.common_edge.*`: No idea.

Figured it out. I was setting the stride wrong.

> * `dEQP-VK.tessellation.user_defined_io.*`: Mysterious failures to link the shader program over a missing `user(locn0)` varying, even though the tessellation evaluation shader writes to it... 😠Bug in Metal?

Gah. I know why this doesn't work. The tessellation evaluation shader writes to a buffer. When you do this in Metal, you can't do rasterization, same as with a vertex shader. So the shader ends up *not* writing to the given varying, and the link fails. I'm not sure this can easily be made to work.

cdavis5e:
> That's good news!
> 
> Once the SPIRV-Cross merge is done...and you add them and your other changes to this PR...are you ready to treat this PR as good-to-go for the upcoming SDK release?

Fine by me. This seems to work OK now, more or less. I'm probably going to try this on a real game tomorrow.

> Regarding `*isoline*` & `dEQP-VK.tessellation.fractional_spacing.*`...looks like there are 90 of those. If this is something that is just not currently supported by Metal...there is the option of adding an `isolineTessellation` member to the [`VkPhysicalDevicePortabilitySubsetFeaturesEXTX`](https://github.com/KhronosGroup/Vulkan-Portability/blob/master/include/vulkan/vk_extx_portability_subset.h#L46) struct. With that in place I can modify the CTS `portability-dev` branch to treat those 90 tests as `Not supported` instead of `Failed`.

Sounds good. I guess we could bring this up at the next meeting.

billhollings:
>Figured it out.
>Gah. I know why this doesn't work.

Nice work!

>Fine by me. This seems to work OK now, more or less. I'm probably going to try this on a real game tomorrow.

Happy to hear that! Once you're happy with the CTS and real-game results...release this PR from _Draft_ status. At that point...I'll run some additional smoke tests to make sure it doesn't ambush existing functionality...and then I'll pull it in.

>Sounds good. I guess we could bring this up at the next meeting.

Yup...that would be good. So we have `isoline` at least. If there are others (perhaps the _"tessellation evaluation shader writes to a buffer"_ issue you mentioned?)...maybe you could raise them as possible feature flags too. When you get a chance...perhaps send an email to the portability mail list suggesting additional feature flags a few days before the next meeting (Mar 8) to prep the discussion.

HansKristian-Work:
Merged the latest trio of PRs in SPIRV-Cross.

billhollings:
Curiouser and curiouser.

Just spent quite a while beating my head against the proverbial wall running this against published Dota2.

`vkCreateDevice()` causes `_MVKStaticCmdShaderSource` to strangely fail to compile with a bunch of:

```
program_source:160:101: error: unknown type name 'MTLDrawPatchIndirectArguments'; did you mean 'MTLDrawPrimitivesIndirectArguments'?
    device char* dest = destBuff + idx * (sizeof(MTLDispatchThreadgroupsIndirectArguments) + sizeof(MTLDrawPatchIndirectArguments));
                                                                                                    ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
                                                                                                    MTLDrawPrimitivesIndirectArguments
```

It turns out that `MVKDevice::getMTLCompileOptions()` is unexpectedly returning a `MTLCompileOptions` instance whose `languageVersion` parameter is set to `MTLLanguageVersion1_1` instead of `MTLLanguageVersion2_1`.

This is probably related to a similar item you [speculated on in the DXVK issue](https://github.com/KhronosGroup/MoltenVK/issues/203#issuecomment-452868875).

So...a couple of thoughts:

1. `_MVKStaticCmdShaderSource` should adapt to MSL versions to conditionally elide MSL artifacts that are MSL version dependent.

2. `MVKDevice::getMTLCompileOptions()` should probably force the `languageVersion` to the latest MSL version available as determined by feature sets. We already track this in `MVKDevice` so it should be an easy change.

I've confirmed that (2) fixes the issue for Dota2...so I'll polish it up and do a separate PR for that. 

Can you take care of (1) as part of this PR...otherwise we might see legitimate crashes if people really run on macOS 10.11 / iOS 9.0?

MoltenVK/MoltenVK/Commands/MVKCommandPipelineStateFactoryShaderSource.h
```diff
@@ -168,5 +168,97 @@ kernel void cmdCopyBufferToImage3DDecompressTempBufferDXTn(constant uint8_t* src
...
+                                          constant uint32_t& srcStride [[buffer(2)]],                           \n\
+                                          constant uint32_t& inControlPointCount [[buffer(3)]],                 \n\
+                                          constant uint32_t& outControlPointCount [[buffer(4)]],                \n\
+                                          constant uint32_t& drawCount [[buffer(5)]],                           \n\
+                                          uint idx [[thread_position_in_grid]]) {                               \n\
+    if (idx >= drawCount) { return; }                                                                           \n\
+    const device auto& src = *reinterpret_cast<const device MTLDrawPrimitivesIndirectArguments*>(srcBuff + idx * srcStride);\n\
+    device char* dest = destBuff + idx * (sizeof(MTLDispatchThreadgroupsIndirectArguments) + sizeof(MTLDrawPatchIndirectArguments));\n\
+#if __METAL_VERSION__ >= 210                                                                                    \n\
```
billhollings: Curiouser and curiouser.

Just spent quite a while beating my head against the proverbial wall running this against published Dota2.

`vkCreateDevice()` causes `_MVKStaticCmdShaderSource` to strangely fail to compile with a bunch of:

```
program_source:160:101: error: unknown type name 'MTLDrawPatchIndirectArguments'; did you mean 'MTLDrawPrimitivesIndirectArguments'?
    device char* dest = destBuff + idx * (sizeof(MTLDispatchThreadgroupsIndirectArguments) + sizeof(MTLDrawPatchIndirectArguments));
                                                                                                    ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
                                                                                                    MTLDrawPrimitivesIndirectArguments
```

It turns out that `MVKDevice::getMTLCompileOptions()` is unexpectedly returning a `MTLCompileOptions` instance whose `languageVersion` parameter is set to `MTLLanguageVersion1_1` instead of `MTLLanguageVersion2_1`.

This is probably related to a similar item you [speculated on in the DXVK issue](https://github.com/KhronosGroup/MoltenVK/issues/203#issuecomment-452868875).

So...a couple of thoughts:

1. `_MVKStaticCmdShaderSource` should adapt to MSL versions to conditionally elide MSL artifacts that are MSL version dependent.

2. `MVKDevice::getMTLCompileOptions()` should probably force the `languageVersion` to the latest MSL version available as determined by feature sets. We already track this in `MVKDevice` so it should be an easy change.

I've confirmed that (2) fixes the issue for Dota2...so I'll polish it up and do a separate PR for that. 

Can you take care of (1) as part of this PR...otherwise we might see legitimate crashes if people really run on macOS 10.11 / iOS 9.0?

cdavis5e: I've added a preprocessor directive to only compile in the tessellation helpers on Metal 1.2+.

billhollings: Great! Thanks.

I'll make the `languageVersion` changes I suggested on another PR.

One other topic of discussion:

When I started looking into this issue...I started wondering about whether we should have a config option / env var that enables tessellation...with it either enabled or disabled by default.

I was initially speculating that Dota2 might now be responding to the `VkPhysicalDeviceFeatures::tessellationShader` feature...and then making use of tessellation functionality that you haven't been able to support yet (or at all...eg isolines).

What do you think? Might some apps detect tessellation and start using it in ways that are not supported yet? Should we provide the option to deselect `VkPhysicalDeviceFeatures::tessellationShader` via an env var? 

And if so...should it be on or off by default?

cdavis5e: > What do you think? Might some apps detect tessellation and start using it in ways that are not supported yet?

For those that use Vulkan natively (e.g. Dota2, DOOM), this is certainly a possibility. But for D3D11 games (via DXVK), we were already faking this feature to make them run at all, so this should help them if anything.

> Should we provide the option to deselect `VkPhysicalDeviceFeatures::tessellationShader` via an env var?

It might be good to have one for now while we work out any remaining bugs. This will enable users to keep running games even if tessellation is broken.

> And if so...should it be on or off by default?

I don't know. I'm leaning towards leaving it on by default, since I want to know if there are bugs that I should go fix.

billhollings:
@cdavis5e
 
> I'm investigating...but nothing obvious yet.

I've narrowed the memory issues down to the [first tessellation commit to SPIRV-Cross](url
https://github.com/KhronosGroup/SPIRV-Cross/commit/eb89c3a428b20627d0f5f02b5f40de2ad7df2a3e). The Dota2 crashes happen with non-tessellation MoltenVK with that SPRIV-Cross commit.

cdavis5e:
Thanks. Unfortunately, that's a huge change. I need to narrow that down some more.

I'm probably going to run the CTS against Address Sanitizer. Hopefully that will turn up something.

billhollings:
> I'm probably going to run the CTS against Address Sanitizer. Hopefully that will turn up something.

I tried running Address Sanitizer against the Cube and Hologram demos with this full PR without finding anything. Unfortunately when I tried doing that against Dota2 it encounters other address issues that trigger before this one.

As is often the case with address issues...the crashes don't occur immediately on the offending code. But as far as I can tell...when it does crash while in SPIRV-Cross...it occurs when several instances of `CompilerMSL` are running simultaneously on different threads. So it appears to be a threading issue during simultaneous shader conversions...which Dota2 uses aggressively.

But I haven't seen a case yet where it uses the same instance of `CompilerMSL`...and that would appear to be impossible anyway because `ComplerMSL` is instantiated dynamically for each conversion by a `SPIRVToMSLConverter`.

cdavis5e:
> Unfortunately when I tried doing that against Dota2 it encounters other address issues that trigger before this one.

Try running with `ASAN_OPTIONS=detect_container_overflow=0`. I get a lot of false positives from that check. (At least I *think* they're false positives... They happen just from the normal operation of `libc++`'s containers.)

cdavis5e:
I wonder if Thread Sanitizer will turn something up...

HansKristian-Work:
A wild shot in the dark here might be the ClassicLocale class in spirv_common.hpp that's used to force a classic C locale. I suspect it might not be thread safe. Never had any issues with it, but worth a shot to remove it.

billhollings:
> A wild shot in the dark here might be the ClassicLocale class in spirv_common.hpp that's used to force a classic C locale. I suspect it might not be thread safe. Never had any issues with it, but worth a shot to remove it. However, it would be very strange if this is only an issue now all of a sudden ...

A good suggestion. But unfortunately removing `ClassicLocale` does not fix the issue.

However...what threading impact does `ClassicLocale` have? It seems to be setting and resetting some kind of global configuration during each run. If so...that does sound potentially dangerous for parallel execution!

HansKristian-Work:
That's a good question, when I added it I believe this is thread local, but that might not actually be true. I'll need to investigate. The reason for it is that we need C locale when using sprintf because some locales use comma instead of point for decimals. I'll investigate if there's a good stateless alternative we can use, although I guess it won't really affect this issue in particular.

cdavis5e:
Thread Sanitizer didn't really turn up anything, either.

cdavis5e:
I did however fix a problem I found with non-tessellated indirect draws. Given that you found that Dota2 crashed with just the tessellation control patch to SPIRV-Cross, I doubt it'll help, but maybe it will?

MoltenVK/MoltenVK/Commands/MVKCommandBuffer.mm
```diff
@@ -452,10 +452,11 @@
 
 const MVKMTLBufferAllocation* MVKCommandEncoder::getTempMTLBuffer(NSUInteger length) {
     const MVKMTLBufferAllocation* mtlBuffAlloc = getCommandEncodingPool()->acquireMTLBufferAllocation(length);
+	MVKMTLBufferAllocationPool* pool = mtlBuffAlloc->getPool();
 
     // Return the MTLBuffer allocation to the pool once the command buffer is done with it
     [_mtlCmdBuffer addCompletedHandler: ^(id<MTLCommandBuffer> mcb) {
-        ((MVKMTLBufferAllocation*)mtlBuffAlloc)->returnToPool();
+        pool->returnObjectSafely((MVKMTLBufferAllocation*)mtlBuffAlloc);
```
billhollings: Strange. I wonder why this matters to Thread Santizer? 

Is it thinking `_pool` could somehow change before the completion handler fires? Sigh.

cdavis5e: Pretty much. It doesn't know that the completion handler can't fire at all until the command buffer is committed.

cdavis5e: Alternatively, it thinks that the completion handler could fire *before* the object is finished being created. Yeah, I know, that should *never* happen. But its idea of "race condition" is "two or more accesses by different threads in close succession, unserialized by a mutex."

billhollings:
@cdavis5e 

Well...I haven't been able to figure this Dota2 memory issue out yet.

I have measured that SPIRV-Cross with the tessellation enhancements seems to run quite a bit slower than previously. Does that sound reasonable?

I'm wondering if what is happening is that the slower performance is affecting the threading in Dota2...to the point where it is highlighting memory/threading issues in the game that are not actually originating in SPIRV-Cross.

Have you run this PR against any other significant games? What are the results?

cdavis5e:
> @cdavis5e
> 
> Well...I haven't been able to figure this Dota2 memory issue out yet.
> 
> I have measured that SPIRV-Cross with the tessellation enhancements seems to run quite a bit slower than previously. Does that sound reasonable?

Quite a bit, you say? That's a problem. I was expecting a slight performance hit, but not that much. Shader compilation is already a massive time sink, even with native Vulkan. I need to do some profiling and see if I can make this go faster.

> 
> I'm wondering if what is happening is that the slower performance is affecting the threading in Dota2...to the point where it is highlighting memory/threading issues in the game that are not actually originating in SPIRV-Cross.
> 
> Have you run this PR against any other significant games? What are the results?

Not yet. I'm trying to run this against Final Fantasy XIV, which is what prompted this in the first place. I've also installed Dota2 from Steam; maybe I'll be able to reproduce the problem locally.

HansKristian-Work:
Btw, there is an --iterations argument in SPIRV-Cross CLI which can be used for some crude profiling. Might be worth a look.

billhollings:
@cdavis5e 
>Quite a bit, you say? That's a problem. I was expecting a slight performance hit, but not that much. Shader compilation is already a massive time sink, even with native Vulkan. I need to do some profiling and see if I can make this go faster.

"Quite a bit" is relative of course. It fluctuates...but seems to take maybe 25%-30% longer. 

With a **Debug** build of MoltenVK, you can use the `MVK_CONFIG_PERFORMANCE_TRACKING` env var at runtime to have MoltenVK log various shader handling performance. Look for lines like this:

    [mvk-debug] Performance to convert SPIR-V to MSL source code curr: 0.506 ms, avg: 7.290 ms, min: 0.506 ms, max: 107.898 ms, count: 252

    [mvk-debug] Performance to convert SPIR-V to MSL source code curr: 8.031 ms, avg: 5.807 ms, min: 1.114 ms, max: 64.846 ms, count: 252

The first is from a run that includes SPIRV-Cross tessellation and which crashes, the second is without SPIRV-Cross tessellation, and does not crash. I'm generally comparing the `avg`...but the `max` is also interesting...but might be a fluke.

@HansKristian-Work 
> Btw, there is an --iterations argument in SPIRV-Cross CLI which can be used for some crude profiling. Might be worth a look.

Thanks!

billhollings:
> Not yet. I'm trying to run this against Final Fantasy XIV, which is what prompted this in the first place. I've also installed Dota2 from Steam; maybe I'll be able to reproduce the problem locally.

We'll need to get to the bottom of this soon. I'm not comfortable pulling in [SPIRV-Cross eb89c3a](https://github.com/KhronosGroup/SPIRV-Cross/commit/eb89c3a428b20627d0f5f02b5f40de2ad7df2a3e) or later until we understand this...which inhibits our ability to include any SPIRV-Cross enhancements.

Essentially we need to figure out if this is a SPIRV-Cross issue with any sufficiently-complex app...or is specific to Dota2.

An alternate option might be to create a branch in SPIRV-Cross that excludes (or includes) the tessellation commits until we figure it out. But I that would be my last choice.

HansKristian-Work:
I added support for ubsan/asan/msan/tsan on my c-api branch, and the test suite runs clean on Linux at least.

HansKristian-Work:
I was able to reproduce the crash with commit: aab4a5632b4953d1ce1d20bd2bc0ed5748fe2d07.

billhollings:
@HansKristian-Work 

> I was able to reproduce the crash with commit: aab4a5632b4953d1ce1d20bd2bc0ed5748fe2d07. As mentioned earlier, it's crashing in a completely weird and unrelated place.

What testing were you doing that caused this crash?

HansKristian-Work:
I started the game totally fresh, then entered the 1. tutorial called "Mechanics". During the loading screen, the error would occur.

billhollings:
> I started the game totally fresh, then entered the 1. tutorial called "Mechanics". During the loading screen, the error would occur.

billhollings:
Sigh. I hit the close button by mistake with a literal finger twitch as I was moving my mouse. Reopened now. Sorry. 🙄

HansKristian-Work:
There seems to have been an update today on DOTA, but it didn't change much, except the shaders seem different. I trigger far fewer warnings.

Been looking into this some more today. This is a serious ghost bug. Some observations.

- If I preload the asan library, I cannot reproduce the issue. I cannot find any issues when MoltenVK and the deps are built with asan/ubsan. I get one error in DOTA on boot though, but it seems like a false positive.
- With tsan, MoltenVK has a lot of weird errors. I think there were some fixes for that in this PR with some false positive workarounds. Either way, I couldn't see any issues in SPIRV-Cross.
- On a crash, no threads are active in SPIRV-Cross. I'm testing this on a dual-core HT macbook pro though.
- To debug with lldb you have to do: `GAME_DEBUGGER=/Applications/Xcode.app/Contents/Developer/usr/bin/lldb bash "$(pwd)/dota.sh"`. Just lldb will block DYLD_LIBRARY_PATH or something silly ...

This is a common backtrace, and I see it almost every time:
```
frame #0: 0x00007fff74277969 libsystem_platform.dylib`_platform_bzero$VARIANT$Haswell + 41
    frame #1: 0x0000000087f0fd61 librendersystemvulkan.dylib`___lldb_unnamed_symbol1018$$librendersystemvulkan.dylib + 353
    frame #2: 0x0000000087f0fbb3 librendersystemvulkan.dylib`___lldb_unnamed_symbol1017$$librendersystemvulkan.dylib + 1395
    frame #3: 0x0000000087f33032 librendersystemvulkan.dylib`___lldb_unnamed_symbol1255$$librendersystemvulkan.dylib + 162
    frame #4: 0x000000009164b34e libscenesystem.dylib`___lldb_unnamed_symbol2365$$libscenesystem.dylib + 286
    frame #5: 0x000000009158e729 libscenesystem.dylib`___lldb_unnamed_symbol1231$$libscenesystem.dylib + 9561
    frame #6: 0x000000009156a4dc libscenesystem.dylib`___lldb_unnamed_symbol1027$$libscenesystem.dylib + 476
```
If I don't get a crash, I get malloc assertions about broken checksums, etc.

billhollings:
> Been looking into this some more today. This is a serious ghost bug. Some observations.

Thanks for keeping on this.

> * If I preload the asan library, I cannot reproduce the issue. I cannot find any issues when MoltenVK and the deps are built with asan/ubsan. I get one error in DOTA on boot though, but it seems like a false positive.
> * With tsan, MoltenVK has a lot of weird errors. I think there were some fixes for that in this PR with some false positive workarounds. Either way, I couldn't see any issues in SPIRV-Cross.

How are you loading as `asan/tsan`? I was finding that difficult. I too hit the false positive...but it wouldn't go any further. Is there a way of having `asan` log but continue on errors?

> * On a crash, no threads are active in SPIRV-Cross. I'm testing this on a dual-core HT macbook pro though.

Historically (I haven't tried with the Dota2 update you mention) I sometimes I'll see it crash either in the middle of SPIRV-Cross...or while MoltenVK is logging the shader conversion. And even in the cases where the crash happens with no SPIRV-Cross active...almost every time, the last thing that appears in the log is a shader conversion.

So it may not be happening within SPRIV-Cross...but there may be some conflict between memory that SPRIV-Cross and Dota2 are using.

> * To debug with lldb you have to do: `GAME_DEBUGGER=/Applications/Xcode.app/Contents/Developer/usr/bin/lldb bash "$(pwd)/dota.sh"`. Just lldb will block DYLD_LIBRARY_PATH or something silly ...
> 
> This is a common backtrace, and I see it almost every time:
> 
> ```
> frame #0: 0x00007fff74277969 libsystem_platform.dylib`_platform_bzero$VARIANT$Haswell + 41
>     frame #1: 0x0000000087f0fd61 librendersystemvulkan.dylib`___lldb_unnamed_symbol1018$$librendersystemvulkan.dylib + 353
>     frame #2: 0x0000000087f0fbb3 librendersystemvulkan.dylib`___lldb_unnamed_symbol1017$$librendersystemvulkan.dylib + 1395
>     frame #3: 0x0000000087f33032 librendersystemvulkan.dylib`___lldb_unnamed_symbol1255$$librendersystemvulkan.dylib + 162
>     frame #4: 0x000000009164b34e libscenesystem.dylib`___lldb_unnamed_symbol2365$$libscenesystem.dylib + 286
>     frame #5: 0x000000009158e729 libscenesystem.dylib`___lldb_unnamed_symbol1231$$libscenesystem.dylib + 9561
>     frame #6: 0x000000009156a4dc libscenesystem.dylib`___lldb_unnamed_symbol1027$$libscenesystem.dylib + 476
> ```
> If I don't get a crash like this, I get malloc assertions about broken checksums, etc, which then crash.

@cdavis5e Have you tried running this PR against other complex games? You mentioned FF, for example.

@danginsburg Are you able to give us a hand looking into this? I'm hoping that you can run this with additional debugging and inspection capabilities within Data2. To summarize...this PR (actually even just [SPIRV-Cross eb89c3a](https://github.com/KhronosGroup/SPIRV-Cross/commit/eb89c3a428b20627d0f5f02b5f40de2ad7df2a3e)) is encountering random memory conflicts with Dota2. It might bet thread related. It doesn't seem to be happening elsewhere.

HansKristian-Work:
@billhollings To disable halt on error, you can pass in:
```
ASAN_OPTIONS=halt_on_error=0 STEAM_DYLD_INSERT_LIBRARIES=/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/lib/clang/10.0.0/lib/darwin/libclang_rt.asan_osx_dynamic.dylib bash "$(pwd)/dota.sh
```
Yeah, I'm not confident I can find this issue. I've reviewed the PR again, and I can't see anything risky with memory, and given I get no asan/msan errors there isn't much to go on.

cdavis5e:
All right, I've tested this with FFXIV, and I'm actually able to see the water now on high tessellation! It's kinda slow, though. I haven't seen any crashes so far; I guess DXVK is a bit less aggressive about compiling shaders on other threads.

There is one other problem though, that predates this change. Our testers report that FFXIV has a tendency to crash after about an hour with no explanation or error message. I'm wondering if this is somehow related to the crashes you saw in Dota2--i.e. this is a preexisting problem, and the change to SPIRV-Cross just exposed it. I need to try playing for longer and seeing if I can replicate the crashes that our testers found. Maybe it'll happen sooner with these changes in place.

cdavis5e:
Well, that crash has nothing to do with the crash in Dota2.

There's got to be a reason for this. I don't want to destabilize Dota2 just so I can use tessellation elsewhere. I wonder if we could get Valve to test a debug version of Dota2 with these changes to MoltenVK.

danginsburg:
@cdavis5e I've been looking at it today.  One thing I pointed out to Bill is that this error:

```[***MoltenVK ERROR***]: VK_ERROR_INITIALIZATION_FAILED: Under Metal, vertex attribute offsets must not exceed the vertex buffer stride.```

is causing pipeline creation to fail for some of our pipelines, which makes the water not render properly in dota.  I wonder if the crash is downstream of this in our handling of pipeline creation failures.  The case in which we're hitting this looks like this:

```(lldb) parray 3 vertexInputCreateInfo.pVertexBindingDescriptions
(const VkVertexInputBindingDescription *) $7 = 0x00007000088a6110 {
  (VkVertexInputBindingDescription) [0] = (binding = 0, stride = 24, inputRate = VK_VERTEX_INPUT_RATE_VERTEX)
  (VkVertexInputBindingDescription) [1] = (binding = 1, stride = 4, inputRate = VK_VERTEX_INPUT_RATE_INSTANCE)
  (VkVertexInputBindingDescription) [2] = (binding = 2, stride = 0, inputRate = VK_VERTEX_INPUT_RATE_INSTANCE)
}
(lldb) parray 4 vertexInputCreateInfo.pVertexAttributeDescriptions
(const VkVertexInputAttributeDescription *) $8 = 0x00007000088a61d0 {
  (VkVertexInputAttributeDescription) [0] = (location = 0, binding = 0, format = VK_FORMAT_R32G32B32_SFLOAT, offset = 0)
  (VkVertexInputAttributeDescription) [1] = (location = 1, binding = 2, format = VK_FORMAT_R32G32B32A32_SFLOAT, offset = 0)
  (VkVertexInputAttributeDescription) [2] = (location = 2, binding = 2, format = VK_FORMAT_R32G32B32A32_SFLOAT, offset = 0)
  (VkVertexInputAttributeDescription) [3] = (location = 3, binding = 1, format = VK_FORMAT_R16G16_SFLOAT, offset = 0)
```

Is this really a case MoltenVK can't handle?  It's worked on Metal for a long time.  Not sure I understand the MoltenVK change here.

I asked Bill to find out if it still crashes if he avoids that check.  

I've run Dota 2 in debug and haven't been able to get it to crash.  The ASAN crash is something I'm looking into, but it looks unrelated to this and is a buffer read overrun so probably shouldn't crash.

cdavis5e:
> @cdavis5e I've been looking at it today. One thing I pointed out to Bill is that this error:
> 
> `[***MoltenVK ERROR***]: VK_ERROR_INITIALIZATION_FAILED: Under Metal, vertex attribute offsets must not exceed the vertex buffer stride.`
> 
> is causing pipeline creation to fail for some of our pipelines, which makes the water not render properly in dota. I wonder if the crash is downstream of this in our handling of pipeline creation failures. The case in which we're hitting this looks like this:
> 
> ```
> (const VkVertexInputBindingDescription *) $7 = 0x00007000088a6110 {
>   (VkVertexInputBindingDescription) [0] = (binding = 0, stride = 24, inputRate = VK_VERTEX_INPUT_RATE_VERTEX)
>   (VkVertexInputBindingDescription) [1] = (binding = 1, stride = 4, inputRate = VK_VERTEX_INPUT_RATE_INSTANCE)
>   (VkVertexInputBindingDescription) [2] = (binding = 2, stride = 0, inputRate = VK_VERTEX_INPUT_RATE_INSTANCE)
> }
> (lldb) parray 4 vertexInputCreateInfo.pVertexAttributeDescriptions
> (const VkVertexInputAttributeDescription *) $8 = 0x00007000088a61d0 {
>   (VkVertexInputAttributeDescription) [0] = (location = 0, binding = 0, format = VK_FORMAT_R32G32B32_SFLOAT, offset = 0)
>   (VkVertexInputAttributeDescription) [1] = (location = 1, binding = 2, format = VK_FORMAT_R32G32B32A32_SFLOAT, offset = 0)
>   (VkVertexInputAttributeDescription) [2] = (location = 2, binding = 2, format = VK_FORMAT_R32G32B32A32_SFLOAT, offset = 0)
>   (VkVertexInputAttributeDescription) [3] = (location = 3, binding = 1, format = VK_FORMAT_R16G16_SFLOAT, offset = 0)
> ```
> Is this really a case MoltenVK can't handle? It's worked on Metal for a long time. Not sure I understand the MoltenVK change here.

I think I got the boundary condition wrong in that check. I added it after encountering some cases in the CTS where offset > stride, which Metal protested loudly against (i.e. killing the process). @billhollings wrote #522 to fix that.

billhollings:
>One thing I pointed out to Bill is that this error:...is causing pipeline creation to fail for some of our pipelines

PR #522 fixes this and the pipeline failures in Dota2. The code that broke this appeared prior to the tessellation PR...but after Dota2's MoltenVK version.

> Is this really a case MoltenVK can't handle? It's worked on Metal for a long time. Not sure I understand the MoltenVK change here.

No...it's fine. The code that tested for offsets and strides was over-aggressive. It should permit a zero offset and stride combo...and now does.

I'm testing the updated fix now against this tessellation PR.

billhollings:
@cdavis5e 

PR #522 conflicts with this PR in `MVKPipeline.mm`. To merge...accept the changes from #522...but return `false` instead of `nil`...and use your `pVI` instead of `pCreateInfo->pVertexInputState`.

billhollings:
Sadly #522 does not fix the primary crashes reported above. So it's unrelated to the pipeline failures that were happening in Dota2. 😭

It may be a twisted hint that no crashes or errors appear while ASAN is running. So however ASAN is altering memory access is smothering the memory errors without detecting them. 🤔

KenThomases:
I didn't mention it before because ASAN would seem to be more powerful, but if it's not helping… Have you tried using MallocScribble=1, MallocCheckHeapStart=1000, and/or CFZombieLevel=65537 (the scribble bits)?

HansKristian-Work:
Valgrind is also an option. However, it does not work yet on macOS 10.14 sadly.

danginsburg:
So I'm unfortunately having trouble reproducing the crash in our development branch, but I did repro the crash in Release.  This was the callstack I saw.

0   libsystem_malloc.dylib        	0x00007fff794ff9ca free_tiny + 330
1   libengine2.dylib              	0x00000000112b8b2d operator delete(void*) + 29
2   libMoltenVK.dylib             	0x0000000086a5ca18 spirv_cross::CompilerGLSL::branch_to_continue(unsigned int, unsigned int) + 280
3   libMoltenVK.dylib             	0x0000000086a5d053 spirv_cross::CompilerGLSL::branch(unsigned int, unsigned int, unsigned int, unsigned int) + 531
4   libMoltenVK.dylib             	0x0000000086a5ac19 spirv_cross::CompilerGLSL::emit_block_chain(spirv_cross::SPIRBlock&) + 2105
5   libMoltenVK.dylib             	0x0000000086a5aa17 spirv_cross::CompilerGLSL::emit_block_chain(spirv_cross::SPIRBlock&) + 1591
6   libMoltenVK.dylib             	0x0000000086a5aa17 spirv_cross::CompilerGLSL::emit_block_chain(spirv_cross::SPIRBlock&) + 1591
7   libMoltenVK.dylib             	0x0000000086a5aa17 spirv_cross::CompilerGLSL::emit_block_chain(spirv_cross::SPIRBlock&) + 1591
8   libMoltenVK.dylib             	0x0000000086a5aa17 spirv_cross::CompilerGLSL::emit_block_chain(spirv_cross::SPIRBlock&) + 1591
9   libMoltenVK.dylib             	0x0000000086a5d005 spirv_cross::CompilerGLSL::branch(unsigned int, unsigned int, unsigned int, unsigned int) + 453
10  libMoltenVK.dylib             	0x0000000086a5ac19 spirv_cross::CompilerGLSL::emit_block_chain(spirv_cross::SPIRBlock&) + 2105
11  libMoltenVK.dylib             	0x0000000086a5aa17 spirv_cross::CompilerGLSL::emit_block_chain(spirv_cross::SPIRBlock&) + 1591
12  libMoltenVK.dylib             	0x0000000086a5aa17 spirv_cross::CompilerGLSL::emit_block_chain(spirv_cross::SPIRBlock&) + 1591
13  libMoltenVK.dylib             	0x0000000086a5aa17 spirv_cross::CompilerGLSL::emit_block_chain(spirv_cross::SPIRBlock&) + 1591
14  libMoltenVK.dylib             	0x0000000086a5aa17 spirv_cross::CompilerGLSL::emit_block_chain(spirv_cross::SPIRBlock&) + 1591
15  libMoltenVK.dylib             	0x0000000086a5aa17 spirv_cross::CompilerGLSL::emit_block_chain(spirv_cross::SPIRBlock&) + 1591
16  libMoltenVK.dylib             	0x0000000086a5aa17 spirv_cross::CompilerGLSL::emit_block_chain(spirv_cross::SPIRBlock&) + 1591
17  libMoltenVK.dylib             	0x0000000086a5aa17 spirv_cross::CompilerGLSL::emit_block_chain(spirv_cross::SPIRBlock&) + 1591
18  libMoltenVK.dylib             	0x0000000086a5aa17 spirv_cross::CompilerGLSL::emit_block_chain(spirv_cross::SPIRBlock&) + 1591
19  libMoltenVK.dylib             	0x0000000086a5aa17 spirv_cross::CompilerGLSL::emit_block_chain(spirv_cross::SPIRBlock&) + 1591
20  libMoltenVK.dylib             	0x0000000086a0571e spirv_cross::CompilerGLSL::emit_function(spirv_cross::SPIRFunction&, spirv_cross::Bitset const&) + 1422
21  libMoltenVK.dylib             	0x000000008697a92c spirv_cross::CompilerMSL::compile() + 2108
22  libMoltenVK.dylib             	0x000000008697ff21 spirv_cross::CompilerMSL::compile(std::__1::vector<spirv_cross::MSLVertexAttr, std::__1::allocator<spirv_cross::MSLVertexAttr> >*, std::__1::vector<spirv_cross::MSLResourceBinding, std::__1::allocator<spirv_cross::MSLResourceBinding> >*) + 449
23  libMoltenVK.dylib             	0x0000000086950189 mvk::SPIRVToMSLConverter::convert(mvk::SPIRVToMSLConverterContext&, bool, bool, bool) + 1625
24  libMoltenVK.dylib             	0x00000000869410eb MVKShaderModule::convert(mvk::SPIRVToMSLConverterContext*) + 75
25  libMoltenVK.dylib             	0x0000000086940f76 MVKShaderLibraryCache::getShaderLibrary(mvk::SPIRVToMSLConverterContext*, MVKShaderModule*, bool*) + 118
26  libMoltenVK.dylib             	0x0000000086920e19 MVKPipelineCache::getShaderLibrary(mvk::SPIRVToMSLConverterContext*, MVKShaderModule*) + 265
27  libMoltenVK.dylib             	0x0000000086941591 MVKShaderModule::getMTLFunction(mvk::SPIRVToMSLConverterContext*, VkSpecializationInfo const*, MVKPipelineCache*) + 113
28  libMoltenVK.dylib             	0x000000008691f315 MVKGraphicsPipeline::getMTLRenderPipelineDescriptor(VkGraphicsPipelineCreateInfo const*) + 677
29  libMoltenVK.dylib             	0x000000008691ee30 MVKGraphicsPipeline::initMTLRenderPipelineState(VkGraphicsPipelineCreateInfo const*) + 32
30  libMoltenVK.dylib             	0x000000008691e25a MVKGraphicsPipeline::MVKGraphicsPipeline(MVKDevice*, MVKPipelineCache*, MVKPipeline*, VkGraphicsPipelineCreateInfo const*) + 810
31  libMoltenVK.dylib             	0x000000008691a68f VkResult MVKDevice::createPipelines<MVKGraphicsPipeline, VkGraphicsPipelineCreateInfo>(VkPipelineCache_T*, unsigned int, VkGraphicsPipelineCreateInfo const*, VkAllocationCallbacks const*, VkPipeline_T**) + 143
32  librendersystemvulkan.dylib   	0x0000000086272b99 0x8621b000 + 359321
33  librendersystemvulkan.dylib   	0x0000000086273b96 0x8621b000 + 363414
34  librendersystemvulkan.dylib   	0x0000000086280fb5 0x8621b000 + 417717
35  librendersystemvulkan.dylib   	0x0000000086280dd8 0x8621b000 + 417240
36  libclient.dylib               	0x0000000093e85050 0x92f1c000 + 16158800
37  libscenesystem.dylib          	0x000000008f2ccd84 0x8f24a000 + 535940
38  libvstdlib.dylib              	0x00000000117989d5 0x11789000 + 63957
39  libscenesystem.dylib          	0x000000008f45091f 0x8f24a000 + 2124063
40  libscenesystem.dylib          	0x000000008f2a9126 0x8f24a000 + 389414
41  libengine2.dylib              	0x00000000110ef98d 0x1107d000 + 469389
42  libengine2.dylib              	0x00000000110e0d2b 0x1107d000 + 408875
43  libengine2.dylib              	0x00000000110e07b2 0x1107d000 + 407474
44  libengine2.dylib              	0x00000000110d41a0 0x1107d000 + 356768
45  libengine2.dylib              	0x00000000110d48d0 0x1107d000 + 358608
46  libengine2.dylib              	0x00000000110d4b5a Source2Main + 298
47  com.dota2                     	0x000000000c2e2d65 0xc2e2000 + 3429
48  libdyld.dylib                 	0x00007fff793523d5 start + 1

danginsburg:
By the way, a thought about what be making this hard to repro: if you want to fully invoke all the MSL compiles run to run you need to delete ~/Library/Application Support/Steam/steamapps/common/dota 2 beta/game/dota/shadercache/vulkan/shaders.cache between runs.

billhollings:
@danginsburg 

Thanks for the results. Is that stack consistent? I have not seen that particular one yet.

Can you try creating a few crashes and see if it is consistent or bounces around the code randomly?

Thanks...yes...I have been starting and stopping Data2 via Xcode...so it never takes the step to save the shader cache. And I've been verifying that the cache is clear.

billhollings:
@danginsburg 

>So I'm unfortunately having trouble reproducing the crash in our development branch, but I did repro the crash in Release.

What is different between your development branch and Release?

danginsburg:
> What is different between your development branch and Release?

A lot.  If need be I can sync up the release branch and try to repro there.

> Can you try creating a few crashes and see if it is consistent or bounces around the code randomly?


Here's one I caught running in Xcode just now.  I did see a couple others that were in more random places, but twice it has failed inside spirv-cross.

```
(lldb) bt
* thread #10, name = 'GlobPool5', stop reason = EXC_BAD_ACCESS (code=1, address=0xf)
  * frame #0: 0x000000008be68680 libMoltenVK.dylib`spirv_cross::Bitset::merge_or(spirv_cross::Bitset const&) + 64
    frame #1: 0x000000008be56152 libMoltenVK.dylib`spirv_cross::Compiler::combined_decoration_for_member(spirv_cross::SPIRType const&, unsigned int) const + 162
    frame #2: 0x000000008bdff0bc libMoltenVK.dylib`spirv_cross::CompilerMSL::member_is_non_native_row_major_matrix(spirv_cross::SPIRType const&, unsigned int) + 60
    frame #3: 0x000000008be7e6c8 libMoltenVK.dylib`spirv_cross::CompilerGLSL::access_chain_internal(unsigned int, unsigned int const*, unsigned int, unsigned int, spirv_cross::AccessChainMeta*) + 3240
    frame #4: 0x000000008be98468 libMoltenVK.dylib`spirv_cross::CompilerGLSL::access_chain(unsigned int, unsigned int const*, unsigned int, spirv_cross::SPIRType const&, spirv_cross::AccessChainMeta*, bool) + 536
    frame #5: 0x000000008be9d0b6 libMoltenVK.dylib`spirv_cross::CompilerGLSL::emit_instruction(spirv_cross::Instruction const&) + 1382
    frame #6: 0x000000008bdeccf8 libMoltenVK.dylib`spirv_cross::CompilerMSL::emit_instruction(spirv_cross::Instruction const&) + 5096
    frame #7: 0x000000008beb645d libMoltenVK.dylib`spirv_cross::CompilerGLSL::emit_block_chain(spirv_cross::SPIRBlock&) + 4221
    frame #8: 0x000000008be6071e libMoltenVK.dylib`spirv_cross::CompilerGLSL::emit_function(spirv_cross::SPIRFunction&, spirv_cross::Bitset const&) + 1422
    frame #9: 0x000000008bdd592c libMoltenVK.dylib`spirv_cross::CompilerMSL::compile() + 2108
    frame #10: 0x000000008bddaf21 libMoltenVK.dylib`spirv_cross::CompilerMSL::compile(std::__1::vector<spirv_cross::MSLVertexAttr, std::__1::allocator<spirv_cross::MSLVertexAttr> >*, std::__1::vector<spirv_cross::MSLResourceBinding, std::__1::allocator<spirv_cross::MSLResourceBinding> >*) + 449
    frame #11: 0x000000008bdab189 libMoltenVK.dylib`mvk::SPIRVToMSLConverter::convert(mvk::SPIRVToMSLConverterContext&, bool, bool, bool) + 1625
    frame #12: 0x000000008bd9c0eb libMoltenVK.dylib`MVKShaderModule::convert(mvk::SPIRVToMSLConverterContext*) + 75
    frame #13: 0x000000008bd9bf76 libMoltenVK.dylib`MVKShaderLibraryCache::getShaderLibrary(mvk::SPIRVToMSLConverterContext*, MVKShaderModule*, bool*) + 118
    frame #14: 0x000000008bd7be19 libMoltenVK.dylib`MVKPipelineCache::getShaderLibrary(mvk::SPIRVToMSLConverterContext*, MVKShaderModule*) + 265
    frame #15: 0x000000008bd9c591 libMoltenVK.dylib`MVKShaderModule::getMTLFunction(mvk::SPIRVToMSLConverterContext*, VkSpecializationInfo const*, MVKPipelineCache*) + 113
    frame #16: 0x000000008bd7a1f0 libMoltenVK.dylib`MVKGraphicsPipeline::getMTLRenderPipelineDescriptor(VkGraphicsPipelineCreateInfo const*) + 384
    frame #17: 0x000000008bd79e30 libMoltenVK.dylib`MVKGraphicsPipeline::initMTLRenderPipelineState(VkGraphicsPipelineCreateInfo const*) + 32
    frame #18: 0x000000008bd7925a libMoltenVK.dylib`MVKGraphicsPipeline::MVKGraphicsPipeline(MVKDevice*, MVKPipelineCache*, MVKPipeline*, VkGraphicsPipelineCreateInfo const*) + 810
    frame #19: 0x000000008bd7568f libMoltenVK.dylib`VkResult MVKDevice::createPipelines<MVKGraphicsPipeline, VkGraphicsPipelineCreateInfo>(VkPipelineCache_T*, unsigned int, VkGraphicsPipelineCreateInfo const*, VkAllocationCallbacks const*, VkPipeline_T**) + 143
    frame #20: 0x000000008e29fb99 librendersystemvulkan.dylib`___lldb_unnamed_symbol983$$librendersystemvulkan.dylib + 2633
    frame #21: 0x000000008e2a0b96 librendersystemvulkan.dylib`___lldb_unnamed_symbol985$$librendersystemvulkan.dylib + 1638
    frame #22: 0x000000008e2adfb5 librendersystemvulkan.dylib`___lldb_unnamed_symbol1074$$librendersystemvulkan.dylib + 405
    frame #23: 0x000000008e2addd8 librendersystemvulkan.dylib`___lldb_unnamed_symbol1073$$librendersystemvulkan.dylib + 24
    frame #24: 0x000000009c43e050 libclient.dylib`___lldb_unnamed_symbol93245$$libclient.dylib + 672
    frame #25: 0x0000000097861d84 libscenesystem.dylib`___lldb_unnamed_symbol1106$$libscenesystem.dylib + 180
    frame #26: 0x00000000022b15ae libvstdlib.dylib`___lldb_unnamed_symbol281$$libvstdlib.dylib + 910
    frame #27: 0x0000000002232f72 libtier0.dylib`CThread::ThreadProc(void*) + 306
    frame #28: 0x000000000222d89c libtier0.dylib`___lldb_unnamed_symbol360$$libtier0.dylib + 412
    frame #29: 0x00000000003c4dc3 libsystem_pthread.dylib`_pthread_body + 126
    frame #30: 0x00000000003c7e8d libsystem_pthread.dylib`_pthread_start + 66
    frame #31: 0x00000000003c3e11 libsystem_pthread.dylib`thread_start + 13
```

HansKristian-Work:
A very likely explanation is that SPIRV-Cross' memory somehow gets corrupted from the outside, address=0xf looks very odd. I haven't been able to repro a crash inside SPIRV-Cross at least. A couple of times it would crash inside the Metal compiler itself, and usually in librendersystemvulkan.dylib.

danginsburg:
I believe I have root caused this and it is a bug in Dota that looks unrelated to this PR, other than bad luck.  It only happened in our RETAIL build, but I was able to get a reliable repro there.  The problem stemmed from some uninitialized memory being used as a count for memset.  The count was clamped to an upper bound, but not a lower bound, so when it went negative it corrupted a potentially large area of memory.

I don't believe this PR needs to be held up for Dota.  I've checked in a fix and I'll integrate it to our release branch for a future Dota update.

Thanks everyone for their diligence and sorry for the inconvenience.

billhollings:
Wonderful! Thanks, @danginsburg !

HansKristian-Work:
Phew. Great that we finally figured it out.

mlfarrell:
Hey @billhollings

 Quick question: does this add tessellation control as well as evaluation?

cdavis5e:
Yes, this adds support for both kinds of tessellation shader.

mlfarrell:
Excellent. Thanks!

https://media1.giphy.com/media/1pofg0cTz5mak/source.gif

PR #528, opened 2019-03-18
Support the VK_EXT_host_query_reset extension.

cdavis5e:
This allows clients to reset query pools on the host, instead of with a
GPU command.

This updates Vulkan headers to 1.1.104... but, there's a problem. This
change has not been merged to `master` in the `Vulkan-Headers` repo yet,
because of an issue building the C++ binding headers. Luckily, we only
need the core C header.


billhollings:
Are you indicating that this is a new Vulkan extension as of 1.1.104, then?

I'm happy to see any extensions added...but I'm curious about what's motivating it? Is someone actually using a 1.1.104 extension? Or did you just come across it and see it as something that could easily be added?

cdavis5e:
> Are you indicating that this is a new Vulkan extension as of 1.1.104, then?

That's correct.

> 
> I'm happy to see any extensions added...but I'm curious about what's motivating it? Is someone actually using a 1.1.104 extension? Or did you just come across it and see it as something that could easily be added?

Yep, DXVK is using this now. In fact, since `VkEvent` isn't implemented, queries don't work at all unless this extension is supported.

PR #532, opened 2019-03-21
 Add correct function entry point handling.

billhollings:
Add correct function entry point handling.
Add support for VK_KHR_get_surface_capabilities2 extension.
Implement newer VK_KHR_swapchain extension functions.
Fix when the following functions return VK_INCOMPLETE:
    vkGetPhysicalDeviceSurfaceFormatsKHR()
    vkGetSwapchainImagesKHR()
    vkEnumeratePhysicalDevices()
    vkEnumerateInstanceLayerProperties()
    vkGetPhysicalDeviceQueueFamilyProperties()
CTS: Pass supported *.macos.surface.* and *.macos.swapchain.* tests.
Update What's New document.


MoltenVK/MoltenVK/GPUObjects/MVKInstance.mm
```diff
@@ -30,6 +30,22 @@
...
+	auto iter = _entryPoints.find(pName);
+	return (iter != _entryPoints.end()) ? &iter->second : nullptr;
+}
+
+// Returns core instance commands, enabled instance extension commands, and all device commands.
+PFN_vkVoidFunction MVKInstance::getProcAddr(const char* pName) {
+	MVKEntryPoint* pMVKPA = getEntryPoint(pName);
+
+	bool isSupported = (pMVKPA &&									// Command exists and...
+						(pMVKPA->isDevice ||						// ...is a device command or...
```
cdavis5e: Are you sure this condition isn't backwards?

billhollings: I agree it does look backwards...but I think it's good.

For `vkGetInstanceProcAddr()` the spec includes returning an FP for:

>available device extension commands for `instance`

and then expands on that with

>An “available device extension” is a device extension supported by any physical device enumerated by `instance`.

I interpreted that to simply mean any device extension supported by MoltenVK (and thereby Metal too).

While developing this I tested against CTS `dEQP-VK.api.version_check.entry_points` and all seems good with that approach.

PR #534, opened 2019-03-21
vkCmdBlitImage: Remove check for invalid behavior.

cdavis5e:
According to the Vulkan spec:

> * If either of `srcImage` or `dstImage` was created with a
>   depth/stencil format, the other **must** have exactly the same format

So this should not happen in well-behaved clients.

Is there a reason this check was added?


billhollings:
Was this causing an issue...or are you just clearing out an unnecessary validation test?

cdavis5e:
> Was this causing an issue...or are you just clearing out an unnecessary validation test?

This is just a drive-by fix while I was working on argument buffers (or rather, a change that will be needed to make argument buffers work).

billhollings:
Okay. Thanks. Good catch.

PR #543, opened 2019-03-27
CTS fixes. MVKPushConstantsCommandEncoderState compute / tessellation conflict.

billhollings:
Fixed crash within MVKPushConstantsCommandEncoderState when accessing absent graphics pipeline during a compute stage.
MVKPushConstantsCommandEncoderState test for tessellation only during graphics stages.
Guard against possible missing graphics pipeline even during graphics stages.

Support null device in MVKBaseDeviceObject and MVKDispatchableDeviceObject instances.

Issue #542, opened 2019-03-27
Wrong MTLTextureDescriptor for texel buffer-backing textures


ch45er:
Hi!

When working with texel buffers in MVK, I observe the following failed Metal validation assertions:
```
-[MTLDebugBuffer newTextureWithDescriptor:offset:bytesPerRow:]:326: failed assertion `resourceOptions (0x10) must match backing buffer resource options (0x20).'
```
```
validateComputeFunctionArguments:892: failed assertion `Compute Function(Main): writes texture (...) whose usage (0x01) doesn't specify MTLTextureUsageShaderWrite (0x02)'
```

The former one occurs upon a call to `vkUpdateDescriptorSets()`, and the latter one relates to `vkQueueSubmit()`.

It seems that `id<MTLTexture> MVKBufferView::getMTLTexture()` function does not set `resourceOptions` and `usage` values in `MTLTextureDescriptor` correctly, when creates a texture:

```c++
id<MTLTexture> MVKBufferView::getMTLTexture() {
// [code ommitted for brevity]
        MTLTextureDescriptor* mtlTexDesc = [MTLTextureDescriptor texture2DDescriptorWithPixelFormat: _mtlPixelFormat
                                                                                              width: _textureSize.width
                                                                                             height: _textureSize.height
                                                                                          mipmapped: NO];
        //mtlTexDesc.resourceOptions = 0x20;
        //mtlTexDesc.usage = MTLTextureUsageShaderWrite | MTLTextureUsageShaderRead | MTLTextureUsagePixelFormatView;
        
		_mtlTexture = [_buffer->getMTLBuffer() newTextureWithDescriptor: mtlTexDesc
																 offset: _mtlBufferOffset
															bytesPerRow: _mtlBytesPerRow];
// [code ommitted for brevity]
}
```

The commented-out lines in a middle "fix" those assertions for my particular use case.

So it seems that the `resourceOptions` and `usage` members of `mtlTexDesc` should also be set up.
 
My MVK version is `v1.0.28-280-ge771f79`, but it seems that this was present in the code since initial public release.

billhollings:
@ch45er Thanks for the research.

I'll look into fixing this properly.

billhollings:
@ch45er 

Can you clarify the use of your options, please? Did you actually need to set them all to get your use case satisfied?

According to Apple docs, setting `resourceOptions` shouldn't have any effect:

[...]

PR #546, opened 2019-03-27
Do not check for tessellation state if there is no graphics pipeline.

HansKristian-Work:
I believe this is an error from the tessellation additions. I got a segfault when using a compute-only command buffer which also used push constants.


cdavis5e:
This is also fixed in #543.

HansKristian-Work:
I see. Then this PR is redundant :) Closing.

billhollings:
We're all crossing paths today! I was halfway through a fix for #542 when @cdavis5e 's PR came in. 😉

Issue #544, opened 2019-03-27
vkCmdFillBuffer does not support VK_WHOLE_SIZE


HansKristian-Work:
I got this error: `[***MoltenVK ERROR***] Buffer fill size must fit into a 32-bit unsigned integer.
`. It seems to originate from a `vkCmdFillBuffer(cmd, ..., VK_WHOLE_SIZE)`.

HansKristian-Work:
Fixed in PR.

PR #547, opened 2019-03-27
Fix wrong offset for vkCmdFillBuffer on VK_WHOLE_SIZE.

HansKristian-Work:
A double offset would be applied by taking the user offset and global offset, causing the word count to wrap around and trigger an assertion and GPU lockup. This seems to be what caused a GPU hang in Granite for some time.

I also added a rounding bump for wordCount. I'm not sure if the internal buffer size is naturally rounded ahead of time, but the case for this rounding would be if a buffer was, say, 3 bytes, and the application calls for VK_WHOLE_SIZE.

This is related to:
https://github.com/KhronosGroup/MoltenVK/issues/544


billhollings:
Does this fix #544?

PR #550, opened 2019-03-29
Override namespace for SPIRV-Cross.

HansKristian-Work:
Some projects also link against SPIRV-Cross statically, and in order to
avoid ABI conflicts, we should use a private namespace for the
SPIRV-Cross dependency to avoid bugs. See SPIRV-Cross issue #902 for
more information. The new namespace is MVK_spirv_cross, and the code
now makes use of the SPIRV_CROSS_NAMESPACE macro rather than spirv_cross.


ExternalDependencies.xcodeproj/project.pbxproj
```diff
@@ -3488,6 +3488,7 @@
 				BITCODE_GENERATION_MODE = bitcode;
 				CODE_SIGN_IDENTITY = "iPhone Developer";
 				"CODE_SIGN_IDENTITY[sdk=iphoneos*]" = "";
+				GCC_PREPROCESSOR_DEFINITIONS = "SPIRV_CROSS_NAMESPACE_OVERRIDE=MVK_spirv_cross";
```
billhollings: All looks good.

I'm curious about why you use an additional define layer here with `SPIRV_CROSS_NAMESPACE_OVERRIDE`.

Was there a preference for why you didn't just use `SPIRV_CROSS_NAMESPACE` directly via something like the following in `spirv_common.hpp`:

```
#ifndef SPIRV_CROSS_NAMESPACE
#define SPIRV_CROSS_NAMESPACE spirv_cross
#endif
```

HansKristian-Work: I wanted to make it more explicit that defining it overrides the namespace.

PR #555, opened 2019-04-01
Workaround GPU Hang on iPhone 6s (iOS 12.2) where setting the renderp…

danginsburg:
…ass descriptor to a size smaller than the rendertarget is causing a GPU hang.  We render with offset renderArea/scissor and for reasons unclear this is inducing a GPU hang with no validation errors.  On iPad Air (3rd Gen) it works fine.

@billhollings I'm not sure what to do with this one - I think it may have some performance implications for cases where the renderArea is smaller than the render target, although it seems necessary for us not to hang.

### DIFF
--- a/MoltenVK/MoltenVK/Commands/MVKCommandBuffer.mm
+++ b/MoltenVK/MoltenVK/Commands/MVKCommandBuffer.mm
@@ -250,8 +250,8 @@ void MVKCommandEncoder::beginMetalRenderPass() {
     getSubpass()->populateMTLRenderPassDescriptor(mtlRPDesc, _framebuffer, _clearValues, _isRenderingEntireAttachment);
     mtlRPDesc.visibilityResultBuffer = _occlusionQueryState.getVisibilityResultMTLBuffer();
 	mtlRPDesc.renderTargetArrayLengthMVK = _framebuffer->getLayerCount();
-	mtlRPDesc.renderTargetWidthMVK = min(_framebuffer->getExtent2D().width, _renderArea.offset.x + _renderArea.extent.width);
-	mtlRPDesc.renderTargetHeightMVK = min(_framebuffer->getExtent2D().height, _renderArea.offset.y + _renderArea.extent.height);
+	mtlRPDesc.renderTargetWidthMVK = _framebuffer->getExtent2D().width;
+	mtlRPDesc.renderTargetHeightMVK = _framebuffer->getExtent2D().height;
 
     _mtlRenderEncoder = [_mtlCmdBuffer renderCommandEncoderWithDescriptor: mtlRPDesc];     // not retained
     _mtlRenderEncoder.label = getMTLRenderCommandEncoderName();


billhollings:
> I'm not sure what to do with this one - I think it may have some performance implications for cases where the renderArea is smaller than the render target, although it seems necessary for us not to hang.

If we think this is a GPU-specific bug...we could also combine your two mods (this one and #554) with a test for `MTLFeatureSet_iOS_GPUFamily4_v1` and pick one of the two accordingly at runtime. That would provide for both options...efficient rendering for GPU's that can handle the setting of this parameter...and crash-avoidance for those that can't.

I'm curious though...

How should we interpret the Metal `renderTargetWidth` and `renderTargetHeight` properties as applied to layered rendering? The Apple docs are short on detail.

- Is it meant to be a scissor-like mask that disallows rendering outside the boundaries...and if so...why is it just extent that is limited and not origin...and why doesn't it just use the scissor limits?

- Or is it just meant to be a performance enhancement option...as you imply in your comment?

- Or something else?

danginsburg:
> Is it meant to be a scissor-like mask that disallows rendering outside the boundaries...and if so...why is it just extent that is limited and not origin...and why doesn't it just use the scissor limits?

I'm not sure - I guessed the load/store ops would be contained by this boundary, but it's true subsequent drawing operations can be constrained by scissor.  I'm not familiar enough with Metal here to say.  The Vulkan renderArea lets the driver restrain the tile boundaries for the load/store ops, but as you say Metal doesn't have an offset so it seems like a lesser feature.

I'm fine with a test for  MTLFeatureSet_iOS_GPUFamily4_v1, although for what it's worth my test matrix included only three devices.  Two of them had a hang (iPhone 6s/A8 and iPad Air 1st Gen/A7) and the other did not (iPad Air 3rd gen/A12).

billhollings:
@cdavis5e Do you have any insights here? The setting of these properties were added by you in #274, and then reduced in size to render area in #287. Do you have use cases that motivated those two PR's?

The Apple docs clearly group `renderTargetWidth` and `renderTargetHeight` under the topic of _Layered Rendering_, right beside `renderTargetArrayLength`. But they give no description about what they are used for...other than to "constrain" the target dimensions...presumably in a way that is different than standard scissors.

Because of this grouping...to solve the issue @danginsburg has highlighted...I'm wondering if we can simply assume that `renderTargetWidth` and `renderTargetHeight` should only be set if layered rendering is supported? That would certainly fix the issue of older GPU's barfing on this value being anything other than the full frame buffer size.

Both `renderTargetWidth` and `renderTargetHeight` are only available on iOS devices.

billhollings:
@cdavis5e Ping re question above.

cdavis5e:
> @cdavis5e Do you have any insights here? The setting of these properties were added by you in #274, and then reduced in size to render area in #287. Do you have use cases that motivated those two PR's?

I initially wrote those when I was working on #284. I was hoping that, when those properties were added on macOS, that rendering with no attachments would work.

> The Apple docs clearly group `renderTargetWidth` and `renderTargetHeight` under the topic of _Layered Rendering_, right beside `renderTargetArrayLength`. But they give no description about what they are used for...other than to "constrain" the target dimensions...presumably in a way that is different than standard scissors.

That's what led me to take the `renderArea` into account in #287.

billhollings:
> I initially wrote those when I was working on #284. I was hoping that, when those properties were added on macOS, that rendering with no attachments would work.
> ...
> That's what led me to take the `renderArea` into account in #287.

Thanks. 

Okay...so we need at least to fix the issue that triggered this PR. If we don't have any particular use case for `renderTargetWidth` and `renderTargetHeight` right now...and don't have a solid understanding of them...options include...

1. Simply ignore them and never set them (ie. pre PR #274).
2. Set them only to the framebuffer extent (as per this PR).
3. Set them only when layered rendering is supported...and to the min of framebuffer extent and render area offset + extent (ie. pre this PR but only with layered rendering).

@cdavis5e @danginsburg Thoughts? Left to my own devices...I'd probably opt for the option 3.

danginsburg:
I think that any of those three options would fix it in our case.  So if option 3 works for @cdavis5e it's fine with me.

Thanks for the follow up.

cdavis5e:
Fine by me, since Mac doesn't support this yet anyway. I may want to revisit this later, though, depending on what the next version of Metal supports.

billhollings:
Replaced by PR #562.

PR #559, opened 2019-04-01
Actually set the renderTargetArrayLength property on iOS 12.

### LINKED ISSUE #558: renderTargetArrayLength not set on iOS
I'm testing on iOS 12.2 (iPad Air 3rd Gen) and hitting this validation error:

-[MTLDebugRenderCommandEncoder validateCommonDrawErrors:instanceCount:baseInstance:maxVertexID:]:3883: failed assertion `the renderPipelineState has a vertex shader that writes to render_target_array_index but renderTargetArrayLength is set to zero’

The reason this is happening is because of the MVK_IOS #ifdef's here:

https://github.com/KhronosGroup/MoltenVK/blob/d95ea7e9a7589a1fc33906c673a6c835f74c716c/MoltenVK/MoltenVK/OS/MTLRenderPassDescriptor%2BMoltenVK.m#L31

and here:

https://github.com/KhronosGroup/MoltenVK/blob/d95ea7e9a7589a1fc33906c673a6c835f74c716c/MoltenVK/MoltenVK/OS/MTLRenderPassDescriptor%2BMoltenVK.m#L39

So either the clear shader needs to not use render_target_array_index on iOS or these would need to change to support iOS.  I locally hacked them to work on iOS and on my particular 12.2 device it did work.

cdavis5e:
Fixes #558.

### DIFF
--- a/MoltenVK/MoltenVK/OS/MTLRenderPassDescriptor+MoltenVK.m
+++ b/MoltenVK/MoltenVK/OS/MTLRenderPassDescriptor+MoltenVK.m
@@ -28,6 +28,8 @@
 	return self.renderTargetArrayLength;
 #endif
 #if MVK_IOS
+	if ( [self respondsToSelector: @selector(renderTargetArrayLength)] ) {
+		return self.renderTargetArrayLength;
 	return 0;
 #endif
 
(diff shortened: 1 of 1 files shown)


MoltenVK/MoltenVK/OS/MTLRenderPassDescriptor+MoltenVK.m
```diff
@@ -38,6 +40,10 @@ -(void) setRenderTargetArrayLengthMVK: (NSUInteger) length {
 #if MVK_MACOS
 	self.renderTargetArrayLength = length;
 #endif
+#if MVK_IOS
+	if ( [self respondsToSelector: @selector(setRenderTargetArrayLength:)] ) {
+		self.renderTargetArrayLength = length;
```
billhollings: This { seems to be unmatched. I think this is what Travis is complaining about.

PR #565, opened 2019-04-09
Several fixes regarding tessellation, especially in indexed and instanced rendering

mbarriault:
This is currently the sum of issues I've been able to find and fix regarding tessellation. These all work to my testing but may require deeper testing due to my limited selection of hardware.


billhollings:
@mbarriault Thanks for the submission!

@cdavis5e Will you be able to confirm these fixes? 

Do we want to try to get these into the SDK release this week?

cdavis5e:
> @cdavis5e Will you be able to confirm these fixes?

LGTM.
> 
> Do we want to try to get these into the SDK release this week?

I think so.

PR #577, opened 2019-04-20
Fix unused attachments terminating loop early.

mbarriault:


MoltenVK/MoltenVK/GPUObjects/MVKRenderPass.mm
```diff
@@ -73,10 +73,11 @@
 													   bool storeOverride) {
 	// Populate the Metal color attachments
 	uint32_t caCnt = getColorAttachmentCount();
+    uint32_t caUsedCnt = caCnt;
 	for (uint32_t caIdx = 0; caIdx < caCnt; caIdx++) {
 		uint32_t clrRPAttIdx = _colorAttachments[caIdx].attachment;
+        MTLRenderPassColorAttachmentDescriptor* mtlColorAttDesc = mtlRPDesc.colorAttachments[caIdx];
         if (clrRPAttIdx != VK_ATTACHMENT_UNUSED) {
-            MTLRenderPassColorAttachmentDescriptor* mtlColorAttDesc = mtlRPDesc.colorAttachments[caIdx];
```
billhollings: Is there a reason for moving accessing the `MTLRenderPassColorAttachmentDescriptor` to above the test for whether it is needed? Nothing is subsequently done with it.

My concern is that accessing it might trigger an internal activity on it within the Metal class (like some form of lazy initialization) that makes it behave a bit differently than if it was left as alone.

mbarriault: My understanding is that has no side effects, however it indeed shouldn't have been moved - that was leftover from my original attempt at fixing the problem. Will update.

MoltenVK/MoltenVK/GPUObjects/MVKRenderPass.mm
```diff
@@ -98,7 +99,7 @@
 			}
 
 		} else {
-			caCnt--;
+			caUsedCnt--;
```
billhollings: One more change. Sorry...I should have thought of this on my first review. 🤦🏻‍♂️

Now that you have a separate variable for tracking this...can we make `caUsedCnt` just start at zero...increment it within the `if()` block...and remove the `else` block entirely. Seems to be a more logical direction.

mbarriault: Done.

billhollings: Thanks!

PR #578, opened 2019-04-22
Fix case where viewport/scissor doesn't get set properly when mixing dynamic and static-configured pipelines in the same command buffer.

mbarriault:
The `mustSetDynamically` and `isSetDynamically` flags clashed when calling `vkCmdSetViewport`/`vkCmdSetScissor` after drawing a pipeline with a static viewport/scissor but before binding a pipeline with a dynamic viewport/scissor.

### DIFF
--- a/MoltenVK/MoltenVK/Commands/MVKCommandEncoderState.mm
+++ b/MoltenVK/MoltenVK/Commands/MVKCommandEncoderState.mm
@@ -56,7 +56,7 @@ void MVKViewportCommandEncoderState::setViewports(const MVKVector<MTLViewport> &
 
 	uint32_t maxViewports = _cmdEncoder->getDevice()->_pProperties->limits.maxViewports;
 	if ((firstViewport + mtlViewports.size() > maxViewports) ||
-		(firstViewport <= maxViewports) ||
+		(firstViewport >= maxViewports) ||
 		(isSettingDynamically && mtlViewports.size() == 0))
 		return;
 
@@ -103,7 +103,7 @@ void MVKScissorCommandEncoderState::setScissors(const MVKVector<MTLScissorRect> 
 
 	uint32_t maxScissors = _cmdEncoder->getDevice()->_pProperties->limits.maxViewports;
 	if ((firstScissor + mtlScissors.size() > maxScissors) ||
-		(firstScissor <= maxScissors) ||
+		(firstScissor >= maxScissors) ||
 		(isSettingDynamically && mtlScissors.size() == 0))
 		return;
 
--- a/MoltenVK/MoltenVK/Commands/MVKCommandEncoderState.mm
+++ b/MoltenVK/MoltenVK/Commands/MVKCommandEncoderState.mm
@@ -55,44 +55,36 @@ void MVKViewportCommandEncoderState::setViewports(const MVKVector<MTLViewport> &
 												  bool isSettingDynamically) {
 
 	uint32_t maxViewports = _cmdEncoder->getDevice()->_pProperties->limits.maxViewports;
-	if (isSettingDynamically && mtlViewports.size() > 0 &&
-		(firstViewport + mtlViewports.size() <= maxViewports) &&
-		(firstViewport < maxViewports)) {
+	if ((firstViewport + mtlViewports.size() > maxViewports) ||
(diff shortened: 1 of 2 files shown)


MoltenVK/MoltenVK/Commands/MVKCommandEncoderState.mm
```diff
@@ -54,34 +54,51 @@
...
+		if (firstViewport + mtlViewports.size() > _mtlDynamicViewports.size()) {
+			_mtlDynamicViewports.resize(firstViewport + mtlViewports.size());
+		}
+
+		std::copy(mtlViewports.begin(), mtlViewports.end(), _mtlDynamicViewports.begin() + firstViewport);
+
+		markDirty();
+	} else if (!isSettingDynamically &&
+		(firstViewport + mtlViewports.size() <= maxViewports) &&
+		(firstViewport < maxViewports)) {
```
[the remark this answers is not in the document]
mbarriault: Good idea, done.

MoltenVK/MoltenVK/Commands/MVKCommandEncoderState.mm
```diff
@@ -54,34 +54,43 @@
...
+		usingMTLViewports.resize(firstViewport + mtlViewports.size());
 	}
+
+	bool mustSetDynamically = _cmdEncoder->supportsDynamicState(VK_DYNAMIC_STATE_VIEWPORT);
+
+	if (isSettingDynamically ||
+		(!isSettingDynamically && !mustSetDynamically && mtlViewports.size() > 0))
+		std::copy(mtlViewports.begin(), mtlViewports.end(), usingMTLViewports.begin() + firstViewport);
+	else if (!isSettingDynamically && (mustSetDynamically || mtlViewports.size() == 0))
+		usingMTLViewports.clear();
```
billhollings: If I'm parsing the boolean logic of the `if` & `else if` conditions, I think it can be simplified:

- The `!isSettingDynamically &&` phrase in both the `if` and `else if` are redundant, since `!isSettingDynamically` is guaranteed to be `true` at that point in each case, otherwise the initial `isSettingDynamically` would have triggered the `if` block.
- The `(mustSetDynamically || mtlViewports.size() == 0)` in the `else if` is redundant as it is the logical negation of the  `!mustSetDynamically && mtlViewports.size() > 0` in the `if` (ie. (A || B) == !(!A & !B) ).

So could this just be simplified to:

    if (isSettingDynamically || (!mustSetDynamically && mtlViewports.size() > 0))
    ...
    else
    ...	

Please confirm my logic! And if I've made a mistake and the two branches are not complimentary...then what is the effect of neither of the `if` of `else if` blocks executing and then `markDirty()` being called?

MoltenVK/MoltenVK/Commands/MVKCommandEncoderState.mm
```diff
@@ -92,26 +101,34 @@
...
+		usingMTLScissors.resize(firstScissor + mtlScissors.size());
 	}
+
+	bool mustSetDynamically = _cmdEncoder->supportsDynamicState(VK_DYNAMIC_STATE_SCISSOR);
+
+	if (isSettingDynamically ||
+		(!isSettingDynamically && !mustSetDynamically && mtlScissors.size() > 0))
+		std::copy(mtlScissors.begin(), mtlScissors.end(), usingMTLScissors.begin() + firstScissor);
+	else if (!isSettingDynamically && (mustSetDynamically || mtlScissors.size() == 0))
+		usingMTLScissors.clear();
```
billhollings: Same here.

MoltenVK/MoltenVK/Commands/MVKCommandEncoderState.mm
mbarriault: I believe you're right. The logical tree here is such that one of those two blocks will always be called, which requires that indeed it can be simplified to `if-else`. Will check and commit.

MoltenVK/MoltenVK/Commands/MVKCommandEncoderState.mm
mbarriault: Resolved.

billhollings:
Thanks for taking care of the update suggestions.

Issue #11, opened 2018-01-10
Feature request: VK_EXT_debug_marker for GPU frame capture


danginsburg:
It would be very useful if MoltenVK could support VK_EXT_debug_marker and pass the markers down to Metal so that they showed up during GPU Frame Capture.  For example, in Source 2 we label all command buffers and many resources and seeing that within the Metal GPU Frame Capture would be very helpful.

billhollings:
PR #588 adds support for `VK_EXT_debug_report`, which is the prerequisite for `VK_EXT_debug_marker`.

billhollings:
PR #608 implements this extension.

One implementation limitation is the labelling of pipelines. Unfortunately, for some strange reason, the labels of Metal pipelines are readonly, and Metal only allows the label of a pipeline to be set during pipeline creation, via the pipeline descriptor. 

I explored potentially delaying the creation of the Metal pipeline to give the app a chance to set the pipeline object name, but we really need to go all the way and create the Metal pipeline during Vulkan pipeline creation in order to provide error feedback to the app. 

In the end…I opted to name the pipeline after the name of the pipeline layout.

There is the option of recompiling the Metal pipeline whenever a name is set for the corresponding Vulkan pipeline. But I didn’t want to assume an app would be happy with that behaviour, since it would obviously double pipeline creation time for any app that innocently sets a debug name for its pipelines. 

But I am happy to hear feedback on that option.

PR #588, opened 2019-05-02
Add support for the VK_EXT_debug_report extension

billhollings:
Add MVKDebugReportCallback class.
Add MVKVulkanAPIObject class as base class of classes exposed in Vulkan API.
Remove MVKRefCountedDeviceObject class and move ref-counting to MVKVulkanAPIObject.
Rename MVKDispatchableObject class to MVKDispatchableVulkanAPIObject.
Introduce multiple inheritance mixin classes into class hierarchy.
Add MVKBaseObject::getVulkanAPIObject() function to find Vulkan API object 
controlling any subobject that needs to report something.
MVKCommandResourceFactory functions pass in calling object for reporting.
Add MVKBaseObject::reportMessage() & reportError() functions.
Replace mvkNotifyErrorWithText() static function with reportError()
instance function to allow access to debug reporting.
Redefine MVKLog*() functions to call reportError() instance method.
Delete MVKLogging.cpp.
Remove MVKLogging.h from MoltenVKShaderConverter project.
Add mvk_datatypes.hpp to redirect errors in mvk_datatypes.h functions to debug 
reporting. Use mvk_datatypes.hpp internally for all references to these functions.
Convert several static methods to instance methods. 
Refactor platform defines in MVKSurface.h.
Explicitly count MVKExtensionList extension count to avoid relying on struct size.
Remove MVKCommandBuffer::recordResult() and use setConfigurationResult().
Change log indication of error in logs from [***MoltenVK ERROR***] to [mvk-error],
for consistency with other log level indications.
Update MoltenVK version to 1.0.35.

Adds the first dependency feature for the functionality requested in issue #11.


billhollings:
@cdavis5e , et al...

This PR is not intrusive from the perspective of Vulkan operational functionality. All Vulkan features and capabilities _**should**_ be functioning as there were already.


However...this is highly intrusive from the following perspectives:
- File updates
- Logging plumbing
- Internal Vulkan class hierarchy

So...I don't expect it to be worthwhile trying to comb through all the code changes...but before merging this I would appreciate some basic testing and validation against existing apps...just in case.

Thanks.

cdavis5e:
I've tested it with Final Fantasy XIV. So far, I'm not seeing any problems with it.

I can see, though, that this is going to interact badly with #591.

billhollings:
Thanks for testing.

>I can see, though, that this is going to interact badly with #591.

I'll merge your outstanding PR's in first and then reconcile this PR with the updated state before merging this one.

PR #590, opened 2019-05-03
Use the new memory barrier methods for pipeline barrier commands.

cdavis5e:
These two new methods introduced in Metal 2.1 (macOS 10.14) are more
granular, providing more control over what exactly must be sync'd. They
also work on buffer memory, meaning that buffer dependencies work
properly here.

### DIFF
--- a/MoltenVK/MoltenVK/Commands/MVKCmdPipeline.mm
+++ b/MoltenVK/MoltenVK/Commands/MVKCmdPipeline.mm
@@ -65,7 +65,7 @@ void MVKCmdPipelineBarrier::encode(MVKCommandEncoder* cmdEncoder) {
 
 #if MVK_MACOS
     // Calls below invoke MTLBlitCommandEncoder so must apply this first
-	if ( [cmdEncoder->_mtlRenderEncoder respondsToSelector: @selector(memoryBarrierWithScope:afterStages:beforeStages:)] ) {
+	if ( _device->_pMetalFeatures->memoryBarriers ) {
 		MTLRenderStages srcStages = mvkMTLRenderStagesFromVkPipelineStageFlags(_srcStageMask);
 		MTLRenderStages dstStages = mvkMTLRenderStagesFromVkPipelineStageFlags(_dstStageMask);
 		for (auto& mb : _memoryBarriers) {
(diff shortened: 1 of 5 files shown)


billhollings:
Travis CI is hitting a code error:

```
/Users/travis/build/KhronosGroup/MoltenVK/MoltenVK/MoltenVK/Vulkan/mvk_datatypes.mm
'MTLBarrierScopeRenderTargets' is unavailable: not available on iOS
                mtlScope |= MTLBarrierScopeRenderTargets;

'MTLBarrierScopeRenderTargets' has been explicitly marked unavailable here
    MTLBarrierScopeRenderTargets API_AVAILABLE(macos(10.14)) API_UNAVAILABLE(ios) = 1 << 2,
```

MoltenVK/MoltenVK/Commands/MVKCmdPipeline.mm
```diff
@@ -65,8 +65,36 @@
 
 #if MVK_MACOS
     // Calls below invoke MTLBlitCommandEncoder so must apply this first
-	if ( !(_memoryBarriers.empty() && _imageMemoryBarriers.empty()) ) {
-		[cmdEncoder->_mtlRenderEncoder textureBarrier];
+	if ( [cmdEncoder->_mtlRenderEncoder respondsToSelector: @selector(memoryBarrierWithScope:afterStages:beforeStages:)] ) {
```
billhollings: I recognize that this is a very localized test...but I would prefer that any Metal feature availability be enabled in `MVKPhysicalDeviceMetalFeatures` in `MVKDevice`...so that we have a single public record of available Metal features.

Can you add a `memoryBarrier` (or equivalent) member to `MVKPhysicalDeviceMetalFeatures`, and test for it here, please?

You don't need to increment `VK_MVK_MOLTENVK_SPEC_VERSION` beyond the `20` it became in PR #591.

Thanks.

(I recognize that we do have some `respondToSelector:` tests in the Metal class extensions...but is localized to the class itself).

billhollings: Thanks!

billhollings:
Another Travis issue:

```
/Users/travis/build/KhronosGroup/MoltenVK/MoltenVK/MoltenVK/Commands/MVKCmdPipeline.mm:68:
error
use of undeclared identifier '_device'
        if ( _device->_pMetalFeatures->memoryBarriers ) {
```

I think you'll need to use `getDevice()` here.

PR #591, opened 2019-05-03
MVKSemaphore: Use MTLEvent for device-side synchronization.

### LINKED ISSUE #438: Implement Vulkan semaphores via MTLEvent or MTLFence if supported on the platform.
Currently Vulkan semaphores are implemented in MoltenVK on the CPU...because Metal originally did not support any form of GPU-GPU sync.

Now that MTLFence is available...investigate whether they can be used to implement Vulkan semaphore GPU-GPU sync functionality.

cdavis5e:
While `MTLFence` could be used for synchronization within a single
queue, it doesn't prevent execution across queues. For this, an
`MTLEvent` is required. Yes, this requires Metal 2.1 (macOS 10.14, iOS
12). Older versions will continue to use the old, CPU-based
implementation.

According to the Vulkan spec, semaphores must either be already signaled
or have a signal operation in progress before waiting for them. This
implementation increments the `MTLEvent`'s expected value after encoding
a wait operation, which requires this assumption to hold.

One thorny bit is acquiring a swapchain image when it is not already in
use. In this case, any semaphore provided must be signaled right away.
To accomplish this, a command buffer is executed which performs the
signal operation on the device. Alternatively, we could use an
`MTLSharedEvent`, which would allow us to signal the event on the host.
But, this could have performance implications that may not be
acceptable, just to handle this one case.

Fixes #438.

### DIFF
--- a/MoltenVK/MoltenVK/API/vk_mvk_moltenvk.h
+++ b/MoltenVK/MoltenVK/API/vk_mvk_moltenvk.h
@@ -55,7 +55,7 @@ typedef unsigned long MTLLanguageVersion;
 #define MVK_MAKE_VERSION(major, minor, patch)    (((major) * 10000) + ((minor) * 100) + (patch))
 #define MVK_VERSION     MVK_MAKE_VERSION(MVK_VERSION_MAJOR, MVK_VERSION_MINOR, MVK_VERSION_PATCH)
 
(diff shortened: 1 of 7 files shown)


billhollings:
Great work! Thanks for taking care of this. It's been on the TODO list a long time...and it's nice to see MoltenVK finally moving towards proper GPU synchronization!

PR #596, opened 2019-05-09
Allow mvkMTLRenderStagesFromVkPipelineStageFlags() to map to all Vulkan stages, by indicating whether the pipeline barrier should come before or after the stages.

billhollings:
@cdavis5e 

`vkCmdPipelineBarrier()` can occur inside or outside a render pass I came across an app whose `vkCmdPipelineBarrier()` call included a source pipeline barrier for `VK_PIPELINE_STAGE_HOST_BIT`...and it caused `memoryBarrierWithScope:afterStages:beforeStages:` to barf.

It looks like the app might include any of the other non-renderpass pipeline stages as well (or even `VK_PIPELINE_STAGE_ALL_GRAPHICS_BIT`). I've modified `mvkMTLRenderStagesFromVkPipelineStageFlags()` to handle these by taking into consideration whether the barrier should be placed before or after the two supported Metal render stages (note the change to using `mvkAreOnlyAnyFlagsEnabled()` instead of `mvkIsAnyFlagEnabled`).

Metal also doesn't allow for multiple stages to be specified (`MTLRenderStageVertex | MTLRenderStageFragment)`...so I've forced one or the other.

I've also included a test for the availability of `cmdEncoder->_mtlRenderEncoder`, to skip that processing outside a render pass (I think Objective-C's do nothing on a method call to a nil object behaviour was saving a crash on this code outside a renderpass).

Can you just have a quick look at my logic to make sure this makes sense, please? I'm not at all certain of it. Perhaps we need more robust logic about whether to call `memoryBarrierWithScope:afterStages:beforeStages:` at all.

BTW...we should probably also look at making use of `MTLComputeCommandEncoder memoryBarrierWithScope:` within `vkCmdPipelineBarrier()` if a `cmdEncoder->_computePipelineState` is available (and in all OS's)...but I've left that for now until we discuss it further.

### DIFF
--- a/MoltenVK/MoltenVK/API/mvk_datatypes.h
+++ b/MoltenVK/MoltenVK/API/mvk_datatypes.h
@@ -402,8 +402,11 @@ MTLWinding mvkMTLWindingFromSpvExecutionMode(uint32_t spvMode);
 /** Returns the Metal MTLTessellationPartitionMode corresponding to the specified SPIR-V spv::ExecutionMode. */
 MTLTessellationPartitionMode mvkMTLTessellationPartitionModeFromSpvExecutionMode(uint32_t spvMode);
 
-/** Returns the combination of Metal MTLRenderStage bits corresponding to the specified Vulkan VkPiplineStageFlags. */
-MTLRenderStages mvkMTLRenderStagesFromVkPipelineStageFlags(VkPipelineStageFlags vkStages);
+/**
+ * Returns the combination of Metal MTLRenderStage bits corresponding to the specified Vulkan VkPiplineStageFlags,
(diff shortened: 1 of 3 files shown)


mbarriault:
I will 👍 the idea of discussing handling compute encoders in pipeline barriers, I use them rather extensively.

cdavis5e:
> I will 👍 the idea of discussing handling compute encoders in pipeline barriers, I use them rather extensively.

That's only necessary if you set it to execute concurrently, which we don't yet. That's one of the things I want to investigate doing.

cdavis5e:
Patch LGTM.

billhollings:
I've added Issue #597 to track the compute pipeline barriers.

PR #598, opened 2019-05-10
texture don't have coherent memory on mac os

mmaldacker:
Since texture can't be coherent on macOS, they need to be synchronized for host read. This PR fixes this.

### DIFF
--- a/MoltenVK/MoltenVK/GPUObjects/MVKImage.h
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKImage.h
@@ -195,7 +195,12 @@ public:
 	/** Returns the Metal CPU cache mode used by this image. */
 	inline MTLCPUCacheMode getMTLCPUCacheMode() { return _deviceMemory->getMTLCPUCacheMode(); }
 
-	
+	/** 
+	 * Returns whether the memory is automatically coherent between device and host. 
+	 * On macOS, this always returns false because textures cannot use Shared storage mode.
+	 */
+	bool isMemoryHostCoherent();
+
 #pragma mark Construction
 
 	MVKImage(MVKDevice* device, const VkImageCreateInfo* pCreateInfo);
--- a/MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
@@ -429,6 +429,8 @@ MTLTextureDescriptor* MVKImage::getMTLTextureDescriptor() {
 }
 
 MTLStorageMode MVKImage::getMTLStorageMode() {
+    if ( !_deviceMemory ) return MTLStorageModePrivate;
+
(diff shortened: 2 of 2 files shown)


MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
```diff
@@ -440,6 +440,10 @@
     return stgMode;
 }
 
+bool MVKImage::isMemoryHostCoherent() {
+    return (getMTLStorageMode() == MTLStorageModeShared);
```
billhollings: The call to `getMTLStorageMode()` will crash if `_deviceMemory` is not yet set for the image.

Can you either add a test for `_deviceMemory` in this method, similar to the overridden `MTLResource::isMemoryHostCoherent()`...or perhaps better...fix `MVKImage::getMTLStorageMode()` to handle a missing `_deviceMemory` (maybe defaulting to `MTLStorageModePrivate` if no `_deviceMemory`?

MoltenVK/MoltenVK/GPUObjects/MVKImage.h
```diff
@@ -195,7 +195,12 @@ class MVKImage : public MVKResource {
 	/** Returns the Metal CPU cache mode used by this image. */
 	inline MTLCPUCacheMode getMTLCPUCacheMode() { return _deviceMemory->getMTLCPUCacheMode(); }
 
-	
+	/** 
+	 * Returns whether the memory is automatically coherent between device and host. 
+	 * On macOS, this always returns false because textures cannot use Shared storage mode.
+	 */
+	bool isMemoryHostCoherent();
```
cdavis5e: Hey wait, doesn't MVKResource already have this method? I find it curious that adding this method here, which hides the method in the base, and doing nothing else, fixes your problem.

billhollings: @cdavis This is a slight variation in that calling `MVKImage:: getMTLStorageMode()` deals with swapping out `MTLStorageModeShared` on macOS.

cdavis5e: Ah, I see. Thanks for clearing that up.

MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
mmaldacker: no problem!

mmaldacker:
> Thanks very much for your submission! It's much appreciated!
> 
> Just a small fix requested to the use of `getMTLStorageMode()`.
> 
> I'm curious...what were the conditions where the code before this change was causing you issues?

Actually, I made a PR over a year ago with the same fix (see [PR#42](https://github.com/KhronosGroup/MoltenVK/pull/42)), but it looks like it was accidentally removed later on.

Issue #584, opened 2019-04-29
Vulkan SDK 1.1.106.0 - Segmentation fault from vkDestroySwapchainKHR


zstuartp:
I encounter a segmentation fault when destroying a swap chain using `vkDestroySwapchainKHR`.  This issue does not seem to occur in SDK versions 1.1.101 and prior.

Stack trace:
<pre>
<code>
* thread #1, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0x18)
  * frame #0: 0x00007fff707eb69d libobjc.A.dylib`objc_msgSend + 29
    frame #1: 0x0000000107075d59 libMoltenVK.dylib`MVKSwapchain::~MVKSwapchain() + 105
    frame #2: 0x0000000107075dae libMoltenVK.dylib`MVKSwapchain::~MVKSwapchain() + 14
</code>
</pre>

Hardware:
2015 Macbook Pro, 13", i7, 16GB

Edit:
macOS version: 10.14.4
Vulkan context: GLFW 3.3

cdavis5e:
I strongly suspect this is my fault, and has to do with #557.

Does [this change](https://github.com/KhronosGroup/MoltenVK/files/3128462/fix-swapchain-crash.diff.txt) help?

zstuartp:
> 
> 
> I strongly suspect this is my fault, and has to do with #557.
> 
> Does [this change](https://github.com/KhronosGroup/MoltenVK/files/3128462/fix-swapchain-crash.diff.txt) help?

Thank you for your response. I'll see if this fixes my issue when I get a chance later today.

zstuartp:
> I strongly suspect this is my fault, and has to do with #557.
> 
> Does [this change](https://github.com/KhronosGroup/MoltenVK/files/3128462/fix-swapchain-crash.diff.txt) help?

This patch does not appear to fix my issue.

mbarriault:
Are you running 10.14.4? I see a crash myself occasionally on exit when it tries to release the CAMetalLayer, and I sometimes see the same crash on work app (Qt + OpenGL). I've found at least one other project reporting it but I can't find the link anymore.

zstuartp:
> Are you running 10.14.4? I see a crash myself occasionally on exit when it tries to release the CAMetalLayer, and I sometimes see the same crash on work app (Qt + OpenGL). I've found at least one other project reporting it but I can't find the link anymore.

[...]

PR #606, opened 2019-05-20
Fix sporadic crash on vkDestroySwapchainKHR().

billhollings:
`MVKSwapchain` & `MVKSurface` init `_layerObserver` to nil.

Fixes issue #584.

### DIFF
--- a/Docs/Whats_New.md
+++ b/Docs/Whats_New.md
@@ -32,6 +32,7 @@ Released TBD
 	- Fix zero local threadgroup size in indirect tessellated rendering.
 - Fix crash when clearing attachments using layered rendering on older macOS devices.
 - Fixes to Metal renderpass layered rendering settings.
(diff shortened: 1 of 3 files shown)


cdavis5e:
Thanks for cleaning up another of my messes. (I thought I did that already... Guess not.)

billhollings:
> Thanks for cleaning up another of my messes. (I thought I did that already... Guess not.)

No worries at all. You actually nailed this fix immediately in the issue thread...and I appreciate that! 

The feedback (or lack thereof) from your initial suggestion was definitely a red-herring for most cases. I expect your fix should actually clear up most of the use cases reported in that issue. It will be interesting to see if there are any outlier cases that still cause a problem. 

If so...a possible next step might be to put a lock around the management of `_layerObserver`...in case there is a race condition between the notification callback and the swapchain destruction...since a change in the window might trigger both the system callback...and the app to destroy the swap-chain.

PR #611, opened 2019-05-27
Avoid calling stringWithUTF8String with NULL.

jozefkucia:
It raises an exception when bytes is NULL.

### DIFF
--- a/MoltenVK/MoltenVK/GPUObjects/MVKVulkanAPIObject.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKVulkanAPIObject.mm
@@ -56,7 +56,7 @@ bool MVKVulkanAPIObject::markDestroyed() {
 
 VkResult MVKVulkanAPIObject::setDebugName(const char* pObjectName) {
 	[_debugName release];
-	_debugName = [[NSString stringWithUTF8String: pObjectName] retain];		// retained
+	_debugName = pObjectName ? [[NSString stringWithUTF8String: pObjectName] retain] : nil;
 	propogateDebugName();
 	return VK_SUCCESS;
 }


MoltenVK/MoltenVK/GPUObjects/MVKVulkanAPIObject.mm
```diff
@@ -56,7 +56,7 @@
 
 VkResult MVKVulkanAPIObject::setDebugName(const char* pObjectName) {
 	[_debugName release];
-	_debugName = [[NSString stringWithUTF8String: pObjectName] retain];		// retained
+	_debugName = pObjectName ? [[NSString stringWithUTF8String: pObjectName] retain] : [@"" retain];
```
billhollings: Is there a reason why you added an empty default string? The design intention was to leave `_debugName` as `nil` if it is not set, for potential performance optimizations...and to allow Metal's default names to continue to appear. Note the use of `setLabelIfNotNil()` everywhere. So the appropriate behaviour here should be `? ... : nil`.

jozefkucia: You're right. It makes more sense to simply set it to `nil`.

jozefkucia:
`setDebugName()` is called with NULL pointer for dedicated allocations in `MVKBuffer::propogateDebugName()`. `VK_EXT_debug_utils` is other reason to allow `NULL` (if you would like to support the extension in future):

> If pObjectName is not NULL, pObjectName must be a null-terminated UTF-8 string

billhollings:
> `setDebugName()` is called with NULL pointer for dedicated allocations in `MVKBuffer::propogateDebugName()`. `VK_EXT_debug_utils` is other reason to allow `NULL` (if you would like to support the extension in future):

Both very good points. Yes...`MVKBuffer::propogateDebugName()` is called from `bindDeviceMemory()`...so we need to accommodate that. And yes...I'm starting to work on `VK_EXT_debug_utils` now. Thanks for pointing out that the spec for that allows `pObjectName` to be nil.

Rather than have you rework this PR...can I suggest that I address this separately as part of the initial work for `VK_EXT_debug_utils` on a separate PR? I'll fix this issue as a first step on that path.

billhollings:
@jozefkucia Thanks for identifying this. As I mentioned above, I've consolidated your change here along with a couple of other null guards into PR #616.

billhollings:
Replaced by #616.

PR #614, opened 2019-05-27
Added missing texelBufferTextureWidth setting in MVKComputePipeline::getMTLFunction.

### LINKED ISSUE #589: Large texel buffer addressing seems broken
Hi!

I am facing an issue with large texel buffers. By large I mean those with length exceeding Metal 1D texture size limit of 16384 texels and causing MVK to use a 2D texture for backing their storage. Below I will be referring to this commit changes in MVK: e959b9d9a6a013a751f1110b831ebf37a917c5e2.

Heres the setup:
1) I have a uint16 texel buffer. The buffer memory resevation is for up to ~117.2k elements. MVK alocates a 16384x8 1-channel texture for it, which is right.
2) then I have a compute shader. It does 191x61 invocations to fill the buffer. For debugging purposes I've stripped out the actual logic, and just write magic number 0xFFFF at each invocation. The write offset comes as a flattened thread index, like this:
```
[numthreads(NUM_THREADS_X, NUM_THREADS_Y, 1)]
void Main(uint3 globalIdx : SV_DispatchThreadID, uint3 localIdx : SV_GroupThreadID, uint3 groupIdx : SV_GroupID)
{
    uint localIdxFlattened = localIdx.x + localIdx.y * NUM_THREADS_X;
    BPerTileLightIndexBufferOut[tileIdxFlattened] = 0xFFFF;
}
```
As the result of this step I expect a line of 191x61=14651 `0xFFFF`-values (I'll call those "marked texels") laid out linearly from the origin of the texture. Which doesn't happen.

Instead, as per XCode graphics debugger, there are 3 lines of 4096 marked texels and one of 2363 marked texels. Those do sum up to expected 14651 marked texels in total, but memory layout is 2D. Okay, presumably I don't get how the things work, but at least there is this [constant](https://github.com/KhronosGroup/MoltenVK/blob/e959b9d9a6a013a751f1110b831ebf37a917c5e2/MoltenVKShaderConverter/MoltenVKSPIRVToMSLConverter/SPIRVToMSLConverter.h#L39) which seems to cause that layout.

3) then I have a debug pixel shader. I've divided my viewport into tiles so that it is entirely covered with 191x61 tiles. For each pixel I take its coords, convert them into a flat index and read the value from the buffer with that offset. I expect to get the magic `0xFFFF` for each pixel on the screen. I check the value and paint pixel red when theres a match, and leave it as is if not. I expect the screen to be entirely painted red, but insted it shows up like [this](https://imgur.com/BO652Bz).

So in the end, it seems like buffer writes are made with a different layout than is expected for reads.

ch45er:
Fixes #589.

### DIFF
--- a/MoltenVK/MoltenVK/GPUObjects/MVKPipeline.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKPipeline.mm
@@ -1196,6 +1196,7 @@ MVKMTLFunction MVKComputePipeline::getMTLFunction(const VkComputePipelineCreateI
 	shaderContext.options.entryPointName = pCreateInfo->stage.pName;
 	shaderContext.options.entryPointStage = spv::ExecutionModelGLCompute;
     shaderContext.options.mslVersion = _device->_pMetalFeatures->mslVersion;
+    shaderContext.options.texelBufferTextureWidth = _device->_pMetalFeatures->maxTextureDimension;
 	shaderContext.options.shouldSwizzleTextureSamples = _fullImageViewSwizzle;
 
     MVKPipelineLayout* layout = (MVKPipelineLayout*)pCreateInfo->layout;


billhollings:
Thanks for the submission!

Nice catch!

PR #640, opened 2019-06-11
Mark stencil reference value as dirty in MVKCmdClearAttachment command.

jozefkucia:
MVKCmdClearAttachment sets the stencil reference value to 0 and the stencil
reference value is not properly restored for subsequent draw calls.

I'm affraid there might be more similar bugs in MoltenVK.


jozefkucia:
I don't know the MoltenVK codebase very well. It's possible that this should be fixed differently.

billhollings:
@jozefkucia 

Good catch! 

Yes...you are probably right...there will be other polluted states...at least for depth and pipeline as well when clearing attachments. I'll have a look at the others.

billhollings:
PR #642 includes additional state reversions.

PR #642, opened 2019-06-13
 Updates for SDK 1.1.108 release.

### LINKED ISSUE #641: vkCmdBeginRenderPass with different depth/stencil load ops may not set clear value correctly
I've noticed a bug with a VkRenderPass that uses `VK_ATTACHMENT_LOAD_OP_CLEAR` for `stencilLoadOp` but `VK_ATTACHMENT_LOAD_OP_LOAD` for `loadOp`; the command buffer is prevented from setting non-default clear values to the MTLRenderPassStencilAttachmentDescriptor.

In MVKRenderPass.mm, `MVKRenderPassAttachment::populateMTLRenderPassAttachmentDescriptor`, there is this block:
```
...} else if ( isRenderingEntireAttachment && (subpass->_subpassIndex == _firstUseSubpassIdx) ) {
    VkAttachmentLoadOp loadOp = isStencil ? _info.stencilLoadOp : _info.loadOp;
    mtlAttDesc.loadAction = mvkMTLLoadActionFromVkAttachmentLoadOp(loadOp);
    willClear = (_info.loadOp == VK_ATTACHMENT_LOAD_OP_CLEAR);
} else...
```

Because `willClear` is being set from `_info.loadOp` instead of `loadOp`, `willClear` is incorrectly getting its value from the _depth_ load op instead of the stencil load op whenever the method is called for a stencil attachment, so if the stencil is set to clear and the depth is not, the stencil is unable to set a clear value other than 0 (the default).

Fix:  the line should be changed to
```
willClear = (loadOp == VK_ATTACHMENT_LOAD_OP_CLEAR);
```

billhollings:
- Updates for SDK 1.1.108 release.
- Fixes to clearing attachments.
- Report image layer limits for attachments in vkGetPhysicalDeviceImageFormatProperties().

Fixes #641.


billhollings:
Travis CI failing on latest `glslang` for some reason. Merging anyway.

PR #643, opened 2019-06-14
Sync Metal-3.0 with master

billhollings:


billhollings:
Build fails because Travis doesn't support Xcode 11. Merge anyway.

PR #665, opened 2019-07-15
Advertise the VK_EXT_shader_stencil_export extension.

cdavis5e:
This is supported by Mac GPU Family 2 starting on macOS 10.14, and Apple
GPU Family 5 starting on iOS 12.

Supporting this is a bit tricky. Because only some devices support this
extension, we now have to keep track of supported device extensions per
device.


billhollings:
Ah, cool! I was thinking as well about the need to introduce a `_supportedExtensions` per physical device...and had it on my ToDo list. Happy to see you bumped into it too and got it done!

PR #666, opened 2019-07-16
Support the VK_EXT_swapchain_colorspace extension.

cdavis5e:
Only some of the color spaces provided by this extension are supported.
macOS 10.12 supports a few more. macOS 10.14 (at least, according to the
*10.15* SDK) supports even more. (But that needs a change to the
Metal-3.0 branch.)

I've chosen to group by color space. That way, programs will find the
all the supported formats early on. Programs that are interested in
the color space can keep looking.


billhollings:
@cdavis5e 

Travis caught the following:

```
/Users/travis/build/KhronosGroup/MoltenVK/MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm:467:6: 
error
: 
value of type 'const MVKExtension' is not contextually convertible to 'bool'
        if (getInstance()->_enabledExtensions.vk_EXT_swapchain_colorspace) {
            ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
1 error generated.
```

Looks like it's missing a `.enabled` on the end.

PR #676, opened 2019-07-17
 Reduce memory leaks when autorelease pools are not available.

billhollings:
Remove use of `autorelease`.
Rename `getXXX()` functions that returned autoreleased objects to `newXXX()` that don't.
Calling functions now responsible for releasing returned objects from these functions.
Move creation and tracking of texture->texture views for copying to `MVKImage`.
`MVKCmdCopyImage` retrieve texture view from `MVKImage`.
`MVKMTLFunction` responsible for releasing its `MTLFunction`.
Add dev notes to `README.md`.
Reduce creation of autoreleased instances of `NSString`, `NSArray` & `NSDictionary`.

Ensure remaining are covered by deliberate autorelease pools.

Cube demo use staging buffers on macOS because host-coherent texture memory unavailable.


galad87:
There is an annotation in clang to tell the compiler (and the static analyzer) that a function/method returns a retained instance: NS_RETURNS_RETAINED , maybe it would be nice to use it.

billhollings:
> There is an annotation in clang to tell the compiler (and the static analyzer) that a function/method returns a retained instance: NS_RETURNS_RETAINED , maybe it would be nice to use it.

Thanks! I'll look into it.

billhollings:
Note to contributors: Future updates should avoid using `autorelease` (direct or implied) in future updates. `README.md` includes guidelines.

PR #677, opened 2019-07-17
Add support for uncached, host-coherent memory.

cdavis5e:
In Metal, this roughly corresponds to `MTLStorageModeShared` and
`MTLCPUCacheModeWriteCombined`. Vulkan requires that host-visible memory
must have either or both of the `HOST_COHERENT` and `HOST_CACHED` bits
set, so no write-combined managed here.

Shared write-combined memory is placed before regular shared memory, as
it must be by the Vulkan rules for sorting memory types; but after
managed memory. Apple warns that write-combined memory "can have
surprising performance pitfalls," so managed memory is preferred.

I notice that `mvkMTLCPUCacheModeFromVkMemoryPropertyFlags()`
already exists, but it just returned `MTLCPUCacheModeDefaultCache`. Is
there a reason for that? Was this supported once, but the support was
removed? (I've marked this draft, because I don't want this merged
until those questions are answered.)

### DIFF
--- a/MoltenVK/MoltenVK/API/mvk_datatypes.h
+++ b/MoltenVK/MoltenVK/API/mvk_datatypes.h
@@ -447,16 +447,19 @@ static inline VkExtent3D mvkVkExtent3DFromMTLSize(MTLSize mtlSize) {
 #pragma mark Memory options
 
 /** Macro indicating the Vulkan memory type bits corresponding to Metal private memory (not host visible). */
-#define MVK_VK_MEMORY_TYPE_METAL_PRIVATE	(VK_MEMORY_PROPERTY_DEVICE_LOCAL_BIT)
+#define MVK_VK_MEMORY_TYPE_METAL_PRIVATE				(VK_MEMORY_PROPERTY_DEVICE_LOCAL_BIT)
 
 /** Macro indicating the Vulkan memory type bits corresponding to Metal managed memory (host visible and non-coherent). */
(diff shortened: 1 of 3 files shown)


billhollings:
> I notice that `mvkMTLCPUCacheModeFromVkMemoryPropertyFlags()`
> already exists, but it just returned `MTLCPUCacheModeDefaultCache`. Is
> there a reason for that?

I never bothered moving beyond `MTLCPUCacheModeDefaultCache` because Apple's description of `MTLCPUCacheModeWriteCombined` didn't seem to map to a specific Vulkan combination:

>Write-combined memory is optimized for resources that the CPU writes into, but never reads. 

And...as you allude to...Apple's own docs don't do a good job of selling this option:

>Applications should use write combining only if writing to normally cached buffers is known to cause performance issues due to cache pollution, as write-combined memory can have surprising performance pitfalls.

There didn't seem to be an upside, so I let it be.

I guess the question is...what benefit will emerge in adding this option? If this option performs poorly in most cases, it would be disappointing to find apps unintentionally triggering poor performance if they assume `VK_MEMORY_PROPERTY_HOST_COHERENT_BIT` is enough, and don't explicitly include `VK_MEMORY_PROPERTY_HOST_CACHED_BIT`.

cdavis5e:
This was prompted by [a change to DXVK](https://github.com/doitsujin/dxvk/commit/f272071d8df5ebd2245f97e8942143687235e3e5), but looking into it, it's apparently intended to work around some Intel Atom GPUs not supporting coherent cached memory. I suspect that right now we really don't want to add this, at least until DXVK supports non-coherent (`Managed`) memory. And even then, we'd probably want to do performance testing to see if this really makes writing to staging buffers faster.

PR #678, opened 2019-07-17
More format feature fixes.

cdavis5e:
Advertise linear filtering of linear textures on macOS. I don't see any
reason this was disabled, and I don't know why I didn't turn it on in
PR #391.

Advertise that texel buffers of `VK_FORMAT_A2R10G10B10_UNORM_PACK32` are
supported. This was another oversight of mine in #391.

*Don't* advertise that sRGB formats can be used as vertex attributes.
I'm not sure what that would even mean--I'm guessing that the
implementation performs sRGB->linear conversion on the attribute before
loading it? But we don't do that and neither does Metal.


cdavis5e:
Xcode crashed.

PR #679, opened 2019-07-19
Enable features added by recently added extensions.

cdavis5e:
Keep track of the enabled status of
`VK_KHR_uniform_buffer_standard_layout`, `VK_EXT_scalar_block_layout`,
and `VK_EXT_texel_buffer_alignment`. These were unintentionally omitted
from #667, #671, and #672.

I've put `VK_EXT_scalar_block_layout` in its own commit, in case it needs
to be reverted because of the regression I caused (cf.
KhronosGroup/SPIRV-Cross#1088).

There are too many places to change to add a new feature struct. I need to
find a way to refactor this the way I did with extensions...


billhollings:
@cdavis5e Travis complaining of compile error.

PR #700, opened 2019-08-05
Add more supported color spaces as of the latest OS versions.

cdavis5e:
Support `VK_EXT_swapchain_colorspace` on iOS as well. The 13.0 headers
add the `colorspace` property needed for this.


billhollings:
@cdavis5e Looks like a colorspace library linking breakdown. Maybe need to add additional framework(s) to dylib build scripts?

PR #708, opened 2019-08-08
Add support for VkEvent.

billhollings:
- Add `MVKEvent` class. `MVKEventNative` subclass uses native `MTLEvent`. `MVKEventEmulated`
subclass uses emulation using CPU blocking and `MTLCommandBuffer` completion handling.
- `MVKConfiguration::synchronousQueueSubmits` now disabled by default if `MTLEvents` are not supported.
- Document updated use of `MVK_ALLOW_METAL_EVENTS` and `MVK_CONFIG_SYNCHRONOUS_QUEUE_SUBMITS`
environment variables and `synchronousQueueSubmits` config setting, in `vk_mvk_moltenvk.h`.

### DIFF
--- a/MoltenVK/MoltenVK/Commands/MVKCmdPipeline.mm
+++ b/MoltenVK/MoltenVK/Commands/MVKCmdPipeline.mm
@@ -347,6 +347,54 @@ MVKCmdPushDescriptorSetWithTemplate::~MVKCmdPushDescriptorSetWithTemplate() {
 
 
 #pragma mark -
+#pragma mark MVKCmdSetResetEvent
+
+void MVKCmdSetResetEvent::setContent(VkEvent event, VkPipelineStageFlags stageMask, bool status) {
+	_mvkEvent = (MVKEvent*)event;
+	_status = status;
+}
+
+void MVKCmdSetResetEvent::encode(MVKCommandEncoder* cmdEncoder) {
+	cmdEncoder->signalEvent(_mvkEvent, _status);
+}
+
+MVKCmdSetResetEvent::MVKCmdSetResetEvent(MVKCommandTypePool<MVKCmdSetResetEvent>* pool)
+	: MVKCommand::MVKCommand((MVKCommandTypePool<MVKCommand>*)pool) {}
+
+
+#pragma mark -
+#pragma mark MVKCmdWaitEvents
+
+void MVKCmdWaitEvents::setContent(uint32_t eventCount,
+							   const VkEvent* pEvents,
+							   VkPipelineStageFlags srcStageMask,
+							   VkPipelineStageFlags dstStageMask,
+							   uint32_t memoryBarrierCount,
+							   const VkMemoryBarrier* pMemoryBarriers,
+							   uint32_t bufferMemoryBarrierCount,
+							   const VkBufferMemoryBarrier* pBufferMemoryBarriers,
+							   uint32_t imageMemoryBarrierCount,
+							   const VkImageMemoryBarrier* pImageMemoryBarriers) {
+
+	_mvkEvents.clear();	// Clear for reuse
+	_mvkEvents.reserve(eventCount);
+	for (uint32_t i = 0; i < eventCount; i++) {
+		_mvkEvents.push_back((MVKEvent*)pEvents[i]);
+	}
+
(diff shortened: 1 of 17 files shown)


MoltenVK/MoltenVK/Commands/MVKCmdPipeline.mm
```diff
@@ -346,6 +346,54 @@
...
+
+#pragma mark -
+#pragma mark MVKCmdWaitEvents
+
+void MVKCmdWaitEvents::setContent(uint32_t eventCount,
+							   const VkEvent* pEvents,
+							   VkPipelineStageFlags srcStageMask,
+							   VkPipelineStageFlags dstStageMask,
+							   uint32_t memoryBarrierCount,
+							   const VkMemoryBarrier* pMemoryBarriers,
```
cdavis5e: You don't seem to be doing anything with these.

billhollings: It wasn't clear what of those we could make use of in Metal's event environment. 

Do you have suggestions?

I was wondering if CTS would highlight differences in handling here...but so far all the CTS event tests pass.

cdavis5e: I was thinking of inserting a barrier, as `vkCmdPipelineBarrier()` does, either before or after the wait.

MoltenVK/MoltenVK/GPUObjects/MVKSync.mm
```diff
@@ -155,6 +160,76 @@
...
+
+void MVKEventNative::signal(bool status) {
+	if (isSet() != status) {
+		_mtlEvent.signaledValue += 1;
+	}
+}
+
+void MVKEventNative::encodeSignal(id<MTLCommandBuffer> mtlCmdBuff, bool status) {
+	if (isSet() != status) {
+		[mtlCmdBuff encodeSignalEvent: _mtlEvent value: _mtlEvent.signaledValue + 1];
```
cdavis5e: Another problem: these methods of `MTLCommandBuffer` may only be called when no command encoders are active. (Try running with the Metal validation layer on.) You'll have to end the current encoder before calling these methods. And if we're in the middle of a render pass, you'll have to start it back up again.

billhollings: > Another problem: these methods of `MTLCommandBuffer` may only be called when no command encoders are active. (Try running with the Metal validation layer on.) You'll have to end the current encoder before calling these methods. And if we're in the middle of a render pass, you'll have to start it back up again.

Thanks. Yup...this is covered in `MVKCommandEncoder::signalEvent()`. And yup...I have been running CTS with Metal validation on. Vulkan `vkCmdRe/SetEvent()` singling is outside render pass scope (matching the Metal requirement you mention).

MoltenVK/MoltenVK/Commands/MVKCmdPipeline.mm
billhollings: Okay. I'll have a closer look at the barrier relationship. I'm not clear on the interaction either.

MoltenVK/MoltenVK/GPUObjects/MVKSync.mm
cdavis5e: I stand corrected. You did indeed end the current encoder there. Also, apparently, `vkCmd{Set,Reset}Event()` cannot be executed inside a render pass.

I also notice that the restriction documented for `-[MTLCommandBuffer encodeSignalEvent:value:]` is not present in the docs at all for `-[MTLCommandBuffer encodeWaitForEvent:value:]`. So I guess I was worried for nothing.

billhollings: Yeah...looks like in both Vulkan & Metal...the event wait can occur within a render pass...but the signalling cannot.

billhollings:
I'm going to pull this in now. The basics are working in CTS...and I have some unrelated fixes to make in order to get more of the CTS event tests working.

As I get them running...I'll see if those further CTS tests reveal anything about the requirements and relationships between events and barriers as they apply to Metal.

PR #709, opened 2019-08-08
Don't prefer dedicated allocations for buffer memory.

cdavis5e:
In these cases, we just use the underlying `MTLBuffer` in the
`MVKDeviceMemory` anyway. Dedicated allocations don't really buy us
anything here.


billhollings:
@cdavis5e Travis choking on compilation at `MVKImage.mm:195:69`.

PR #733, opened 2019-09-06
Avoid MTLDevice minimumLinearTextureAlignmentForPixelFormat failed assertion

aerofly:
On iOS 13 and iPad OS 13 MTLDevice minimumLinearTextureAlignmentForPixelFormat will cause a failed assertion if parameter is MTLPixelFormatInvalid.

### DIFF
--- a/MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
@@ -1159,7 +1159,7 @@ void MVKPhysicalDevice::initProperties() {
         uint32_t maxStorage = 0, maxUniform = 0;
         bool singleTexelStorage = true, singleTexelUniform = true;
         mvkEnumerateSupportedFormats({0, 0, VK_FORMAT_FEATURE_UNIFORM_TEXEL_BUFFER_BIT | VK_FORMAT_FEATURE_STORAGE_TEXEL_BUFFER_BIT}, true, [&](VkFormat vk) {
-            // MTLDevice minimumLinearTextureAlignmentForPixelFormat with MTLPixelFormatInvalid will cause as a failed assertion on iOS 13 and iPad OS 13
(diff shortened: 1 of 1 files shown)


MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
```diff
@@ -1159,6 +1159,10 @@
         uint32_t maxStorage = 0, maxUniform = 0;
         bool singleTexelStorage = true, singleTexelUniform = true;
         mvkEnumerateSupportedFormats({0, 0, VK_FORMAT_FEATURE_UNIFORM_TEXEL_BUFFER_BIT | VK_FORMAT_FEATURE_STORAGE_TEXEL_BUFFER_BIT}, true, [&](VkFormat vk) {
+            // MTLDevice minimumLinearTextureAlignmentForPixelFormat with MTLPixelFormatInvalid will cause as a failed assertion on iOS 13.x and iPad OS 13.x
+            if ( vk == VK_FORMAT_UNDEFINED )
+              return false;
+
```
billhollings: MoltenVK doesn't use hanging if and for statements. I'll pull in anyway and fix it here.

billhollings:
Thanks for the submission!

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
aerofly: Sorry for that, I could have added this check at the call site, feel free to modify so it follows you standards. We just had a few customers with iOS Beta 13 and they reported a crash on application start. This line fixed it. It seems like Metal has extra assertions on iOS 13 when used with wrong parameters.

billhollings: No worries. Yeah...Apple often adds new validation operations on each release of Metal. Thanks for catching and fixing this.

Issue #755, opened 2019-10-12
Cannot use glslang compiler linking capability with the GLSLToSPIRVConverter


w103csh:
Hello,

I have been lugging around a couple of really small changes to the [GLSLToSPIRVConverter](https://github.com/KhronosGroup/MoltenVK/blob/master/MoltenVKShaderConverter/MoltenVKGLSLToSPIRVConverter/GLSLToSPIRVConverter.h) in a fork of this repo, and I thought I would reach out and see if you all are interested in adding something similar.

In the renderer I am developing I use the glslang glsl to spirv conversion quite a bit to do hot swapping of shaders, and other things at runtime. I have a bunch of smaller glsl shader files that I link together based on dynamic runtime circumstances during shader program compilation. This all works great when I am directly using the glslang compiler on windows, but I cannot do this using the current state of the [GLSLToSPIRVConverter](https://github.com/KhronosGroup/MoltenVK/blob/master/MoltenVKShaderConverter/MoltenVKGLSLToSPIRVConverter/GLSLToSPIRVConverter.h) here in MoltenVK.

I am just proposing that instead of being able to set one one string:

>`void setGLSL(const char* glslSrc, size_t length) { _glsl.assign(glslSrc, length); }`
>`...`
>`protected:`
>`...`
>`std::string _glsl;`

You could instead set a list of shader text strings:

>`void setGLSL(const std::vector<std::string>& glslSrcs) {`
>`_glsls.clear();`
>`_glsls.insert(_glsls.end(), glslSrcs.begin(), glslSrcs.end());`
>`}`
>`...`
>`protected:`
>`...`
>`std::vector<std::string> _glsls;`

The `convert` function would also need to be updated to make this work. I have a fork with these changes implemented [here](https://github.com/w103csh/MoltenVK). The specific commits are [here](https://github.com/w103csh/MoltenVK/commit/5ff86565237e9a44c1f0daa988fb94f472b20766), and [here](https://github.com/w103csh/MoltenVK/commit/f1dd1c57e311093c86e5bde627935d1cfab9b7d5).

I did not make a pull request because some of the formatting is wonky in those commits, I included `<memory>` because I was really into unique_ptr's at the time, and I am not really sure what your process is.  I could potentially clean up those commits if needed. Let me know what you think.

billhollings:
@w103csh 

Thanks for your suggestion...and the PoC work already.

[...]

PR #757, opened 2019-10-16
Add shader linking capability to GLSLToSPIRVConverter

w103csh:
This is a PR for this [issue](https://github.com/KhronosGroup/MoltenVK/issues/755). Simple change to make it so that you can use the linking feature in [glslang ](https://github.com/KhronosGroup/glslang)

Thanks

### DIFF
--- a/MoltenVKShaderConverter/MoltenVKGLSLToSPIRVConverter/GLSLToSPIRVConverter.cpp
+++ b/MoltenVKShaderConverter/MoltenVKGLSLToSPIRVConverter/GLSLToSPIRVConverter.cpp
@@ -70,8 +70,8 @@ MVK_PUBLIC_SYMBOL bool GLSLToSPIRVConverter::convert(MVKGLSLConversionShaderStag
 	configureGLSLCompilerResources(&glslCompilerResources);
 	std::vector<std::unique_ptr<glslang::TShader>> glslShaders;
 	const char *glslStrings[1];
-    glslang::TProgram glslProgram;
-    
+	glslang::TProgram glslProgram;
+
 	for (const auto& glsl : _glsls) {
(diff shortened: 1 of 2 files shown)


MoltenVKShaderConverter/MoltenVKGLSLToSPIRVConverter/GLSLToSPIRVConverter.cpp
```diff
@@ -125,8 +147,8 @@ bool GLSLToSPIRVConverter::validateSPIRV() {
 void GLSLToSPIRVConverter::logGLSL(const char* opDesc) {
 	_resultLog += opDesc;
 	_resultLog += " GLSL:\n";
-	_resultLog += _glsl;
-	_resultLog += "\nEnd GLSL\n\n";
+	for (const auto& glsl : _glsls) _resultLog += glsl + "\n";
```
billhollings: Can you just wrap this in `{}` to avoid the hanging-`for`, please?

w103csh: @billhollings  I wrapped a couple statements in brackets as you requested [here](https://github.com/w103csh/MoltenVK/commit/5c3b265888b02fd83dc81cfe9c322fab6295f654). I wasn't sure if you wanted the  statements inside the braces on a separate line or not. I saw some `if`'s in the function above with a similar style so I went with that. Let me know if its what you want.

billhollings: Yup. Looks great. And thanks for noticing and fixing the additional hanging-if statements.

PR #765, opened 2019-10-26
Fix tessellation when using resolve attachments

mbarriault:
I believe this also gets hit when using subpasses. Store action wasn't being set right when resolve action was set. I believe this will still fail on macOS GPU Family 1 v1, and iOS GPU Family 2 and earlier devices.


cdavis5e:
It looks like every device that supports tessellation also supports simultaneous store-and-resolve.

Issue #756, opened 2019-10-13
Resizing the window makes the app run very slow


DiligentGraphics:
After updating MacOS to Catalina and XCode to 11.1, I started to see the following issue: after resizing the window, applications start to run extremely slow (~1fps) and become unresponsive. They don't crash though and can exit properly. The cube demo from the SDK works fine, so I think it may be related to the DisplayLink that I am using in my applications.
Funny enough, before the update, resize in Vulkan mode worked fine, but resizing in OpenGL mode caused crashes. After the update, resizing in OpenGL mode works fine, but there is issue with Vulkan.
SDK version: 1.1.121.0. 
Maybe you have an insight into what could be wrong here?

QuantumDeveloper:
I am also observing almost the same issue, but in my case on Mojave it was also reproducible.

billhollings:
As mentioned...this doesn't occur on the Cube demo...which also uses DisplayLink.

I'm wondering what options we have to debug this.

Where is this occurring? In an available app?

Or is there a relatively simple app in which this occurs that we can use to replicate the situation here for debugging? If so...can you ZIP it up and post it somewhere in the cloud?

DiligentGraphics:
Here is the [link to the pre-built executable](https://www.dropbox.com/s/11gc1katwokycjm/Tutorial02_Cube.zip?dl=0)
The source code of the OSX application is [here](https://github.com/DiligentGraphics/DiligentEngine/tree/master/Common/NativeApp/Apple/Source/Classes/OSX)

billhollings:
@DiligentGraphics 

>Here is the link to the pre-built executable

I've tried running this...but it's failing to find libraries.

Where is the app expecting to find `libvulkan.1.dylib`, `libMoltenVK.dylib` & `MoltenVK_icd.json`?

DiligentGraphics:
https://github.com/DiligentGraphics/DiligentEngine/blob/master/README.md#configuring-vulkan-build-environment

DiligentGraphics:
I confirm the problem is fixed now. Thanks!

BTW, it would be great if there was a way to download pre-built libraries. It took forever to build dependencies...

PR #769, opened 2019-10-30
Fix severe FPS degradation caused by certain swapchain recreation situations.

billhollings:
Ensure swapchain clears all cached `CAMetalDrawables` when destroyed.

Fixes issue #756.


billhollings:
@QuantumDeveloper @jeremyk-lunarg 

This fix was made after the initial SDK cutoff. We have requested that it be included in the upcoming SDK, but  It will depend on whether the SDK process can fit it in at this point.

DiligentGraphics:
Ah, it will be a bummer if it does not make it into the next SDK.

billhollings:
@QuantumDeveloper @DiligentGraphics  @jeremyk-lunarg

LunarG has indicated that they should be able to fit this fix into the SDK release. 

I've re-released MoltenVK [v1.0.38](https://github.com/KhronosGroup/MoltenVK/releases/tag/v1.0.38) to include this in anticipation of that.

QuantumDeveloper:
@billhollings 
Great news! Thanks a lot.

Issue #780, opened 2019-11-18
Device memory leaks MTLHeap


ystreet:
Using SDK 1.1.126.0:

I have an application that currently creates and destroys images every frame (yes I know that is terrible) that has a memory leak growing on the order of all of the image contents.

I believe this was introduced by https://github.com/KhronosGroup/MoltenVK/pull/711 by a missing `[_mtlHeap release]` in MVKDeviceMemory's destructor.

Unfortunately I do not currently have the build and runtime setup to test out such a change locally.

billhollings:
@ystreet 

Thanks for reporting this and doing the detective work. PR #781 implements your suggestion. Please retest and close this issue if it fixes the problem for you.

ystreet:
The leak is gone now! Thanks.

PR #781, opened 2019-11-18
Fix MTLHeap memory leak in MVKDeviceMemory.

billhollings:
Fixes issue #780.


ystreet:
I think this is missing a `nil` check on `_mtlHeap` before `release` as the heap is not always created.  One case the heap is not created is for dedicated buffers/images.

cdavis5e:
> I think this is missing a `nil` check on `_mtlHeap` before `release` as the heap is not always created. One case the heap is not created is for dedicated buffers/images.

Shouldn't matter. In Objective-C, sending a message (== calling a method) to `nil` has no effect.

Issue #782, opened 2019-11-24
Request support VK_EXT_inline_uniform_block


mbarriault:
From the [spec](https://www.khronos.org/registry/vulkan/specs/1.1-extensions/html/vkspec.html#VK_EXT_inline_uniform_block), looks like it would be very handy for small descriptors (personally looking at using a storage buffer for dynamic lighting, and would like to use this for the light count). Suspect this should be fairly straightforward to implement via `set[Vertex/Fragment/Compute]Bytes`, which I know is used elsewhere in MoltenVK for various quantities.

mbarriault:
To head anyone off who might start this, think I might have this implemented. Just a few gaps to fill and to test.

PR #783, opened 2019-11-25
Support for EXT_inline_uniform_block

mbarriault:
Implements #782. 

I initially started implementing checking for `VkDescriptorPoolInlineUniformBlockCreateInfoEXT` in `VkDescriptorPoolCreateInfo`'s `pNext` chain, but that only serves to set additional allocation limits on descriptors that are already not honoured from the descriptor pool sizes, so seemed unnecessary and inconsistent to implement it at the moment.

Apologies for extra commits on tabbing. MVK isn't globally consistent but tried to make the additions consistent with the code around it.


billhollings:
Thanks!

Issue #772, opened 2019-10-31
Tessellation shader produces incorrect results


DiligentGraphics:
[Sample application](https://www.dropbox.com/s/5hqxnjvjlk51jpe/Tutorial08_Tessellation.zip?dl=0)

Tried with the latest release (1.0.38)

OpenGL:
![image](https://user-images.githubusercontent.com/15135271/67916683-bfc92a80-fb53-11e9-8c45-2634ef86b2ff.png)

Vulkan:
![image](https://user-images.githubusercontent.com/15135271/67916749-f010c900-fb53-11e9-88a2-c5640ec6ab45.png)

billhollings:
@cdavis5e Can you have a look at this when you get a chance, please?

cdavis5e:
Sorry. I can't look at this right now because I'm busy with critical bugs in CrossOver 19.

billhollings:
@cdavis5e  I've started to look at this now.

billhollings:
@DiligentGraphics PR #785 fixes this issue. Please re-test and close this issue if it fixes your issue.

DiligentGraphics:
@billhollings Thanks! Will test the fix soon.

DiligentGraphics:
Verified with 1.1.130.0 sdk - everything works correctly.
Thanks!

Issue #784, opened 2019-11-26
Current spirv-cross dependency


o1lo01ol1o:
Hi,

  Just wondering when the effects of https://github.com/KhronosGroup/SPIRV-Cross/pull/1217 could be officially felt here in master.  

Thanks.

billhollings:
@o1lo01ol1o  PR #785 includes a pull of the latest SPIRV-Cross version, which includes your request.

PR #785, opened 2019-11-26
Fix tessellation break when control stage declares but does not use position builtin.

billhollings:
- Add `SPIRVShaderOutput::isUsed` retrieved from shader reflection.
- `mvk::sizeOfOutput()` returns zero if output var is not used.
- Make `mvkCeilingDivide()` a template function.
- Update to latest SPIRV-Cross version.

Fixes issues #772 & #784.


billhollings:
@cdavis5e 

Please review.

This PR fixes the use-case from #772 (where the SPIR-V defined a `gl_PerVertex::gl_Position` struct member that was not referenced and therefore not taken into consideration during tess buffer input/output...and will also fix the same issue for any declared-but-unused builtin.

But I am wondering if there are more generic use cases where a shader declares but does not use non-builtin outputs, which might trigger the same I/O impedance mismatches. The current SPIRV-Cross does not track shader outputs that are not used...and in fact `Compiler::get_active_interface_variables()` deliberately includes outputs that are declared but never used in the shader.

PR #787, opened 2019-12-02
Cube demo use VK_EXT_metal_surface extension.

billhollings:


Valakor:
I'm getting a validation layer warning followed by a crash when attempting to use the recommended `vkCreateMetalSurfaceEXT` rather than `vkCreateMacOSSurfaceMVK` like in this PR in conjunction with `VK_LAYER_KHRONOS_validation`. Disabling the validation layers causes everything to work fine.

Message: `[Warn ] [Validation] [VUID_Undefined] Instance Extension VK_EXT_metal_surface is not supported by this layer.  Using this extension may adversely affect validation results and/or produce undefined behavior.`
This is then followed by a crash in one of the layers (presumably the validation layer) the first time the returned VKSurfaceKHR is actually used.

Everything seems to work fine when I don't enable the validation layers - is this expected behavior for SDK 1.1.126.0 (should I just wait to perform this conversion until a later SDK)? Or should I report this to the vulkan loader repository?

billhollings:
@Valakor 

>is this expected behavior for SDK 1.1.126.0 (should I just wait to perform this conversion until a later SDK)? Or should I report this to the vulkan loader repository?

The answer to both questions is "yes". 😉

_Yes_...it looks like _Vulkan SDK 1.1.126_ does not support the `VK_EXT_metal_surface` extension yet...so it is expected behaviour. And _yes_...you should request this from the _Vulkan Loader & Layers_ team to help them prioritize.

Issue #786, opened 2019-11-28
Wrong memory types reported


starbucksDave:
On M395x and Radeon VII the following memory/heap types are reported:
memtypes:
```
0 - device_local
	heapindex 0
1 - device_local / host_visible / host_cached
	heapindex 0
2 - device_local / host_visible / host_coherent / host_cached
	heapindex 1
```
heaps:
```
0 - device_local
1 - no flags
```

It reports that memory type 2 is device_local but it goes on a heap that is not device_local.
However the spec states: 
```
VK_MEMORY_PROPERTY_DEVICE_LOCAL_BIT bit specifies that memory allocated with this type is the most efficient for device access. This property will be set if and only if the memory type belongs to a heap with the VK_MEMORY_HEAP_DEVICE_LOCAL_BIT set.
```

billhollings:
@starbucksDave This is fixed in PR #789. Please retest and close if it fixes your issue.

PR #789, opened 2019-12-06
Fix inconsistency in reporting device local memory between type and heap on macOS.

billhollings:
Fixes issue #786.

### DIFF
--- a/MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
@@ -1844,7 +1844,7 @@ void MVKPhysicalDevice::initMemoryProperties() {
 		// This means we really have two heaps. The second heap is system memory.
 		_memoryProperties.memoryHeapCount = 2;
 		_memoryProperties.memoryHeaps[1].size = mvkGetSystemMemorySize();
-		_memoryProperties.memoryHeaps[1].flags = 0;
+		_memoryProperties.memoryHeaps[1].flags = VK_MEMORY_HEAP_DEVICE_LOCAL_BIT;
 		_memoryProperties.memoryTypes[2].heapIndex = 1;	// Shared memory in the shared heap
 	}
 #endif
--- a/MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
@@ -791,7 +791,7 @@ void MVKPhysicalDevice::initMetalFeatures() {
 
 	_metalFeatures.maxPerStageBufferCount = 31;
     _metalFeatures.maxMTLBufferSize = (256 * MEBI);
-    _metalFeatures.dynamicMTLBuffers = false;
+    _metalFeatures.dynamicMTLBufferSize = 0;
 
     _metalFeatures.maxPerStageSamplerCount = 16;
     _metalFeatures.maxQueryBufferSize = (64 * KIBI);
@@ -812,7 +812,7 @@ void MVKPhysicalDevice::initMetalFeatures() {
 
     if ( [_mtlDevice supportsFeatureSet: MTLFeatureSet_iOS_GPUFamily1_v2] ) {
 		_metalFeatures.mslVersionEnum = MTLLanguageVersion1_1;
-        _metalFeatures.dynamicMTLBuffers = true;
+        _metalFeatures.dynamicMTLBufferSize = (4 * KIBI);
 		_metalFeatures.maxTextureDimension = (8 * KIBI);
     }
 
@@ -881,7 +881,7 @@ void MVKPhysicalDevice::initMetalFeatures() {
 
     if ( [_mtlDevice supportsFeatureSet: MTLFeatureSet_macOS_GPUFamily1_v2] ) {
 		_metalFeatures.mslVersionEnum = MTLLanguageVersion1_2;
-        _metalFeatures.dynamicMTLBuffers = true;
+        _metalFeatures.dynamicMTLBufferSize = (4 * KIBI);
         _metalFeatures.shaderSpecialization = true;
         _metalFeatures.stencilViews = true;
         _metalFeatures.samplerClampToBorder = true;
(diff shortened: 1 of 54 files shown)


MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
```diff
@@ -1844,7 +1844,7 @@ static uint32_t mvkGetEntryProperty(io_registry_entry_t entry, CFStringRef prope
 		// This means we really have two heaps. The second heap is system memory.
 		_memoryProperties.memoryHeapCount = 2;
 		_memoryProperties.memoryHeaps[1].size = mvkGetSystemMemorySize();
-		_memoryProperties.memoryHeaps[1].flags = 0;
+		_memoryProperties.memoryHeaps[1].flags = VK_MEMORY_HEAP_DEVICE_LOCAL_BIT;
 		_memoryProperties.memoryTypes[2].heapIndex = 1;	// Shared memory in the shared heap
```
cdavis5e: I'm wondering if we should actually *clear* the `VK_MEMORY_PROPERTY_DEVICE_LOCAL_BIT` from the memory type corresponding to `MTLStorageModeShared`.

billhollings: @cdavis5e 

I thought along those lines at first...but then I realized that it's shared memory we're talking about. It is device local as far as both Vulkan and Metal are concerned.

From an Apple [doc](https://developer.apple.com/documentation/metal/setting_resource_storage_modes/choosing_a_resource_storage_mode_in_macos?language=objc) on the matter:

>In both [_unified_ and _discrete_] memory models, a resource with a MTLStorageModeShared mode resides in system memory accessible to both the CPU and the GPU.

BTW...what was the original problem that the second heap was trying to solve? The original code seems to be trying to say that shared memory is not accessible to the GPU.

cdavis5e: `vkd3d` checks how many heaps there are to see if the device is a UMA device.

mbarriault: I was always of the understanding that `MTLStorageModeShared` was equivalent to, for example, `cudaMallocHost` - memory that is device visible, but still on the host. That would suggest that it shouldn't be marked device local in Vulkan parlance, as that suggests a certain level of performance that device visible but host local memory doesn't have.

billhollings: @cdavis5e 

>if at least one is not host-visible

Do you mean not GPU-visible? How do you test for the host-visibility of a heap?

@mbarriault 

>At least one heap must include `VK_MEMORY_HEAP_DEVICE_LOCAL_BIT` in `VkMemoryHeap::flags`. If there are multiple heaps that all have similar performance characteristics, they may all include `VK_MEMORY_HEAP_DEVICE_LOCAL_BIT`. In a unified memory architecture (UMA) system there is often only a single memory heap which is considered to be equally “local” to the host and to the device, and such an implementation must advertise the heap as device-local.

So a UMA device must have it's single heap listed as device-local...implying (to me at least) that the device local flags don't imply  special fast device-only memory...just memory that the GPU can access.

However...the spec also states that...

>Device memory is memory that is visible to the device 

...which would imply that ALL heaps and memory types (which are collectively discussed under the Device Memory banner) are assumed to be device-visible...implying also that the requirement for device-local flags is almost redundant in the UMA case above.

_...[sigh]..._

billhollings: >...[sigh]...

I've [posted a question](https://gitlab.khronos.org/vulkan/vulkan/issues/1930) to the Vulkan team about this.

billhollings: @cdavis5e @mbarriault 

I've updated this PR...and the logic should be right now.

Because this PR has been sitting open for a while, there is a bunch of other semi-related stuff in it now...but, if you are reviewing, the areas of interest for the memory logic are:

- `MVKPhysicalDevice::initMemoryProperties()`, which I've made much more procedural, organized, and traceable.
- The value of `MVK_VK_MEMORY_TYPE_METAL_SHARED` no longer includes `VK_MEMORY_PROPERTY_DEVICE_LOCAL_BIT` or `VK_MEMORY_PROPERTY_HOST_CACHED_BIT`...but `VK_MEMORY_PROPERTY_DEVICE_LOCAL_BIT` is added dynamically for unified memory.
- `MVKPhysicalDevice::getHasUnifiedMemory()` adds a bit of additional logic to test for the low-power GPU and device count in the case the `MTLDevice` does not respond to `hasUnifiedMemory`.

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
```diff
@@ -1749,127 +1749,131 @@ static uint32_t mvkGetEntryProperty(io_registry_entry_t entry, CFStringRef prope
...
 
 bool MVKPhysicalDevice::getHasUnifiedMemory() {
 #if MVK_IOS
 	return true;
 #endif
 #if MVK_MACOS
-	return [_mtlDevice respondsToSelector: @selector(hasUnifiedMemory)] && _mtlDevice.hasUnifiedMemory;
+	return (([_mtlDevice respondsToSelector: @selector(hasUnifiedMemory)] && _mtlDevice.hasUnifiedMemory)
+			|| _mtlDevice.isLowPower
+			|| getInstance()->getPhysicalDeviceCount() == 1);
```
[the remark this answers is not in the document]
billhollings: >Won't that erroneously include Macs with one discrete card and no integrated graphics?

Hmmm...you might be right. It was intended to cover the case where there was only integrated graphics which might not be listed as low-power (I figured a Mac with only one integrated GPU might not list that one GPU as low-power...and I don't have a machine to test that on).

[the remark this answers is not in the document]
billhollings: > I figured a Mac with only one integrated GPU might not list that one GPU as low-power

Hmmm...looks like my assumption was wrong. From Apple's docs on `isLowPower`...

>This property returns YES for integrated GPUs and NO for discrete GPUs.

Pretty straightforward. I'll remove the single-GPU test...but leave the `isLowPower` test...on the assumption that if `hasUnifiedMemory` is not available...then `isLowPower` is a valid indication of integrated, and therefore unified, memory.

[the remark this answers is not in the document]
billhollings: > I'll remove the single-GPU test...but leave the isLowPower test...on the assumption that if hasUnifiedMemory is not available...then isLowPower is a valid indication of integrated, and therefore unified, memory.

Implemented in PR #795

Issue #792, opened 2019-12-15
Depth-only render pass doesn't clear depth if first draw is tessellation


mbarriault:
This doesn't seem to happen if there's a color attachment, but does if it's depth-only (such as a shadows pre-pass). Possible the issue comes in `MVKCommandEncodingPool::getMTLDepthStencilState`. When stepping through, `useDepth` was false. Note I first discovered this with a tessellated draw call, but putting a non-tessellated draw call first appeared to have the same issue.

Just in case I have my shadow pass configured incorrectly, the render pass is created with

    static const vk::AttachmentDescription attachmentDescription {{}, vk::Format::eD32Sfloat, vk::SampleCountFlagBits::e1, vk::AttachmentLoadOp::eClear, vk::AttachmentStoreOp::eStore, vk::AttachmentLoadOp::eDontCare, vk::AttachmentStoreOp::eDontCare, vk::ImageLayout::eUndefined, vk::ImageLayout::eTransferSrcOptimal};
    static const vk::AttachmentReference attachmentReference {0, vk::ImageLayout::eDepthAttachmentStencilReadOnlyOptimal};
    static const vk::SubpassDependency subpassDependency {0, VK_SUBPASS_EXTERNAL, vk::PipelineStageFlagBits::eLateFragmentTests, vk::PipelineStageFlagBits::eFragmentShader, vk::AccessFlagBits::eDepthStencilAttachmentWrite, vk::AccessFlagBits::eShaderRead, vk::DependencyFlagBits::eDeviceGroup};
    static const vk::SubpassDescription subpassDescription {{}, vk::PipelineBindPoint::eGraphics, 0, nullptr, 0, nullptr, nullptr, &attachmentReference, 0, nullptr};
    shadowRenderPass = device.createRenderPassUnique({{}, 1, &attachmentDescription, 1, &subpassDescription, 1, &subpassDependency});

mbarriault:
Still investigating, I think I'm getting a lack of clear with a non-tessellated draw call too. This particular draw is full-screen so it's harder to measure.

billhollings:
>Still investigating

Any progress? Is it possible to replicate what you're seeing in a sample app?

mbarriault:
Not yet. I got distracted fixing the inline uniform, and generally busy with holidays. I'll try to get a reproducible demo over the next couple days.

billhollings:
> I'll try to get a reproducible demo over the next couple days.

[...]

Issue #832, opened 2020-03-02
Image Corruption


DiegoAce:
I'm getting what looks like an image corruption. I was able to reproduce this in the Cube demo by adding VK_IMAGE_USAGE_SAMPLED_BIT to the depth image, using VK_IMAGE_TILING_OPTIMAL for depth and texture_object images, and allocating the depth and texture_object images with VulkanMemoryAllocator (https://github.com/GPUOpen-LibrariesAndSDKs/VulkanMemoryAllocator). This could be a bug in VMA, but I thought I'd bring it here first. 

This does not occur when running the Cube demo on iOS. Only Mac. My MacOS is 10.15.3 with Radeon Pro 570 4 GB.

I'm testing with the latest commit bfde618. I've attached the modified cube.c. I attached vk_mem_alloc.h and vma.cpp which I add to the Cube demo project. I also attached a screenshot of the corruption I'm seeing when I run the Cube demo.

In my own similar demo it seems like the corruption happens when I begin/end the related renderpass. If I submit the command buffer with just the begin and end of the renderpass, with nothing in between, and afterwards transfer the texture_object image to a file I can view, then it will be corrupted.

[Archive.zip](https://github.com/KhronosGroup/MoltenVK/files/4277974/Archive.zip)

billhollings:
Thanks for submitting the sample app demonstrating the issue. The error appears in my environment too.

I've done a fair bit of digging into it. The error is triggered by a long combination of things:

1. VMA uses a single large underlying `VkDeviceMemory` to bind to both the depth image and the samplable texture being loaded from file (via a buffer copy). The native Cube demo uses separate device memory instances. Changing back fixes this issue.

2. MoltenVK uses a placement `MTLHeap` within the device memory to create the underlying `MTLTextures` for each image. Configuring MoltenVK to avoid using `MTLHeaps` also fixes this issue.

3. The issue only appears if the the depth image includes samplable `usage`. Changing this back also fixes the issue.

There doesn't seem to be a clear reason why this is happening. I've verified that VMA is picking byte offsets and alignments for the `MTLTextures` that should avoid conflicts within the placement `MTLHeap` memory space. And it's hard to see why modifying the `usage` property should naturally have any effect on where in the heap the memory is allocated from.

At this point, I'm suspecting a possible bug within the Metal driver that triggers a layout conflict inside the placement `MTLHeap` when using samplable depth images.

[...]

PR #845, opened 2020-03-30
Do not create depth/stencil MTLTextures from MTLHeap.

billhollings:
Depth/stencil textures issued from `MTLHeap` can be flaky in some situations, including depth-only rendering on Intel GPUs, and corruption when DS texture with sampling usage is combined with non-DS texture on a single placement heap.

Fixes issues #792 and #832.


billhollings:
Cube demo failing until Vulkan-Tools updated to sync with other dependencies. Merging anyway

PR #852, opened 2020-04-06
Various fixes resulting from testing for SDK release.

billhollings:
- Disable use of `MTLHeap` by default.
- Swapchain fixes.
- Disable `API-Samples` demos.


billhollings:
Travis failed for an internal reason, and is not making a Restart build option available. Merging anyway.

PR #856, opened 2020-04-14
 MVKPipeline: Disable fragment shader outputs for unused attachments.

cdavis5e:
Naturally, Metal complains when a fragment shader write to a
`[[depth]]` output, but there be no depth attachment. I imagine that it
will also complain if the shader write to ``[[stencil]]`` with no
stencil attachment, or it write to a ``[[color(n)]]`` output with no
corresponding attachment.


billhollings:
Thanks for remembering `SPIRVToMSLConversionOptions::matches()` and `serialize(Archive & archive, CompilerMSL::Options& opt)`! They're easy to forget!

PR #861, opened 2020-04-15
MVKBuffer: Support texel buffers in "host-coherent" memory on Mac.

cdavis5e:
According to the Vulkan spec:

> * If `buffer` is a `VkBuffer` not created with the
>   `VK_BUFFER_CREATE_SPARSE_BINDING_BIT` bit set[...] then the
>   `memoryTypeBits` member always contains at least one bit set
>   corresponding to a `VkMemoryType` with a `propertyFlags` that has
>   both the `VK_MEMORY_PROPERTY_HOST_VISIBLE_BIT` bit and the
>   `VK_MEMORY_PROPERTY_HOST_COHERENT_BIT` bit set. In other words,
>   mappable coherent memory **can** always be attached to these
>   objects.

There is no exception for texel buffers. Even though desktop Metal
disallows textures in shared memory, even linear textures created from a
buffer, we have to advertise host-coherent memory for texel buffers.
Some applications actually depend on this behavior, so it's not just a
theoretical concern.

To support host-coherent texel buffers, we implicitly create a managed
buffer and copy data between the device memory and the managed buffer,
just like for a linear image.

### DIFF
--- a/MoltenVK/MoltenVK/GPUObjects/MVKBuffer.h
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKBuffer.h
@@ -72,7 +72,7 @@ public:
 	id<MTLBuffer> getMTLBuffer();
 
 	/** Returns the offset at which the contents of this instance starts within the underlying Metal buffer. */
-	inline NSUInteger getMTLBufferOffset() { return _deviceMemory && _deviceMemory->getMTLHeap() ? 0 : _deviceMemoryOffset; }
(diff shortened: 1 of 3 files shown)


cdavis5e:
Bah. It broke because of some failure to clean itself up in preparation to build.

billhollings:
> Bah. It broke because of some failure to clean itself up in preparation to build.

Travis has been breaking regularly.

billhollings:
> Travis has been breaking regularly recently.

And it is often not possible to restart the CI run. Very frustrating.

PR #867, opened 2020-04-23
MVKPixelFormats enables atomic capabilities only when applicable.

billhollings:
Consolidates setting VkFormatProperties, and only sets atomic features in `linearTilingFeatures` and `bufferFeatures` for `MTLPixelFormatR32Uint` and `MTLPixelFormatR32Sint`.

This setting logic was already being done, but was being done by setting the atomic features in all formats by default, and then stripping them away for everything other than those two formats.

In addition, previously, the two formats allowed atomic operations for `optimalTilingFeatures` too. This has been move to explicitly set only for `linearTilingFeatures` now.


billhollings:
@cdavis5e 

Your previous update triggered me to do a redesign of the way format properties were being set. In particular, since you identified that only two formats should permit atomic writes, I flipped the logic to assume atomic feature is not set by default, and enable only for those two formats using the same enabling mechanism for the other feature flags.

In addition, your description of how atomics are handled on your previous change indicated that in Metal the atomic feature only applies to linear, and not optimal tiling. But in your original code, it appeared to enable for both linear and optimal for those two formats. I've modified the logic to explicitly enable for only linear tiling now.

Everything should still operate as you described in your previous PR. When you get a chance, can you have a look, test, and verify that I haven't broken your initial logic, please? I can modify the linear/optimal enablement if I've messed it up.

Thanks.

MoltenVK/MoltenVK/GPUObjects/MVKPixelFormats.mm
```diff
@@ -1104,8 +1108,10 @@
 
 	addMTLPixelFormatCapabilities( iOS_GPUFamily2_v1, RG8Snorm, All );
 
-	addMTLPixelFormatCapabilities( iOS_GPUFamily1_v2, R32Uint, RWC );
-	addMTLPixelFormatCapabilities( iOS_GPUFamily1_v2, R32Sint, RWC );
+	addMTLPixelFormatCapabilities( iOS_GPUFamily1_v2, R32Uint, Write );
+	addMTLPixelFormatCapabilities( iOS_GPUFamily1_v2, R32Sint, Write );
```
cdavis5e, line 1112: Was this intentional? You're no longer setting the `Read` and `ColorAtt` caps.

billhollings: It's safe. The `addMTLPixelFormatCapabilities()` function adds (`or`'s) the additional flags to what is there already. The other flags come from the base setting for the format. 

Actually all of the similar `addMTLPixelFormatCapabilities()` calls could do the same, but almost all the other calls are about redefining the capabilities for a Metal feature set table entry it was easier to simply redefine (and visually verify from the Apple docs) the entire entry.

Meh. But having said that...I'll revert the `Write` entry itself to a full redefinition so it's consistent with all the other feature-set-based mods...and then add the `Atomic` part as it's being done here.

cdavis5e: Oh, you set it above. Never mind.

PR #874, opened 2020-05-08
Malloc redoBuff in reportMessage to prevent stack overflow

js6i:
Happened to me while logging a big shader.


billhollings:
@js6i  Thanks for your fix!

PR #876, opened 2020-05-11
Mark computePipelineState as dirty when destroying mtlComputeEncoder

js6i:
Consider the following (approximate) sequence of commands:
1. vkCmdBindPipeline (compute)
2. vkCmdDispatch: will call setComputePipelineState:_ and mark state as clean
3. vkCmdBeginRenderPass .. draw .. vkCmdEndRenderPass: will endCurrentMetalEncoding()
4. vkCmdDispatch: will recreate mtlComputeEncoder, but won't setComputePipelineState:_ as state is still marked clean, and will crash as a result

AFAIU this is valid usage and binding made in 1. should hold. The patch makes endCurrentMetalEncoding() also mark compute pipeline state as dirty.

### DIFF
--- a/MoltenVK/MoltenVK/Commands/MVKCommandBuffer.mm
+++ b/MoltenVK/MoltenVK/Commands/MVKCommandBuffer.mm
@@ -449,6 +449,10 @@ void MVKCommandEncoder::endMetalRenderEncoding() {
 void MVKCommandEncoder::endCurrentMetalEncoding() {
 	endMetalRenderEncoding();
 
+	_computePipelineState.markDirty();
+	_computeResourcesState.markDirty();
+	_computePushConstants.markDirty();
+
 	[_mtlComputeEncoder endEncoding];
 	_mtlComputeEncoder = nil;       // not retained
 	_mtlComputeEncoderUse = kMVKCommandUseNone;
--- a/MoltenVK/MoltenVK/GPUObjects/MVKBuffer.h
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKBuffer.h
@@ -75,7 +75,7 @@ public:
 	id<MTLBuffer> getMTLBuffer();
 
 	/** Returns the offset at which the contents of this instance starts within the underlying Metal buffer. */
-	inline NSUInteger getMTLBufferOffset() { return _deviceMemory && _deviceMemory->getMTLHeap() && !_isHostCoherentTexelBuffer ? 0 : _deviceMemoryOffset; }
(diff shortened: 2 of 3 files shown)


js6i:
Looking at MVKBuffer::getMTLBuffer(), it seems that we should be returning _deviceMemoryOffset from MVKBuffer::getMTLBufferOffset() only when we return the underlying allocation buffer (_deviceMemory->getMTLBuffer(), last else branch). The second commit changes logic in getMTLBufferOffset to do that.

MoltenVK/MoltenVK/Commands/MVKCommandBuffer.mm
```diff
@@ -449,6 +449,8 @@
 void MVKCommandEncoder::endCurrentMetalEncoding() {
 	endMetalRenderEncoding();
 
+	_computePipelineState.markDirty();
```
billhollings, line 452: Did you investigate whether the other two state trackers in `MVKCommandEncoder::finalizeDispatchState()` (`_computeResourcesState` and `_computePushConstants`) need to be marked dirty too?

MoltenVK/MoltenVK/GPUObjects/MVKBuffer.h
```diff
@@ -75,7 +75,7 @@ class MVKBuffer : public MVKResource {
 	id<MTLBuffer> getMTLBuffer();
 
 	/** Returns the offset at which the contents of this instance starts within the underlying Metal buffer. */
-	inline NSUInteger getMTLBufferOffset() { return _deviceMemory && _deviceMemory->getMTLHeap() && !_isHostCoherentTexelBuffer ? 0 : _deviceMemoryOffset; }
+	inline NSUInteger getMTLBufferOffset() { return !_deviceMemory || _deviceMemory->getMTLHeap() || _isHostCoherentTexelBuffer ? 0 : _deviceMemoryOffset; }
```
billhollings, line 78: Good catch! Thanks.

MoltenVK/MoltenVK/Commands/MVKCommandBuffer.mm
js6i: No, I didn't. It seems likely that they do, though honestly I'm not familiar with the code enough to be sure. This made sense and works for my case.

billhollings: Okay. Can you add them here anyway, please?

js6i: Sure, I pushed changed commits

PR #881, opened 2020-05-21
Add support for VK_GOOGLE_display_timing.  Some notes:

danginsburg:
* vkGetRefreshCycleDurationGOOGLE is implemented for iOS using UIScreen maximumFramesPerSecond.  I did not implement it for macOS (just hard-coded to 60).  It looks like doing it will require using CGDisplayModeGetRefreshRate
* vkGetPastPresentationTimingGOOGLE fills in the actualPresentationTime, but Metal does not provide the earlierPresentTime or the presentMargin as far as I can tell so earliestPresentTime is set to actualPresentTime and presentMargin is zero.
* Tested mostly on iOS - confirmed with Metal System Trace that desiredPresentTime works properly aligning the presentation correctly with vsync.


billhollings:
Nice work! Thanks for implementing this!

PR #898, opened 2020-06-01
 Check for stencil only image view in identity swizzle case 

js6i:
The buffer binding size looks unused now (?), but the other commit sets it to the specified value.

### DIFF
--- a/MoltenVK/MoltenVK/GPUObjects/MVKDescriptor.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKDescriptor.mm
@@ -148,7 +148,11 @@ void MVKDescriptorSetLayoutBinding::push(MVKCommandEncoder* cmdEncoder,
                 MVKBuffer* buffer = (MVKBuffer*)bufferInfo.buffer;
                 bb.mtlBuffer = buffer->getMTLBuffer();
                 bb.offset = buffer->getMTLBufferOffset() + bufferInfo.offset;
-				bb.size = (uint32_t)buffer->getByteCount();
+                if (bufferInfo.range == VK_WHOLE_SIZE)
+                    bb.size = (uint32_t)(buffer->getByteCount() - bb.offset);
+                else
+                    bb.size = (uint32_t)bufferInfo.range;
+
                 for (uint32_t i = kMVKShaderStageVertex; i < kMVKShaderStageMax; i++) {
                     if (_applyToStage[i]) {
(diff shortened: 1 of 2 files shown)


MoltenVK/MoltenVK/GPUObjects/MVKDescriptor.mm
```diff
@@ -148,7 +148,11 @@
                 MVKBuffer* buffer = (MVKBuffer*)bufferInfo.buffer;
                 bb.mtlBuffer = buffer->getMTLBuffer();
                 bb.offset = buffer->getMTLBufferOffset() + bufferInfo.offset;
-				bb.size = (uint32_t)buffer->getByteCount();
```
billhollings, line 151: @js6i Is there a use-case where this is actually causing problems?

When it hits Metal, `MVKMTLBufferBinding.size` is only used when setting inline bytes, not when using an actual `VkBuffer/MTLBuffer`, so `VK_WHOLE_SIZE` should have no applicability.

@cdavis5e `MVKMTLBufferBinding.size` was introduced in support of SPIRV `OpArrayLength`. Your thoughts on this?

MoltenVK/MoltenVK/GPUObjects/MVKDescriptor.mm
```diff
@@ -493,7 +497,10 @@
 			if (_mvkBuffer) {
 				bb.mtlBuffer = _mvkBuffer->getMTLBuffer();
 				bb.offset = _mvkBuffer->getMTLBufferOffset() + _buffOffset + bufferDynamicOffset;
-				bb.size = (uint32_t)_mvkBuffer->getByteCount();
```
billhollings, line 496: Ditto.

MoltenVK/MoltenVK/GPUObjects/MVKDescriptor.mm
js6i: Not that I know of. I was confused by this line, but didn't see the value used anywhere. If it's intended to stay for some reason, I'd put a comment there.

cdavis5e: That is used to set the buffer size passed down to the shader. See MoltenVK/MoltenVK/Commands/MVKCommandEncoderState.mm:585 (or thereabouts). This change appears to be correct.

PR #904, opened 2020-06-02
Add new MVKConfiguration option allowImageCopyFormatReinterpretation that indicates the application won't require format reinterpreting conversions so MTLTextureUsagePixelFormatView can be disabled

danginsburg:
* The issue here is that on iOS 5+ GPU family, MTLTextureUsagePixelFormatView disables lossless compression.
* MoltenVK sets MTLTextureUsagePixelFormatView for two reasons: 1) to handle format reinterpreting vkCmdCopyImage and 2) to handle texture swizzles
* For case 2), MTLTextureUsagePixelFormatView is not needed if native swizzles are supported by the device because we can always create views with a swizzle instead of a different format
* For case 1), we have no way of knowing whether an image marked a VK_IMAGE_USAGE_TRANSFER_SRC_BIT will require image conversion, so this MVKConfiguration flag indicates that.

If the two conditions are met (allowImageCopyFormatReinterpretation is false and native swizzles are supported), then MTLTextureUsagePixelFormatView is not set.

One upshot of this change is that on any device supporting native swizzles, MoltenVK will now always do a view with a swizzle instead of a format reinterpretations (i.e. handling R8 + swizzle as A8 will instead just become a swizzle).  I presume there is no reason not to do this, but this will happen now regardless of how allowImageCopyFormatReinterpretation is set.


MoltenVK/MoltenVK/GPUObjects/MVKPixelFormats.mm
```diff
@@ -431,17 +433,25 @@
...
 
-		mvkEnableFlags(mtlUsage, MTLTextureUsagePixelFormatView);
+		// MTLTextureUsagePixelFormatView is required only for reinterpretation of formats.  vkCmdCopyImage
+		// allows format reinterpretations for formats of the same bits, so this must be set for VK_IMAGE_USAGE_TRANSFER_SRC_BIT.
+		// However, on iOS GPU family 5 and later, this will disable lossless compression causing a performance impact,
+		// so there is an MVKConfig option to disable this (allowImageCopyReinterpretation).  It's the application's
+		// way of telling MoltenVK that it doesn't require format converting copies.  This also requires nativeTextureSwizzle
+		// in order to not need format reinterpretation for swizzling
+		if (!allowImageCopyReinterpretation || !nativeTextureSwizzle) {
+			mvkEnableFlags(mtlUsage, MTLTextureUsagePixelFormatView);
```
billhollings, line 453: Is this logic correct? From the description, it seems that if I set `allowImageCopyReinterpretation` to `false`, I am intending that I DON'T want `MTLTextureUsagePixelFormatView` enabled (unless native swizzle is missing)...but this logic seems to be doing the opposite. Am I missing something?

danginsburg: Ahh, thank you I made a mistake here both on Source 2 and MVK where I inverted the meaning in my head.  Will fix.

kvark:
If this is so important to performance, shouldn't we raise this issue in VkPI call / mailing list?

danginsburg:
> If this is so important to performance, shouldn't we raise this issue in VkPI call / mailing list?

Sure, I will do that.

billhollings:
Closing for redesign.

PR #906, opened 2020-06-04
Reduce about 4% CPU cost on Intel Core i7 processor

Michael-Lfx:


MoltenVK/MoltenVK/GPUObjects/MVKVulkanAPIObject.h
```diff
@@ -103,7 +103,7 @@ class MVKVulkanAPIObject : public MVKConfigurableObject {
 protected:
 	virtual void propagateDebugName() = 0;
 
-	std::atomic<uint32_t> _refCount;
+	std::atomic<uint64_t> _refCount;
```
billhollings, line 106: Is this change a requirement for the performance boost? Does the use of 64 bits improve access performance?

This change bumps the size of the base `MVKVulkanAPIObject` class up by 8 bytes (and leaves a 4-byte gap), which affects a lot of other classes and their instances. 

Not a huge deal, since these are API classes, so there will not be enormous quantities, and a 4% performance boost is definitely more important...but nevertheless, we've just finished a round of memory optimization on the class hierarchy, and I'd only like to regress on that if necessary.

billhollings:
@Michael-Lfx Can you agree to the CLA, please? We can't pull your PR in unless you do so.

MoltenVK/MoltenVK/GPUObjects/MVKVulkanAPIObject.h
billhollings: I tried testing this, and found no difference with 32-bit vs 64-bit.

billhollings:
Without the CLA in place, we can't accept this PR.

I implemented these changes separately and compared them against the existing implementation using the following test code:

	uint32_t rrCnt = 1e7;
	uint64_t rrStartTime = mvkGetTimestamp();
	for (uint32_t i = 0; i < rrCnt; i++) {
		mvkDev->retain();
		mvkDev->release();
	}
	printf("[mvk-trace] Time for %d retain/release cycles: [%.1f ms]\n", rrCnt, mvkGetElapsedMilliseconds(rrStartTime));

I didn't find any consistent difference in performance beyond differences due to noise. I attribute this to a good implementation of atomic `++` and `--` in the STL.

However, I DID find a consistent 30% improvement by inlining the `retain()` and `release()`functions, which I've included in PR #928 

During testing, I also didn't find any performance difference between using `uint32_t` or `uint64_t` for `_refCount`, so I've left it at `uint32_t`, since it saves 8-bytes per instance, as I described in my analysis in a note above.

I'm closing this PR. If it can be demonstrated that improvements can be found by changing the incrementing implementation, we can reopen.

PR #907, opened 2020-06-04
Handle vertex attributes with zero stride

js6i:
6df700f Avoids a div by zero crash below on %= stride line.
I don't have a good test case for 011f029, but searching github for `MTLVertexStepFunctionConstant` reveals that UE4 and Dawn do something similar.

### DIFF
--- a/MoltenVK/MoltenVK/GPUObjects/MVKPipeline.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKPipeline.mm
@@ -1009,7 +1009,7 @@ bool MVKGraphicsPipeline::addVertexInputToPipeline(MTLRenderPipelineDescriptor* 
 				const VkVertexInputBindingDescription* pVKVB = pVI->pVertexBindingDescriptions;
 				for (uint32_t j = 0; j < vbCnt; j++, pVKVB++) {
 					if (pVKVB->binding == pVKVA->binding) {
-						if (vaOffset >= pVKVB->stride) {
+						if (pVKVB->stride && vaOffset >= pVKVB->stride) {
 							// Move vertex attribute offset into the stride. This vertex attribute may be
 							// combined with other vertex attributes into the same translated buffer binding.
 							// But if the reduced offset combined with the vertex attribute size still won't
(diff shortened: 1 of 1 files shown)


MoltenVK/MoltenVK/GPUObjects/MVKPipeline.mm
```diff
@@ -970,9 +970,15 @@ static VkFormat mvkFormatFromOutput(const SPIRVShaderOutput& output) {
 			maxBinding = max(pVKVB->binding, maxBinding);
 			uint32_t vbIdx = getMetalBufferIndexForVertexAttributeBinding(pVKVB->binding);
 			MTLVertexBufferLayoutDescriptor* vbDesc = plDesc.vertexDescriptor.layouts[vbIdx];
-			vbDesc.stride = (pVKVB->stride == 0) ? sizeof(simd::float4) : pVKVB->stride;      // Vulkan allows zero stride but Metal doesn't. Default to float4
-            vbDesc.stepFunction = mvkMTLVertexStepFunctionFromVkVertexInputRate(pVKVB->inputRate);
-            vbDesc.stepRate = 1;
+			if (pVKVB->stride == 0) {
+				vbDesc.stride = 4000; // Made up value, can't be 0, and must be bigger than any attribute's offset + size.
```
billhollings, line 974: Can you change 4000 to `kMVKUndefinedLargeUInt32`, to make it clearer that its purpose is as an undefined large number, please?

js6i: Not really, Metal errors out saying that it's not aligned to 4 bytes.
`kMVKUndefinedLargeUInt32 & ~3` works, but that's not great too because then I get lots of warnings saying roughly "buffer has space for N bytes, but argument has a length 1073741820". To be fair same could happen for 4000 or any other arbitrary value.
My rationale was that 4000 kind-of stands out and is below the limit we advertise (`maxVertexInputBindingStride = 4095`, though not sure where that's coming from).

Maybe the least bad solution would be to iterate over attributes and look for the largest value of offset + size?

js6i: Pushed it, I think that should do the job.

Issue #913, opened 2020-06-10
Undefined symbol `___isPlatformVersionAtLeast` while linking with libmoltenvk.a


MaikKlein:
Since version 1.0.42 we are getting the symbol `___isPlatformVersionAtLeast` in `libmoltenvk.a`, which is failing our builds.

>   = note: Undefined symbols for architecture x86_64:
>             "___isPlatformVersionAtLeast", referenced from:
>                 MVKPresentableSwapchainImage::presentCAMetalDrawable(id<MTLCommandBuffer>, bool, unsigned int, unsigned long long) in libash_molten-f9a8f3466dcc62c4.rlib(libMoltenVK.a-x86_64-master.o)
>           ld: symbol(s) not found for architecture x86_64
>           clang: error: linker command failed with exit code 1 (use -v to see invocation)

For context we have a project that statically links moltenvk https://github.com/EmbarkStudios/ash-molten

The error is reproducible on multiple local machines, and it also fails on the CI https://github.com/EmbarkStudios/ash-molten/runs/757461072#step:5:13

```
OSX 10.15.5
Xcode 11.5
```

Do you have any idea where `___isPlatformVersionAtLeast` is coming from and how to resolve it? 
I have read somewhere that it is supposed to be defined in `compiler-rt`  and re-exported in `libSystem`, but at least that is not the case for me.

billhollings:
@MaikKlein 

I tried downloading and building your environment...and I too see the error shown above.

I'm [fairly certain](https://forums.developer.apple.com/thread/123003) this has to do with the use of `@available()` in`MVKPresentableSwapchainImage::presentCAMetalDrawable()`.

It does not happen in other uses of MoltenVK `libMoltenVK.a`, specifically in the MoltenVK Demos.

In order to work on a fix, I tried to figure out how to re-jig your build environment (`build.rs`) so that it would attempt to link to a different `libMoltenVK.a` file via a symbolic link...so that I can plug in a locally-built version rather than always fetching `v1.0.43` from the public repo...but I couldn't figure out your build environment and build sequence well enough to make that happen.

Can you provide me with an alternate `build.rs` that would allow me to identify the location of a `libMoltenVK.a` outside of `ash-molten` (maybe via a sym-link inside `ash-molten`) so that I can work on figuring out what's going on, and fixing this, please? 

[...]

PR #914, opened 2020-06-11
Remove use of @available() directive as it can cause issues in some build environments.

billhollings:
- Remove use of `@available()` directive as it can cause issues in some build environments.
- Refactor Xcode build architectures.
- Demo `API-Samples generateSPIRVShaders` no longer builds `MoltenVKShaderConveter` tool.

Fixes issue #913.


MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
```diff
@@ -1476,7 +1476,7 @@
     _properties.limits.maxComputeWorkGroupCount[1] = kMVKUndefinedLargeUInt32;
     _properties.limits.maxComputeWorkGroupCount[2] = kMVKUndefinedLargeUInt32;
 
-    _properties.limits.maxDrawIndexedIndexValue = numeric_limits<uint32_t>::max();
+    _properties.limits.maxDrawIndexedIndexValue = kMVKUndefinedLargeUInt32;
```
cdavis5e, line 1479: Why? If the `fullDrawIndexUint32` is supported, this is required to be the maximum `uint32_t` value. That's why I changed it to set it to `numeric_limits<uint32_t>::max()` in the first place. DXVK requires this for feature level 10 support, because all D3D10+ hardware supports it, and I'm pretty sure Metal (at least, on desktop) requires D3D11+ hardware.

billhollings: Shoot! My bad. I just interpreted it as a typo...without fully absorbing how that limit is special.

Fixing it now...

billhollings: Thanks for catching that! 🤦🏻‍♂️😏

PR #922, opened 2020-06-14
Fixes 4 bugs in VK_EXT_inline_uniform_block

Lichtso:
* descriptorCount measures bytes
* array indices are offsets into the buffer (allow partial read & write)
* missing isInline flag in bind
* missing struct VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_INLINE_UNIFORM_BLOCK_PROPERTIES_EXT

### DIFF
--- a/MoltenVK/MoltenVK/GPUObjects/MVKDescriptor.h
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKDescriptor.h
@@ -71,7 +71,7 @@ public:
 	inline uint32_t getBinding() { return _info.binding; }
 
 	/** Returns the number of descriptors in this layout. */
-	inline uint32_t getDescriptorCount() { return _info.descriptorCount; }
+    inline uint32_t getDescriptorCount() { return (_info.descriptorType == VK_DESCRIPTOR_TYPE_INLINE_UNIFORM_BLOCK_EXT) ? 1 : _info.descriptorCount; }
 
 	/** Returns the descriptor type of this layout. */
 	inline VkDescriptorType getDescriptorType() { return _info.descriptorType; }
@@ -284,17 +284,19 @@ public:
 
 	void write(MVKDescriptorSet* mvkDescSet,
 			   VkDescriptorType descriptorType,
-			   uint32_t srcIndex,
+			   uint32_t dstIndex,
 			   size_t stride,
 			   const void* pData) override;
 
 	void read(MVKDescriptorSet* mvkDescSet,
 			  VkDescriptorType descriptorType,
-			  uint32_t dstIndex,
+			  uint32_t srcIndex,
 			  VkDescriptorImageInfo* pImageInfo,
 			  VkDescriptorBufferInfo* pBufferInfo,
 			  VkBufferView* pTexelBufferView,
 			  VkWriteDescriptorSetInlineUniformBlockEXT* inlineUniformBlock) override;
+    
+    void setLayout(MVKDescriptorSetLayoutBinding* dslBinding, uint32_t index) override;
 
 	void reset() override;
 
@@ -302,7 +304,6 @@ public:
 
 protected:
 	id<MTLBuffer> _mtlBuffer = nil;
-	uint32_t _dataSize = 0;
 };
 
 
--- a/MoltenVK/MoltenVK/GPUObjects/MVKDescriptor.h
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKDescriptor.h
@@ -115,6 +115,7 @@ public:
 	~MVKDescriptorSetLayoutBinding() override;
 
 protected:
+    friend class MVKInlineUniformBlockDescriptor;
 	void initMetalResourceIndexOffsets(MVKShaderStageResourceBinding* pBindingIndexes,
 									   MVKShaderStageResourceBinding* pDescSetCounts,
 									   const VkDescriptorSetLayoutBinding* pBinding);
@@ -152,11 +153,12 @@ public:
 
 	/**
 	 * Updates the internal binding from the specified content. The format of the content depends
-	 * on the descriptor type, and is extracted from pData at the location given by srcIndex * stride.
+	 * on the descriptor type, and is extracted from pData at the location given by index * stride.
+	 * MVKInlineUniformBlockDescriptor uses the index as byte offset to write to.
 	 */
 	virtual void write(MVKDescriptorSet* mvkDescSet,
 					   VkDescriptorType descriptorType,
-					   uint32_t srcIndex,
+					   uint32_t index,
 					   size_t stride,
 					   const void* pData) = 0;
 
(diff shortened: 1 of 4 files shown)


MoltenVK/MoltenVK/GPUObjects/MVKDescriptor.mm
```diff
@@ -605,18 +614,19 @@
...
-					if (i == kMVKShaderStageCompute) {
-						if (cmdEncoder) { cmdEncoder->_computeResourcesState.bindBuffer(bb); }
-					} else {
-						if (cmdEncoder) { cmdEncoder->_graphicsResourcesState.bindBuffer(MVKShaderStage(i), bb); }
-					}
-				}
-			}
+            bb.mtlBytes = _mtlBuffer.contents;
+            bb.size = (uint32_t)_mtlBuffer.length;
+            bb.isInline = true;
```
billhollings, line 619: If you are using `mtlBytes` and `isInline`, the effect during encoding will be to set the bytes directly into the `MTLRender/ComputeCommandEncoder`, without reference to the `MTLBuffer` in `_mtlBuffer`. Is there a reason why we would hold these bytes here in a `MTLBuffer` in `_mtlBuffer`? Could we just have `setLayout()` perform a general CPU `malloc()`?

MoltenVK/MoltenVK/GPUObjects/MVKDescriptor.mm
```diff
@@ -626,25 +636,18 @@
 }
 
 void MVKInlineUniformBlockDescriptor::write(MVKDescriptorSet* mvkDescSet,
-									   VkDescriptorType descriptorType,
-									   uint32_t srcIndex,
-									   size_t stride,
-									   const void* pData) {
+                                            VkDescriptorType descriptorType,
+                                            uint32_t dstIndex,
```
billhollings, line 640: See above comment about the param name.

MoltenVK/MoltenVK/GPUObjects/MVKDescriptor.mm
```diff
@@ -626,25 +636,18 @@
...
-			} else {
-				_mtlBuffer = nil;
-			}
-
+            const auto& pInlineUniformBlock = *(VkWriteDescriptorSetInlineUniformBlockEXT*)pData;
+			uint8_t* pDstData = (uint8_t*)_mtlBuffer.contents;
+            uint8_t* pSrcData = (uint8_t*)pInlineUniformBlock.pData;
+            if (pSrcData && pDstData) {
+                memcpy(pDstData + dstIndex, pSrcData, pInlineUniformBlock.dataSize);
+            }
```
billhollings, line 650: Moving the creation of the `MTLBuffer` out of this function was a really good catch. See my comment above about whether we need it at all, given your change to how the bytes are transferred to the encoder.

MoltenVK/MoltenVK/GPUObjects/MVKDescriptor.mm
```diff
@@ -654,23 +657,19 @@
...
 
 void MVKInlineUniformBlockDescriptor::read(MVKDescriptorSet* mvkDescSet,
-									  VkDescriptorType descriptorType,
-									  uint32_t dstIndex,
-									  VkDescriptorImageInfo* pImageInfo,
-									  VkDescriptorBufferInfo* pBufferInfo,
-									  VkBufferView* pTexelBufferView,
-									  VkWriteDescriptorSetInlineUniformBlockEXT* pInlineUniformBlock) {
+                                           VkDescriptorType descriptorType,
+                                           uint32_t srcIndex,
```
billhollings, line 661: As above.

MoltenVK/MoltenVK/GPUObjects/MVKDescriptorSet.mm
```diff
@@ -249,15 +259,23 @@
 	uint32_t bindCnt = (uint32_t)layout->_bindings.size();
 	for (uint32_t bindIdx = 0; bindIdx < bindCnt; bindIdx++) {
 		MVKDescriptorSetLayoutBinding* mvkDSLBind = &layout->_bindings[bindIdx];
-		uint32_t descCnt = mvkDSLBind->getDescriptorCount();
```
billhollings, line 252: Given your change above to `getDescriptorCount()`, could this function remain unchanged, and rely on the count always being `1` in the case of `VK_DESCRIPTOR_TYPE_INLINE_UNIFORM_BLOCK_EXT`?

MoltenVK/MoltenVK/GPUObjects/MVKDescriptor.mm
Lichtso: I was wondering about that too. Inlining the buffer is the purpose of the extension after all. So if there is no other reason to keep the `MTLBuffer` then I will replace it with a `malloc` and `free`.

MoltenVK/MoltenVK/GPUObjects/MVKDescriptor.mm
Lichtso: I had to move it out because pInlineUniformBlock.dataSize can be less than the length of the buffer (thus we don't know the total length here). This is used for partial updates (with an offset) into the buffer as far as I can tell.

MoltenVK/MoltenVK/GPUObjects/MVKDescriptor.mm
```diff
@@ -89,13 +89,18 @@
 	// Establish the resource indices to use, by combining the offsets of the DSL and this DSL binding.
     MVKShaderResourceBinding mtlIdxs = _mtlResourceIndexOffsets + dslMTLRezIdxOffsets;
 
-	uint32_t descCnt = _info.descriptorCount;
```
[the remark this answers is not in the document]
Lichtso: Good idea.

MoltenVK/MoltenVK/GPUObjects/MVKDescriptor.mm
```diff
@@ -679,10 +678,14 @@
 	}
 }
 
+void MVKInlineUniformBlockDescriptor::setLayout(MVKDescriptorSetLayoutBinding* dslBinding, uint32_t index) {
+    _mtlBuffer = [dslBinding->getMTLDevice() newBufferWithLength: dslBinding->getDescriptorCount()
```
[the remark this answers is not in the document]
Lichtso: Good catch! You are right, the allocation would be to small resulting in a buffer overflow later.

MoltenVK/MoltenVK/GPUObjects/MVKDescriptorSet.mm
Lichtso: It can't for two reasons:
* `getDescriptorIndex()` needs `0` as second argument (not `pDescriptorAction->dstArrayElement`)
* `_descriptors[dstStartIdx + descIdx]->write()` needs `pDescriptorAction->dstArrayElement` as third argument (not `descIdx`)

So the if branch of `VK_DESCRIPTOR_TYPE_INLINE_UNIFORM_BLOCK_EXT` uses the parameters differently than the else loop does. The same goes for `read()`.

MoltenVK/MoltenVK/GPUObjects/MVKDescriptor.h
```diff
@@ -284,25 +284,26 @@ class MVKInlineUniformBlockDescriptor : public MVKDescriptor {
 
 	void write(MVKDescriptorSet* mvkDescSet,
 			   VkDescriptorType descriptorType,
-			   uint32_t srcIndex,
+			   uint32_t dstIndex,
```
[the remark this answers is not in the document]
Lichtso: I changed the name, because it is actually used for something different in this case (and only this one, not the other `MVKDescriptor` subclasses). I changed the caller of `write() `and `read()` as well (see below).

So basically these are two different parameters:
* byte offset in case of a `VK_DESCRIPTOR_TYPE_INLINE_UNIFORM_BLOCK_EXT`
* descriptor index for all others

They just happen to be both `uint32_t` and the third parameter. But yes, this should be documented better to explain the exception here and clear the confusion.

PR #927, opened 2020-06-16
Advertise the VK_EXT_robustness2 extension.

cdavis5e:
We only support the `robustImageAccess2` feature for now. Metal's
guarantees around out-of-bounds accesses to textures give us this for
free. `nullDescriptor` should also be possible to implement, but it
needs testing. Null image descriptors will probably just work, but null
buffer pointers probably not. We may need help, either from SPIRV-Cross
or a pass in SPIRV-Tools. `robustBufferAccess2` definitely needs an
assist from SPIRV-Tools. All three are useful for Direct3D compatibility
layers (DXVK, wined3d).


cdavis5e:
When is that (what I assume is a) race condition in Travis going to get fixed?

billhollings:
>When is that (what I assume is a) race condition in Travis going to get fixed?

I talked to James (Khronos webmaster) about that a little while ago...and he indicated I'll need to reach out to Travis with a complaint. I just haven't got around to it yet.

jakejaylee123:
Hi @billhollings and @cdavis5e! Hope all is well.

Is there an issue open for building support for the `nullDescriptor` and `robustBufferAccess2` features of the `VK_EXT_robustness2` extension? Would love to follow the progress on them.

jakejaylee123:
> Hi @billhollings and @cdavis5e! Hope all is well.
> 
> Is there an issue open for building support for the `nullDescriptor` and `robustBufferAccess2` features of the `VK_EXT_robustness2` extension? Would love to follow the progress on them.

Hi @billhollings! Updating this because I just realized this is a pull request rather than an open issue. I guess I meant to ask if there is already a branch to tackle the implementation of `nullDescriptor` and `robustBufferAccess`?

cdavis5e:
> > Hi @billhollings and @cdavis5e! Hope all is well.
> > Is there an issue open for building support for the `nullDescriptor` and `robustBufferAccess2` features of the `VK_EXT_robustness2` extension? Would love to follow the progress on them.
> 
> Hi @billhollings! Updating this because I just realized this is a pull request rather than an issue. I guess I meant to ask if there is already a branch to tackle the implementation of `nullDescriptor` and `robustBufferAccess`?

Since I wrote my commit message, I looked into it, and I'm afraid it's more complicated than what I wrote. The problem is for `robustBufferAccess2`--and actually, the original `robustBufferAccess` as well--is that it also applies to buffers we can't easily bounds-check from a shader, like the vertex attribute and index buffers. We'll need help from Metal for that.

`nullDescriptors`, meanwhile, won't fly with Metal validation--having `nil` shader resources that are used isn't permitted. We'll need help from Metal here, too.

jakejaylee123:
> Since I wrote my commit message, I looked into it, and I'm afraid it's more complicated than what I wrote. The problem is for `robustBufferAccess2`--and actually, the original `robustBufferAccess` as well--is that it also applies to buffers we can't easily bounds-check from a shader, like the vertex attribute and index buffers. We'll need help from Metal for that.
> 
> 
> 
> `nullDescriptors`, meanwhile, won't fly with Metal validation--having `nil` shader resources that are used isn't permitted. We'll need help from Metal here, too.

@cdavis5e woops, totally meant to tag you instead of Bill. But thanks for the reply! Hopefully Metal/Apple will indeed help.

PR #930, opened 2020-06-19
Host-coherent texel buffers

js6i:
MVKDeviceBuffer cache management code calls the `flushToDevice` and `pullFromDevice` methods on MVKBuffers without regard to their location and overlap with given range. In the first commit I changed those methods to check for overlap and adjust the range, similarly to how it's done in MVKImage.

The other one allows host-coherent texel buffers to be bound as plain buffers using proper host-coherent memory directly, and not the auxiliary buffer. It is only used for buffer views, when a Metal texture needs to be created.

This fixes a problem which arised when using dxvk, which creates buffers with VK_BUFFER_USAGE_UNIFORM_TEXEL_BUFFER_BIT set when the more general D3D11_BIND_SHADER_RESOURCE is used. The root cause of the problem is cache management, or rather the lack of it, for persistently mapped buffers, which I'm not sure how to fix. Even if it was fixed though, this way is still more efficient.

I'm unsure about thread safety here. Should not `MVKBuffer::getMTLBuffer` (and `MVKBuffer::getMTLBufferCache` alike) be protected with a lock?

### DIFF
--- a/MoltenVK/MoltenVK/GPUObjects/MVKBuffer.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKBuffer.mm
@@ -156,9 +156,9 @@ bool MVKBuffer::shouldFlushHostMemory() { return _isHostCoherentTexelBuffer; }
 VkResult MVKBuffer::flushToDevice(VkDeviceSize offset, VkDeviceSize size) {
 #if MVK_MACOS
     VkDeviceSize flushOffset, flushSize;
-	if (shouldFlushHostMemory() && overlaps(offset, size, flushOffset, flushSize)) {
-		memcpy(getMTLBuffer().contents, reinterpret_cast<const char *>(_deviceMemory->getHostMemoryAddress()) + flushOffset, flushSize);
-		[getMTLBuffer() didModifyRange: NSMakeRange(flushOffset - _deviceMemoryOffset, flushSize)];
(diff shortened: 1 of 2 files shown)


MoltenVK/MoltenVK/GPUObjects/MVKBuffer.mm
```diff
@@ -187,22 +187,21 @@
...
 			return _deviceMemory->getMTLBuffer();
 		}
 	}
 	return nil;
 }
 
+id<MTLBuffer> MVKBuffer::getMTLBufferCache() {
+    if (_isHostCoherentTexelBuffer && !_mtlBufferCache) {
+        _mtlBufferCache = [_device->getMTLDevice() newBufferWithLength: getByteCount()
+                                                               options: MTLResourceStorageModeManaged];    // retained
```
billhollings, line 200: `MTLResourceStorageModeManaged` breaks iOS & tvOS builds. Needs to be wrapped in `#if MVK_MACOS` conditional build.

js6i:
> Are you comfortable adding these locks here...following the pattern for `getMTLTexture()` implementations in `MVKBufferView` and `MVKImagePlane`?

I think for double-checking to be bullet proof, one also needs to use atomic operations with proper memory barriers. I'm not sure how it's handled in Objective-C++. Might be worth looking into to prevent heisenbugs..

billhollings:
> to be bullet proof...use atomic operations with proper memory barriers.

Good point. I'll look into this later on. Thanks for adding the basics for now.

Issue #918, opened 2020-06-12
VkPhysicalDeviceSubgroupProperties


Lichtso:
Based on https://github.com/KhronosGroup/SPIRV-Cross/pull/982 I would like to add the following struct to VkPhysicalDeviceProperties2:
```
case VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_SUBGROUP_PROPERTIES: {
	auto* physicalDeviceSubgroupProps = (VkPhysicalDeviceSubgroupProperties*)next;
    physicalDeviceSubgroupProps->supportedStages =
        VK_SHADER_STAGE_COMPUTE_BIT |
        VK_SHADER_STAGE_ALL_GRAPHICS;
    physicalDeviceSubgroupProps->supportedOperations =
        VK_SUBGROUP_FEATURE_BASIC_BIT |
        // VK_SUBGROUP_FEATURE_VOTE_BIT |
        VK_SUBGROUP_FEATURE_ARITHMETIC_BIT |
        VK_SUBGROUP_FEATURE_BALLOT_BIT |
        VK_SUBGROUP_FEATURE_SHUFFLE_BIT |
        VK_SUBGROUP_FEATURE_SHUFFLE_RELATIVE_BIT |
        VK_SUBGROUP_FEATURE_CLUSTERED_BIT |
        VK_SUBGROUP_FEATURE_QUAD_BIT;
        // VK_SUBGROUP_FEATURE_PARTITIONED_BIT_NV;
    physicalDeviceSubgroupProps->quadOperationsInAllStages = true;
	break;
}
```
@cdavis5e I understand there are some Metal specific limitations here, so are these bit-fields correct(ish)?

### subgroupSize
Of course, the biggest problem is the `subgroupSize` field.

One way would be to compile a minimal compute shader and ask it for its `MTLComputePipelineState.threadExecutionWidth`. However, I don't understand how this is shader code dependent and why it is not part of the physical device in Metal.

The other way is the hacky one:
According to http://gpuinfo.org all subgroupSizes (of GPUs supported by Metal) are 32 for Intel and Nvidia and 64 for AMD. So it could simply be done like this:
`physicalDeviceSubgroupProps->subgroupSize = (strstr(getName(), "AMD") || strstr(getName(), "Radeon")) ? 64 : 32;`

Edit: We can also just check the vendor ID instead of a string compare:
[...]

Issue #822, opened 2020-01-25
Support Vulkan 1.1


cdavis5e:
As of version 1.5.2, DXVK now requires a Vulkan 1.1 implementation. This means that until MoltenVK supports 1.1, we won't be able to use it with DXVK. I expect that the number of applications requiring Vulkan 1.1 is going to increase in the future as 1.0 ICDs become rarer. Much has been done to move towards this, but there's still much to be done.

Remaining 1.1 tasks:
- [x] `VK_KHR_external_fence` and `VK_KHR_external_fence_capabilities`
- [x] `VK_KHR_external_memory` and `VK_KHR_external_memory_capabilities`
- [x] `VK_KHR_external_semaphore` and `VK_KHR_external_semaphore_capabilities`
- [x] [`VK_KHR_multiview`](/KhronosGroup/MoltenVK/issues/347) - SPIRV-Cross work is already done
- [x] `VK_KHR_sampler_ycbcr_conversion` - SPIRV-Cross work is already done
- [ ] Feature struct for `VK_KHR_shader_draw_parameters`
- [x] Subgroup operations - SPIRV-Cross work is already done
- [ ] Protected memory - Metal exposes no API for this, but we still need to fill in the feature struct, if only to say we don't support it
- [ ] `vkEnumerateInstanceVersion()`

oscarbg:
just reposting here from dxvk issue:
also slightly related noticed recently two Vulkan projects have recently updated needing/benefiting also from Vulkan 1.1 support:
*VKD3D DXIL branch:
https://www.winehq.org/pipermail/wine-devel/2020-January/158470.html
Vulkan 1.1 for wave ops:
"Vulkan 1.1 is enabled if active, because subgroup operations requires it. SM 6.0 support is only activated if subgroup operations are sufficiently supported and DXIL is enabled.
There are other features required for SM 6.0 such as 16-bit arithmetic and storage, but that is left for later."
*VkBasalt (since 1.3) "Vulkan 1.1 is now required"
https://github.com/DadSchoorse/vkBasalt/releases/tag/v0.3.0

741g:
This is awesome news! Android/Fuchsia Emulator on macOS has heretofore used the Vulkan 1.0 or Swiftshader Vulkan 1.1 on macOS, but looks like we can use 1.1 there now which lets us ship with something supporting Vulkan 1.1 on all 3 platforms (macos/windows/linux) :)

PR #932, opened 2020-06-21
Adds VkPhysicalDeviceSubgroupProperties.

Lichtso:
As discussed in #918.
This comes with a fixed subgroupSize per physical device until VK_EXT_subgroup_size_control is implemented.

### DIFF
--- a/MoltenVK/MoltenVK/API/vk_mvk_moltenvk.h
+++ b/MoltenVK/MoltenVK/API/vk_mvk_moltenvk.h
@@ -611,7 +611,7 @@ typedef struct {
 	VkBool32 placementHeaps;					/**< If true, MTLHeap objects support placement of resources. */
 	VkDeviceSize pushConstantSizeAlignment;		/**< The alignment used internally when allocating memory for push constants. Must be PoT. */
 	uint32_t maxTextureLayers;					/**< The maximum number of layers in an array texture. */
-    uint32_t subgroupSize;			            /**< The number of threads in a wavefront. */
+    uint32_t subgroupSize;			            /**< The number of threads in a SIMD-group. */
 } MVKPhysicalDeviceMetalFeatures;
 
 /** MoltenVK performance of a particular type of activity. */
--- a/MoltenVK/MoltenVK/API/vk_mvk_moltenvk.h
+++ b/MoltenVK/MoltenVK/API/vk_mvk_moltenvk.h
@@ -611,6 +611,7 @@ typedef struct {
 	VkBool32 placementHeaps;					/**< If true, MTLHeap objects support placement of resources. */
 	VkDeviceSize pushConstantSizeAlignment;		/**< The alignment used internally when allocating memory for push constants. Must be PoT. */
 	uint32_t maxTextureLayers;					/**< The maximum number of layers in an array texture. */
+    uint32_t subgroupSize;			            /**< The number of threads in a wavefront. */
 } MVKPhysicalDeviceMetalFeatures;
 
 /** MoltenVK performance of a particular type of activity. */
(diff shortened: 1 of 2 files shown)


MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
```diff
@@ -233,6 +233,29 @@
...
 			}
+#if MVK_MACOS
+            case VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_SUBGROUP_PROPERTIES:
+                if (mvkOSVersionIsAtLeast(10.14)) {
+                    auto* subgroupProps = (VkPhysicalDeviceSubgroupProperties*)next;
+                    subgroupProps->subgroupSize = _metalFeatures.subgroupSize;
+                    subgroupProps->supportedStages =
+                        VK_SHADER_STAGE_COMPUTE_BIT |
+                        VK_SHADER_STAGE_TESSELLATION_CONTROL_BIT |
+                        VK_SHADER_STAGE_TESSELLATION_EVALUATION_BIT |
```
cdavis5e, line 244: This bit shouldn't be set, so just `COMPUTE | TESSELLATION_CONTROL | FRAGMENT`, then.

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
```diff
@@ -1094,6 +1117,9 @@
         }
     }
 
+    static const uint32_t kAMDVendorId = 0x1002;
+    _metalFeatures.subgroupSize = (_properties.vendorID == kAMDVendorId) ? 64 : 32;
+
```
cdavis5e, line 1122: It's possible to fetch the `_metalFeatures` struct from user code using `VK_MVK_moltenvk`. We probably shouldn't set this unconditionally... or really, at all on iOS or tvOS until we figure out what to do there.

MoltenVK/MoltenVK/API/vk_mvk_moltenvk.h
```diff
@@ -611,6 +611,7 @@ typedef struct {
 	VkBool32 placementHeaps;					/**< If true, MTLHeap objects support placement of resources. */
 	VkDeviceSize pushConstantSizeAlignment;		/**< The alignment used internally when allocating memory for push constants. Must be PoT. */
 	uint32_t maxTextureLayers;					/**< The maximum number of layers in an array texture. */
+    uint32_t subgroupSize;			            /**< The number of threads in a wavefront. */
```
cdavis5e, line 614: "Wavefront" is a term specific to Direct3D and to AMD cards. This should probably use the Metal equivalent term "SIMD-group."

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
```diff
@@ -1094,6 +1116,13 @@
         }
     }
 
+#if MVK_MACOS
+    if (mvkOSVersionIsAtLeast(10.14)) {
+        static const uint32_t kAMDVendorId = 0x1002;
+        _metalFeatures.subgroupSize = (_properties.vendorID == kAMDVendorId) ? 64 : 32;
```
billhollings, line 1122: This feels a bit hacky and subject to future unexpected behaviour. 

Is creating a `MTLComputePipelineState` and using `threadExecutionWidth` the only way to get this programmatically? 

Or is it possible that this test could be done with reference to `MTLFeatureSet` instead (AMD referencing a particular GPU type)?

cdavis5e: > Is creating a `MTLComputePipelineState` and using `threadExecutionWidth` the only way to get this value programmatically?

Currently, yes. I've asked them to provide queries for the minimum and maximum thread execution width, but I think I asked for it too late for macOS 11.

billhollings: > Currently, yes

And 64/32 it can't realistically be differentiated via a `MTLFeatureSet` (ie. GPU family)?

cdavis5e: > > Currently, yes
> 
> And 64/32 it can't realistically be differentiated via a `MTLFeatureSet` (ie. GPU family)?

No. Modern AMD and Intel chips both support `MTLGPUFamilyMac2`, but Intel uses 32 while AMD uses 64.

billhollings:
@cdavis5e Can you have a look at the updates after your first review, and approve if good now, please?

Lichtso:
Afterwards #822 can be updated again, or is there something missing for the "Subgroup operations" task?

PR #939, opened 2020-06-25
MVKDevice: Advertise VK_FORMAT_A2B10G10R10_UNORM_PACK32 as a surface format.

cdavis5e:
The corresponding Metal format is now documented as supported for
`CAMetalLayer`, confirming my findings in #673.

Only support the 10-10-10-2 formats on macOS. iOS and tvOS instead have
extended-range formats, which unfortunately have no Vulkan equivalents.
I've added them anyway for completeness. I suppose if Vulkan ever does
get these formats, we'll then be ready to expose them as surface
formats.

### DIFF
--- a/MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
@@ -619,7 +619,16 @@ VkResult MVKPhysicalDevice::getSurfaceFormats(MVKSurface* surface,
 		MTLPixelFormatBGRA8Unorm,
 		MTLPixelFormatBGRA8Unorm_sRGB,
 		MTLPixelFormatRGBA16Float,
+#if MVK_MACOS
+		MTLPixelFormatRGB10A2Unorm,
 		MTLPixelFormatBGR10A2Unorm,
+#endif
+#if MVK_IOS_OR_TVOS
+		MTLPixelFormatBGRA10_XR,
+		MTLPixelFormatBGRA10_XR_sRGB,
+		MTLPixelFormatBGR10_XR,
+		MTLPixelFormatBGR10_XR_sRGB,
+#endif
 	};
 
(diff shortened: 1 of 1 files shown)


tmm1:
Looks good!

How about `MTLPixelFormatRGBA16Unorm` (i.e. `VK_FORMAT_R16G16B16A16_UNORM`). Is that supported as a surface format anywhere (doesn't appear to be on tvOS atleast).

cdavis5e:
> How about `MTLPixelFormatRGBA16Unorm` (i.e. `VK_FORMAT_R16G16B16A16_UNORM`). Is that supported as a surface format anywhere (doesn't appear to be on tvOS atleast).

Nope.

cdavis5e:
What happened? This time, it didn't even fail to build the external dependencies. It just abruptly aborted with no output after finishing that.

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
```diff
@@ -670,7 +679,12 @@
 	}
 
 	uint mtlFmtsCnt = sizeof(mtlFormats) / sizeof(MTLPixelFormat);
+#if MVK_MACOS
```
billhollings, line 682: I know this is not new in this update...but the way that the dynamic nature of the value of `mtlFmsCnt` is used to define indexing into `mtlFormats` depends on the last consecutive elements of `mtlFormats` being the only ones that are dynamically elided. 

This works for now, but feels fragile. For example...if a device in the future removed any except the last of the four `XR` formats...it would trigger undefined behaviour.

billhollings:
>This works for now, but feels fragile.

Fixed in #943.

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
billhollings: Fixed in #943.

Issue #933, opened 2020-06-23
Add support for VK_KHR_driver_properties


flibitijibibo:
At a surface level this shouldn't be too hard to implement, it's one case to add to this function...

https://github.com/KhronosGroup/MoltenVK/blob/2dd34de8ee8823b9facedbc071a225191b57b6df/MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm#L175

I don't know what `driverID` would be though. We would be happy to have the driver just be "MoltenVK 1.2.141.1" for example, but I don't know how much Metal driver info we could get from the OS.

cdavis5e:
> I don't know what `driverID` would be though.

Probably something specific to MoltenVK. DXVK also asked for this, and they specifically wanted this to be able to ID MoltenVK and enable driver-specific hacks for it.

flibitijibibo:
This should be resolved once the headers have been updated to 1.x.148 and `VK_DRIVER_ID_MOLTENVK` is filled in here:

https://github.com/KhronosGroup/MoltenVK/blob/master/MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm#L262

billhollings:
> This should be resolved once the headers have been updated to 1.x.148 and `VK_DRIVER_ID_MOLTENVK` is filled in here:
> 
> https://github.com/KhronosGroup/MoltenVK/blob/master/MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm#L262

Yup. PR #966 is in flight. 😉

PR #948, opened 2020-07-07
Implement VK_KHR_driver_properties extension

TheSpydog:
This implements the `VK_KHR_driver_properties` extension, per [this issue](https://github.com/KhronosGroup/MoltenVK/issues/933). Thankfully it was pretty straightforward to add. However, there are a couple problems that need to be resolved before merging:

1. MoltenVK needs a dedicated Driver ID. I couldn't find any instructions on how to add one, but I suspect the process is similar to [registering Vendor IDs and Author IDs](https://www.khronos.org/registry/vulkan/specs/1.2/styleguide.html#extensions-author-ID). I think it makes sense for an MVK maintainer to do this step instead of me.
2. I'm not sure what to put for the "subminor" section in the properties. I would assume this should be 0, but please correct me if I'm wrong.


MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
```diff
@@ -255,6 +255,19 @@
...
+				auto* physicalDeviceDriverProps = (VkPhysicalDeviceDriverPropertiesKHR*)next;
+				char mvkVersion[16], vkVersion[16];
+				vkGetVersionStringsMVK(mvkVersion, sizeof(mvkVersion), vkVersion, sizeof(vkVersion));
+				strcpy(physicalDeviceDriverProps->driverName, "MoltenVK");
+				strcpy(physicalDeviceDriverProps->driverInfo, mvkVersion);
+				physicalDeviceDriverProps->driverID = (VkDriverId) 0; /* FIXME: VK_DRIVER_ID_MOLTENVK_KHR */
+				physicalDeviceDriverProps->conformanceVersion.major = VK_VERSION_MAJOR(VK_API_VERSION_1_0);
+				physicalDeviceDriverProps->conformanceVersion.minor = VK_VERSION_MINOR(VK_API_VERSION_1_0);
+				physicalDeviceDriverProps->conformanceVersion.subminor = 0; /* FIXME: What is this? */
+				physicalDeviceDriverProps->conformanceVersion.patch = VK_VERSION_PATCH(VK_HEADER_VERSION);
```
cdavis5e, line 268: So this is actually supposed to be the version of the [Vulkan conformance test suite](/KhronosGroup/VK-GL-CTS) that the driver has been validated against. To my knowledge, MoltenVK has not yet passed verification against any version of the CTS... so I'm not sure what to actually put here.

TheSpydog: Hmm. Maybe 0.0.0.0 then?

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
```diff
@@ -255,6 +255,19 @@
                 }
 				break;
 #endif
+			case VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_DRIVER_PROPERTIES: {
+				auto* physicalDeviceDriverProps = (VkPhysicalDeviceDriverPropertiesKHR*)next;
+				char mvkVersion[16], vkVersion[16];
+				vkGetVersionStringsMVK(mvkVersion, sizeof(mvkVersion), vkVersion, sizeof(vkVersion));
+				strcpy(physicalDeviceDriverProps->driverName, "MoltenVK");
+				strcpy(physicalDeviceDriverProps->driverInfo, mvkVersion);
+				physicalDeviceDriverProps->driverID = (VkDriverId) 0; /* FIXME: VK_DRIVER_ID_MOLTENVK_KHR */
```
billhollings, line 264: I've submitted a request to the Vulkan Working Group to add `VK_DRIVER_ID_MOLTENVK` to `VkDriverId`. 

Once it gets approved and released into the Vulkan headers, I'll let you know, and you can modify this accordingly, and update the value in `Vulkan-Headers_repo_revision` to use the updated Vulkan headers.

We'll hold this PR here until that happens.

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
```diff
@@ -255,6 +255,19 @@
                 }
 				break;
 #endif
+			case VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_DRIVER_PROPERTIES: {
+				auto* physicalDeviceDriverProps = (VkPhysicalDeviceDriverPropertiesKHR*)next;
+				char mvkVersion[16], vkVersion[16];
+				vkGetVersionStringsMVK(mvkVersion, sizeof(mvkVersion), vkVersion, sizeof(vkVersion));
+				strcpy(physicalDeviceDriverProps->driverName, "MoltenVK");
+				strcpy(physicalDeviceDriverProps->driverInfo, mvkVersion);
```
billhollings, line 263: Might as well just use `mvkGetMoltenVKVersionString(MVK_VERSION).c_str()` right in the `strcpy()` arg, and skip `mvkVersion` and `vkGetVersionStringsMVK()` altogether.

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
billhollings: `VK_DRIVER_ID_MOLTENVK ` is available now. I'm going to pull this in and add that.

PR #969, opened 2020-07-22
Add support for USCALED/SSCALED vertex formats

TheSpydog:
It appears the `USCALED` and `SSCALED` vertex formats should be the exact same as their `UINT/SINT` counterparts, just with `ColorFloat` instead of `ColorInt8/UInt8`. The only parts I'm not sure about are the "alt" formats (`UChar2` for `R8_USCALED`, for example) -- for those I just followed what the int formats did.

I tested this patch with the work-in-progress Vulkan backend for the game Murder Miners, which extensively uses *SCALED vertex formats.

### DIFF
--- a/MoltenVK/MoltenVK/GPUObjects/MVKPixelFormats.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKPixelFormats.mm
@@ -580,24 +580,24 @@ void MVKPixelFormats::initVkFormatCapabilities() {
 
 	addVkFormatDesc( R8_UNORM, R8Unorm, Invalid, UCharNormalized, UChar2Normalized, 1, 1, 1, ColorFloat );
 	addVkFormatDesc( R8_SNORM, R8Snorm, Invalid, CharNormalized, Char2Normalized, 1, 1, 1, ColorFloat );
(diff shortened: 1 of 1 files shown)


flibitijibibo:
Our Vulkan renderer's use of the `SCALED` formats, for reference:

https://github.com/FNA-XNA/FNA3D/blob/master/src/FNA3D_Driver_Vulkan.c#L1065

This will most likely be used with DXVK as well, as this is part of reimplementing support for D3D9 vertex data.

billhollings:
I'm glad it's working successfully. But I'm curious how this is handled in the MSL shader.

With a vertex attribute set to say `MTLVertexFormatUChar`, the attribute would be treated as a single byte of memory. But presumably the SPIR-V `OpTypeFloat` would cause the shader to declare this attribute as a `float`.

Without a specific _scaled_ `MTLVertexFormat` variant, how does Metal know that this single byte should be converted to a float value between `0.0` and `255.0`? I guess as a C++ variant, MSL just reads it as a `0` - `255` `uint8_t` and then automatically casts it to the `float`?

billhollings:
> I'm not sure about are the "alt" formats

You chose the right option! 😉

The alternates exist because the scalar `MTLVertexFormats` (`Char`/`Uchar`/`Short`/`Ushort`) were added later to Metal (macOS 10.13 and iOS 11). So the fall-back on older OS's is to treat the attribute as a two-element vector where the second element is not used.

PR #972, opened 2020-07-23
Various fixes

### LINKED ISSUE #962: Linear textures created with MTLTextureUsageRenderTarget usage
When running my application I get the following assertion from Metal validation: 
`[MTLDebugBuffer newTextureWithDescriptor:offset:bytesPerRow:]:326: failed assertion 'linear textures cannot be created with usage MTLTextureUsageRenderTarget'`.

It seems that in https://github.com/KhronosGroup/MoltenVK/blob/f967fb6aebcce02d02ad9bc249f45cb5cf4a11f6/MoltenVK/MoltenVK/GPUObjects/MVKImage.mm#L101

at this line
https://github.com/KhronosGroup/MoltenVK/blob/f967fb6aebcce02d02ad9bc249f45cb5cf4a11f6/MoltenVK/MoltenVK/GPUObjects/MVKImage.mm#L123

it is determined that the image should have `MTLTextureUsageRenderTarget`. This seems to be due to the fact that this block
https://github.com/KhronosGroup/MoltenVK/blob/21630b59bb65ae6a329186786b5eb02c2a37f881/MoltenVK/MoltenVK/GPUObjects/MVKPixelFormats.mm#L482-L487

checks if my buffer has `VK_IMAGE_USAGE_TRANSFER_DST_BIT` usage bit (which it does) and determines it should use `MTLTextureUsageRenderTarget`.

which is not valid when
https://github.com/KhronosGroup/MoltenVK/blob/f967fb6aebcce02d02ad9bc249f45cb5cf4a11f6/MoltenVK/MoltenVK/GPUObjects/MVKImage.mm#L55-L58

is called. 

As stated by the documentation: https://developer.apple.com/documentation/metal/mtlbuffer/1613852-newtexturewithdescriptor

```
On macOS, usage must not include the MTLTextureUsageRenderTarget option. The texture cannot be used as a render target.
```

Note that I am reproducing this on the last stable release (e9fc62b27dcd964526fd0c0746679c44bc161701), not the latest master. Looking at the code is issue appears to still be present, but I cannot confirm.

BearishSun:
This PR is split into 4 commits, fixing four separate issues:
1. Fixes #962 . Linear textures shouldn't be created with MTLTextureUsageRenderTarget on macOS, otherwise Metal validation complains.
2. `MVKImagePlane::overlaps` did not account for device memory offset. Instead it was just using the subresource layout offset, which is relative to the image memory binding (usually 0).
3. `MVKImage._byteCount` was incorrect in some cases. In my case I had a 16x16 RGBA16F texture that reported row pitch of 256 bytes, yet the size of the entire image was 2048 bytes, which is too small (it should be 16x256=4096). 
This was because `MVKImage::getBytesPerLayer` did not account for the row byte alignment, which was 256 in my case - the actual row size was 128 bytes, rounded to byte alignment it was 256, as reported by `MVKImage::getBytesPerRow`. Yet `MVKImage::getBytesPerLayer` returned 2048. 
4. This is a minor one but hopefully it doesn't hurt. When calculating mip level size we were always multiplying depth pitch by the 0th level mip (`_extents.depth`), rather than the current mip level's. Unsure if this was done for a reason.

### DIFF
--- a/MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
@@ -536,7 +536,7 @@ VkDeviceSize MVKImage::getBytesPerRow(uint8_t planeIndex, uint32_t mipLevel) {
 
 VkDeviceSize MVKImage::getBytesPerLayer(uint8_t planeIndex, uint32_t mipLevel) {
     VkExtent3D extent = getExtent3D(planeIndex, mipLevel);
-    size_t bytesPerRow = mvkAlignByteCount(getPixelFormats()->getBytesPerRow(_vkFormat, extent.width), _rowByteAlignment);
+    size_t bytesPerRow = getBytesPerRow(planeIndex, mipLevel);
     return getPixelFormats()->getBytesPerLayer(_vkFormat, bytesPerRow, extent.height);
 }
 
--- a/MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
@@ -120,7 +120,7 @@ MTLTextureDescriptor* MVKImagePlane::newMTLTextureDescriptor() {
     mtlTexDesc.mipmapLevelCount = _image->_mipLevels;
     mtlTexDesc.sampleCount = mvkSampleCountFromVkSampleCountFlagBits(_image->_samples);
     mtlTexDesc.arrayLength = _image->_arrayLayers;
-    mtlTexDesc.usageMVK = _image->getPixelFormats()->getMTLTextureUsage(_image->_usage, mtlPixFmt, minUsage);
+    mtlTexDesc.usageMVK = _image->getPixelFormats()->getMTLTextureUsage(_image->_usage, mtlPixFmt, minUsage, _image->_isLinear);
     mtlTexDesc.storageModeMVK = _image->getMTLStorageMode();
     mtlTexDesc.cpuCacheMode = _image->getMTLCPUCacheMode();
 
--- a/MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
@@ -533,7 +533,7 @@ VkDeviceSize MVKImage::getBytesPerRow(uint8_t planeIndex, uint32_t mipLevel) {
 
 VkDeviceSize MVKImage::getBytesPerLayer(uint8_t planeIndex, uint32_t mipLevel) {
     VkExtent3D extent = getExtent3D(planeIndex, mipLevel);
-    size_t bytesPerRow = getPixelFormats()->getBytesPerRow(_vkFormat, extent.width);
+    size_t bytesPerRow = mvkAlignByteCount(getPixelFormats()->getBytesPerRow(_vkFormat, extent.width), _rowByteAlignment);
     return getPixelFormats()->getBytesPerLayer(_vkFormat, bytesPerRow, extent.height);
 }
 
--- a/MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
@@ -267,8 +267,8 @@ void MVKImagePlane::getMTLTextureContent(MVKImageSubresource& subresource,
 bool MVKImagePlane::overlaps(VkSubresourceLayout& imgLayout, VkDeviceSize offset, VkDeviceSize size) {
 	VkDeviceSize memStart = offset;
 	VkDeviceSize memEnd = offset + size;
-	VkDeviceSize imgStart = imgLayout.offset;
-	VkDeviceSize imgEnd = imgLayout.offset + imgLayout.size;
+	VkDeviceSize imgStart = getMemoryBinding()->_deviceMemoryOffset + imgLayout.offset;
+	VkDeviceSize imgEnd = imgStart + imgLayout.size;
 	return imgStart < memEnd && imgEnd > memStart;
 }
 
--- a/MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
@@ -147,12 +147,15 @@ void MVKImagePlane::initSubresources(const VkImageCreateInfo* pCreateInfo) {
         VkDeviceSize rowPitch = _image->getBytesPerRow(_planeIndex, mipLvl);
         VkDeviceSize depthPitch = _image->getBytesPerLayer(_planeIndex, mipLvl);
 
+        VkExtent3D mipExtent = _image->getExtent3D(_planeIndex, mipLvl);
+        
         for (uint32_t layer = 0; layer < _image->_arrayLayers; layer++) {
             subRez.subresource.arrayLayer = layer;
 
             VkSubresourceLayout& layout = subRez.layout;
             layout.offset = offset;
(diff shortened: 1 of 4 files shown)


MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
```diff
@@ -533,7 +536,7 @@
 
 VkDeviceSize MVKImage::getBytesPerLayer(uint8_t planeIndex, uint32_t mipLevel) {
     VkExtent3D extent = getExtent3D(planeIndex, mipLevel);
-    size_t bytesPerRow = getPixelFormats()->getBytesPerRow(_vkFormat, extent.width);
+    size_t bytesPerRow = mvkAlignByteCount(getPixelFormats()->getBytesPerRow(_vkFormat, extent.width), _rowByteAlignment);
```
billhollings, line 539: Could this line be consolidated as:

    size_t bytesPerRow = getBytesPerRow(planeIndex, mipLevel);

which would consolidate format and alignment management into `getBytesPerRow()`. That might reduce fragility moving forward.

MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
```diff
@@ -147,12 +147,15 @@
...
 
+        VkExtent3D mipExtent = _image->getExtent3D(_planeIndex, mipLvl);
+        
         for (uint32_t layer = 0; layer < _image->_arrayLayers; layer++) {
             subRez.subresource.arrayLayer = layer;
 
             VkSubresourceLayout& layout = subRez.layout;
             layout.offset = offset;
-            layout.size = depthPitch * _image->_extent.depth;
+            layout.size = depthPitch * mipExtent.depth;
```
billhollings, line 157: Oops! 🤦🏻‍♂️ Nice catch!

MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
```diff
@@ -267,8 +270,8 @@
 bool MVKImagePlane::overlaps(VkSubresourceLayout& imgLayout, VkDeviceSize offset, VkDeviceSize size) {
 	VkDeviceSize memStart = offset;
 	VkDeviceSize memEnd = offset + size;
-	VkDeviceSize imgStart = imgLayout.offset;
-	VkDeviceSize imgEnd = imgLayout.offset + imgLayout.size;
+	VkDeviceSize imgStart = getMemoryBinding()->_deviceMemoryOffset + imgLayout.offset;
+	VkDeviceSize imgEnd = imgStart + imgLayout.size;
```
billhollings, line 274: Thanks!

MoltenVK/MoltenVK/GPUObjects/MVKPixelFormats.mm
```diff
@@ -484,7 +485,13 @@
...
 		mvkIsAnyFlagEnabled(mtlFmtCaps, (kMVKMTLFmtCapsColorAtt | kMVKMTLFmtCapsDSAtt))) {
 
-		mvkEnableFlags(mtlUsage, MTLTextureUsageRenderTarget);
+#if MVK_MACOS
+        if(!isLinear) {
+            mvkEnableFlags(mtlUsage, MTLTextureUsageRenderTarget);
+        }
+#else
+        mvkEnableFlags(mtlUsage, MTLTextureUsageRenderTarget);
+#endif
```
billhollings, line 494: This is an interesting solution. And there are examples of us handling other OS variations along these lines.

I am a bit concerned that silently handling this in `MVKPixelFormats::getMTLTextureUsage()` introduces a functional change that might surprise the app with unexpected behaviour. It might just defer the problem, as any attempt by the app to use the linear texture as a BLIT destination will cause a Metal error.

In `MVKImage::validateLinear()`, we perform the equivalent test to see if the linear image might be used as an attachment, and we report an error if so. 

I'm wondering if adding a similar check and error report to `MVKImage::validateLinear()` if `VK_IMAGE_USAGE_TRANSFER_DST_BIT` is set would be an appropriate answer.

The spec says one of the restrictions for linear images is:

>usage only includes `VK_IMAGE_USAGE_TRANSFER_SRC_BIT` and/or `VK_IMAGE_USAGE_TRANSFER_DST_BIT`

And I recognize that `VK_IMAGE_USAGE_TRANSFER_DST_BIT` covers both BLIT and copy...so perhaps leaving the image usage at only `VK_IMAGE_USAGE_TRANSFER_SRC_BIT` usage might be too restrictive.

Thoughts?

billhollings: If silently eliminating `MTLTextureUsageRenderTarget` as is done here does prove useful and necessary, we should add a validation error test to `MVKCmdBlitImage<N>::setContent()` similar to what is done for a depth/stencil format for the same reasons (but only for macOS).

I can add that if needed.

MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
BearishSun: I have updated the code as you requested.

MoltenVK/MoltenVK/GPUObjects/MVKPixelFormats.mm
BearishSun: Adding a check to `MVKImage::validateLinear()` for `VK_IMAGE_USAGE_TRANSFER_DST_BIT` would just move the error to a different place for me. 

My issue is that I have a HOST_VISIBLE | HOST_COHERENT 2D texture that I map/unmap in order to write from the CPU, and it then gets SHADER_READ by a shader. This texture gets assigned a render target usage in MVK and triggers the Metal validation error.

I'm not sure what is the correct solution for MVK's design, I just know the above is causing problems for my use case.

billhollings: >HOST_VISIBLE | HOST_COHERENT 2D texture that I map/unmap in order to write from the CPU, and it then gets SHADER_READ by a shader

Yes...that's a typical use of linear textures. With that requirement, why does the image need `VK_IMAGE_USAGE_TRANSFER_DST_BIT`, which would imply you intend to execute a [transfer command](https://www.khronos.org/registry/vulkan/specs/1.0-extensions/html/vkspec.html#synchronization-pipeline-stages-transfer) to it (see definition of `VK_PIPELINE_STAGE_TRANSFER_BIT`)?

Sorry for belabouring this...and I'm not questioning your use of `VK_IMAGE_USAGE_TRANSFER_DST_BIT` per se, because as I mention above, Vulkan indeed allows both `VK_IMAGE_USAGE_TRANSFER_SRC_BIT` and `VK_IMAGE_USAGE_TRANSFER_DST_BIT` for linear textures.

But in this case, the use case matters...because I'm trying to determine what the appropriate use cases should be for a linear texture, given the restrictions of Metal's transfer command capabilities, and whether it would be a fair restriction to disallow `VK_IMAGE_USAGE_TRANSFER_DST_BIT` for linear textures on macOS.

@cdavis5e I'm happy to hear your thoughts on these options too.

BearishSun: We don't actually use the `VK_IMAGE_USAGE_TRANSFER_DST_BIT`. But our user-facing API is higher level and we assign the flag for the eventuality the user decides to execute the transfer operation.

billhollings: Okay. Accepting this, and then I'll add the [suggested validation code](https://github.com/KhronosGroup/MoltenVK/pull/972#discussion_r459593524).

Issue #959, opened 2020-07-14
Issue with buffer bindings for multiple SBOs


sungiant:
Hi, I'm having trouble with bindings for compute shaders that use multiple storage buffers, perhaps I've found an issue?

I consistently get the following type of error when queueing up work from compute pipelines with descriptor set layouts and associated shader modules that use more than one storage buffer:

```
validateComputeFunctionArguments:762: failed assertion `Compute Function(main0): missing buffer binding at index 5 for spvBufferSizeConstants[0].'
```
- I'm using OSX and MoltenVK.
- I've tested on multiple machines.
- On Windows with Vulkan the same examples do not have issues.
- Being concerned that I had something wrong with API usage in my test case I created another test case based on [this](https://github.com/SaschaWillems/Vulkan/blob/master/examples/computeraytracing/computeraytracing.cpp) example - I continue to experience the same kind of error with MoltenVK.

Any help would be much appreciated.  Thanks.

billhollings:
This is fixed in PR #979. Please retest and close this issue if it fixes it for you.

sungiant:
Brilliant, thank you.  I've just tested my project with these changes and the issue looks to be resolved.

PR #979, opened 2020-07-28
Fix issue where expected buffer-sizes buffer not bound to Metal compute encoder.

billhollings:
`MVKComputeResourcesCommandEncoderState` update buffer-size value before buffer
bindings are encoded into Metal and are no longer marked as dirty.

Fixes issue #959.


cdavis5e:
We've been seeing issues running Dark Souls III with CrossOver on Mojave, and I traced them back to this change. Specifically, we were seeing a weird effect where a multicolored overlay would make the game nigh unplayable. I'm not sure why this change causes it. Further, we've only seen this on Mojave; we don't see it on Catalina or Big Sur. Unfortunately, because Mojave doesn't support GPU capture without Xcode, I haven't been able to analyze a GPU trace to figure out why this change causes the issue.

PR #980, opened 2020-07-28
dynamicOffsets are supposed to be ordered by binding index

### LINKED ISSUE #978: Dynamic Offsets in vkCmdBindDescriptorSet seem incorrect
Dynamic offsets are supposed to be ordered in this way: "entries are ordered by the binding numbers in the descriptor set layouts". That is, if my layout has two buffers, the first using binding=1 with desired offset 1024 and the second using binding=0 with desired offset 256, the dynamic offsets should be provided as [256, 1024].
MoltenVK seems to read through the dynamic offsets in the order of the elements in the layout, not by their binding order.
Am I correct that this behavior is a bug?

mbechard:
not by layout element index. Fixes #978

**Note**: I'm not sure if something different needs to be done for push descriptors though. Unfamiliar with those.

### DIFF
--- a/MoltenVK/MoltenVK/GPUObjects/MVKDescriptorSet.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKDescriptorSet.mm
@@ -39,7 +39,7 @@ void MVKDescriptorSetLayout::bindDescriptorSet(MVKCommandEncoder* cmdEncoder,
                                                MVKDescriptorSet* descSet,
                                                MVKShaderResourceBinding& dslMTLRezIdxOffsets,
                                                MVKArrayRef<uint32_t> dynamicOffsets,
-                                               uint32_t* pDynamicOffsetIndex) {
+                                               uint32_t baseDynamicOffsetIndex) {
     if (_isPushDescriptorLayout) return;
 
 	clearConfigurationResult();
@@ -47,7 +47,7 @@ void MVKDescriptorSetLayout::bindDescriptorSet(MVKCommandEncoder* cmdEncoder,
     for (uint32_t descIdx = 0, bindIdx = 0; bindIdx < bindCnt; bindIdx++) {
 		descIdx += _bindings[bindIdx].bind(cmdEncoder, descSet, descIdx,
 										   dslMTLRezIdxOffsets, dynamicOffsets,
-										   pDynamicOffsetIndex);
+										   baseDynamicOffsetIndex);
     }
 }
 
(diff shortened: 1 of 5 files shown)


MoltenVK/MoltenVK/GPUObjects/MVKDescriptorSet.mm
```diff
@@ -39,15 +39,15 @@
...
+                                               uint32_t baseDynamicOffsetIndex) {
     if (_isPushDescriptorLayout) return;
 
 	clearConfigurationResult();
     size_t bindCnt = _bindings.size();
     for (uint32_t descIdx = 0, bindIdx = 0; bindIdx < bindCnt; bindIdx++) {
 		descIdx += _bindings[bindIdx].bind(cmdEncoder, descSet, descIdx,
 										   dslMTLRezIdxOffsets, dynamicOffsets,
-										   pDynamicOffsetIndex);
+										   baseDynamicOffsetIndex);
```
billhollings, line 50: I'm wondering if we can simplify this overall solution to a single change to this `MVKDescriptorSetLayout::bindDescriptorSet()` function. Something like...

Creating a temp copy of the portion of `dynamicOffsets` that is covered by this descriptor set here, and that is sorted by descriptor order instead of binding number, and then submitting that re-sorted array to the individual bindings here.

Each `MVKBufferDescriptor` would then just do what it currently does, using the next value, and incrementing the overall `pDynamicOffsetIndex` as it does now.

Making the `_bindingToIndex` an ordered `map` instead of `unordered_map` might be a help in the task of re-sorting of the elements of `dynamicOffsets` on the fly (ie- iterate through `_bindingToIndex` and for each dynamic buffer descriptor found, move the next value in `dynamicOffsets` into the temp array).

Thoughts?

billhollings:
Thanks for catching the spec requirement!

>entries are ordered by the binding numbers in the descriptor set layouts

What a strange requirement, to have part of the incoming data sorted differently than the main descriptor elements themselves!

And thanks for submitting this solution!

If dealing with the above offset-ordering is the primary goal of this fix, I've made a suggestion above about how we might simplify accomplishing that. Have a look and let me know if that makes sense. I may have missed something. 

I can take on the task of working on that alternate solution if you'd prefer.

MoltenVK/MoltenVK/GPUObjects/MVKDescriptorSet.mm
mbechard: Thanks for the quick response. Totally your call here. I designed it this way to avoid any memory allocation or sorting during the bind* operation since that is on the critical fast path. I was worried about cases with large amounts of descriptors, and since this solution is just an extra uint32_t stored per descriptor/descriptorArray it seems reasonable.

billhollings: Okay. Yeah...that does make sense.

PR #982, opened 2020-08-01
Enable building on xcode 12 beta

jherico:
Quiet warnings and errors building on Xcode 12 Beta.

### DIFF
--- a/MoltenVK/MoltenVK/Commands/MVKCommandResourceFactory.h
+++ b/MoltenVK/MoltenVK/Commands/MVKCommandResourceFactory.h
@@ -180,7 +180,7 @@ typedef struct MVKMTLStencilDescriptorData {
         // even if the structure contains alignment gaps.
         mvkClear(this);
 
-        enabled = false,
+        enabled = false;
         stencilCompareFunction = MTLCompareFunctionAlways;
         stencilFailureOperation = MTLStencilOperationKeep;
         depthFailureOperation = MTLStencilOperationKeep;
--- a/MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
@@ -1162,6 +1162,9 @@ void MVKPhysicalDevice::initMetalFeatures() {
 	_metalFeatures.mslVersion = SPIRV_CROSS_NAMESPACE::CompilerMSL::Options::make_msl_version(maj, min);
 
 	switch (_metalFeatures.mslVersionEnum) {
+		case MTLLanguageVersion2_3:
+			setMSLVersion(2, 3);
+			break;
 		case MTLLanguageVersion2_2:
 			setMSLVersion(2, 2);
 			break;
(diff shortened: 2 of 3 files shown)


MoltenVK/MoltenVK/Utility/MVKFoundation.h
```diff
@@ -24,6 +24,7 @@
 #include "mvk_vulkan.h"
 #include <algorithm>
 #include <string>
+#include <cassert>
```
jherico, line 27: The assert later in this file fails to compile on Xcode-beta 12 without the inclusion of this header.

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
```diff
@@ -1162,6 +1162,7 @@
 	_metalFeatures.mslVersion = SPIRV_CROSS_NAMESPACE::CompilerMSL::Options::make_msl_version(maj, min);
 
 	switch (_metalFeatures.mslVersionEnum) {
+		default:
```
jherico, line 1165: Present a warning about missing case for `MTLLanguageVersion2_3`

MoltenVK/MoltenVK/Commands/MVKCommandResourceFactory.h
```diff
@@ -180,7 +180,7 @@ typedef struct MVKMTLStencilDescriptorData {
         // even if the structure contains alignment gaps.
         mvkClear(this);
 
-        enabled = false,
+        enabled = false;
```
jherico, line 183: prevent warning about improper use of comma

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
billhollings: The `default:` case is deliberately left off here for exactly this reason...to get warnings when new MSL versions are added, so that we remember to include them.

With that in mind, I've created a new [`xcode12`](https://github.com/KhronosGroup/MoltenVK/tree/xcode12) branch to this repo, where we can add all sorts of new functionality using the API's supported by Xcode 12, including `MTLLanguageVersion2_3`.

Later this year, once Xcode 12 and the OS versions it supports are released, I'll merge that branch with master for general use.

Can you add `MTLLanguageVersion2_3` appropriately to the top of the switch statement in this function, and submit a new pull request targeting the `xcode12` branch of this repo, please?

If you want to be thorough, also update `MVKPhysicalDevice::initMetalFeatures()` to include the setting of `_metalFeatures.mslVersionEnum` to `MTLLanguageVersion2_3`, following the pattern used to set it to `MTLLanguageVersion2_2` using `mvkOSVersionIsAtLeast()`. You'll need to set it in 3 places, one for each OS. However, this action is not strictly necessary for this PR, since your goal is just to get Xcode 12 working, and we can set `MTLLanguageVersion2_3 ` in a future PR.

jherico: fixed

PR #983, opened 2020-08-03
Process multiple patches per workgroup in a tessellation control shader.

cdavis5e:
This should hopefully reduce underutilization of the GPU, especially on
GPUs where the thread execution width is greater than the number of
control points.

This also eliminates the extra invocations previously needed to read the
varyings from the vertex shader into the tessellation shader. The number
of threads per workgroup is now lcm(SIMD-size, output control points).
This should ensure we always process a whole number of patches per
workgroup, and further reduce underutilization of the GPU's SIMD units.

To avoid complexity handling indices in the tessellation control shader,
I've also changed the way vertex shaders for tessellation are handled.
They are now compute kernels using Metal's support for vertex-style
stage input. This lets us always emit vertices into the buffer in order
of vertex shader execution. Now we no longer have to deal with indexing
in the tessellation control shader, nor do we always have to duplicate
the index buffer to insert gaps. This also fixes a long-standing issue
where if an index were greater than the number of vertices to draw, the
vertex shader would wind up writing outside the buffer, and the vertex
would be lost.

### DIFF
--- a/MoltenVK/MoltenVK/Commands/MVKCmdDraw.mm
+++ b/MoltenVK/MoltenVK/Commands/MVKCmdDraw.mm
@@ -112,21 +112,21 @@ void MVKCmdDraw::encode(MVKCommandEncoder* cmdEncoder) {
     pipeline->getStages(stages);
 
     const MVKMTLBufferAllocation* vtxOutBuff = nullptr;
-    const MVKMTLBufferAllocation* vtxParamsBuff = nullptr;
     const MVKMTLBufferAllocation* tcOutBuff = nullptr;
     const MVKMTLBufferAllocation* tcPatchOutBuff = nullptr;
     const MVKMTLBufferAllocation* tcLevelBuff = nullptr;
-    uint32_t patchCount = 0;
-    uint32_t inControlPointCount = 0, outControlPointCount = 0;
+	struct {
+		uint32_t inControlPointCount = 0;
+		uint32_t patchCount = 0;
+	} tessParams;
+    uint32_t outControlPointCount = 0;
     if (pipeline->isTessellationPipeline()) {
-        inControlPointCount = pipeline->getInputControlPointCount();
+        tessParams.inControlPointCount = pipeline->getInputControlPointCount();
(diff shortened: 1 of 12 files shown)


MoltenVK/MoltenVK/Commands/MVKCmdDraw.mm
```diff
@@ -112,112 +112,99 @@
...
-                                                    instanceCount: _instanceCount];
+                    [mtlTessCtlEncoder setBuffer: vtxOutBuff->_mtlBuffer
+                                          offset: vtxOutBuff->_offset
+                                         atIndex: pipeline->getOutputBufferIndex().stages[kMVKShaderStageVertex]];
                 }
+				[mtlTessCtlEncoder setStageInRegion: MTLRegionMake2D(_firstVertex, _firstInstance, _vertexCount, _instanceCount)];
+				id<MTLComputePipelineState> vtxState = pipeline->getTessVertexStageState();
+				if (cmdEncoder->getDevice()->_pMetalFeatures->nonUniformThreadgroups) {
+					[mtlTessCtlEncoder dispatchThreads: MTLSizeMake(_vertexCount, _instanceCount, 1)
+                                 threadsPerThreadgroup: MTLSizeMake(vtxState.threadExecutionWidth, 1, 1)];
```
billhollings, line 149: To shut the compiler up for tvOS, can you wrap this in `#if MVK_MACOS_OR_IOS`, similar to what we did with `drawPatches:patchIndexBuffer:patchIndexBufferOffset:indirectBuffer:indirectBufferOffset:` in the same file?

There are 5 occurrences of this in this file. I won't bother marking them all.

billhollings: Thanks.

Sorry to be pedantic...but wouldn't wrapping `dispatchThreads:threadsPerThreadgroup:` by itself work the same, since `nonUniformThreadgroups = false` already guards the missing flow for tvOS?

If so, in doing so, we could get rid of the redundant calls to `dispatchThreadgroups:threadsPerThreadgroup:`.

PR #984, opened 2020-08-05
Images in coherent memory

js6i:
It looks like pretending to have images in coherent memory strikes again. Here I add flushing to source images of a image to image transfer. That's probably not great, but it fixes some games having invalid data loaded into textures. I'm not sure to what extent this is a dxvk specific problem in my cases.
Maybe we should also flush (and queue a pull from destination images) in other kinds of transfers? Or do something else entirely?

### DIFF
--- a/MoltenVK/MoltenVK/Commands/MVKCmdTransfer.mm
+++ b/MoltenVK/MoltenVK/Commands/MVKCmdTransfer.mm
@@ -93,6 +93,8 @@ void MVKCmdCopyImage<N>::encode(MVKCommandEncoder* cmdEncoder, MVKCommandUse com
     VkBufferImageCopy vkDstCopies[copyCnt];
     size_t tmpBuffSize = 0;
 
+    _srcImage->flushToDevice(0, VK_WHOLE_SIZE);
+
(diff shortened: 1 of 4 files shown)


cdavis5e:
As an aside, we may not need any of these kluges for Catalina anymore. Apple claims in [FB5398992](https://feedbackassistant.apple.com/feedback/5398992) that 10.15.5 now supports linear textures and texture buffers in shared memory. I haven't been able to test this yet, but if so, this is good news.

(If you can't see the issue, ask Ken to put you on our team in Apple Developer.)

billhollings:
Good catch.

I must admit...at first, calling the CPU command, `replaceRegion:...`, during the encoding step of a GPU command felt odd...but I can't think of a way this might break things from a coherence perspective.

>Apple claims in FB5398992 that 10.15.5 now supports linear textures and texture buffers in shared memory

That will be helpful news for the future.

js6i:
In general, if we always flush on encoding, and pull when the command buffer finishes, any bound images that are supposed to be coherent but really aren't, we should be sort-of fine - I think. It would be wonderful if we can skip that dance for newer macOS's.

PR #988, opened 2020-08-07
Validation warning fixes

js6i:
There's one more thing - apparently now it's OK, at least on recent macOS, to not have any attachments. It could be useful to take advantage of that and not allocate a dummy texture - not just to save memory, I've seen one game specify a crazy framebuffer size (16k x 16k x 2k) and then spit out validation warnings when the dummy texture wasn't created. I'm not sure how to detect for that though, and if it deserves a dedicated feature bit.
Maybe `MTLRenderPassDescriptor respondsToSelector: defaultRasterSampleCount` check would do the trick?

### DIFF
--- a/MoltenVK/MoltenVK/Commands/MVKCommandBuffer.mm
+++ b/MoltenVK/MoltenVK/Commands/MVKCommandBuffer.mm
@@ -303,13 +303,10 @@ void MVKCommandEncoder::beginMetalRenderPass(bool loadOverride, bool storeOverri
     getSubpass()->populateMTLRenderPassDescriptor(mtlRPDesc, _framebuffer, _clearValues.contents(), _isRenderingEntireAttachment, loadOverride, storeOverride);
     mtlRPDesc.visibilityResultBuffer = _occlusionQueryState.getVisibilityResultMTLBuffer();
 
-	// Only set the layered rendering properties if layered rendering is supported and the framebuffer really has multiple layers
-	if (_isUsingLayeredRendering) {
-		VkExtent2D fbExtent = _framebuffer->getExtent2D();
-		mtlRPDesc.renderTargetWidthMVK = min(_renderArea.offset.x + _renderArea.extent.width, fbExtent.width);
-		mtlRPDesc.renderTargetHeightMVK = min(_renderArea.offset.y + _renderArea.extent.height, fbExtent.height);
-		mtlRPDesc.renderTargetArrayLengthMVK = _framebuffer->getLayerCount();
-	}
+    VkExtent2D fbExtent = _framebuffer->getExtent2D();
+    mtlRPDesc.renderTargetWidthMVK = min(_renderArea.offset.x + _renderArea.extent.width, fbExtent.width);
+    mtlRPDesc.renderTargetHeightMVK = min(_renderArea.offset.y + _renderArea.extent.height, fbExtent.height);
+    mtlRPDesc.renderTargetArrayLengthMVK = _framebuffer->getLayerCount();
 
     _mtlRenderEncoder = [_mtlCmdBuffer renderCommandEncoderWithDescriptor: mtlRPDesc];     // not retained
 	setLabelIfNotNil(_mtlRenderEncoder, getMTLRenderCommandEncoderName());
--- a/MoltenVK/MoltenVK/GPUObjects/MVKPipeline.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKPipeline.mm
@@ -1275,9 +1275,11 @@ void MVKGraphicsPipeline::addFragmentOutputToPipeline(MTLRenderPipelineDescripto
     if (pixFmts->isDepthFormat(mtlDSFormat)) { plDesc.depthAttachmentPixelFormat = mtlDSFormat; }
     if (pixFmts->isStencilFormat(mtlDSFormat)) { plDesc.stencilAttachmentPixelFormat = mtlDSFormat; }
 
-    // In Vulkan, it's perfectly valid to render with no attachments. Not so in Metal.
-	// If we have no attachments, then we'll have to add a dummy attachment.
-    if (!caCnt && !pixFmts->isDepthFormat(mtlDSFormat) && !pixFmts->isStencilFormat(mtlDSFormat)) {
+    // In Vulkan, it's perfectly valid to render with no attachments. In Metal we need to check for
+    // support for it. If we have no attachments, then we may have to add a dummy attachment.
+    if (!caCnt && !pixFmts->isDepthFormat(mtlDSFormat) && !pixFmts->isStencilFormat(mtlDSFormat) &&
+        !getDevice()->_pMetalFeatures->renderWithoutAttachments) {
+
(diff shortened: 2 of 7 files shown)


js6i:
OK, I updated the pull request to include a check for support of having no attachments in a render pass.

MoltenVK/MoltenVK/OS/MTLRenderPassDescriptor+MoltenVK.m
```diff
@@ -54,10 +54,10 @@ -(void) setRenderTargetArrayLengthMVK: (NSUInteger) length {
 
 -(NSUInteger) renderTargetWidthMVK {
 
-#if MVK_MACOS || MVK_TVOS
+#if MVK_TVOS
 	return 0;
 #endif
-#if MVK_IOS
+#if MVK_MACOS || MVK_IOS
```
billhollings, line 60: Not important, but there is a `MVK_MACOS_OR_IOS` declaration available for use too.

MoltenVK/MoltenVK/OS/MTLRenderPassDescriptor+MoltenVK.m
```diff
@@ -67,7 +67,7 @@ -(NSUInteger) renderTargetWidthMVK {
 
 -(void) setRenderTargetWidthMVK: (NSUInteger) width {
 
-#if MVK_IOS
+#if MVK_MACOS || MVK_IOS
```
billhollings, line 70: As above.

MoltenVK/MoltenVK/OS/MTLRenderPassDescriptor+MoltenVK.m
```diff
@@ -76,10 +76,10 @@ -(void) setRenderTargetWidthMVK: (NSUInteger) width {
 
 -(NSUInteger) renderTargetHeightMVK {
 
-#if MVK_MACOS || MVK_TVOS
+#if MVK_TVOS
 	return 0;
 #endif
-#if MVK_IOS
+#if MVK_MACOS || MVK_IOS
```
billhollings, line 82: As above.

MoltenVK/MoltenVK/OS/MTLRenderPassDescriptor+MoltenVK.m
```diff
@@ -89,7 +89,7 @@ -(NSUInteger) renderTargetHeightMVK {
 
 -(void) setRenderTargetHeightMVK: (NSUInteger) height {
 
-#if MVK_IOS
+#if MVK_MACOS || MVK_IOS
```
billhollings, line 92: As above.

MoltenVK/MoltenVK/GPUObjects/MVKPipeline.mm
```diff
@@ -1275,12 +1275,16 @@ static VkFormat mvkFormatFromOutput(const SPIRVShaderOutput& output) {
...
-	// If we have no attachments, then we'll have to add a dummy attachment.
-    if (!caCnt && !pixFmts->isDepthFormat(mtlDSFormat) && !pixFmts->isStencilFormat(mtlDSFormat)) {
-        MTLRenderPipelineColorAttachmentDescriptor* colorDesc = plDesc.colorAttachments[0];
-        colorDesc.pixelFormat = MTLPixelFormatR8Unorm;
-        colorDesc.writeMask = MTLColorWriteMaskNone;
+    // In Vulkan, it's perfectly valid to render with no attachments. In Metal we need to detect
+    // support for it. If we have no attachments, then we may have to add a dummy attachment.
+    @autoreleasepool {
+        if (!caCnt && !pixFmts->isDepthFormat(mtlDSFormat) && !pixFmts->isStencilFormat(mtlDSFormat) &&
+            ![[MTLRenderPassDescriptor renderPassDescriptor] respondsToSelector: @selector(defaultRasterSampleCount)]) {
```
billhollings, line 1282: This would be more consistently handled by creating an entry at the **_end_** of `MVKPhysicalDeviceMetalFeatures` (maybe `renderWithoutAttachments`?), and set its it to `true` in `MVKPhysicalDevice::initMetalFeatures()`, in the `#if MVK_IOS` and `#if MVK_MACOS` sections, using `mvkOSVersionIsAtLeast(11.0)` and `mvkOSVersionIsAtLeast(10.15)` (which already exists there), , respectively...and then testing for that flag here.

MoltenVK/MoltenVK/GPUObjects/MVKRenderPass.mm
```diff
@@ -138,6 +138,12 @@
 
 	_mtlDummyTex = nil;
 	if (caUsedCnt == 0 && dsRPAttIdx == VK_ATTACHMENT_UNUSED) {
+        if ([mtlRPDesc respondsToSelector: @selector(defaultRasterSampleCount)]) {
```
billhollings, line 141: Test for the proposed new `MVKPhysicalDeviceMetalFeatures` flag instead.

MoltenVK/MoltenVK/GPUObjects/MVKRenderPass.mm
```diff
@@ -138,6 +138,12 @@
...
+            // We support having no attachments.
+            mtlRPDesc.defaultRasterSampleCount = 1;
```
billhollings, line 143: This line needs to be wrapped in a `#if MVK_MACOS_OR_IOS`, to avoid compilation errors on `tvOS`, where  `defaultRasterSampleCount ` is not supported at all. See examples elsewhere.

MoltenVK/MoltenVK/GPUObjects/MVKPipeline.mm
js6i: Agreed, that's better.

MoltenVK/MoltenVK/Commands/MVKCommandBuffer.mm
```diff
@@ -303,13 +303,10 @@
...
-	if (_isUsingLayeredRendering) {
-		VkExtent2D fbExtent = _framebuffer->getExtent2D();
-		mtlRPDesc.renderTargetWidthMVK = min(_renderArea.offset.x + _renderArea.extent.width, fbExtent.width);
-		mtlRPDesc.renderTargetHeightMVK = min(_renderArea.offset.y + _renderArea.extent.height, fbExtent.height);
-		mtlRPDesc.renderTargetArrayLengthMVK = _framebuffer->getLayerCount();
-	}
+    VkExtent2D fbExtent = _framebuffer->getExtent2D();
+    mtlRPDesc.renderTargetWidthMVK = min(_renderArea.offset.x + _renderArea.extent.width, fbExtent.width);
+    mtlRPDesc.renderTargetHeightMVK = min(_renderArea.offset.y + _renderArea.extent.height, fbExtent.height);
+    mtlRPDesc.renderTargetArrayLengthMVK = _framebuffer->getLayerCount();
```
cdavis5e, line 309: Unfortunately, removing this check entirely caused #991, which is why it was there in the first place. I've added it back and limited it to only setting `renderTargetArrayLength` in #992.

PR #993, opened 2020-08-14
Initialize tessellation related variables conditionally in indirect draws

### LINKED ISSUE #989: Commit 3db2cbff6 introduces a crash during shader compilation
Attn: @cdavis5e 
Tracked down the crash to this commit. When starting up my app with this commit included I get a crash here:

    id<MTLComputePipelineState>
    MVKComputePipelineCompiler::newMTLComputePipelineState(MTLComputePipelineDescriptor* plDesc) {
	unique_lock<mutex> lock(_completionLock);

	compile(lock, ^{
		[_owner->getMTLDevice() newComputePipelineStateWithDescriptor: plDesc
								  options: MTLPipelineOptionNone
								completionHandler: ^(id<MTLComputePipelineState> ps,

crash due to plDesc being nil

js6i:
Calling `MVKGraphicsPipeline::getTessVertexStage*State` causes a crash for non-tessellated draws due to the lack of corresponding pipeline descriptor. I moved it, together with other tessellation related variables, inside the `isTessellationPipeline` check. This should fix #989.

### DIFF
--- a/MoltenVK/MoltenVK/Commands/MVKCmdDraw.mm
+++ b/MoltenVK/MoltenVK/Commands/MVKCmdDraw.mm
@@ -510,6 +510,11 @@ void MVKCmdDrawIndirect::encode(MVKCommandEncoder* cmdEncoder) {
     uint32_t inControlPointCount = 0, outControlPointCount = 0;
 	VkDeviceSize paramsIncr = 0;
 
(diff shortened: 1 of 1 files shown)


billhollings:
@cdavis5e Since this is a fix to tess behavior, before I pull this in, can you review and approve, please?

PR #998, opened 2020-09-01
Use deferred store actions instead of tracking multi-pass draws.

cdavis5e:
The current code does not handle multiple subpasses, nor does it handle
secondary command buffers. Handling subpasses was easy enough. The
problem came with secondary command buffers. Tracking them became
extremely complicated, particularly since pipelines may be set either
inside or outside a render pass, and further, a pipeline set in one
buffer might be used in another.

I then realized a simpler and more elegant solution: Metal's deferred
store actions feature. This allows you to defer setting the store action
for a render pass until encoding time. This is exactly what we need,
since we won't know what store action we actually want until we start
encoding draws. This solution should now work with multiple subpasses
and secondary command buffers, with much less code.


cdavis5e:
@mbarriault IIRC you were the one who contributed this in the first place. Can you make sure this doesn't break anything for you?

mbarriault (approved):
All looks fine to me. The main complexity as I recall was handling multisampled store actions, and looks like that's still covered. I think this leaves the case where deferred store actions aren't supported, but this detail is only needed to my knowledge to support tessellation, which requires the same or later hardware support across the board anyway. This functionality may be used in the future if and when someone implements mesh and/or geometry shaders, which I suspect will be done via compute as well, but that'll be easy enough to restrict to the same hardware.

billhollings:
Nice elegant solution!

PR #1000, opened 2020-09-01
MVKCmdClearAttachments: Fix vertex count for clearing multiple layers.

cdavis5e:
Prior to this, we were assuming one layer per clear rect. When that
assumption were broken, we wound up smashing the stack. This will be
more important once multiview support lands, since all views must be
cleared by the clear command.


billhollings:
Congratulations! Yours is the 1,000th submission to the MoltenVK repo! 

You win...the honour of telling everyone this interesting fact!

PR #1006, opened 2020-09-03
Support the VK_KHR_multiview extension.

### LINKED ISSUE #347: Multiview support
I just thought I should mention that I'm using Multiview, with gl_ViewIndex a lot in my game, and of course that means I can't use MoltenVK currently without creating an alternative path.

I know this is a 1.1 feature, so I guess this is just a general request for full support of that feature when you have 1.1 support, and to let you know at least one person is looking forward to it. Thanks.

cdavis5e:
Originally, Metal did not support this directly, and still largely
doesn't on GPUs other than Apple family 6. Therefore, this
implementation uses vertex instancing to draw the needed views. To
support the Vulkan requirement that only the layers for the enabled
views are loaded and stored in a multiview render pass, this
implementation uses multiple Metal render passes for multiple "clumps"
of enabled views.

For indirect draws, as with tessellation, we must adjust the draw
parameters at execution time to account for the extra views, so we need
to use deferred store actions here. Without them, tracking the state
becomes too involved.

If the implementation doesn't support either layered rendering or
deferred store actions, multiview render passes are instead unrolled and
rendered one view at a time. This will enable us to support the
extension even on older devices and OSes, but at the cost of additional
command buffer memory and (possibly) worse performance.

Eventually, we should consider using vertex amplification to accelerate
this, particularly since indirect multiview draws are terrible and
currently require a compute pass to adjust the instance count. Also,
instanced drawing in itself is terrible due to its subpar performance.
But, since vertex amplification on family 6 only supports two views,
when `VK_KHR_multiview` mandates a minimum of 6, we'll still need to use
instancing to support more than two views.

I have tested this extensively against the CTS. I'm very confident in
its correctness. The only failing tests are
`dEQP-VK.multiview.queries.*`, due to our inadequate implementation of
timestamp queries; and `dEQP-VK.multiview.depth.*`, due to what I assume
is a bug in the way Metal handles arrayed packed depth/stencil textures,
and which may only be a problem on Mojave. I need to test this on
Catalina and Big Sur.

Update SPIRV-Cross to pull in some fixes necessary for this to work.

Fixes #347.


MoltenVK/MoltenVK/GPUObjects/MVKRenderPass.mm
```diff
@@ -67,7 +68,95 @@
...
+	if (!_renderPass->getDevice()->getPhysicalDevice()->canUseInstancingForMultiview()) {
+		// If we can't use instanced drawing for this, we'll have to unroll the render pass.
+		return __builtin_popcount(_viewMask);
+	}
+	uint32_t mask = _viewMask;
+	uint32_t count;
+	for (count = 0; mask != 0; ++count) {
+		mask = getNextViewMaskGroup(mask, nullptr, nullptr);
+	}
+	return count;
```
billhollings, line 108: When `canUseInstancingForMultiview()` is `true`, I'm not clear on how the "clumping" count that drives subpass rewind is established from this logic.

I'm sure it works...but it's obscure to me.

Can you add some comments in this section (and/or in `getNextViewMaskGroup()`) that clarifies how and why we use view masking to derive "clumps" that drive subpass rewinding when `canUseInstancingForMultiview()` is `true`, please?

billhollings (changes requested):
Holy moly! This is an incredibly creative and comprehensive solution (as was your tessellation)! 

I'm really impressed once again! Thanks so much for taking this on and producing such a sophisticated and creative solution to a really challenging problem!

Just one small request for additional explanatory commentary. Feel free to add even more if appropriate for long-term maintenance.

MoltenVK/MoltenVK/GPUObjects/MVKRenderPass.mm
cdavis5e: Added. Does that help?

billhollings: Looks great. Thanks.

billhollings (approved):

PR #1010, opened 2020-09-05
MVKCommandBuffer: Fix a crash on starting a query outside a render pass.

### LINKED ISSUE #1007: Crash in vkCmdWriteTimestamp executed outside of a render pass
The crash occurs here:
https://github.com/KhronosGroup/MoltenVK/blob/412c9ddb6763b5b80a60fcad90366b4eefea5ce3/MoltenVK/MoltenVK/Commands/MVKCommandBuffer.mm#L638
with the render pass related data uninitialized. I guess I could add some sort of inRenderpass() check as a fix?

cdavis5e:
This was introduced by #1006.

Fixes #1007.


billhollings (approved):
Looks good.

PR #1011, opened 2020-09-06
Add basic support for VK_KHR_external_semaphore{,_capabilities}.

cdavis5e:
Also a non-functional base for future extensions. We can't implement it
anyway until all remaining bugs in `MTLEvent`-based semaphores are
fixed.

This is the last of the extensions that was promoted to core for Vulkan
1.1. We're almost there!


billhollings (approved):
Thanks. Looks good. 

Yes...almost there for 1.1! Thanks for all your hard work on making that happen!

PR #1014, opened 2020-09-08
Store action validation warnings

### LINKED ISSUE #1013: Store action validation warnings
I ran into some new warnings related to e2ba444aa6cef44f1150f8778e0802c403444b58. One is `store actions known at render command encoder init time may not be mutated`, happening when, apparently mistakenly, `setColorStoreAction` was called for a depth-only attachment.

With that fixed I still got `endEncoding called with store action for color attachment index 0 still set to MTLStoreActionUnknown` in a tessellated draw. Always calling `encodeStoreActions` after encoding seems to fix this one, but it may be overkill. 

Linked will be a drafty PR with these changes. @cdavis5e

js6i:
Fixes #1013.

### DIFF
--- a/MoltenVK/MoltenVK/GPUObjects/MVKRenderPass.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKRenderPass.mm
@@ -486,13 +486,19 @@ void MVKRenderPassAttachment::encodeStoreAction(MVKCommandEncoder* cmdEncoder,
                                                 bool storeOverride) {
     MTLStoreAction storeAction = getMTLStoreAction(subpass, isRenderingEntireAttachment, hasResolveAttachment, isStencil, storeOverride);
     MVKPixelFormats* pixFmts = _renderPass->getPixelFormats();
-    if (pixFmts->isDepthFormat(pixFmts->getMTLPixelFormat(_info.format)) && !isStencil) {
-        [cmdEncoder->_mtlRenderEncoder setDepthStoreAction: storeAction];
-    } else if (pixFmts->isStencilFormat(pixFmts->getMTLPixelFormat(_info.format)) && isStencil) {
-        [cmdEncoder->_mtlRenderEncoder setStencilStoreAction: storeAction];
-    } else {
-        [cmdEncoder->_mtlRenderEncoder setColorStoreAction: storeAction atIndex: caIdx];
-    }
+
+	MTLPixelFormat mtlFmt = pixFmts->getMTLPixelFormat(_info.format);
+	bool isDepthFormat = pixFmts->isDepthFormat(mtlFmt);
+	bool isStencilFormat = pixFmts->isStencilFormat(mtlFmt);
+	bool isColorFormat = !(isDepthFormat || isStencilFormat);
+
+	if (isColorFormat) {
+		[cmdEncoder->_mtlRenderEncoder setColorStoreAction: storeAction atIndex: caIdx];
+	} else if (isDepthFormat && !isStencil) {
+		[cmdEncoder->_mtlRenderEncoder setDepthStoreAction: storeAction];
+	} else if (isStencilFormat && isStencil) {
+		[cmdEncoder->_mtlRenderEncoder setStencilStoreAction: storeAction];
+	}
 }
 
 void MVKRenderPassAttachment::populateMultiviewClearRects(MVKSmallVector<VkClearRect, 1>& clearRects, MVKCommandEncoder* cmdEncoder) {


MoltenVK/MoltenVK/GPUObjects/MVKRenderPass.mm
```diff
@@ -486,11 +486,11 @@ static uint32_t getNextViewMaskGroup(uint32_t viewMask, uint32_t* startView, uin
                                                 bool storeOverride) {
     MTLStoreAction storeAction = getMTLStoreAction(subpass, isRenderingEntireAttachment, hasResolveAttachment, isStencil, storeOverride);
     MVKPixelFormats* pixFmts = _renderPass->getPixelFormats();
-    if (pixFmts->isDepthFormat(pixFmts->getMTLPixelFormat(_info.format)) && !isStencil) {
-        [cmdEncoder->_mtlRenderEncoder setDepthStoreAction: storeAction];
-    } else if (pixFmts->isStencilFormat(pixFmts->getMTLPixelFormat(_info.format)) && isStencil) {
+    if (pixFmts->isStencilFormat(pixFmts->getMTLPixelFormat(_info.format)) && isStencil) {
```
billhollings, line 489: Unfortunately, I think this logic is now going to cause `setColorStoreAction:` to be called on a stencil-only format when processing the depth attachment (`isStencil == false`).

Perhaps we should move away from an `else` chain, and implement something more explicit, like:

	MTLPixelFormat mtlFmt = pixFmts->getMTLPixelFormat(_info.format);
	bool isDepthFormat = pixFmts->isDepthFormat(mtlFmt);
	bool isStencilFormat = pixFmts->isStencilFormat(mtlFmt);
	bool isColorFormat = !(isDepthFormat || isStencilFormat);

    if (isColorFormat) {
        [cmdEncoder->_mtlRenderEncoder setColorStoreAction: storeAction atIndex: caIdx];
    }
    if (isDepthFormat && !isStencil) {
        [cmdEncoder->_mtlRenderEncoder setDepthStoreAction: storeAction];
    }
    if (isStencilFormat && isStencil) {
        [cmdEncoder->_mtlRenderEncoder setStencilStoreAction: storeAction];
    }

billhollings (changes requested):
I think there is still a logic issue on stencil-only formats now. See inline comments.

And @cdavis5e, since this is a fix to your recent code, and you also review the other aspects.

js6i:
> I think there is still a logic issue on stencil-only formats now. See inline comments.

You're right, I "forgot" that stencil only was a thing and didn't check. I'll fix it in the morning. Thanks!

cdavis5e:
I thought I hit all the places where `endMetalRenderEncoding()` could be called, but I guess I didn't. I do think it's overkill to set it when any Metal encoder ends encoding. Let me see if I can find the places I missed where we should be setting a store action...

cdavis5e (changes requested):
I also agree with @billhollings that the logic in `getMTLStoreAction()` needs more work.

cdavis5e:
> I thought I hit all the places where `endMetalRenderEncoding()` could be called, but I guess I didn't. I do think it's overkill to set it when any Metal encoder ends encoding. Let me see if I can find the places I missed where we should be setting a store action...

I think I figured it out. In all the tessellation draw calls at the beginning of the vertex shader stage, there is a call to `encodeStoreActions(true)`; this is supposed to set the store actions for the Metal render pass before ending it. But, there's also a call to `finalizeDrawState(stage)`. I think this call is triggering a switch to compute, and this is what caused the validation problems you were seeing. If you move the call to `encodeStoreActions(true)` before the call to `finalizeDrawState(stage)`, or possibly even sink it into `finalizeDrawState(stage)`, the validation problem should be fixed without setting store actions on ending every encoder.

billhollings:
@cdavis5e 

>If you move the call to `encodeStoreActions(true)` before the call to `finalizeDrawState(stage)`, or possibly even sink it into `finalizeDrawState(stage)`

This will cause `encodeStoreActions(true)` to be called on all stages. Is that your intention?

If so...I agree, it would probably make sense to sink it into `finalizeDrawState(stage)`, as you suggest, particularly since they are both `MVKCommandEncoder` functions.

Since that represents a fairly significant design change to this PR, if your approach does make sense, can you capture it in a different PR, and then reference what **_doesn't_** then need to be included in this PR (some or all of commit 1414143 I assume), please?

cdavis5e:
> @cdavis5e
> 
> > If you move the call to `encodeStoreActions(true)` before the call to `finalizeDrawState(stage)`, or possibly even sink it into `finalizeDrawState(stage)`
> 
> This will cause `encodeStoreActions(true)` to be called on all stages. Is that your intention?

No, just during the Vertex Shader stage.

> 
> If so...I agree, it would probably make sense to sink it into `finalizeDrawState(stage)`, as you suggest, particularly since they are both `MVKCommandEncoder` functions.
> 
> Since that represents a fairly significant design change to this PR, if your approach does make sense, can you capture it in a different PR, and then reference what **_doesn't_** then need to be included in this PR (some or all of commit [1414143](https://github.com/KhronosGroup/MoltenVK/commit/1414143b5a82bca01eb004d28054dd307b68e0bd) I assume), please?

#1018

MoltenVK/MoltenVK/GPUObjects/MVKRenderPass.mm
js6i: I added the check for color and nice named intermediates, but kept the `else`'s as they make it clearer that at most one case will occur.
One little usability thing is maybe it should do nothing for color formats if isStencil is set true?

aerofly:
just a quick comment here: The current release of MoltenVK causes a validation issue on iOS and on Mac OS one of our framebuffers no longer has valid content. With the fixes by js6i both issues are fixed. So besides the validation issues, it seems to cause some real problems on Mac OS.

billhollings (approved):

billhollings:
> I also agree with @billhollings that the logic in `getMTLStoreAction()` needs more work.

@cdavis5e If this looks good relative to #1018, can you approve your review here, so I can pull this in, please?

cdavis5e (approved):
Looks good. And thanks for cleaning up my mess.

PR #1015, opened 2020-09-08
Support Vulkan 1.1.

### LINKED ISSUE #822: Support Vulkan 1.1

cdavis5e:
These changes add the remaining missing features of Vulkan 1.1.

Fixes #822.

### DIFF
--- a/Common/MVKCommonEnvironment.h
+++ b/Common/MVKCommonEnvironment.h
@@ -75,6 +75,9 @@ extern "C" {
 /** Directive to identify public symbols. */
 #define MVK_PUBLIC_SYMBOL        __attribute__((visibility("default")))
 
+/** Directive to make a public alias of another symbol. */
+#define MVK_PUBLIC_ALIAS(a, t)   asm(".globl _" #a "; _" #a " = _" #t "\n")
+
 
 #ifdef __cplusplus
 }
--- a/Docs/MoltenVK_Runtime_UserGuide.md
+++ b/Docs/MoltenVK_Runtime_UserGuide.md
@@ -53,7 +53,7 @@ distribution package, see the main [`README.md`](../README.md) document in the `
 About **MoltenVK**
 ------------------
 
-**MoltenVK** is a layered implementation of [*Vulkan 1.0*](https://www.khronos.org/vulkan) 
+**MoltenVK** is a layered implementation of [*Vulkan 1.1*](https://www.khronos.org/vulkan) 
 graphics and compute functionality, that is built on Apple's [*Metal*](https://developer.apple.com/metal) 
 graphics and compute framework on *macOS*, *iOS*, and *tvOS*. **MoltenVK** allows you to use *Vulkan* 
 graphics and compute functionality to develop modern, cross-platform, high-performance graphical games 
--- a/Docs/Whats_New.md
+++ b/Docs/Whats_New.md
@@ -18,6 +18,12 @@ MoltenVK 1.0.45
 
 Released TBD
 
+- Add support for Vulkan 1.1, including:
+	- The `vkEnumerateInstanceVersion()` function
+	- The `vkGetDeviceQueue2()` function
+	- Protected memory (non-functional)
+	- A feature struct for `VK_KHR_shader_draw_parameters`
+	- All extensions that were promoted to core in Vulkan 1.1
 - Add support for extensions:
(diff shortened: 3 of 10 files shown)


Common/MVKCommonEnvironment.h
```diff
@@ -75,6 +75,9 @@ extern "C" {
 /** Directive to identify public symbols. */
 #define MVK_PUBLIC_SYMBOL        __attribute__((visibility("default")))
 
+/** Directive to make a public alias of another symbol. */
+#define MVK_PUBLIC_ALIAS(a, t)   asm(".globl _" #a "; _" #a " = _" #t "\n")
```
billhollings, line 79: Hmmm...no way to alias C function name symbols without invoking `asm()`?

All function aliasing references I can find are for C++, which I assume is why you couldn't use that? And the `__attribute__((alias))` seems to be available...but it seem to require the declaration to include function params.

I'm amazed you uncovered this technique. Where did you find it?

billhollings, line 79: Hmmm...no way to alias C function name symbols without invoking `asm()`?

All function aliasing references I can find are for C++, which I assume is why you couldn't use that? And the `__attribute__((alias))` seems to be available...but it seem to require the declaration to include function params.

I'm amazed you uncovered this technique. Where did you find it?

billhollings (approved):
Wow! What a milestone accomplishment! Great work!

In this review, I can't anticipate everything that might be impacted for use cases depending on 1.1, but everything looks good at first scan. We'll see how this falls out in practice as we move forward.

Thanks for taking this on, and congratulations! 🥳🎉

Common/MVKCommonEnvironment.h
cdavis5e: > Hmmm...no way to alias C function name symbols without invoking `asm()`?
> 
> All function aliasing references I can find are for C++, which I assume is why you couldn't use that? And the `__attribute__((alias))` seems to be available...but it seem to require the declaration to include function params.

The other thing is that `__attribute__((alias))` doesn't work on Darwin. Clang won't let you use it--even though Mach-O has `N_INDR`. Go figure.

`using` only works on functions in a different `namespace`, and it just makes that name available in the current namespace. Defining a variable with `auto` would've defined a *function pointer*, when what I wanted was to make a *symbol* alias.

> I'm amazed you uncovered this technique. Where did you find it?

I came up with it myself.

repi:
Awesome work!

Common/MVKCommonEnvironment.h
billhollings: @cdavis5e 

Building for iOS chokes on a number of identical build errors in the _Create Dynamic Library_ build phase:

    Undefined symbols for architecture arm64:
      "_vkBindBufferMemory2KHR", referenced from:
          MVKInstance::initProcAddrs() in libMoltenVK.a(libMoltenVK.a-arm64-master.o)
          ...

which presumably means those symbols don't appear in the built library. I'm wondering if the `asm()` technique doesn't work on iOS either.

cdavis5e: It's the semicolon. On x86, it's a statement separator, but for ARM it's the comment character. I'll make a PR to fix it.

cdavis5e: #1022

PR #1016, opened 2020-09-08
MVKDescriptorPool: Only free descriptor sets it knows about.

cdavis5e:
Fixes a crash in `dEQP-VK.api.null_handle.free_descriptor_sets`.


billhollings (approved):
👍🏻

PR #1018, opened 2020-09-08
MVKCommandEncoder: Set store override actions before finalizing draw state.

cdavis5e:
Otherwise, they could be left unset when we switch to compute in order
to set up the state for the draw call.

This supersedes part of #1014. The fix to the logic in
`MVKRenderPassAttachment::encodeStoreAction()`, however, is
still necessary.


cdavis5e:
@js6i Can you test this with your change to `MVKRenderPassAttachment::getMTLStoreAction()` and see if this fixes the validation problems you were seeing?

js6i:
Yes, that works fine (I assume you meant change to `MVKRenderPassAttachment::encodeStoreAction`).

billhollings (approved):
Looks good. Thanks!

PR #1022, opened 2020-09-11
Fix aliases on ARM64.

cdavis5e:
The semicolon is the comment character in ARM64 assembly. Just put the
alias symbol definition on another line.


billhollings (approved):
Thanks!

PR #1029, opened 2020-09-14
MVKCmdResolveImage: Fix resolving sub-regions.

cdavis5e:
When doing multisample resolution in Metal, the dimensions of the MSAA
RT and the resolve destination must be the same. Therefore, if the
resolve region does not cover the entire destination, we must use a
temporary transfer image. This fixes a validation error in the differing
image size tests from the CTS
(`dEQP-VK.api.copy_and_blit.*.resolve_image.diff_image_size.*`).

Because the temporary transfer image has the same dimensions as the
destination and is intended to be resolved to it, copies from the source
should use the destination's parameters for the temp image. That way,
the regions show up in the correct place in the destination. This fixes
the remaining resolve tests.

Don't do expansion blits if the resolve region covers the entire
destination. This should reduce the amount of needless work we do in
that case.

### DIFF
--- a/MoltenVK/MoltenVK/Commands/MVKCmdTransfer.h
+++ b/MoltenVK/MoltenVK/Commands/MVKCmdTransfer.h
@@ -125,9 +125,8 @@ typedef MVKCmdBlitImage<4> MVKCmdBlitImageMulti;
 
 /** Describes Metal texture resolve parameters. */
 typedef struct {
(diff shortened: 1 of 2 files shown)


billhollings (approved):
Thanks for fixing this relative to the CTS tests. 

I had thought I'd worked through the logic of this correctly...but it's a tough nut. 🤦🏻‍♂️

PR #1030, opened 2020-09-14
MVKCmdClearColorImage: Clear linear images on Mac with a shader.

cdavis5e:
Linear textures on Mac family GPUs aren't renderable, so we cannot use
a `Clear`/`Store` `MTLRenderPass` to clear them. Instead, use a compute
shader to clear them.

I haven't expanded this to all color images, because the
`MTLTextureUsageShaderWrite` usage disables lossless compression on
Apple GPUs, but `RenderTarget` usage does not. Also, multisample
textures do not yet support writing.

### DIFF
--- a/MoltenVK/MoltenVK/Commands/MVKCommandEncodingPool.h
+++ b/MoltenVK/MoltenVK/Commands/MVKCommandEncodingPool.h
@@ -109,6 +109,11 @@ public:
 	/** Returns a MTLComputePipelineState for filling a buffer. */
 	id<MTLComputePipelineState> getCmdFillBufferMTLComputePipelineState();
 
+#if MVK_MACOS
+	/** Returns a MTLComputePipelineState for clearing an image. Currently only used for 2D linear images on Mac. */
+	id<MTLComputePipelineState> getCmdClearColorImageMTLComputePipelineState(MVKFormatType type);
+#endif
+
 	/** Returns a MTLComputePipelineState for decompressing a buffer into a 3D image. */
 	id<MTLComputePipelineState> getCmdCopyBufferToImage3DDecompressMTLComputePipelineState(bool needsTempBuff);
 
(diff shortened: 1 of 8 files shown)


billhollings (approved):
Looks good. Thanks for considering impact on lossless compression textures.

cdavis5e:
> Thanks for considering impact on lossless compression textures.

I know it's important to Valve, since I remember Dan Ginsburg isn't happy about the need to set `PixelFormatView` usage in a lot of cases. Also, it's going to become important even on Mac going forward as the Apple Silicon transition moves forward.

billhollings:
>I know it's important to Valve, since I remember Dan Ginsburg isn't happy about the need to set PixelFormatView usage in a lot of cases

Indeed. I'm going to start to address `PixelFormatView` and implement `VK_KHR_image_format_list` later this week.

MoltenVK/MoltenVK/Commands/MVKCommandEncodingPool.h
```diff
@@ -151,6 +156,9 @@ class MVKCommandEncodingPool : public MVKBaseObject {
     id<MTLDepthStencilState> _cmdClearDefaultDepthStencilState = nil;
     id<MTLComputePipelineState> _mtlCopyBufferBytesComputePipelineState = nil;
 	id<MTLComputePipelineState> _mtlFillBufferComputePipelineState = nil;
+#if MVK_MACOS
+	id<MTLComputePipelineState> _mtlClearColorImageComputePipelineState[3] = {nil, nil, nil};
+#endif
```
billhollings, line 161: @cdavis5e This declaration is guarded with `MVK_MACOS` but not references to this variable in `MVKCommandEncodingPool::destroyMetalResources()` are not. Compilation on iOS fails as a result.

billhollings (commented):
Compilation fails on iOS. See inline comments.

MoltenVK/MoltenVK/Commands/MVKCommandEncodingPool.h
cdavis5e: Sorry. #1040 should fix it.

PR #1031, opened 2020-09-14
MVKCmdBindDescriptorSets: Retain given objects.

cdavis5e:
Fixes a use-after-free bug when the pipeline layout is destroyed after
recording--e.g. in the
`dEQP-VK.api.pipeline_layout.lifetime.destroy_after_end` test.

### DIFF
--- a/MoltenVK/MoltenVK/Commands/MVKCmdPipeline.mm
+++ b/MoltenVK/MoltenVK/Commands/MVKCmdPipeline.mm
@@ -193,9 +193,13 @@ VkResult MVKCmdBindDescriptorSetsStatic<N>::setContent(MVKCommandBuffer* cmdBuff
 													   uint32_t firstSet,
 													   uint32_t setCount,
 													   const VkDescriptorSet* pDescriptorSets) {
+	if (_pipelineLayout) { _pipelineLayout->release(); }
+
 	_pipelineBindPoint = pipelineBindPoint;
 	_pipelineLayout = (MVKPipelineLayout*)layout;
 	_firstSet = firstSet;
+
+	_pipelineLayout->retain();
 
 	// Add the descriptor sets
 	_descriptorSets.clear();	// Clear for reuse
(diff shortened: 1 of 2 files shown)


billhollings (changes requested):
Intersting.

`MVKCmdBindDescriptorSetsStatic` destructor will need to similarly release the retained `_pipelineLayout ` and `_descriptorSets`.

Also...would a similar issue of `VkPipelineLayout` lifetime also affect the following?

`MVKCmdPushConstants` (although we probably don't even need to be holding a reference to it).
`MVKCmdPushDescriptorSet`
`MVKCmdPushDescriptorSetWithTemplate`

cdavis5e:
> Interesting (and a pain in the butt).
> 
> `MVKCmdBindDescriptorSetsStatic` destructor will need to similarly release the retained `_pipelineLayout ` and `_descriptorSets`.

Yep, found the leaks.

> 
> Also...would a similar issue of `VkPipelineLayout` lifetime also affect the following?
> 
> `MVKCmdPushConstants` (although we probably don't even need to be holding a reference to it).

No. I will remove that reference.

> `MVKCmdPushDescriptorSet`
> `MVKCmdPushDescriptorSetWithTemplate`

Yes.

I think I can also dispense with the retains of the `MVKDescriptorSet`s, since `VkDescriptorSet` objects can't be destroyed until the command buffer is done executing; or otherwise, we'll also need to have them retain their `MVKDescriptorPool`s.

MoltenVK/MoltenVK/Commands/MVKCmdPipeline.mm
```diff
@@ -193,10 +193,14 @@
 													   uint32_t firstSet,
 													   uint32_t setCount,
 													   const VkDescriptorSet* pDescriptorSets) {
+	if (_pipelineLayout) { _pipelineLayout->release(); }
```
billhollings, line 196: Meh. Sorry...I should have thought of this on the first review pass. 

We probably need to init all `_pipelineLayout` member vars in `MVKCmdPipeline.h` to `nullptr` to guard against this line possibly crashing on first invocation if `_pipelineLayout` contains memory garbage?

billhollings (changes requested):
See comment.

MoltenVK/MoltenVK/Commands/MVKCmdPipeline.mm
cdavis5e: Done.

billhollings (approved):
👍🏻

PR #1032, opened 2020-09-14
Destroy/free objects when configuration fails.

cdavis5e:
Prior to this, we were leaking objects after failing to configure them.
The sole exception was `VkDescriptorSet`; failed descriptor sets are
automatically returned to the pool. Now all objects are destroyed or
freed when creation fails.

### DIFF
--- a/MoltenVK/MoltenVK/Commands/MVKCommandPool.mm
+++ b/MoltenVK/MoltenVK/Commands/MVKCommandPool.mm
@@ -59,7 +59,10 @@ VkResult MVKCommandPool::allocateCommandBuffers(const VkCommandBufferAllocateInf
 
 		// Command buffers start out in a VK_NOT_READY config result
 		VkResult cbRslt = mvkCmdBuff->getConfigurationResult();
-		if (rslt == VK_SUCCESS && cbRslt != VK_NOT_READY) { rslt = cbRslt; }
+		if (cbRslt != VK_NOT_READY) {
+			if (rslt == VK_SUCCESS) { rslt = cbRslt; }
+			freeCommandBuffers(1, &pCmdBuffer[cbIdx]);
+		}
 	}
 	return rslt;
 }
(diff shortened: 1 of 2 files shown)


billhollings (commented):
I can't find an explicit reference in the spec to correct implementation behaviour on object creation failure. The wording I find in the spec is vague on the subject. Best I can find is this:

> The reserved values `VK_NULL_HANDLE` and `NULL` can be used in place of valid non-dispatchable handles and dispatchable handles, respectively, when explicitly called out in the Specification. Any command that creates an object successfully must not return these values.

But it doesn't explicitly state the opposite...that an object creation error _**must**_ result in `VK_NULL_HANDLE` being returned.

Do you have a spec reference that clarifies that an error returned during object creation will result in the handle being set to `VK_NULL_HANDLE`?

cdavis5e:
I don't. A lot of the tests, however, assume that this is the case. I noticed about 20 or so `MVKDevice` objects leaked, for example, after the `dEQP-VK.api.device_init.create_device_unsupported_features` test. And a 120 were leaked after the external memory tests.

billhollings:
Okay...I've [posted a question](https://gitlab.khronos.org/vulkan/vulkan/-/issues/2351) about this to the Vulkan WG.

I'm a little worried about apps starting to crash if they don't expect a `VK_NULL_HANDLE` to be returned, so it will be good to get confirmation about expected behaviour.

cdavis5e:
Another data point: prior to this change, some of the `dEQP-VK.api.device_init.*` tests were failing because MoltenVK returned an object even though it returned an error. With this change, those tests pass. This would suggest that the expected behavior is for failed creation/allocation not to return an object.

billhollings:
Looks like the consensus from [feedback](https://gitlab.khronos.org/vulkan/vulkan/-/issues/2351#note_281081) from my post to the Vulkan WG is that the pointer can either be left alone or nulled. So this implementation should be fine.

PR #1033, opened 2020-09-15
MVKCmdCopyImage: Support copies between 3D and 2D images.

cdavis5e:
This can be done by copying between each slice of the 2D image and each
plane of the 3D image individually.

This was actually quite simple to implement. I don't know why I punted
on this.


billhollings (approved):
👍🏻

PR #1034, opened 2020-09-15
MVKSampler: Wrap MTLSamplerState creation in an autorelease pool.

cdavis5e:
Or the `MTLSamplerDescriptor` could be leaked.


billhollings (approved):
Good catch.

PR #1035, opened 2020-09-15
MVKCmdBlitImage: Support blits with non-2D images.

cdavis5e:
This is needed to get us past the 3D blit tests that were recently added
to the CTS. It *almost* passes all these new tests; the 3D format tests
fail for some reason.


billhollings (approved):
👍🏻

PR #1036, opened 2020-09-15
MVKSampler: Ignore addressModeW with unnormalized coordinates.

cdavis5e:
Some tests intentionally pass an invalid address mode here when
unnormalized coordinates are in use, ostensibly to test that it is
ignored. Metal's validation layer, however, complains if you set
`rAddressMode` to an invalid value, even if `normalizedCoordinates` is
`false`. To avoid this, don't set the `rAddressMode`, since it can't be
used with unnormalized coordinates anyway.


billhollings (approved):
👍🏻

PR #1037, opened 2020-09-16
MVKCmdClearImage: Support clears of 3D images.

cdavis5e:
For these, we must use the `depthPlane` property to set the layer to
clear, and we must iterate over the image's depth, since the layer count
will be 1 in this case.


billhollings (approved):
👍🏻

PR #1038, opened 2020-09-16
MVKGraphicsPipeline: Fail triangle fan pipelines.

cdavis5e:
Otherwise, when we try to draw with them, Metal's validation layer will
complain that the primitive type and primitive topology don't match.


billhollings (approved):
This is cool. I guess it means the CTS tests will fail even so...but at least not via an abort. And once `VK_KHR_portability_subset` is in place and CTS is updated, those tests will move to Not Supported.

PR #1039, opened 2020-09-16
Disable the variableMultisampleRate feature for now.

cdavis5e:
We actually can't support this. Metal's validation layer complains if
a pipeline has a different raster sample count from that of the
framebuffer, even in the no-attachment case. This means that the
`defaultRasterSampleCount` property must be set correctly if Metal
supports no-attachment rendering, and the sample count of the dummy
texture must be set properly otherwise.


billhollings (approved):
👍🏻

PR #1042, opened 2020-09-17
MVKGraphicsPipeline: Fix applying divisors to instanced vertex attributes.

### LINKED ISSUE #1041: Vertex attribute binding mismatch for custom step rate/divisors
Hey, I'm having trouble getting custom step rates/divisors to work when using instanced rendering with MoltenVK version `1.0.43` from the Vulkan SDK version `1.2.141.2`.

In `MVKGraphicsPipeline::addVertexInputToPipeline` where `pVertexInputDivisorState` is handled `getMetalBufferIndexForVertexAttributeBinding` is called which ends up returning the Metal buffer index as `maxPerStageBufferCount - binding` ([link](https://github.com/KhronosGroup/MoltenVK/blob/v.1.0.43/MoltenVK/MoltenVK/GPUObjects/MVKPipeline.mm#L992)). This means that the binding index gets "inverted" and the subsequent call to `SPIRVToMSLConversionConfiguration::isVertexBufferUsed` fails because the bindings (which for me are in the range `0, 1, ...`) gets compared to the "inverted" bindings/indices (in the range `30, 29, ...`).

If I skip that `isVertexBufferUsed` call and thus force-set the `vbDesc.stepRate = pVKVB->divisor` the rest works as expected. Otherwise the shader vertex data never gets "repeated" and only the first few rendered instances get the data.

So I basically don't understand the `maxPerStageBufferCount - binding` logic. This could just be me setting up the data incorrectly but everything else seems to work fine.

cdavis5e:
Use the Vulkan binding index to check that the binding is used.
Previously, we were erroneously using the Metal binding index. This
allows vertex binding divisor info to actually be used, and fixes some
tests in the CTS.

Fixes #1041.

### DIFF
--- a/MoltenVK/MoltenVK/GPUObjects/MVKPipeline.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKPipeline.mm
@@ -1139,8 +1139,8 @@ bool MVKGraphicsPipeline::addVertexInputToPipeline(T* inputDesc,
         vbCnt = pVertexInputDivisorState->vertexBindingDivisorCount;
         for (uint32_t i = 0; i < vbCnt; i++) {
             const VkVertexInputBindingDivisorDescriptionEXT* pVKVB = &pVertexInputDivisorState->pVertexBindingDivisors[i];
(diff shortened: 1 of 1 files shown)


billhollings (approved):
👍🏻

PR #1043, opened 2020-09-21
Support linear images and buffer views in shared memory on macOS 10.15.5+.

cdavis5e:
Apple assured me that, starting in 10.15.5, linear textures and texture
buffers can be created from buffers with `Shared` storage. I have tested
this and can confirm that it works at least on Big Sur, and probably on
Catalina as well.

### DIFF
--- a/MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
@@ -1026,6 +1026,7 @@ void MVKPhysicalDevice::initMetalFeatures() {
     _metalFeatures.texelBuffers = true;
 	_metalFeatures.maxTextureDimension = (8 * KIBI);
     _metalFeatures.dynamicMTLBufferSize = (4 * KIBI);
+    _metalFeatures.sharedLinearTextures = true;
 
     if (supportsMTLFeatureSet(tvOS_GPUFamily1_v2)) {
 		_metalFeatures.mslVersionEnum = MTLLanguageVersion1_2;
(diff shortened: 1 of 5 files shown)


MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
```diff
@@ -1186,6 +1188,7 @@
 		_metalFeatures.mslVersionEnum = MTLLanguageVersion2_2;
 		_metalFeatures.native3DCompressedTextures = true;
         _metalFeatures.renderWithoutAttachments = true;
+        _metalFeatures.sharedLinearTextures = true; // XXX Only in 10.15.5
```
billhollings, line 1191: For accuracy, move this to a similar but separate test:

```
if (mvkOSVersionIsAtLeast(mvkMakeOSVersion(10, 15, 5))) {
    _metalFeatures.sharedLinearTextures = true;
}
```

billhollings (changes requested):
Minor change to OS version test for shared linear textures.

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
cdavis5e: Huh. I didn't know we had that. Thanks.

billhollings (approved):
👍🏻

MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
```diff
@@ -761,10 +758,10 @@
 
     if (_ioSurface && stgMode == MTLStorageModePrivate) { stgMode = MTLStorageModeShared; }
 
-#if MVK_MACOS
-	// For macOS, textures cannot use Shared storage mode, so change to Managed storage mode.
-    if (stgMode == MTLStorageModeShared) { stgMode = MTLStorageModeManaged; }
-#endif
+	// For macOS prior to 10.15.5, textures cannot use Shared storage mode, so change to Managed storage mode.
+    if (stgMode == MTLStorageModeShared && !_device->_pMetalFeatures->sharedLinearTextures) {
+        stgMode = MTLStorageModeManaged;
```
billhollings, line 763: Oops. `MTLStorageModeManaged` not supported on iOS. Needs some guard code.

billhollings (commented):
@cdavis5e Needs an update for iOS build.

MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
cdavis5e: #1049

PR #1044, opened 2020-09-21
MVKImage: Reject attempts to create compressed linear images.

cdavis5e:
This doesn't sit well with Metal, which will immediately abort the
program on the attempt.


billhollings (approved):
👍🏻

PR #1045, opened 2020-09-21
MVKPipeline: Don't skip setting the vertex stage pipeline.

cdavis5e:
In an indirect draw, we wouldn't set the pipeline for the vertex stage
to the command encoder, because we had already switched encoders. This
caused havoc down the line, as we wound up running the "convert indirect
buffers" pipeline again, which smashed memory randomly--possibly causing
the GPU to crash, and bringing down the `WindowServer` with it!

Why is it so easy to bring down the `WindowServer` like this?


cdavis5e:
This one *really* needs to be in the SDK. It fixes a problem that could cause a `WindowServer` crash.

billhollings (approved):
Thanks. Yes...this will make it into the SDK.

PR #1047, opened 2020-09-21
MVKImage: Fix broken getMemoryRequirements2().

cdavis5e:
We were looking in the wrong chain for the
`VkImagePlaneMemoryRequirementsInfo` struct. We were also failing to
call through to the `MVKImageMemoryBinding` to get dedicated
requirements, which broke dedicated memory.

Clearly, I failed to properly review the patch which refactored
`MVKImage` for `VK_KHR_sampler_ycbcr_conversion`.


billhollings (approved):
👍🏻

PR #1050, opened 2020-09-21
MVKImage: Handle VK_IMAGE_CREATE_EXTENDED_USAGE_BIT.

cdavis5e:
If this bit is set, add `MTLTextureUsage` bits for features supported by
all possible view formats, not just the image format. Happily, Metal's
API validator layer is OK with this.

This is needed for the mutable format tests, which use this bit with a
format that doesn't support writing, then casts the image to a format
which does.

### DIFF
--- a/MoltenVK/MoltenVK/GPUObjects/MVKPixelFormats.h
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKPixelFormats.h
@@ -77,6 +77,61 @@ typedef enum : uint16_t {
 	kMVKMTLFmtCapsMultiPlanar = kMVKMTLFmtCapsChromaSubsampling,
 } MVKMTLFmtCaps;
 
+inline MVKMTLFmtCaps operator|(MVKMTLFmtCaps leftCaps, MVKMTLFmtCaps rightCaps) {
+	return static_cast<MVKMTLFmtCaps>(static_cast<uint32_t>(leftCaps) | rightCaps);
+}
+
+inline MVKMTLFmtCaps& operator|=(MVKMTLFmtCaps& leftCaps, MVKMTLFmtCaps rightCaps) {
+	return (leftCaps = leftCaps | rightCaps);
+}
+
+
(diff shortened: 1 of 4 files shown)


MoltenVK/MoltenVK/GPUObjects/MVKPixelFormats.h
```diff
@@ -77,6 +77,61 @@ typedef enum : uint16_t {
...
+	BC4_R,
+	BC5_RG,
+	BC6H_RGB,
+	BC7_RGBA,
+	Depth24_Stencil8,
+	Depth32_Stencil8,
+	BGRA10_XR,
+	BGR10_XR
+};
+
```
billhollings, line 134: Looks like `VK_IMAGE_CREATE_EXTENDED_USAGE_BIT` was introduced in `VK_KHR_maintenance2`, which also introduced uncompressed image views of compressed images.

Does this classification scheme allow, say, a 64-bit compressed block to adopt the capabilities of a 64-bit color format? Or what about between two compressed formats of the same block size? It seems more divided than that.

Or am I missing something in my understanding?

billhollings (commented):
Strange feature. Not sure when it would be used in practise. But there it is in the Vulkan spec.

Question about the classification scheme.

MoltenVK/MoltenVK/GPUObjects/MVKPixelFormats.h
cdavis5e: > Does this classification scheme allow, say, a 64-bit compressed block to adopt the capabilities of a 64-bit color format? Or what about between two compressed formats of the same block size? It seems more divided than that.

Deliberately not, since Metal still doesn't allow reinterpreting a compressed texture with an uncompressed format.

billhollings: Meh. You're right of course. I brain-farted between Metal and Vulkan capabilities.

PR #1052, opened 2020-09-21
MVKImage: Handle VK_IMAGE_CREATE_MUTABLE_FORMAT_BIT.

cdavis5e:
Only set `MTLTextureUsagePixelFormatView` if this bit is set. This
should reduce the usage of this bit, which disables lossless compression
on Apple GPUs, in many cases.

We continue to set it anyway for `VK_IMAGE_USAGE_TRANSFER_SRC_BIT`; this
is because we create those texture views on the application's behalf, to
implement `vkCopyImage()` where the source and destination formats do
not agree. We also continue to set it for depth/stencil formats; Metal
requires it in order to use only the stencil aspect in a view.

### DIFF
--- a/MoltenVK/MoltenVK/GPUObjects/MVKPixelFormats.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKPixelFormats.mm
@@ -465,6 +465,7 @@ MTLTextureUsage MVKPixelFormats::getMTLTextureUsage(VkImageUsageFlags vkImageUsa
 													MTLPixelFormat mtlFormat,
 													MTLTextureUsage minUsage,
                                                     bool isLinear,
+                                                    bool isMutableFormat,
                                                     bool isExtended) {
 	bool isDepthFmt = isDepthFormat(mtlFormat);
 	bool isStencilFmt = isStencilFormat(mtlFormat);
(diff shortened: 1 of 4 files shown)


MoltenVK/MoltenVK/GPUObjects/MVKPixelFormats.mm
```diff
@@ -465,6 +465,7 @@
 													MTLPixelFormat mtlFormat,
 													MTLTextureUsage minUsage,
                                                     bool isLinear,
+                                                    bool isMutableFormat,
                                                     bool isExtended) {
```
billhollings, line 469: I'm concerned about the number of image-specific customization elements we pass into this function. Continuing this approach will only get worse with `VK_KHR_image_format_list`.

I'm wondering if we need to start thinking about refactoring this function into the basic conversion which would be handled in `MVKPixelFormats`, and a per-Image overlay function, which would tweak the `MTLTextureUsage` results returned by this function, based on the customization needs of the Image.

Let's go with this for now, but I'll look into that when addressing `VK_KHR_image_format_list`.

billhollings (approved):
Looks good. I've made note of something to look for in future expansions on `MVKPixelFormats::getMTLTextureUsage`.

PR #1053, opened 2020-09-21
MVKDevice: Limit the maximum number of inline uniform blocks on Mac.

cdavis5e:
Mac GPUs only support 14 dynamic buffers. One of those is used for push
constants, leaving 13 for inline uniform blocks.


billhollings (approved):
👍🏻

Issue #1027, opened 2020-09-14
Generate to a wrong MSL from SPIRV


daemyung:
This issue is that SPIRV-Cross generates wrong MSL. The error message is below
```
validateFunctionArguments:3476: failed assertion `Fragment Function(main0): missing buffer binding at index 2 for headIndexImage_atomic[0].'
```
However a resource at index 2 is not buffer. It's image. Below is shader's resources.
```
layout (set = 0, binding = 1) buffer GeometrySBO
{
    uint count;
    uint maxNodeCount;
};

layout (set = 0, binding = 2, r32ui) uniform uimage2D headIndexImage;

layout (set = 0, binding = 3) buffer LinkedListSBO
{
    Node nodes[];
};
```
You can find this shader at Vulkan/data/shaders/glsl/oit

You can reproduce this issue very easily. Follow below steps to reproduce this issue.
1. git clone --recursive https://github.com/SaschaWillems/Vulkan.git
2. python download_assets.py
3. Generate a project using cmake.
```
mkdir build
cd build
cmake -G "Xcode .."
```
4. Build `oit` example  and run.

My VulkanSDK is 1.2.148.
```
Revisions used to build this SDK on Tue Aug 11 20:08:23 2020
fc87182aab33a63b43db41d7a424eed27283aebf : https://github.com/KhronosGroup/Vulkan-Headers.git
2ad4e990f5a47d3556a5912acfa9fc7dc475fc0e : https://github.com/KhronosGroup/Vulkan-Loader.git
91cb71f2fa0c4d710cf188e89b4f7d1a2a0e5879 : https://github.com/KhronosGroup/Vulkan-ValidationLayers.git
3f7ccf2cce264525d14361ac563e7a6ba9a54f54 : https://github.com/KhronosGroup/Vulkan-Tools.git
3ce31c1c29919f0695d2d4d4990f29a90e39bcf6 : https://github.com/LunarG/VulkanTools.git
8f6edef8ba54503f70064ea94b0fb0020dcca819 : https://www.github.com/KhronosGroup/MoltenVK.git
3ee5f2f1d3316e228916788b300d786bb574d337 : https://github.com/KhronosGroup/Glslang.git
895927bd3f2d653f40cebab55aa6c7eabde30a86 : https://github.com/KhronosGroup/SPIRV-Tools
979924c8bc839e4cb1b69d03d48398551f369ce7 : https://github.com/KhronosGroup/SPIRV-Headers
[...]

PR #1055, opened 2020-09-22
MVKImage: Always use a texel buffer for atomic linear storage images.

cdavis5e:
If the image has a format that supports atomic access, or can be cast to
a format which supports atomic access, then use a texel buffer,
regardless of the memory type. If we can't use the `MTLBuffer` from the
device memory, then create our own.

For #1027.


billhollings (approved):
👍🏻

PR #1056, opened 2020-09-22
MVKImage: Only create a stencil view for reading.

cdavis5e:
We don't want to do this for stencil attachment views, because we use
the original packed depth/stencil format in render pipelines, and
Metal's validation layer for some reason doesn't consider packed formats
and their corresponding stencil view formats to match. So only do this
if the image view usage includes `SAMPLED` or `INPUT_ATTACHMENT`.


billhollings (approved):
👍🏻

PR #1057, opened 2020-09-22
MVKDevice: Fix reported image format properties for 1D images as 2D.

cdavis5e:
Fall through to the 2D case, so all the special handling for 2D is used
for 1D as well. Also, make sure 1D doesn't report multisampling or
support for 420 subsampled formats. Vulkan doesn't support 1D
multisampled textures anyway.

Also, clear the `VkImageFormatProperties` struct if the format is not
supported with the given parameters. Some tests seem to expect this.


billhollings (approved):
👍🏻

PR #1059, opened 2020-09-22
MVKImage: Correct offset for texel buffers and heaps.

cdavis5e:
For planes other than 0, we must add the offset of the first
subresource, so the plane is at the correct place in memory.


billhollings (approved):
👍🏻

PR #1060, opened 2020-09-22
MVKCmdClearImage: Use layered rendering when possible.

cdavis5e:
This makes the clearing tests from the CTS marginally faster, by about 30 ms in many cases.

### DIFF
--- a/MoltenVK/MoltenVK/Commands/MVKCmdTransfer.mm
+++ b/MoltenVK/MoltenVK/Commands/MVKCmdTransfer.mm
@@ -1345,20 +1345,41 @@ void MVKCmdClearImage<N>::encode(MVKCommandEncoder* cmdEncoder) {
 				layerEnd = layerStart + layerCnt;
 			}
 
-            for (uint32_t layer = layerStart; layer < layerEnd; layer++) {
+            // If we can do layered rendering, I can clear all the layers at once.
+            if (cmdEncoder->getDevice()->_pMetalFeatures->layeredRendering &&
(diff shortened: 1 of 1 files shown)


billhollings (approved):
Elegant. Nice idea!

MoltenVK/MoltenVK/Commands/MVKCmdTransfer.mm
```diff
@@ -1345,20 +1345,41 @@ static inline MTLSize mvkClampMTLSize(MTLSize size, MTLOrigin origin, MTLSize ma
...
+                    mtlRPSADesc.depthPlane = layerStart;
                 } else {
-                    mtlRPCADesc.slice = layer;
-                    mtlRPDADesc.slice = layer;
-                    mtlRPSADesc.slice = layer;
+                    mtlRPCADesc.slice = layerStart;
+                    mtlRPDADesc.slice = layerStart;
+                    mtlRPSADesc.slice = layerStart;
                 }
+                mtlRPDesc.renderTargetArrayLength = (layerCnt == VK_REMAINING_ARRAY_LAYERS
```
billhollings, line 1360: @cdavis5e `renderTargetArrayLength` not available on tvOS. Use `renderTargetArrayLengthMVK` instead.

billhollings (commented):
Build fail on tvOS.

MoltenVK/MoltenVK/Commands/MVKCmdTransfer.mm
cdavis5e: #1063

PR #1061, opened 2020-09-23
MVKCmdClearAttachments: Account for the renderArea.

cdavis5e:
When calculating the vertices, we need to use the render area's
extent--but only if the implementation supports constraining the render
area using `renderTargetWidth` and `renderTargetHeight`. Otherwise, the
quad will be stretched and/or squashed because of the render area
constraint.

### DIFF
--- a/MoltenVK/MoltenVK/Commands/MVKCmdTransfer.mm
+++ b/MoltenVK/MoltenVK/Commands/MVKCmdTransfer.mm
@@ -1130,6 +1130,15 @@ void MVKCmdClearAttachments<N>::encode(MVKCommandEncoder* cmdEncoder) {
 	simd::float4 clearColors[kMVKClearAttachmentCount];
 
 	VkExtent2D fbExtent = cmdEncoder->_framebuffer->getExtent2D();
+	// I need to know if the 'renderTargetWidth' and 'renderTargetHeight' properties
+	// actually do something, but [MTLRenderPassDescriptor instancesRespondToSelector: @selector(renderTargetWidth)]
+	// returns NO even on systems that do support it. So we have to check an actual instance.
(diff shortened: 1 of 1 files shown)


billhollings (approved):
👍🏻

MoltenVK/MoltenVK/Commands/MVKCmdTransfer.mm
```diff
@@ -1130,6 +1130,15 @@ static inline MTLSize mvkClampMTLSize(MTLSize size, MTLOrigin origin, MTLSize ma
 	simd::float4 clearColors[kMVKClearAttachmentCount];
 
 	VkExtent2D fbExtent = cmdEncoder->_framebuffer->getExtent2D();
+	// I need to know if the 'renderTargetWidth' and 'renderTargetHeight' properties
+	// actually do something, but [MTLRenderPassDescriptor instancesRespondToSelector: @selector(renderTargetWidth)]
+	// returns NO even on systems that do support it. So we have to check an actual instance.
+	MTLRenderPassDescriptor* tempRPDesc = [MTLRenderPassDescriptor new];	// temp retain
+	if ([tempRPDesc respondsToSelector: @selector(renderTargetWidth)]) {
```
billhollings, line 1137: This is causing build warnings on tvOS. Need to wrap this section in `#if MVK_MACOS_OR_IOS`

billhollings (commented):
@cdavis5e Build warnings on tvOS.

MoltenVK/MoltenVK/Commands/MVKCmdTransfer.mm
cdavis5e: #1063

PR #1062, opened 2020-09-23
MVKCmdResolveImage: Use layered rendering where possible.

cdavis5e:


cdavis5e:
Surprisingly, this is actually a bit slower in the CTS, by about 10-30 ms, than resolving layers one at a time--at least, after the first test which resolves multiple layers, which takes about the same amount of time. I wonder if I should just withdraw this change.

billhollings (approved):
👍🏻

billhollings:
>I wonder if I should just withdraw this change.

Let me know whether to pull this in.

cdavis5e:
I'm not sure it'll hurt in practice--I don't think we'll see a string of arrayed `vkCmdResolveImage()` calls in practice. And maybe it'll actually help in some cases. Feel free to pull it in.

PR #1063, opened 2020-09-23
Fix build errors on tvOS.

cdavis5e:


MoltenVK/MoltenVK/Commands/MVKCmdTransfer.mm
```diff
@@ -1138,6 +1139,7 @@ static inline MTLSize mvkClampMTLSize(MTLSize size, MTLOrigin origin, MTLSize ma
 		VkRect2D renderArea = cmdEncoder->clipToRenderArea({{0, 0}, fbExtent});
 		fbExtent = {renderArea.offset.x + renderArea.extent.width, renderArea.offset.y + renderArea.extent.height};
 	}
+#endif
 	[tempRPDesc release];													// temp release
```
billhollings, line 1143: Need to include `[tempRPDesc release]` before `#endif`.

billhollings (commented):
@cdavis5e One more minor build issue.

MoltenVK/MoltenVK/Commands/MVKCmdTransfer.mm
cdavis5e: Dammit, I thought I did that. #1065

PR #1064, opened 2020-09-23
MVKCmdCopyImage: Use the method to copy multiple slices when possible.

cdavis5e:
When the extent covers both the source and destination images
completely, we can use the copy method on `MTLBlitCommandEncoder` which
can copy multiple slices at once. This should hopefully reduce CPU
overhead and command buffer memory usage.


billhollings (approved):
👍🏻

PR #1068, opened 2020-09-24
Offset vertex buffers for attribute bindings with zero divisors.

cdavis5e:
One aspect of the `VK_EXT_vertex_attribute_divisor` spec that I
apparently missed is that, for a vertex buffer binding with a divisor of
0, the base instance determines where the attributes are read from. This
cannot be expressed in Metal, but we can emulate it by offseting the
buffer by `firstInstance * stride`.

Unfortunately, we can't do this for indirect draws. If a program tries
this, we're hosed. Maybe I should turn off
`vertexAttributeInstanceRateZeroDivisor` because of this.


billhollings:
Damn Travis.

BTW & FYI...I learned we can trigger a restart of Travis by closing the PR are then reopening it.

billhollings (approved):
👍🏻

PR #1070, opened 2020-09-24
Fix memory overrun when enabling VkPhysicalDevicePortabilitySubsetFeaturesKHR features.

billhollings:
Plus update MoltenVK_icd.json to tell Vulkan Loader and Layers to use Vulkan 1.1.

### DIFF
--- a/MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
@@ -3368,9 +3368,9 @@ void MVKDevice::enableFeatures(const VkDeviceCreateInfo* pCreateInfo) {
 			}
 			case VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_PORTABILITY_SUBSET_FEATURES_KHR: {
 				auto* requestedFeatures = (VkPhysicalDevicePortabilitySubsetFeaturesKHR*)next;
-				enableFeatures(&_enabledPortabilityFeatures.triangleFans,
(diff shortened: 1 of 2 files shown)


cdavis5e (approved):
Yet another bit I missed when I enabled 1.1 support. Good catch.

billhollings:
@cdavis5e 

Just want your input on the change to 1.1 in `MoltenVK_icd.json`. 

I came across this because Vulkan Layers were changing the instance `apiVersion` to 1.0, which then broke CTS because CTS blindly uses the `apiVersion` from the physical device properties to decide whether to call the `PFN_vkGetPhysicalDeviceFeatures2` proc addr. But we rightly don't create a proc addr for that becuase `MVKInstance` thinks the app is only going to use the Vulkan 1.0 API...triggering an empty pointer crash.

CTS shouldn't be doing this. The spec is pretty explicit about the app not calling functionality beyond what it ask for the `apiVersion` it sets. But I guess it wasn't aware that the Layers were overriding it. 

I toyed with other options for how to handle this internally...like having `MVKPhysicalDevice` limit itself to the `MVKInstance apiVersion`...but again...the Vulkan spec specifically allows physical devices to support API versions above what the app asks for.

Anyway...in the end, giving the Layers the updated supported API version fixed it in the end.

I just want to make sure that updating `MoltenVK_icd.json` to 1.1 is appropriate from your perspective.

billhollings:
>Yet another bit I missed when I enabled 1.1 support.

Okay. Thanks for the confirmation.

PR #1071, opened 2020-09-24
Ensure there is at least one pixel in the render area.

cdavis5e:
If there are no attachments and `renderTargetWidth` and
`renderTargetHeight` are zero, the Metal validation layer complains. To
prevent this, ensure both are at least 1.


billhollings (approved):
👍🏻

PR #1072, opened 2020-09-25
MVKPhysicalDevice: Get the vendor ID before setting the subgroup size.

cdavis5e:
At this point in device initialization, the device properties have not
yet been initialized. Unfortunately, this includes the vendor ID, on
which the maximum SIMD-group size depends. Initialize that property so
we can use it to set the subgroup size correctly.

### DIFF
--- a/MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
@@ -1243,6 +1243,7 @@ void MVKPhysicalDevice::initMetalFeatures() {
 #if MVK_MACOS
     if (mvkOSVersionIsAtLeast(10.14)) {
         static const uint32_t kAMDVendorId = 0x1002;
+        initGPUInfoProperties();    // Need to know the vendor ID for this.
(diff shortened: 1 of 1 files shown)


MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
```diff
@@ -1243,6 +1243,7 @@
 #if MVK_MACOS
     if (mvkOSVersionIsAtLeast(10.14)) {
         static const uint32_t kAMDVendorId = 0x1002;
+        initGPUInfoProperties();    // Need to know the vendor ID for this.
```
billhollings, line 1246: Hmmm...calling `initGPUProperties()` twice, with one of them out-of-place, rubs me the wrong way.

But alternates that I can think of, such as renaming `initGPUProperties()` to `retrieveGPUInfo()`, having it store the values in separate member variables and then copy them into `_properties` within `initProperties()` are also less than optimal and would be intrusive at this point in the dev cycle given there are three implementations of `initGPUProperties()`.

I might put that step on my TODO list at some point.

billhollings (approved):
Good tactical fix for this release...but I hope we can clean this up at some point. See inline comments.

billhollings:
>I might put that step on my TODO list at some point.

Done in #1094.

PR #1073, opened 2020-09-25
Allow multiple images for dedicated allocs of aliasabled images.

cdavis5e:
Believe it or not, this is valid usage. If an image is aliasable and it
has a dedicated alloc, it is valid for multiple images to bind to the
dedicated memory. Some tests actually try this--for example, the
`dEQP-VK.device_group.afr_dedicated` test.

### DIFF
--- a/MoltenVK/MoltenVK/Utility/MVKFoundation.h
+++ b/MoltenVK/MoltenVK/Utility/MVKFoundation.h
@@ -470,7 +470,7 @@ void mvkReleaseContainerContents(C& container) {
 
 /** Returns whether the container contains an item equal to the value. */
 template<class C, class T>
-bool contains(const C& container, const T& val) {
+bool contains(C& container, const T& val) {
 	for (const T& cVal : container) { if (cVal == val) { return true; } }
 	return false;
 }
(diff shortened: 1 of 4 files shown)


MoltenVK/MoltenVK/Utility/MVKFoundation.h
```diff
@@ -475,6 +475,13 @@ bool contains(const C& container, const T& val) {
 	return false;
 }
 
+/** Returns whether the container contains an item equal to the value. */
+template<class C, class T>
+bool contains(C& container, const T& val) {
+	for (const T& cVal : container) { if (cVal == val) { return true; } }
```
billhollings, line 481: So this is a repeat of the `const` version.

Can we delete the `const` version? Or more to the point, just remove the `const` declaration on it?

I probably just brain-farted on that one. None of the other container iterator templates declare the container as `const`. And in any case, it doesn't look to be used by current code anyway.

billhollings (changes requested):
Can we remove duplicate `contains()` functions?

MoltenVK/MoltenVK/Utility/MVKFoundation.h
cdavis5e: I actually wanted to use the `const` version, but I couldn't make it work with the pointer specialization of `MVKSmallVector`. Done.

billhollings (approved):
👍🏻

PR #1075, opened 2020-09-25
Advertise the VK_KHR_shader_subgroup_extended_types extension.

cdavis5e:
I anticipated this, and I tried to design the support in SPIRV-Cross so
that this would just work. And they do... well, they work as well as
32-bit types currently do, which is to say, there's plenty of room for
improvement.


billhollings (approved):
👍🏻

PR #1076, opened 2020-09-25
MVKDevice: Only force a mux switch when a swapchain is created.

cdavis5e:
We only want the window server to use the high-performance GPU if we
will use it to present to the display. If we won't use it to present, we
can save some battery life by not using the display. I had hoped this
would help window server stability in case something goes horribly
wrong while using the GPU, but my experience has sadly not borne this
out.

My testing shows that the device returned by
`MTLCreateSystemDefaultDevice()` is exactly equal (i.e. has the same
pointer value) to one of the devices returned by `MTLCopyAllDevices()`,
so we should see no problems from doing this at swapchain create time
instead of device create time.

### DIFF
--- a/MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
@@ -2590,6 +2590,20 @@ void MVKDevice::destroyImageView(MVKImageView* mvkImgView,
 
 MVKSwapchain* MVKDevice::createSwapchain(const VkSwapchainCreateInfoKHR* pCreateInfo,
 										 const VkAllocationCallbacks* pAllocator) {
+#if MVK_MACOS
+	// If we have selected a high-power GPU and want to force the window system
+	// to use it, force the window system to use a high-power GPU by calling the
+	// MTLCreateSystemDefaultDevice function, and if that GPU is the same as the
+	// selected GPU, update the MTLDevice instance used by the MVKPhysicalDevice.
+	id<MTLDevice> mtlDevice = _physicalDevice->getMTLDevice();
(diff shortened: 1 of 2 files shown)


cdavis5e:
I'm actually not sure I want this to go in at all. It might be causing a hang in the WSI tests.

cdavis5e:
> I'm actually not sure I want this to go in at all. It might be causing a hang in the WSI tests.

It just occurred to me that it couldn't be causing that hang, because starting in 10.15, `MTLCreateSystemDefaultDevice()` returns `nil` unless for some reason you link to `CoreGraphics.framework`. Maybe we should start linking to it ourselves?

billhollings:
>I'm actually not sure I want this to go in at all. It might be causing a hang in the WSI tests.

>...it couldn't be causing that hang

Does this mean you are recommending pulling this PR in after all?

billhollings:
@cdavis5e Also...we need to cut a release for the SDK today. Your rapid series of fixes are great...but unfortunately we need to know when is a good cutoff point for the SDK build.

Are you able to indicate a finalization you are aiming for...or are you constantly going back to the CTS well for more issues...and so the cutoff is really an arbitrary action?

cdavis5e:
> Are you able to indicate a finalization you are aiming for...or are you constantly going back to the CTS well for more issues...and so the cutoff is really an arbitrary action?

I have been going back to the CTS to find more problems, but I think this is it for now.

> Does this mean you are recommending pulling this PR in after all?

Yes.

billhollings:
> > Are you able to indicate a finalization you are aiming for...or are you constantly going back to the CTS well for more issues...and so the cutoff is really an arbitrary action?
> 
> I have been going back to the CTS to find more problems, but I think this is it for now.
> 
> > Does this mean you are recommending pulling this PR in after all?
> 
> Yes.

Okay, great! I'll pull the current PR's in, update the external dependency libs (including latest SPIRV-Cross)...and we'll call it a release!

PR #1077, opened 2020-09-25
MVKQueryPool: Fix ending a query outside a render pass.

cdavis5e:
I missed this when I fixed occlusion queries begun outside a render
pass.


cdavis5e:
This fixes a crash that was introduced by #1006. It really needs to go in the SDK.

billhollings (approved):
👍🏻

PR #1078, opened 2020-09-26
MVKCmdFillBuffer: Round size down, not up.

cdavis5e:
According to the Vulkan spec:

> If `VK_WHOLE_SIZE` is used and the remaining size of the buffer is not
> a multiple of 4, then the nearest **smaller** multiple is used.
> [emphasis added]

Therefore, we should round down when calculating the number of words to
write.


billhollings (approved):
Good catch!

PR #1079, opened 2020-09-28
MVKPipeline: Don't rely on the index buffer to tell apart indexed draws.

cdavis5e:
This was a good heuristic, because the index buffer must be bound for
indexed draws. However, it may also be bound for non-indexed draws--for
example, if an indexed draw were immediately followed by a non-indexed
draw, as happens in some `dEQP-VK.synchronization.*` tests. Therefore,
we can't tell from the presence or absence of the index buffer what kind
of draw we're in. We'll have to keep track of this state in the command
encoder.


billhollings (approved):
👍🏻

PR #1082, opened 2020-09-28
MVKCmdCopyImage: Fix a copy-pasto.

cdavis5e:
I know I said I wouldn't have any more, but this fixes a problem with
my change #1064 that could trip Metal's validation layer and/or result in
no data being copied.


billhollings (approved):
Thanks for the catch!

cdavis5e:
Ugh, there's another copy-pasto on that line which I missed, and now it's too late for the release.

billhollings:
@cdavis5e 

>Ugh, there's another copy-pasto on that line which I missed, and now it's too late for the release.

Submit it as soon as you can this morning and there might be time to switch the MoltenVK commit in the SDK build!

PR #1086, opened 2020-10-01
MVKCmdCopyImage: Only use the multi-layer copy method if available.

cdavis5e:
It is only available starting in Metal 3.0 (macOS 10.15, iOS/tvOS 13).


cdavis5e:
This one fixes a crash introduced by #1064 that could occur on macOS 10.14 and lower if copying between two images and the extent covers both images fully. I know it's a little late, but can this one go in the SDK?

@richard-lunarg

billhollings (approved):
Thursday is usually their SDK cut date...so likely too late...but I'll check with them on the off chance they are delayed.

PR #1087, opened 2020-10-01
Use VK_KHR_image_format_list to disable MTLTextureUsagePixelFormatView where possible.

billhollings:
Don't use `MTLTextureUsagePixelFormatView` if image will only allow views that use same format or its sRGB/linear variation.

- `MVKImage` track `VkImageFormatListCreateInfo::pViewFormats` and validate `MVKImageView` format against them.
- Add `MVKImage::getMTLTextureUsage()` and move tests for dedicated aliasables, compressed formats, and minimum image usage into it.
- Add `MVKMTLFormatDesc::mtlPixelFormatLinear` and `MVKPixelFormats::compatibleAsLinearOrSRGB()` to track linear counterparts to each Metal `sRGB` format.
- Don't use `MTLTextureUsagePixelFormatView` if image will only allow views that use same format or its sRGB/linear variation.
- Propagate `VkImageFormatListCreateInfo` from mutable format swapchain to swapchain images.

### DIFF
--- a/MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
@@ -1564,11 +1564,6 @@ MVKImageView::MVKImageView(MVKDevice* device,
 		}
 	}
 
-    // Validate whether the image view configuration can be supported
-	if ( !_image->getIsValidViewFormat(pCreateInfo->format) ) {
-		setConfigurationResult(reportError(VK_ERROR_FORMAT_NOT_SUPPORTED, "vkCreateImageView(): The format is not in the list of allowed view formats declared in the VkImageFormatListCreateInfo provided in vkCreateImage()."));
-	}
-
 	VkImageType imgType = _image->getImageType();
 	VkImageViewType viewType = pCreateInfo->viewType;
 
--- a/MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
@@ -1559,46 +1559,42 @@ MVKImageView::MVKImageView(MVKDevice* device,
 	}
 
     // Validate whether the image view configuration can be supported
-    if ( _image ) {
+	if ( !_image->getIsValidViewFormat(pCreateInfo->format) ) {
+		setConfigurationResult(reportError(VK_ERROR_FORMAT_NOT_SUPPORTED, "vkCreateImageView(): The format is not in the list of allowed view formats declared in the VkImageFormatListCreateInfo provided in vkCreateImage()."));
+	}
 
-		if ( !_image->getIsValidViewFormat(pCreateInfo->format) ) {
-			setConfigurationResult(reportError(VK_ERROR_FORMAT_NOT_SUPPORTED, "vkCreateImageView(): The format is not in the list of allowed view formats declared in the VkImageFormatListCreateInfo provided in vkCreateImage()."));
+	VkImageType imgType = _image->getImageType();
+	VkImageViewType viewType = pCreateInfo->viewType;
+
+	// VK_KHR_maintenance1 supports taking 2D image views of 3D slices for sampling.
+	// No dice in Metal. But we are able to fake out a 3D render attachment by making the Metal view
+	// itself a 3D texture (when we create it), and setting the rendering depthPlane appropriately.
+	if ((viewType == VK_IMAGE_VIEW_TYPE_2D || viewType == VK_IMAGE_VIEW_TYPE_2D_ARRAY) && (imgType == VK_IMAGE_TYPE_3D)) {
+		if (!mvkIsOnlyAnyFlagEnabled(_usage, VK_IMAGE_USAGE_COLOR_ATTACHMENT_BIT)) {
+			setConfigurationResult(reportError(VK_ERROR_FEATURE_NOT_PRESENT, "vkCreateImageView(): 2D views on 3D images can only be used as color attachments."));
 		}
+	}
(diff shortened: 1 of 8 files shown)


billhollings:
@cdavis5e 

I think the logic for `MTLTextureUsagePixelFormatView` makes sense, and is probably the best we can do...but a sanity check from you would be helpful.

In particular, I'm not sure about impact or suitability for compressed sRGB texture variations, and specifically, ASTC. I'm left wondering a little about why Apple's naming syntax is different for their `ASTC_*_LDR` linear variations, relative to naming their other linear formats.

Also, my movement of logic for dedicated aliasable images, and compressed 3D image format usages.

MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
```diff
@@ -1531,41 +1559,42 @@
...
-		// No dice in Metal. But we are able to fake out a 3D render attachment by making the Metal view
-		// itself a 3D texture (when we create it), and setting the rendering depthPlane appropriately.
-        if ((viewType == VK_IMAGE_VIEW_TYPE_2D || viewType == VK_IMAGE_VIEW_TYPE_2D_ARRAY) && (imgType == VK_IMAGE_TYPE_3D)) {
-            if (!mvkIsOnlyAnyFlagEnabled(_usage, VK_IMAGE_USAGE_COLOR_ATTACHMENT_BIT)) {
-                setConfigurationResult(reportError(VK_ERROR_FEATURE_NOT_PRESENT, "vkCreateImageView(): 2D views on 3D images can only be used as color attachments."));
-            }
-        }
+	if ( !_image->getIsValidViewFormat(pCreateInfo->format) ) {
+		setConfigurationResult(reportError(VK_ERROR_FORMAT_NOT_SUPPORTED, "vkCreateImageView(): The format is not in the list of allowed view formats declared in the VkImageFormatListCreateInfo provided in vkCreateImage()."));
+	}
```
cdavis5e, line 1564: Isn't it [just undefined behavior](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/html/chap12.html#VUID-VkImageViewCreateInfo-pNext-01585) to use view formats outside that list?

cdavis5e (changes requested):
A couple of things.

cdavis5e:
> I think the logic for `MTLTextureUsagePixelFormatView` makes sense, and is probably the best we can do...but a sanity check from you would be helpful.

Looks fine. 👍

MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
billhollings: >Isn't it just undefined behavior to use view formats outside that list?

Hmmm...that rule has explicit normative terminology. I always read normative VU's to return an error, rather than undefined behavior. 

Admittedly, monitoring VU's is what the validation layer is doing, and drivers don't necessarily need to repeat that validation.

Creating a view is set-up code, so leaving a simple test like this in doesn't really cause a performance hit. But this does raise the issue that we are fairly arbitrary with our VU validations within MoltenVK. Sometimes we do them, sometimes not. 

Perhaps we should discuss a policy that says MoltenVK should only perform validation that is needed because of Metal limitations that cause non-conformance, but **_not_** perform validation that is defined by the spec, because the validation layers are responsible for that?

MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
```diff
@@ -852,7 +874,6 @@
     // If this is a storage image of format R32_UINT or R32_SINT, or MUTABLE_FORMAT is set
     // and R32_UINT is in the set of possible view formats, then we must use a texel buffer,
     // or image atomics won't work.
-    // TODO: Also add handling for VK_KHR_image_format_list here.
     _isLinearForAtomics = _isLinear && mvkIsAnyFlagEnabled(_usage, VK_IMAGE_USAGE_STORAGE_BIT) &&
                           ((_vkFormat == VK_FORMAT_R32_UINT || _vkFormat == VK_FORMAT_R32_SINT) ||
                            (_hasMutableFormat && pixFmts->getViewClass(_vkFormat) == MVKMTLViewClass::Color32));
```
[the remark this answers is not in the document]
billhollings: Done. Please verify my logic.

cdavis5e (approved):
👍

MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
cdavis5e: > Perhaps we should discuss a policy that says MoltenVK should only perform validation that is needed because of Metal limitations that cause non-conformance, but **_not_** perform validation that is defined by the spec, because the validation layers are responsible for that?

That sounds reasonable. It's what I was thinking we should do.

MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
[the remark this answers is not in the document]
cdavis5e: Looks OK.

MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
billhollings: >That sounds reasonable. It's what I was thinking we should do.

Okay...let's adopt that from now on. I've removed the code above, and I've added text regarding a validation code policy to the _Contributing_ section of the `README.md` document. Let me know if that wording looks good.

Moving forward, we can each remove excess validation code by attrition as we come across it during other development.

cdavis5e (approved):
👍

PR #1088, opened 2020-10-03
MVKImagePlane: Reset _mtlTexture after releasing.

gnoliyil:
For an `MVKImage`, when we call `vkUseIOSurfaceMVK()`, the `_mtlTexture` object of each plane will be released but not reset. And later when we destroy the `MVKImage`, the destructor of `MVKImagePlane` calls `releaseMTLTexture()` again, which will cause a crash because of double releasing of an object.

To solve this we need to reset `_mtlTexture` to `nil` after it is successfully released.


gnoliyil:
@dreveman @cdotstout

PR #1089, opened 2020-10-03
MVKImage: Check plane count in IOSurface only if it has chroma subsampling

gnoliyil:
`useIOSurface()` sets the plane information in `IOSurface` only if the image format has chroma subsampling. Whenever we import an `IOSurface`, we should only check count of planes when the format has chroma subsampling.

This resolves a crash in the following case to let two `VkImage`s share the same `IOSurface`:

```
VkImage image1, image2;

// the format in |info| is a non-chroma-subsampling format,
// e.g. R8G8B8A8_UNORM.
vkCreateImage(info, &image1);
vkCreateImage(info, &image2);

IOSurfaceRef ioSurface;
vkUseIOSurfaceMVK(image1, nil);
vkGetIOSurfaceMVK(image1, &ioSurface);
vkUseIOSurfaceMVK(image2, ioSurface);
```


gnoliyil:
@dreveman @cdotstout FYI

Issue #929, opened 2020-06-17
Using multiple swap chain kills performance on macOS


dolphineye:
Hi guys,

Today, I tried to use multiple swap chains for multiple windows rendering.

When I create my second window with my secondary swap chain, my performance drops by 50%. Note that I use GLFW as my window management system and I use MoltenVK 1.2.135 (LunarG). I also use triple buffering (3 swap chain images).

The Xcode profiler results indicate that my main thread is spending 32 msec waiting for the next drawables (2 x 16 msec to be accurate).

I'm pretty sure I'm missing something, so I've decided to post an issue in case someone would have a clue.

Thanks for helping!

dolphineye:
Hey, just bumping this issue with additional information. We ran a test with our app in a Linux environment  with Vulkan, and everything's fine. So, it's likely to be something related to MoltenVK.

What information would you need for a more detailed report to address that issue?

Thanks

FunMiles:
Could this be the same issue as in https://github.com/KhronosGroup/MoltenVK/issues/581 ? I have had a drop from 60fps to 30fps in general when having more than one window ever since I've started playing with more than one window. 
I've observed stranger result still more recently but haven't had time to look into them.

billhollings:
They might indeed be related...but #581 discusses a wide range of situations.

> I have had a drop from 60fps to 30fps in general when having more than one window ever since I've started playing with more than one window.

Per this thread, are you using more than one swapchain to do this?

dolphineye:
Hi guys,

Thanks for helping 👍 Just back from vacation, so back to business now :)

> With your triple buffering, are you using 3 images per swapchain? So 6 swapchain images total?

Yes, I'm using 3 swap chain images for each window. Hence N x 3 swap chain images depending on how many of them are open.

> Do you have a sample app that you can upload to the cloud so we can investigate?

Working on it. I'll get back to you as soon as I have it for you.

dolphineye:
I've just stumbled on this post: https://developer.apple.com/forums/thread/651571

billhollings:
> I've just stumbled on this post, if that helps: https://developer.apple.com/forums/thread/651571

Interesting. Thanks for highlighting that.

Hmmm...we can certainly experiment with using `-[MTLDrawable present]` and `-[MTLDrawable presentAtTime:]` instead of `-[MTLCommandBuffer presentDrawable:]` and `-[MTLCommandBuffer presentDrawable:atTime:]`.

[...]

PR #1090, opened 2020-10-03
Move Metal drawable presentation from MTLCommandBuffer to MTLDrawable.

billhollings:
Update `MVKPresentableSwapchainImage::presentCAMetalDrawable()` to create a
`MTLCommandBuffer` scheduled-handler and present the `MTLDrawable` from there.

According to feedback from Apple, it is more performant to call `MTLDrawable` present from within a
`MTLCommandBuffer` scheduled-handler than it is to call `MTLCommandBuffer presentDrawable:`.

Pass presentation timing info as a struct to simplify calls.

This may fix issue #929.

### DIFF
--- a/Docs/Whats_New.md
+++ b/Docs/Whats_New.md
@@ -21,6 +21,8 @@ Released TBD
 - Use `VK_KHR_image_format_list` to disable `MTLTextureUsagePixelFormatView` 
   if only swizzles or `sRGB` conversion will be used for image views, improving 
   performance on *iOS* by allowing Metal to use lossless texture compression.
+- Move *Metal* drawable presentation from `MTLCommandBuffer` to `MTLDrawable`
+  to improve performance and reduce blocking.
 
 
 
--- a/MoltenVK/MoltenVK/GPUObjects/MVKImage.h
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKImage.h
@@ -429,6 +429,14 @@ typedef struct MVKSwapchainImageAvailability {
 	bool operator< (const MVKSwapchainImageAvailability& rhs) const;
 } MVKSwapchainImageAvailability;
 
+/** VK_GOOGLE_display_timing extension info */
+typedef struct  {
+	MVKPresentableSwapchainImage* presentableImage;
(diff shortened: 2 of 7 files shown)


billhollings:
@danginsburg Can you verify `VK_GOOGLE_display_timing` still works as expected wrt your app requirements, please?

mbarriault:
> According to feedback from Apple, it is more performant to call MTLDrawable present from within a
MTLCommandBuffer scheduled-handler than it is to call MTLCommandBuffer presentDrawable:.

Just a curious aside, are you able to expand on this? The documentation says

> This is a convenience method that calls the drawable’s present() method after the command queue schedules the command buffer for execution. You can’t call this method after you commit the command buffer.

which suggests that the two methods should be identical. (I don't doubt that there's important implementation details beyond what the documentation suggests, just curious to know them.)

danginsburg:
> @danginsburg Can you verify `VK_GOOGLE_display_timing` still works as expected wrt your app requirements, please?

@billhollings I reviewed the commit and it looks like it would behave in an equivalent manner with VK_GOOGLE_display_timing, but I'll have to test it out to be sure.

billhollings:
>looks like it would behave in an equivalent manner...but I'll have to test it out to be sure

Okay. Thanks. I agree. I'll pull it in and if subsequent testing indicates a problem, I can address it with a fix.

billhollings:
@mbarriault 

> are you able to expand on this?

This was a recommendation from an Apple engineer during discussions about possible performance improvements within MoltenVK. As I understand the feedback, there may be some internal retentions and lockings within `MTLCommandBuffer presentDrawable:` that can interfere with drawable reuse.

I've asked them for some background documentation or validation of this, and if I get anything back, I'll report it here.

mbarriault:
@billhollings 

> This was a recommendation from an Apple engineer during discussions about possible performance improvements within MoltenVK. As I understand the feedback, there may be some internal retentions and lockings within `MTLCommandBuffer presentDrawable:` that can interfere with drawable reuse.
> 
> In addition, I came across [this discussion](https://developer.apple.com/forums/thread/651571), which corroborated the recommendation.
> 
> I've asked Apple for some more background documentation or validation of this, and if I get anything back, I'll report it here.

Ok cool, figured it was something along these lines. Thanks!

PR #1094, opened 2020-10-07
MVKPhysicalDevice remove need to call initGPUInfoProperties() twice.

billhollings:
Move setting of limits from `MVKPhysicalDevice::initProperties()` into separate `initLimits()` function. Call `initProperties()` before `initMetalFeatures()` and `initLimits()` after `initMetalFeatures()`.


billhollings:
@cdavis5e This fixes the leftover from #1072. I think I've worked out all the dependencies correctly, and I've tested using `vulkaninfo`, and all properties and limits seem to be unchanged, but I wouldn't mind you checking my logic sanity.

cdavis5e (approved):

PR #1095, opened 2020-10-08
MVKPixelFormats: Add 0.5 ULP to clear values for normalized formats.

cdavis5e:
I've noticed that in some tests, the clear values seem to be off by
about one ULP, and the tests subsequently fail. This should cause Metal
to round up instead of down and fix those tests.


billhollings (approved):
Uggh! Talk about the devil in the details! Thanks for sorting this out.

PR #1104, opened 2020-10-14
MVKPipeline: Fix calculation of atomic image buffer addresses.

cdavis5e:
Align the row address using the linear texture alignment for an
`R32Uint` texture.

Pull in the SPIRV-Cross changes needed to make this work.


cdavis5e:
I wonder if this should wait until KhronosGroup/SPIRV-Cross#1494 is in, so I can pull that in, too.

billhollings (approved):

billhollings:
>I wonder if this should wait until KhronosGroup/SPIRV-Cross#1494 is in, so I can pull that in, too.

Let me know. Holding off pulling this in until hearing from you.

billhollings (changes requested):
Marking as waiting for feedback regarding bumping SPIRV-Cross version so I don't pull this in by mistake.

billhollings (approved):

PR #1109, opened 2020-10-16
MVKPhysicalDevice: Correct fragment input component limit.

cdavis5e:
Starting in iOS/tvOS 13, Apple family 4 and up support 124 fragment
input components, matching macOS. Reduce Mac max to 124 to prevent Metal
from complaining when using more than 124 components.

Tessellation control shaders must support at least 120 per-patch output
components, so make sure the limit is that value.

### DIFF
--- a/MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
@@ -1643,7 +1643,11 @@ void MVKPhysicalDevice::initLimits() {
     }
 
 #if MVK_TVOS
-    _properties.limits.maxFragmentInputComponents = 60;
+    if (mvkOSVersionIsAtLeast(13.0) && supportsMTLGPUFamily(Apple4)) {
+        _properties.limits.maxFragmentInputComponents = 124;
+    } else {
+        _properties.limits.maxFragmentInputComponents = 60;
+    }
 
     if (supportsMTLFeatureSet(tvOS_GPUFamily2_v1)) {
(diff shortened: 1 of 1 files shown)


MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
```diff
@@ -1693,7 +1701,7 @@
         _properties.limits.maxTessellationControlPerVertexInputComponents = _properties.limits.maxVertexOutputComponents;
         _properties.limits.maxTessellationControlPerVertexOutputComponents = _properties.limits.maxTessellationControlPerVertexInputComponents;
         // Reserve a few for the tessellation levels.
-        _properties.limits.maxTessellationControlPerPatchOutputComponents = _properties.limits.maxFragmentInputComponents - 8;
+        _properties.limits.maxTessellationControlPerPatchOutputComponents = std::max(_properties.limits.maxFragmentInputComponents - 8, 120u);
```
billhollings, line 1704: @cdavis5e 

Since the current max for `maxFragmentInputComponents` is 124, looks like this is always going to be 120.

Are you just future-proofing this in case future OS's bump `maxFragmentInputComponents` higher?

billhollings (approved):

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
cdavis5e: > Are you just future-proofing this in case future OS's bump `maxFragmentInputComponents` higher?

Yes.

PR #1111, opened 2020-10-16
MVKGPUCapture: Make sure the MTLCaptureScope has only one reference.

cdavis5e:
For some reason, when we create the `MTLCaptureScope` with capture
enabled, in addition to the reference from object creation, Metal
retains the object for no apparent reason in the base `MTLCaptureScope`
initializer. This causes us to leak the capture scope--and with it, the
queue. Let's hope Apple isn't doing anything with that extra reference.

I admit this is a horrible kluge and we shouldn't need it. But without this,
I can't get through the CTS without it hanging.

### DIFF
--- a/MoltenVK/MoltenVK/OS/MVKGPUCapture.mm
+++ b/MoltenVK/MoltenVK/OS/MVKGPUCapture.mm
@@ -61,6 +61,14 @@ MVKGPUCaptureScope::MVKGPUCaptureScope(MVKQueue* mvkQueue) : _queue(mvkQueue) {
 	if (mvkOSVersionIsAtLeast(kMinOSVersionMTLCaptureScope)) {
 		_mtlCaptureScope = [[MTLCaptureManager sharedCaptureManager] newCaptureScopeWithCommandQueue: _mtlQueue];	// retained
 		_mtlCaptureScope.label = @(_queue->getName().c_str());
+		// Due to a retain bug in Metal when the capture layer is installed, capture scopes
+		// can have too many references on them. Release the excess references so the scope--
+		// and the command queue--aren't leaked. This is a horrible kludge that depends on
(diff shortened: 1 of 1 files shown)


billhollings (approved):
Ugh! Good catch!

billhollings:
@cdavis5e 

>I can't get through the CTS without it hanging.

You've clearly been doing a fair bit of CTS lately. Are you working based on a coverage plan...or just applying it during other bug fixes?

Asking because I should have some time coming up to get back to significant CTS efforts...and it would probably help our collective efforts to coordinate on that to make sure our work is spread out from a coverage perspective.

Any CTS areas you want to focus on...and other areas you recommend me focusing on as I get back into it?

cdavis5e:
> Are you working based on a coverage plan...or just applying it during other bug fixes?

A little of both.

> Any CTS areas you want to focus on...?

I've mostly been trying to get it to run all the way through with the validation layers on. Right now, I know of a few tests where this breaks: `dEQP-VK.spirv_assembly.instruction.graphics.image_sampler.imageread.storage_image.optypeimage_mismatch.*` and `dEQP-VK.binding_model.descriptorset_random.*`. I think I'll focus on those for now.

cdavis5e:
Actually, right now, with the upcoming Apple Silicon release, I'm probably going to focus more on bringing up MoltenVK on Apple Silicon for the time being.

billhollings:
>`VK.spirv_assembly.instruction...`

Yeah...those are oddballs. I noticed many of them breaking Metal validations a while ago. If you're down to only a few of them...that's great progress.

>I'm probably going to focus more on bringing up MoltenVK on Apple Silicon for the time being.

That would be helpful. I don't have access to an Apple Silicon device right now.

I'll start working on the general CTS run through that you described.

PR #1113, opened 2020-10-19
MVKImage: Always use texel buffers for linear images in MTLHeaps.

cdavis5e:
This is particularly important if the resource will be aliased with
another resource; the memory in this case must have a linear layout.


billhollings (approved):
Hmmm...the logic of this function is getting rather convoluted. Hopefully we can clean it up soon.

PR #1115, opened 2020-10-19
MVKImage: Make sure plane heap offsets are properly aligned.

cdavis5e:
If the texture is not properly aligned, creation will fail. Make sure
we ask for enough memory to hold all planes with alignment, or a plane
could wind up beyond the bounds of the heap.


billhollings (approved):
Whew!

PR #1120, opened 2020-10-22
MVKPipeline: shorten vertex attribute format's length when stride < size

js6i:
Metal currently does not support overlapping attribute loads and asserts on
them. Prevent it by shortening format's vector length to fit within the stride,
but print a cautionary message.

An example of affected program is Kingdom Come: Deliverance, which sets attribute
format to RGBA32 in IA and binding stride to 12, while the shader input is just
a float3.

### DIFF
--- a/MoltenVK/MoltenVK/GPUObjects/MVKPipeline.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKPipeline.mm
@@ -688,6 +688,33 @@ static VkFormat mvkFormatFromOutput(const SPIRVShaderOutput& output) {
 	return VK_FORMAT_UNDEFINED;
 }
 
+// Returns a format of the same base type with vector length adjusted to fit size.
+static MTLVertexFormat mvkAdjustFormatVectorToSize(MTLVertexFormat format, uint32_t size) {
+#define MVK_ADJUST_FORMAT_CASE(size_1, type, suffix) \
+	case MTLVertexFormat##type##4##suffix: if (size >= 4 * (size_1)) { return MTLVertexFormat##type##4##suffix; } \
+	case MTLVertexFormat##type##3##suffix: if (size >= 3 * (size_1)) { return MTLVertexFormat##type##3##suffix; } \
+	case MTLVertexFormat##type##2##suffix: if (size >= 2 * (size_1)) { return MTLVertexFormat##type##2##suffix; } \
+	case MTLVertexFormat##type##suffix:    if (size >= 1 * (size_1)) { return MTLVertexFormat##type##suffix; } \
+	return MTLVertexFormatInvalid;
+
+	switch (format) {
+		MVK_ADJUST_FORMAT_CASE(1, UChar, )
+		MVK_ADJUST_FORMAT_CASE(1, Char, )
+		MVK_ADJUST_FORMAT_CASE(1, UChar, Normalized)
+		MVK_ADJUST_FORMAT_CASE(1, Char, Normalized)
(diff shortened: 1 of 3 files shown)


MoltenVK/MoltenVK/GPUObjects/MVKPipeline.mm
```diff
@@ -1177,17 +1204,18 @@ static VkFormat mvkFormatFromOutput(const SPIRVShaderOutput& output) {
...
+					attrSize = getPixelFormats()->getBytesPerBlock(pVKVA->format);
 					if (pVKVB->stride == 0) {
 						// The step is set to constant, but we need to change stride to be non-zero for metal.
 						// Look for the maximum offset + size to set as the stride.
 						uint32_t vbIdx = getMetalBufferIndexForVertexAttributeBinding(pVKVB->binding);
 						auto vbDesc = inputDesc.layouts[vbIdx];
 						uint32_t strideLowBound = vaOffset + attrSize;
 						if (vbDesc.stride < strideLowBound) vbDesc.stride = strideLowBound;
-					} else if (vaOffset + attrSize > pVKVB->stride) {
+					} else if (vaOffset && vaOffset + attrSize > pVKVB->stride) {
```
billhollings, line 1218: Ah! Good. When I was reviewing this PR earlier...I noticed that `attrSize > pVKVB->stride` would trigger an alternate buffer to be created and was going to ask you about that. Glad you caught it too and updated it.

MoltenVK/MoltenVK/GPUObjects/MVKPipeline.mm
```diff
@@ -1208,7 +1236,16 @@ static VkFormat mvkFormatFromOutput(const SPIRVShaderOutput& output) {
...
 			auto vaDesc = inputDesc.attributes[pVKVA->location];
-			vaDesc.format = (decltype(vaDesc.format))getPixelFormats()->getMTLVertexFormat(pVKVA->format);
+			auto mtlFormat = (decltype(vaDesc.format))getPixelFormats()->getMTLVertexFormat(pVKVA->format);
+			if (pVKVB->stride && attrSize > pVKVB->stride) {
+				/* Metal does not support overlapping loads. Truncate format vector length to prevent an assertion
+				 * and hope it's not used by the shader. */
+				auto newFormat = (decltype(vaDesc.format))mvkAdjustFormatVectorToSize((MTLVertexFormat)mtlFormat, pVKVB->stride);
+				reportError(VK_SUCCESS, "Found attribute with size (%u) larger than it's binding's stride (%u). Changing descriptor format from %u to %u.",
+					attrSize, pVKVB->stride, (uint32_t)mtlFormat, (uint32_t)newFormat);
+				mtlFormat = newFormat;
```
billhollings, line 1246: To make this a little more user-friendly, instead of reporting the formats as integers, can you add to `MVKPixelFormats`:

```
const char* MVKPixelFormats::getName(MTLVertexFormat mtlFormat) {
    return getMTLVertexFormatDesc(mtlFormat).name;
}
```

and use it in the `reportError("...format from %s to %s")`?

See the similar `MVKPixelFormats` functions for the names of `VkFormat` and `MTLPixelFormat`.

billhollings (changes requested):
Looks good. Just a small change request on the reporting part.

MoltenVK/MoltenVK/GPUObjects/MVKPipeline.mm
js6i: Good point, added.

billhollings (approved):

PR #1123, opened 2020-10-26
MVKPhysicalDevice: Enable strictLines for Intel and NVIDIA.

cdavis5e:
Both Intel and NV Vulkan drivers enable `strictLines`. My testing shows
in fact that AMD and Apple Silicon exhibit the non-strict-line behavior,
while Intel shows the strict-line behavior.

### DIFF
--- a/MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
@@ -1726,8 +1726,10 @@ void MVKPhysicalDevice::initLimits() {
     _properties.limits.lineWidthRange[1] = 1;
     _properties.limits.lineWidthGranularity = 1;
 
+    static const uint32_t kIntelVendorId = 0x8086;
+    static const uint32_t kNVVendorId = 0x10de;
     _properties.limits.standardSampleLocations = VK_TRUE;
-    _properties.limits.strictLines = VK_FALSE;
+    _properties.limits.strictLines = _properties.vendorID == kIntelVendorId || _properties.vendorID == kNVVendorId;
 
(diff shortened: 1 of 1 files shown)


billhollings (approved):
Interesting. Since Metal doesn't support line widths, how can you even tell? 🤔

mbechard:
I'm not sure if this is correct. Testing on an Intel Iris Pro Graphics and line rendering definitely seem to be following the Diamond Exit Rule for rasterization.
Does Metal define what their line primitive rasterization rules are anywhere? I couldn't seem to find it.

mbechard:
For my test case, I'm rendering in a 5x1 1xAA buffer using ortho projection with coordinates matching pixels. (0= left edge, 5 = right edge).
If my line start point on the left is located at (0.99, 0.5), then the first pixel is rasterized. While (0.99, 0.49) will not be rasterized.
(0.5, 0.2) is rasterized
(0.7, 0.2) is not rasterized.

Similarly for the endpoint.
(4.99, 0.5) is not rasterized.
(4.99, 0.35) is rasterized.
This all follows the diamond exit rule, and thus not strictLines.

cdavis5e:
> Does Metal define what their line primitive rasterization rules are anywhere?

It doesn't.

Great. Now we may have to contend with `strictLines` differing not just between vendors, but between devices by a specific vendor.

mbechard:
What Intel GPU were you testing on?

cdavis5e:
Intel HD Graphics 530.

Let's see what vulkan.gpuinfo.org has to say...

Well I'll be darned. It has `strictLines` as `VK_FALSE`. I guess I was wrong there.

mbechard:
Yeah, I would be surprised if Intel doesn't rasterize using Bresenham, since that's what OpenGL and DirectX use.

Issue #625, opened 2019-05-30
Dota 2 GPU Hang with MTLEvent change 


danginsburg:
I've been getting GPU hangs on startup in Dota 2 on Macbook Pro (Mid 2012) NV GT 650M on macOS 10.14.5.  I was able to use git bisect and identify that the commit that is causing the hang is b02d354be576fffb4b01501e504fe341473c75e8 "MVKSemaphore: Use MTLEvent for device-side synchronization."

@cdavis5e

cdavis5e:
This is a known issue. See #602.

billhollings:
PR #633 provides an automated workaround for this issue.

PR #1124, opened 2020-10-27
Support the VK_KHR_timeline_semaphore extension.

cdavis5e:
This implementation uses `MTLSharedEvent` where possible, and emulates
it on the host otherwise. Unlike binary semaphores, `MTLSharedEvent`s
map well to timeline semaphores; there should be no problems using them
when they're available.

I'm extremely confident in the `MTLSharedEvent`-based implementation. It
passes nearly all the synchronization tests. I'm less confident in the
emulated implementation.


MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
```diff
@@ -2668,12 +2678,30 @@ static uint32_t mvkGetEntryProperty(io_registry_entry_t entry, CFStringRef prope
...
+		}
 	} else {
-		return new MVKSemaphoreEmulated(this, pCreateInfo);
+		if (_useMTLFenceForSemaphores) {
+			return new MVKSemaphoreMTLFence(this, pCreateInfo);
+		} else if (_useMTLEventForSemaphores) {
+			return new MVKSemaphoreMTLEvent(this, pCreateInfo);
+		} else {
+			return new MVKSemaphoreEmulated(this, pCreateInfo);
+		}
```
billhollings, line 2704: Currently, by default, `_useMTLFenceForSemaphores` is enabled and `_useMTLEventForSemaphores` is disabled, which means regular `VkSemaphores` will prefer using `MTLFence` over `MTLEvent`.

Given that if a timeline semaphore is requested, we must use `MTLSharedEvent`, should we re-evaluate and possibly flip this? Is there any drive or appetite to prefer using `MTLEvent` over `MTLFence` for regular `VkSemaphores`?

billhollings (approved):
Looks great. Just one question.

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
cdavis5e: IIRC the reason for preferring `MTLFence` for binary semaphores is that we observed problems when `MTLEvent`s were used with presentation, where waits would time out and subsequently cause the system to revoke access to the device. The presentation code has changed significantly since then, but I'm still investigating whether or not it's safe to start using `MTLEvent` again.

(I keep forgetting that most people don't have memories as long as mine. How ironic.)

cdavis5e: Now where was that issue again...?

Oh yeah, there were two, #602 and #625.

billhollings: >I keep forgetting that most people don't have memories as long as mine. How ironic. (🤦🏻‍♂️...for the irony joke).

Yeah...raw memory is definitely not my strong suit. I'd not survive long in medical school. 😉

Sorry...I didn't expect you to troll back into the previous change triggers...but was more looking for your thoughts on whether either our code has evolved...or Apple's `MTLEvent` code has...based on the sync testing you were doing.

>The presentation code has changed significantly since then, but I'm still investigating whether or not it's safe to start using MTLEvent again.

Okay. Thanks for the review. If you do continue to investigate and it reveals that `MTLEvents` make sense, we can flip back.

PR #1126, opened 2020-10-28
MVKDevice: Set properties for Apple Silicon GPUs on macOS.

cdavis5e:


MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
```diff
@@ -1851,6 +1879,22 @@ static uint32_t mvkGetEntryProperty(io_registry_entry_t entry, CFStringRef prope
 	_properties.deviceType = isIntegrated ? VK_PHYSICAL_DEVICE_TYPE_INTEGRATED_GPU : VK_PHYSICAL_DEVICE_TYPE_DISCRETE_GPU;
 	strlcpy(_properties.deviceName, _mtlDevice.name.UTF8String, VK_MAX_PHYSICAL_DEVICE_NAME_SIZE);
 
+	if (supportsMTLGPUFamily(Apple5)) {
+		// This is an Apple GPU. It won't have a 'device-id' property, so fill it in
+		// like on iOS/tvOS.
+		_properties.vendorID = 0x106b;	// Apple's PCI ID
+#if (defined(__MAC_OS_X_VERSION_MAX_ALLOWED) && __MAC_OS_X_VERSION_MAX_ALLOWED >= 101600)
+		if (supportsMTLGPUFamily(Apple6)) {
```
billhollings, line 1887: There are several other places in `MVKDevice.mm` that need to be updated to support `Apple6`, both in `MVK_MACOS` and `MVK_IOS` conditional areas. For example, the iOS version of this code section...plus search for `Apple5` for a couple of others.

Curiously, Apple [makes reference](https://developer.apple.com/metal/Metal-Feature-Set-Tables.pdf) to `Apple7` (A14) as well...but it doesn't look to be supported in the SDK in Xcode 12.2b3 yet.

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
```diff
@@ -727,6 +727,14 @@
 #if MVK_MACOS
 	addSurfFmt(MTLPixelFormatRGB10A2Unorm);
 	addSurfFmt(MTLPixelFormatBGR10A2Unorm);
+#if (defined(__MAC_OS_X_VERSION_MAX_ALLOWED) && __MAC_OS_X_VERSION_MAX_ALLOWED >= 101600)
```
billhollings, line 730: Ugh. In a pinch I threw similar guard code in to support `MTLLanguageVersion2_3`, but it's ugly and I wouldn't want it to propagate too far. And I guess when Xcode 12.2 comes out of beta, and becomes the standard, we can remove all of these.

If we don't think we can strip this out soon (if we need to support Xcode 12.0 or 11.7), and it might propagate even further (see other review note), it might be worth wrapping

     (defined(__MAC_OS_X_VERSION_MAX_ALLOWED) && __MAC_OS_X_VERSION_MAX_ALLOWED >= 101600)

into something like a `MVK_MACOS_ARM` or `MVK_MACOS_APPLE_SILICON` macro?

That way, the intention would be clearer, and it would be easier to search and replace later.

billhollings (changes requested):

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
cdavis5e: > Curiously, Apple [makes reference](https://developer.apple.com/metal/Metal-Feature-Set-Tables.pdf) to `Apple7` (A14) as well...but it doesn't look to be supported in the SDK in Xcode 12.2b3 yet.

Not the macOS SDK, or the tvOS SDK for that matter, but it is present in the iOS SDK.

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
cdavis5e: Agreed. That looks much cleaner. I also added one for the iOS conditional.

cdavis5e: Huh. So using the `defined` operator in a macro produces a warning, but not checking if the symbol is defined doesn't. Go figure.

billhollings (approved):

PR #1130, opened 2020-10-30
Support Apple GPU pixel formats with Apple Silicon on Mac.

cdavis5e:
Conversely, disable capabilities supported by desktop GPUs that are not
supported on Apple GPUs.

Sadly, no support for 16-bit packed pixel formats on non-Apple GPUs.


cdavis5e:
Travis is down?

billhollings (approved):
Thanks for handling all this bookkeeping.

PR #1131, opened 2020-11-03
Initial support for the VK_EXT_descriptor_indexing extension

billhollings:
This PR provides initial support for `VK_EXT_descriptor_indexing` extension, using discrete resource handling, and limited to arrays of 96/128 (iOS/macOS) textures and 16 samplers per stage.

A second version of this implementation will move to using Metal Argument buffers, which will dramatically increase the number of available texture and samplers on Tier 2 devices.

CTS descriptor indexing tests do well. There are a few outstanding failures due to SPIRV handling of arrays of atomics.

@cdavis5e Please review.


MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
```diff
@@ -315,6 +339,33 @@
...
+				pDescIdxProps->shaderStorageImageArrayNonUniformIndexingNative		= _metalFeatures.arrayOfTextures;
+				pDescIdxProps->shaderInputAttachmentArrayNonUniformIndexingNative	= _metalFeatures.arrayOfTextures;
+				pDescIdxProps->robustBufferAccessUpdateAfterBind					= _features.robustBufferAccess;
+				pDescIdxProps->quadDivergentImplicitLod								= false;
+				pDescIdxProps->maxPerStageDescriptorUpdateAfterBindSamplers			= _properties.limits.maxPerStageDescriptorSamplers;
+				pDescIdxProps->maxPerStageDescriptorUpdateAfterBindUniformBuffers	= _properties.limits.maxPerStageDescriptorUniformBuffers;
+				pDescIdxProps->maxPerStageDescriptorUpdateAfterBindStorageBuffers	= _properties.limits.maxPerStageDescriptorStorageBuffers;
+				pDescIdxProps->maxPerStageDescriptorUpdateAfterBindSampledImages	= _properties.limits.maxPerStageDescriptorSampledImages;
+				pDescIdxProps->maxPerStageDescriptorUpdateAfterBindStorageImages	= _properties.limits.maxPerStageDescriptorStorageImages;
+				pDescIdxProps->maxPerStageDescriptorUpdateAfterBindInputAttachments	= _properties.limits.maxPerStageDescriptorInputAttachments;
```
cdavis5e, line 357: Don't many of these have a minimum limit of 500k?

MoltenVK/MoltenVK/GPUObjects/MVKPipeline.mm
```diff
@@ -102,6 +101,7 @@
 										  models[i],
 										  kPushConstDescSet,
 										  kPushConstBinding,
+										  1,
```
cdavis5e, line 104: Wait, I don't get it. Isn't this where you're supposed to set the variable descriptor count so the array has the right size?

cdavis5e (changes requested):
Looks mostly good. Just a few concerns.

MoltenVK/MoltenVK/GPUObjects/MVKPipeline.mm
cdavis5e: No of course not, it's the call in MVKDescriptor.mm.

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
billhollings: In the spec, yes. And once I get Argument Buffers working, we should be there or close to there. But for this initial release, the limits are the same as the non-UAB quantities. 

It might also be that, with Argument Buffers, Metal allows both UAB and non-UAB values to use the same larger numbers.

MoltenVK/MoltenVK/GPUObjects/MVKDescriptorSet.mm
```diff
@@ -24,18 +24,23 @@
 #pragma mark -
 #pragma mark MVKDescriptorSetLayout
 
+// Dynamnic offsets passed in are in order of binding number, not binding index.
```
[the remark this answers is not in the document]
billhollings: Fixed.

cdavis5e (approved):

PR #1139, opened 2020-11-05
Fix tvOS builds and fix supportsGPUFamily: typo.

billhollings:
- Add macros to set `MTLPixelFormatASTC_*_HDR` to `MTLPixelFormatInvalid` on tvOS.
- Use `MTLRenderPipelineDescriptor::inputPrimitiveTopologyMVK` instead of native method.
- Replace use of `MTLDevice::supportsGPUFamily:` with `supportsFamily:`.
- Wrap unused variable of tvOS with conditional compile macro.


cdavis5e (approved):
Thanks for cleaning up the messes I made.

Issue #1129, opened 2020-10-30
Change the order of MVKDescriptorSetLayoutBindings in MVKDescriptorSetLayout?


billhollings:
In the Vulkan `vkCreateDescriptorSetLayout()` call, the `VkDescriptorSetLayoutBindings` in `VkDescriptorSetLayoutCreateInfo` do not need to be supplied in binding number order, and currently MoltenVK populates the corresponding `MVKDescriptorSetLayoutBindings` in the `MVKDescriptorSetLayout` in the same order as the bindings appear in `VkDescriptorSetLayoutCreateInfo`.

However, most Vulkan operations on a descriptor set layout's bindings (for example, specifying dynamic offsets, or declaring a variable-count binding) require processing bindings in order of binding number, not positional order. This requires some mapping logic within MoltenVK.

So, there appears to be some simplification and streamlining optimizations we can make in MoltenVK, by simply sorting and holding the `MVKDescriptorSetLayoutBindings` in `MVKDescriptorSetLayout` in binding number order.

In looking at the code, I can't see any obvious processing that depends on the `MVKDescriptorSetLayoutBindings` being ordered in the order provided in `VkDescriptorSetLayoutCreateInfo`...but I may be missing something.

Can anyone determine a reason **_not_** to sort and hold the `MVKDescriptorSetLayoutBindings` in `MVKDescriptorSetLayout` in binding number order when `MVKDescriptorSetLayout` is created?

@cdavis5e

billhollings:
This is implemented in #1141.

PR #1141, opened 2020-11-05
Descriptor code cleanups

billhollings:
This streamlines the class and runtime structure of descriptors, partly in prep for supporting Argument Buffers, and fixes a couple of bugs in descriptor code.

- Fix crash in `vkUpdateDescriptorSets()` when copying inline block descriptors.
- Use temporary buffer to hold inline block data between read and write stages while copying between descriptor sets.
- `MVKDescriptorSetLayout` sort and hold bindings by binding number.
- `MVKBufferDescriptor` increment dynamic offset index as consumed during binding.
- Remove need to track dynamic offset indexes in `MVKDescriptorSetLayoutBinding`.
- Simplify `MVKDescriptor` subclass implementations.
- Don't pass descriptor type to `MVKDescriptor` functions, which already know their descriptor type, and remove redundant switch tests in each subclass.
- Move sanity tests for correct descriptor type to callers.
- Update What's New and `MoltenVK_Runtime_UserGuide.md` documents.

Implements #1129.


MoltenVK/MoltenVK/GPUObjects/MVKPipeline.mm
```diff
@@ -48,12 +48,9 @@
...
 		MVKDescriptorSetLayout* dsl = _descriptorSetLayouts[dslIdx];
-		dynamicOffsetIndex += dsl->bindDescriptorSet(cmdEncoder, descSet,
-													 _dslMTLResourceIndexOffsets[dslIdx],
-													 dynamicOffsets, dynamicOffsetIndex);
-		if (!cmdEncoder && dsl->getConfigurationResult() != VK_SUCCESS) {
-			setConfigurationResult(dsl->getConfigurationResult());
-		}
+		dsl->bindDescriptorSet(cmdEncoder, descSet, _dslMTLResourceIndexOffsets[dslIdx],
+							   dynamicOffsets, dynamicOffsetIndex);
+		if (!cmdEncoder) { setConfigurationResult(dsl->getConfigurationResult()); }
```
cdavis5e, line 53: Wait, if one fails, but a later one doesn't, won't that cause the whole thing to report success? There's a reason I put that `dsl->getConfigurationResult() != VK_SUCCESS` check in.

cdavis5e (changes requested):
Looks good. Just one minor nit.

MoltenVK/MoltenVK/GPUObjects/MVKPipeline.mm
billhollings: Yeah...if figured the loop was why you added that test.

`setConfigurationResult()` is a latch function. It will only move from `VK_SUCCESS` to something else. The design idea is intentionally to catch and hold onto the first error that occurs (loop or not).

cdavis5e: Oh, OK. I did not know that.

cdavis5e (approved):

PR #1142, opened 2020-11-05
Device fallback and fixed typo

MarnixKuijs:
Added device fallback to make sure that projects that depend on MoltenVK keep working in certain situations like for CI. Also fixed a typo that prevented the project from building

### DIFF
--- a/MoltenVK/MoltenVK/Commands/MVKCmdTransfer.mm
+++ b/MoltenVK/MoltenVK/Commands/MVKCmdTransfer.mm
@@ -1346,7 +1346,7 @@ VkResult MVKCmdClearImage<N>::setContent(MVKCommandBuffer* cmdBuff,
 
         // Validate
         MVKMTLFmtCaps mtlFmtCaps = cmdBuff->getPixelFormats()->getCapabilities(_image->getMTLPixelFormat(planeIndex));
-		bool isDestUnwritableLinear = MVK_MACOS && !cmdBuff->getDevice()->_pMetalFeatures->renderLinearTextures && _dstImage->getIsLinear();
+		bool isDestUnwritableLinear = MVK_MACOS && !cmdBuff->getDevice()->_pMetalFeatures->renderLinearTextures && _image->getIsLinear();
 		uint32_t reqCap = isDS ? kMVKMTLFmtCapsDSAtt : (isDestUnwritableLinear ? kMVKMTLFmtCapsWrite : kMVKMTLFmtCapsColorAtt);
         if (!mvkAreAllFlagsEnabled(mtlFmtCaps, reqCap)) {
             return cmdBuff->reportError(VK_ERROR_FEATURE_NOT_PRESENT, "vkCmdClear%sImage(): Format %s cannot be cleared on this device.", (isDS ? "DepthStencil" : "Color"), cmdBuff->getPixelFormats()->getName(_image->getVkFormat()));
--- a/MoltenVK/MoltenVK/GPUObjects/MVKPixelFormats.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKPixelFormats.mm
@@ -1425,7 +1425,7 @@ void MVKPixelFormats::modifyMTLFormatCapabilities() {
 #endif
 #if MVK_MACOS
 		NSArray<id<MTLDevice>>* mtlDevices = MTLCopyAllDevices();	// temp retained
-		id<MTLDevice> mtlDevice = [mtlDevices[0] retain];			// temp retained
+		id<MTLDevice> mtlDevice = [mtlDevices count] > 0 ? [mtlDevices[0] retain] : MTLCreateSystemDefaultDevice();			// temp retained
 		[mtlDevices release];										// temp release
 #endif
 		modifyMTLFormatCapabilities(mtlDevice);


MoltenVK/MoltenVK/Commands/MVKCmdTransfer.mm
```diff
@@ -1346,7 +1346,7 @@ static inline MTLSize mvkClampMTLSize(MTLSize size, MTLOrigin origin, MTLSize ma
 
         // Validate
         MVKMTLFmtCaps mtlFmtCaps = cmdBuff->getPixelFormats()->getCapabilities(_image->getMTLPixelFormat(planeIndex));
-		bool isDestUnwritableLinear = MVK_MACOS && !cmdBuff->getDevice()->_pMetalFeatures->renderLinearTextures && _dstImage->getIsLinear();
+		bool isDestUnwritableLinear = MVK_MACOS && !cmdBuff->getDevice()->_pMetalFeatures->renderLinearTextures && _image->getIsLinear();
```
cdavis5e, line 1349: Ugh, I swear I fixed this. It must have been in another change of mine.

Travis would've caught this. I wish it were back up.

MoltenVK/MoltenVK/GPUObjects/MVKPixelFormats.mm
```diff
@@ -1425,7 +1425,7 @@
 #endif
 #if MVK_MACOS
 		NSArray<id<MTLDevice>>* mtlDevices = MTLCopyAllDevices();	// temp retained
-		id<MTLDevice> mtlDevice = [mtlDevices[0] retain];			// temp retained
+		id<MTLDevice> mtlDevice = [mtlDevices count] > 0 ? [mtlDevices[0] retain] : MTLCreateSystemDefaultDevice();			// temp retained
```
cdavis5e, line 1428: This doesn't make sense. `MTLCopyAllDevices()` is supposed to copy all known `MTLDevice`s in the system. Have you actually encountered a situation where, on macOS, this function can return `nil` or an empty array?

cdavis5e (changes requested):

MoltenVK/MoltenVK/GPUObjects/MVKPixelFormats.mm
MarnixKuijs: On one of our projects we had an issue [upgrading to Molten-VK 1.1.0](https://github.com/EmbarkStudios/ash-molten/pull/25) and our CI kept failing because of an uncaught exception of indexing into an array of size 0. This change has seemed to fixed that issue.

billhollings: >This doesn't make sense.

I agree. 

This is arising in the `MVKPixelFormats` that is statically declared in `mvk_datatypes.mm`. This is not actually the primary invocation of `MTLCopyAllDevices()` for creating and supporting `VkPhysicalDevices`, which occurs in `MVKInstance`. There, Vulkan would report no physical devices available, making it useless for an actual running Vulkan app.

Hmmm...assuming `MTLCopyAllDevices()` and `MTLCreateSystemDefaultDevice()` are consistent with one another, presumably `MTLCreateSystemDefaultDevice()` is returning `nil`, and all the method calls to the `nil` `MTLDevice` in `MVKPixelFormats` to evaluate device capabilities simply fail silently. But at least they don't error out on an array bound failure.

@MarnixKuijs What type of platform is your CI running on? And does your CI just build and start up the app without actually running Vulkan code that would detect no `VkPhysicalDevices` if the invocation of `MTLCopyAllDevices()` in `MVKInstance` also returned an empty array?

In any case...the code fix here doesn't look to hurt anything. And looking at it now...perhaps the thing to do for this routine is simply to use `MTLCreateSystemDefaultDevice()` on all devices, since the purpose inside the static instance of `MVKPixelFormats` is to simply grab any device, and the default device would make sense for that anyway.

I'll pull this PR in, and then make a note to review and make that change later.

billhollings: @cdavis5e If that makes sense, are you good to approve your change request? Or was there something else?

MarnixKuijs: > @MarnixKuijs What type of platform is your CI running on? And does your CI just build and start up the app without actually running Vulkan code that would detect no VkPhysicalDevices if the invocation of MTLCopyAllDevices() in MVKInstance also returned an empty array?

We are running on the latest version of MacOS simply through Github actions. Because of the crash we don't even get to the part that would call the primary MTLCopyAllDevices when creating a VKPhysicalDevice. I don't think I can't prevent the initialization of the static variable to avoid this crash.

I agree that it doesn't really make sense why one works and the other one doesn't like you said: "_presumably MTLCreateSystemDefaultDevice() is returning nil, and all the method calls to the nil MTLDevice in MVKPixelFormats to evaluate device capabilities simply fail silently. But at least they don't error out on an array bound failure._" is probably the reason why.

cdavis5e: The reason I did it that way was to stop us from always triggering a mux switch on startup. I guess this is OK for now.

cdavis5e (approved):

billhollings:
>triggering a mux switch on startup

Ah! Damn! Thanks for the reminder that `MTLCreateSystemDefaultDevice()` forces a GPU switch.

With that in mind I won't make the change that I suggested above, and we'll leave the implementation as it appears in this PR.

PR #1144, opened 2020-11-09
MVKCommandEncoder: Don't set renderTargetArrayLength for mixed 2D/3D renders.

cdavis5e:
Metal's validation layer doesn't like it when we render to both a 2D and
a 3D texture with layered rendering. Work around this by not setting the
`renderTargetArrayLength` if it would be one and we are rendering to
mixed 2D and 3D textures.

If the pipeline sets `[[render_target_array_index]]`, we're hosed either
way. Perhaps this is why Vulkan has a special flag,
`VK_IMAGE_CREATE_2D_ARRAY_COMPATIBLE_BIT`, intended for rendering to 3D
textures.


billhollings (approved):

billhollings:
Whoa! Sorry for the delay in bringing this in! I don't know how I lost track of it!

PR #1146, opened 2020-11-11
Disable culling for the duration of vkCmdClearAttachments

TheSpydog:
This fixes a bug where the quad drawn by `vkCmdClearAttachments` can be culled if the winding order of the quad happens to match the current pipeline's cull mode. Note that the original cull mode will be restored immediately after the quad is drawn via the existing `cmdEncoder->_graphicsPipelineState.markDirty();` call, so no pushing and popping of state should be necessary.

### DIFF
--- a/MoltenVK/MoltenVK/Commands/MVKCmdTransfer.mm
+++ b/MoltenVK/MoltenVK/Commands/MVKCmdTransfer.mm
@@ -1300,6 +1300,10 @@ void MVKCmdClearAttachments<N>::encode(MVKCommandEncoder* cmdEncoder) {
     [mtlRendEnc setDepthStencilState: cmdEncPool->getMTLDepthStencilState(isClearingDepth, isClearingStencil)];
     [mtlRendEnc setStencilReferenceValue: _mtlStencilValue];
     [mtlRendEnc setCullMode:MTLCullModeNone];
+    [mtlRendEnc setTriangleFillMode:MTLTriangleFillModeFill];
+    [mtlRendEnc setDepthBias:0 slopeScale:0 clamp:0];
+    cmdEncoder->_viewportState.reset();
+    cmdEncoder->_scissorState.reset();
 
     cmdEncoder->setVertexBytes(mtlRendEnc, clearColors, sizeof(clearColors), 0);
     cmdEncoder->setFragmentBytes(mtlRendEnc, clearColors, sizeof(clearColors), 0);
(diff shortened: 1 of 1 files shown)


cdavis5e (changes requested):
@billhollings doesn't seem to be around, and this change looks simple enough. Good catch.

But, there are other states which could interfere with this. I'm thinking in particular of the fill mode, which could be set to `Lines`; or the depth bias possibly interfering with a depth clear.  We should probably reset those two, too. The viewport and scissor rect... hmm, the spec doesn't specify anything. But since this command doesn't depend on a pipeline being set, we might want to reset those, too. (You'll have to add a `markDirty()` call for those states as well.) I think the blend color can be left alone, since we have our own pipeline state which disables blending.

And maybe I should go and lobby Khronos to put tests for all this into the Vulkan CTS.

TheSpydog:
> But, there are other states which could interfere with this.

Good point, thanks! I've made the changes you requested. I'm not 100% sure I've set/reset the viewport and scissor states correctly, so please let me know if I need to change anything there.

MoltenVK/MoltenVK/Commands/MVKCmdTransfer.mm
```diff
@@ -1299,6 +1299,11 @@ static inline MTLSize mvkClampMTLSize(MTLSize size, MTLOrigin origin, MTLSize ma
     [mtlRendEnc setRenderPipelineState: cmdEncPool->getCmdClearMTLRenderPipelineState(_rpsKey)];
     [mtlRendEnc setDepthStencilState: cmdEncPool->getMTLDepthStencilState(isClearingDepth, isClearingStencil)];
     [mtlRendEnc setStencilReferenceValue: _mtlStencilValue];
+    [mtlRendEnc setCullMode:MTLCullModeNone];
+    [mtlRendEnc setTriangleFillMode:MTLTriangleFillModeFill];
+    [mtlRendEnc setDepthBias:0 slopeScale:0 clamp:0];
+    cmdEncoder->_viewportState.reset();
+    cmdEncoder->_scissorState.reset();
```
cdavis5e, line 1306: This will cause MoltenVK to forget whatever viewports and scissors were set from the Vulkan API, if they were set beforehand. It won't even affect the current render encoder, since MoltenVK doesn't apply states until a draw call. You want to set the viewport and scissor state on the `MTLRenderCommandEncoder` instead, using `-[MTLRenderCommandEncoder setViewport:]` and `-[MTLRenderCommand setScissorRect:]`; then the `markDirty()` call will cause MoltenVK to restore the settings from the pipeline (or `vkCmdSetViewports()`/`vkCmdSetScissorRects()`) the next time a draw call is made.

cdavis5e (changes requested):
Almost...

cdavis5e (approved):
Looks good.

PR #1149, opened 2020-11-15
MVKPhysicalDevice: Enable 3D compressed textures on iOS/tvOS.

cdavis5e:
Forbid ETC2 and EAC 3D textures on all platforms. Apple GPUs do not
support 3D for those.

### DIFF
--- a/MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
@@ -1017,21 +1017,6 @@ void MVKImage::validateConfig(const VkImageCreateInfo* pCreateInfo, bool isAttac
 	bool isCompressed = pixFmts->getFormatType(pCreateInfo->format) == kMVKFormatCompressed;
 	bool isChromaSubsampled = pixFmts->getChromaSubsamplingPlaneCount(pCreateInfo->format) > 0;
 
-#if MVK_IOS_OR_TVOS
-	if (isCompressed && !is2D) {
-		setConfigurationResult(reportError(VK_ERROR_FEATURE_NOT_PRESENT, "vkCreateImage() : Under Metal, compressed formats may only be used with 2D images."));
-	}
-#endif
-#if MVK_MACOS
-	if (isCompressed && !is2D) {
-		if (getImageType() != VK_IMAGE_TYPE_3D) {
-			setConfigurationResult(reportError(VK_ERROR_FEATURE_NOT_PRESENT, "vkCreateImage() : Under Metal, compressed formats may only be used with 2D or 3D images."));
-		} else if (!_device->_pMetalFeatures->native3DCompressedTextures && !mvkCanDecodeFormat(pCreateInfo->format)) {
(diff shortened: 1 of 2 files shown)


MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
```diff
@@ -1017,21 +1017,6 @@
...
-	if (isCompressed && !is2D) {
-		if (getImageType() != VK_IMAGE_TYPE_3D) {
-			setConfigurationResult(reportError(VK_ERROR_FEATURE_NOT_PRESENT, "vkCreateImage() : Under Metal, compressed formats may only be used with 2D or 3D images."));
-		} else if (!_device->_pMetalFeatures->native3DCompressedTextures && !mvkCanDecodeFormat(pCreateInfo->format)) {
-			setConfigurationResult(reportError(VK_ERROR_FEATURE_NOT_PRESENT, "vkCreateImage() : Under Metal, the %s compressed format may only be used with 2D images.", getPixelFormats()->getName(pCreateInfo->format)));
-		}
-	}
-#endif
-
 	if (isChromaSubsampled && !is2D) {
```
billhollings, line 1020: I'm not sure I understand why these validations should be removed. Shouldn't it at least include a test for `native3DCompressedTextures`...and maybe also `mvkTreatTexture1DAs2D()` come to think of it?

cdavis5e: I thought we agreed to remove parameter validation for invalid usage.

billhollings: >I thought we agreed to remove parameter validation for invalid usage.

Yes indeed.

Is that the case here? That's my question. It looks like those validations and errors would still be useful if `native3DCompressedTextures` or `mvkTreatTexture1DAs2D()` are `false`? 

Perhaps I'm missing something. Is it that we wouldn't get here because of `vkGetPhysicalDeviceImageFormatProperties()`? If so...then removing these makes sense.

cdavis5e: > > I thought we agreed to remove parameter validation for invalid usage.
> 
> Yes indeed.
> 
> Is that the case here? That's my question. It looks like those validations and errors would still be useful if `native3DCompressedTextures` or `mvkTreatTexture1DAs2D()` are `false`?
> 
> Perhaps I'm missing something. Is it that we wouldn't get here because of `vkGetPhysicalDeviceImageFormatProperties()`? If so...then removing these makes sense.

I would assume so, since if `vkGetPhysicalDeviceImageFormatProperties()` returns `VK_ERROR_FORMAT_NOT_SUPPORTED` for a given combination of parameters, that means that using those parameters to create an image isn't supported.

billhollings: Okay. Sounds good. Thanks for cleaning up the validations. 

And sorry for the delay as I mentally absorbed the redundancy. 😉

Issue #803, opened 2019-12-19
vkAcquireNextImageKHR do not properly signal the VkSemaphore


apayen:
Hello,

I was checking whether our engine had sane use of VkSemaphore. so I modified the MTLEvent implementation to have separates signal and wait counters, and assert when they are out of sync (they should be equal before a signal; and have signal == wait + 1 before a wait).

A first (minor) issue arose from here:

https://github.com/KhronosGroup/MoltenVK/blob/9e09ece372b2fa383aaf60316483e5cd9243b12f/MoltenVK/MoltenVK/GPUObjects/MVKSync.mm#L111-L114

Even if the command buffer is nil, the counter is being incremented.

I then ending up breaking the assert when inserting the wait after having acquired the image from the swapchain:

Our swapchain pattern is as follow (with two images):
```
Acquire, signals S0 -> returns index 0
wait on S0
frame commands
signal S0
Present image 0, wait S0
Acquire, signals S1 -> returns index 1
wait on S1
frame commands
signal S1
Present image 1, wait S1
... repeat ...
```

Acquire's behavior is:
- if the image is readily available, a command buffer is inserted in the first queue, and signals the semaphore
- If the image is not available, the semaphore is queued as a signaler for the next to be available image, and will be signaled when presenting that image.

The first case updates the counter (although the signal operation may happen much later than it should be)
In the second case, the counter is not updated and they get out of sync. The signal call will happen on the next present of this buffer. This means that, it will signal that value after the wait has been issued, and deadlock the GPU.

I notice the break would happen at the same places we sometimes had GPU crashes.

By refactoring our code to call Acquire before present, we got rid of the issue:
```
Init time:
Acquire, signals SA0 -> returns index 0
Acquire, signals SA1 -> returns index 1

Frame:
wait on SA0
frame commands
signal SP0
Acquire, signals SA0 -> returns index 0
Present image 0, wait SP0
wait on SA1
frame commands
signal SP1
Acquire, signals SA1 -> returns index 1
Present image 1, wait SP1
... repeat ...
```

Idealy, the event should be signaled by the CPU (using a MTLSharedEvent, it seems to have the same SDK requirements as MTLEvent), when the image is readily available.
When it is not, the counter would be backed up, without actually signaling (say, a DelayedSignal function). The semaphore can then be queued as is currently the case, along with the signal value. The Present already has a completion handler, where the delayed signal would actually happen (from the CPU).

[...]

PR #1165, opened 2020-11-28
MVKSync: Miscellaneous fixes.

cdavis5e:
Fix reversed condition in `MVKTimelineSemaphoreEmulated`, which probably
caused some test failures.

Don't increment the `MTLEvent` binary semaphore's counter unless a
command buffer is actually present to schedule a wait on.

Defer signal operations when a swapchain image is not yet available.
That way, the correct value will be signaled when the image is ready,
instead of causing GPU lockups and timeouts.

When a drawable is presented, immediately mark it available, instead of
waiting until the command buffer finishes. Otherwise, the wrong
semaphore could be signaled when an image is used twice in a row.

This should fix the problems using `MTLEvent` binary semaphores with
presentation, which was preventing us from enabling them by default when
available.

Special thanks to @apayen, whose
[idea](https://github.com/KhronosGroup/MoltenVK/issues/803) and
[change](https://github.com/apayen/MoltenVK/commit/a4ac715975ea3aa4cb95e91973b8f9010516304f)
were the basis for this.

### DIFF
--- a/MoltenVK/MoltenVK/GPUObjects/MVKImage.h
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKImage.h
@@ -438,7 +438,11 @@ typedef struct  {
 } MVKPresentTimingInfo;
 
 /** Tracks a semaphore and fence for later signaling. */
-typedef std::pair<MVKSemaphore*, MVKFence*> MVKSwapchainSignaler;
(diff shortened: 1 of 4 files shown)


billhollings (approved):
Good catches!

PR #1167, opened 2020-11-30
Support the VK_EXT_subgroup_size_control extension.

cdavis5e:
This extension allows the subgroup size to vary between draw/dispatch
calls, and even allows clients to declare that full subgroups must
always be dispatched. It corresponds better to how Metal actually works.

No support for declaring a required subgroup size, unfortunately.


billhollings (approved):

MoltenVK/MoltenVK/Layers/MVKExtensions.def
```diff
@@ -95,6 +95,7 @@ MVK_EXTENSION(EXT_robustness2, EXT_ROBUSTNESS_2, DEVICE)
 MVK_EXTENSION(EXT_scalar_block_layout, EXT_SCALAR_BLOCK_LAYOUT, DEVICE)
 MVK_EXTENSION(EXT_shader_stencil_export, EXT_SHADER_STENCIL_EXPORT, DEVICE)
 MVK_EXTENSION(EXT_shader_viewport_index_layer, EXT_SHADER_VIEWPORT_INDEX_LAYER, DEVICE)
+MVK_EXTENSION(EXT_subgroup_size_control, EXT_SUBGROUP_SIZE_CONTROL, DEVICE)
```
billhollings, line 98: Add to `MoltenVK_Runtime_UserGuide.md` and `Whats_New.md` as well?

billhollings (changes requested):

billhollings (approved):

billhollings:
Meh. Looks like Travis decided to return to action. And looks like we'll have to update to using Xcode 12.2. I'll take care of that at some point soon.

PR #1168, opened 2020-12-02
Support Mac Catalyst on macOS 11.0+

billhollings:
Define `MVK_MACCAT` build macro and use it to conditionally compile code to align 
with build features and capabilities of Mac Catalyst platform on macOS 11.0+.
Treat Mac Catalyst as minor variation of macOS 11.0.
Update documentation.

Currently only support Mac Catalyst on macOS 11.0+, to avoid complexities of 
deselecting iOS features and capabilities for Mac Catalyst on previous macOS versions.

Mac Catalyst (and Simulators) require use of XCFrameworks. 
Currently unable to generate a dylib for Mac Catalyst.

Update build scripts to build for Mac Catalyst.

### DIFF
--- a/MoltenVK/MoltenVK/GPUObjects/MVKPixelFormats.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKPixelFormats.mm
@@ -1434,8 +1434,22 @@ void MVKPixelFormats::modifyMTLFormatCapabilities() {
 }
 
 
+// Mac Catalyst does not support feature sets, so we redefine them to GPU families in MVKDevice.h.
+#if MVK_MACCAT
+#define addFeatSetMTLPixFmtCaps(FEAT_SET, MTL_FMT, CAPS)  \
+	addMTLPixelFormatCapabilities(mtlDevice, MTLFeatureSet_ ##FEAT_SET, 10.16, MTLPixelFormat ##MTL_FMT, kMVKMTLFmtCaps ##CAPS)
+
+#define addFeatSetMTLVtxFmtCaps(FEAT_SET, MTL_FMT, CAPS)  \
+	addMTLVertexFormatCapabilities(mtlDevice, MTLFeatureSet_ ##FEAT_SET, 10.16, MTLVertexFormat ##MTL_FMT, kMVKMTLFmtCaps ##CAPS)
+
+#else
 #define addFeatSetMTLPixFmtCaps(FEAT_SET, MTL_FMT, CAPS)  \
 	addMTLPixelFormatCapabilities(mtlDevice, MTLFeatureSet_ ##FEAT_SET, MTLPixelFormat ##MTL_FMT, kMVKMTLFmtCaps ##CAPS)
+
+#define addFeatSetMTLVtxFmtCaps(FEAT_SET, MTL_FMT, CAPS)  \
+	addMTLVertexFormatCapabilities(mtlDevice, MTLFeatureSet_ ##FEAT_SET, MTLVertexFormat ##MTL_FMT, kMVKMTLFmtCaps ##CAPS)
+
+#endif
 
 #define addGPUOSMTLPixFmtCaps(GPU_FAM, OS_VER, MTL_FMT, CAPS)  \
 	addMTLPixelFormatCapabilities(mtlDevice, MTLGPUFamily ##GPU_FAM, OS_VER, MTLPixelFormat ##MTL_FMT, kMVKMTLFmtCaps ##CAPS)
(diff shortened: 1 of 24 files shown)


billhollings:
@cdavis5e 

Please review.

With the intro of Big Sur and Apple Silicon, I felt it would be helpful to provide this functionality in the SDK release next week. However, I am a little wary about pushing this so close to the deadline, in particular, the possible ramifications of the redefinition of `MVK_MACOS` and `MVK_IOS` (and a couple of other related platform test changes) on native platform builds.

It looks safe to me at this point, but I'd definitely appreciate a second set of eyes on it. And if we're not comfortable and feel it needs more testing before releasing to the SDK we could hold off pulling this in until after that.

I do not have an Apple Silicon device to test that part of this. I don't expect any issues, but if you have that capability, I'd appreciate you running a smoke test or two.

Finally, as mentioned above, adding this support was made much easier by limiting it to macOS 11+, mostly to make use of all the iOS pixel formats and other compatibilities that were introduced in macOS as a result. But let me know if you are more optimistic than me that we could extend this to macOS 10.15 too in a straightforward manner, and that it would be useful to do so.

MoltenVK/MoltenVK/GPUObjects/MVKPixelFormats.mm
```diff
@@ -2039,7 +2050,8 @@
 		// XXX We really want to use the device's Metal features instead of duplicating the
 		// logic from MVKPhysicalDevice, but those may not have been initialized yet.
 #if MVK_MACOS
-		 (isStencilFormat(vkDesc.mtlPixelFormat) && (!_physicalDevice || ![mtlDev supportsFeatureSet: MTLFeatureSet_macOS_GPUFamily2_v1]))
+		(isStencilFormat(vkDesc.mtlPixelFormat) && (!_physicalDevice || !([mtlDev supportsFamily: MTLGPUFamilyMac2] ||
```
cdavis5e, line 2053: This won't work on Mojave. It doesn't yet have `-[MTLDevice supportsFamily:]`.

MoltenVK/MoltenVK/GPUObjects/MVKRenderPass.mm
```diff
@@ -327,15 +327,11 @@ static uint32_t getNextViewMaskGroup(uint32_t viewMask, uint32_t* startView, uin
 			mtlTexDesc.textureType = MTLTextureType2DMultisample;
 			mtlTexDesc.sampleCount = sampleCount;
 		}
-#if MVK_IOS
-		if ([_renderPass->getMTLDevice() supportsFeatureSet: MTLFeatureSet_iOS_GPUFamily1_v3]) {
+		if (mvkOSVersionIsAtLeast(10.0, 11.0, 14.0)) {
```
cdavis5e, line 330: This will break on macOS 11 with a non-Apple GPU.

cdavis5e (changes requested):
A couple of problems with this; looks good otherwise.

MoltenVK/MoltenVK/GPUObjects/MVKRenderPass.mm
cdavis5e: Then again, we shouldn't even hit that codepath on macOS 11, since it supports render passes with no render targets. It's why I didn't bother updating that for Apple Silicon.

MoltenVK/MoltenVK/GPUObjects/MVKPixelFormats.mm
billhollings: Good catch. I've refactored this section. Please review.

MoltenVK/MoltenVK/GPUObjects/MVKRenderPass.mm
billhollings: Another good catch. 

Actually in review, I realized that `mvkOSVersionIsAtLeast(macOS, iOS, macCat)` won't actually work, so I've reverted it. 

It turns out that `mvkOSVersion()` reports the same value for macOS and MacCat (eg. 11.0001), and reports the same value for iOS and iOS simulator (eg. 14.2), so there is no distinct MacCat OS version to act on.

billhollings: Also. The original code was just checking for iOS anyway, and is not affected by Mac Cat, so I've reverted it to the original.

I think I made this original change early on when I was originally thinking of MacCat as a "variation of iOS", instead of the current "variation of macOS", and missed backing it out. 🤦🏻‍♂️

cdavis5e (approved):

MoltenVK/MoltenVK/GPUObjects/MVKPixelFormats.mm
cdavis5e: Looks a lot cleaner. Thanks.

PR #1169, opened 2020-12-02
Handle device loss.

cdavis5e:
If a command buffer fails in Metal, its `status` becomes
`MTLCommandBufferStatusError`, and its `error` property is populated. In
Vulkan, command buffer failure usually triggers device loss. This is
because most drivers can't guarantee that resources weren't affected in
an undefined fashion. We can't make that guarantee, either, so when a
Metal command buffer has an error, mark the device lost. The error is
also logged. All waits are immediately signaled; those that were client
requests instead of internal implementation details immediately return
`VK_ERROR_DEVICE_LOST`.

The app may be able to recreate the logical device and reconstruct its
state. However, some Metal errors are severe enough that all subsequent
command buffers will fail--in those cases, the physical device becomes
lost as well, indicating that the device cannot be recreated.

Only certain commands are allowed to report device loss. These commands
test for device loss before continuing, and will immediately return if
the device is lost.

### DIFF
--- a/MoltenVK/MoltenVK/GPUObjects/MVKQueue.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKQueue.mm
@@ -76,6 +76,8 @@ static inline void execute(MVKQueueSubmission* qSubmit) { @autoreleasepool { qSu
 // Relying on the dispatch queue to find time to drain the autoreleasepool can
 // result in significant memory creep under heavy workloads.
 VkResult MVKQueue::submit(MVKQueueSubmission* qSubmit) {
+	if (_device->getConfigurationResult() != VK_SUCCESS) { return _device->getConfigurationResult(); }
+
 	if ( !qSubmit ) { return VK_SUCCESS; }     // Ignore nils
 
 	VkResult rslt = qSubmit->getConfigurationResult();     // Extract result before submission to avoid race condition with early destruction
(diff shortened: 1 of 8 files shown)


MoltenVK/MoltenVK/GPUObjects/MVKQueue.mm
```diff
@@ -292,8 +296,27 @@
...
+			device->reportError(device->markLost(), "Command buffer %p \"%s\" execution failed (code %li): %s", mtlCmdBuff, mtlCmdBuff.label ? mtlCmdBuff.label.UTF8String : "", mtlCmdBuff.error.code, mtlCmdBuff.error.localizedDescription.UTF8String);
+			// Some errors indicate we lost the physical device as well.
+			switch (mtlCmdBuff.error.code) {
+				case MTLCommandBufferErrorBlacklisted:
+				// XXX This may also be used for command buffers executed in the background without the right entitlement.
+				case MTLCommandBufferErrorNotPermitted:
+#if MVK_MACOS
+				case MTLCommandBufferErrorDeviceRemoved:
+#endif
+					device->getPhysicalDevice()->setConfigurationResult(VK_ERROR_DEVICE_LOST);
```
billhollings, line 314: Setting configuration of `MVKPhysicalDevice` here. Most tests above seem to be testing against `MVKDevice`. The only place I can see where the config result of `MVKDevice` is set is in the constructor. Should we not be setting that value dynamically too? Perhaps I'm not seeing it...or missing something in the logic?

cdavis5e: `MVKDevice::markLost()` sets the configuration result to `VK_ERROR_DEVICE_LOST`.

billhollings: D'oh. Right. I don't know how I missed that.

billhollings (approved):

PR #1170, opened 2020-12-02
Re-enable MTLEvent-based binary semaphores.

cdavis5e:
I haven't seen any problems yet. It passes all the tests--even the
`signal_order` tests. I have tried it with some games, and it seems OK
there, too.


billhollings (changes requested):
Please update `Whats_New.md`, and the appropriate documentation in `vk_mvk_moltenvk.h`:

>By default, MVK_ALLOW_METAL_FENCES is enabled and MVK_ALLOW_METAL_EVENTS is disabled, meaning MoltenVK will use MTLFences, if they are available, to control VkSemaphore synchronization behaviour, by default.

billhollings (approved):

PR #1171, opened 2020-12-03
MVKSwapchain: Allow images whose size doesn't match the CAMetalLayer.

cdavis5e:
The game NieR: Automata (via DXVK) attempts to create a 1680x1050
swapchain on a 1600x900 window. It then attempts to render to this
swapchain with a 1680x1050 framebuffer. But we created the textures at
1600x900, matching the window. The render area is thus too big, which
triggers a Metal validation failure. Apparently, DXVK doesn't check the
surface caps before creating the Vulkan swapchain.

Rather than expecting the swapchain to be the same size as the layer, we
can actually support any swapchain size, up to the maximum size of a
texture supported by the device. The system will just scale the texture
when rendering it if it doesn't match the layer size. If the sizes don't
match up, we return `VK_SUBOPTIMAL_KHR`, instead of
`VK_ERROR_OUT_OF_DATE_KHR`, indicating that presentation is still
possible, but performance may suffer. This is good enough to let the
game continue under the validation layers.

This really needs a corresponding change to the CTS, because it
currently assumes that we can't do this.

### DIFF
--- a/MoltenVK/MoltenVK/GPUObjects/MVKSwapchain.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKSwapchain.mm
@@ -332,7 +332,7 @@ void MVKSwapchain::initCAMetalLayer(const VkSwapchainCreateInfoKHR* pCreateInfo,
 			// Nothing - the default is not to do color matching.
 			break;
 	}
-	_mtlLayerOrigDrawSize = _mtlLayer.updatedDrawableSizeMVK;
+	_mtlLayer.drawableSize = mvkCGSizeFromVkExtent2D(pCreateInfo->imageExtent);
 
 	// TODO: set additional CAMetalLayer properties before extracting drawables:
 	//	- presentsWithTransaction
@@ -370,7 +370,7 @@ void MVKSwapchain::initSurfaceImages(const VkSwapchainCreateInfoKHR* pCreateInfo
 		}
 	}
 
-    VkExtent2D imgExtent = mvkVkExtent2DFromCGSize(_mtlLayerOrigDrawSize);
+    VkExtent2D imgExtent = pCreateInfo->imageExtent;
 
     VkImageCreateInfo imgInfo = {
         .sType = VK_STRUCTURE_TYPE_IMAGE_CREATE_INFO,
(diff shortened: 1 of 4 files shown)


MoltenVK/MoltenVK/GPUObjects/MVKSwapchain.mm
```diff
@@ -332,7 +332,7 @@ static inline CIE1931XY VkXYColorEXTToCIE1931XY(VkXYColorEXT xy) {
 			// Nothing - the default is not to do color matching.
 			break;
 	}
-	_mtlLayerOrigDrawSize = _mtlLayer.updatedDrawableSizeMVK;
+	_mtlLayer.drawableSize = mvkCGSizeFromVkExtent2D(pCreateInfo->imageExtent);
```
billhollings, line 335: Interesting. Does this force an internal adjustment to the `contentsScale` value of the `CAMetalLayer`, and the system compositor just handles the scaling accordingly?

billhollings (commented):
Just a confirmation question above. I want to make sure we're working with content scaling properly (maybe we weren't in the past...or at least we were treating this with more complexity than we needed to).

MoltenVK/MoltenVK/GPUObjects/MVKSwapchain.mm
cdavis5e: No and yes. According to the docs:

> By default, a layer creates textures sized to match its content—that is, this property’s value is the layer’s `bounds` size multiplied by its `contentsScale` factor.

This property just affects the dimensions of the `MTLTexture`s produced. The system then handles scaling to the layer's content area. That's my understanding.

billhollings: Okay. Makes sense.

Just one more thing. Have you tested this on a Retina display? Part of the use of `contentsScale` handling was to handle Retina. But again...we may not have been interpreting its use correctly in the past.

cdavis5e: My laptop has a Retina display.

I think you might be confused. `bounds` and `contentsScale` just affect the coordinate system of the layer itself. Unless `drawableSize` is unset explicitly, they have no bearing on the size of the drawables that the layer produces.

This default size is returned as the current surface size, so this should work OK for Retina.

billhollings: Yeah. The original behaviour might have conflated the scaling of the layer coordinate system with that of the drawables.

Thanks for clearing this up.

PR #1173, opened 2020-12-04
Let SPIRV-Cross assign MSL resource indices.

js6i:
There's an issue with assigning Metal resource indices upfront that this PR attempts to address.
Suppose that a client creates a pipeline with more resources in the layout than is allowed for a shader to use, say, more than 16 samplers. In the current scheme, even if no single stage uses too many samplers by itself, they will get numbered globally for the entire pipeline and crash when an index is > 15.
The proposed solution is to let SPIRV-Cross allocate the resource indices on demand. There's already a mechanism to do that, although I made some changes to make use of it more easily.
Apart from fixing that problem, there are some other upsides to it: we can stop binding unused resources and maybe handle implicit buffers more nicely.
Currently this PR is in a work in progress, not quite working yet state. Although it does seem to work with some games I was testing. @billhollings please let me know if that's a direction you like, or should the problem be addressed in some other way?

### DIFF
--- a/MoltenVK/MoltenVK/Commands/MVKCmdDraw.mm
+++ b/MoltenVK/MoltenVK/Commands/MVKCmdDraw.mm
@@ -44,6 +44,7 @@ VkResult MVKCmdBindVertexBuffers<N>::setContent(MVKCommandBuffer* cmdBuff,
         b.index = mvkDvc->getMetalBufferIndexForVertexAttributeBinding(startBinding + bindIdx);
         b.mtlBuffer = mvkBuffer->getMTLBuffer();
         b.offset = mvkBuffer->getMTLBufferOffset() + pOffsets[bindIdx];
+        b.isFixed = true;
         _bindings.push_back(b);
     }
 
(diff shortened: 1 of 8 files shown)


billhollings:
@js6i Thanks for the ideas.

I and @cdavis5e can review this in more detail after the SDK is out next week. Ideas like this might also help when native MSL is submitted too.

I'm also currently working on argument buffers, which also won't make it into this SDK release, unfortunately. That will also impact how descriptor sets are handled.

oscarbg:
@js6i can share with what games are you testing this patch? just curious..

js6i:
Not many so far, I tried Hitman 2, Risk of Rain 2 and Wolfenstein II. Of which the latter has this problem.

oscarbg:
@js6i any update on this patch? seems like it not will ready for next week "freeze" for next Vulkan SDK release, right?

js6i:
No, it on pause until the argument buffer situation is clear.

oscarbg:
> No, it on pause until the argument buffer situation is clear.

now that support is in VK SDK 1.2.176 (at least enabled on Big Sur, and Catalina Intel GPU only), can share some update?
thanks..

js6i:
There's no updates yet. That said, as Intel GPU's are not that relevant for graphically advanced games, if support on other GPU's is blocked by something longer term, there's still motivation for these changes. Otherwise I'm not sure there is as, if I remember correctly, the resource limits are significantly raised when argument buffers are used.

Issue #1127, opened 2020-10-29
Question: interest on adding artifacts support to exisiting CI?


oscarbg:
Hi,
don't know if requested before by me (bad memory :-)) or others, but can't seem to find a similar open or closed issue so just asking.. MoltenVK supports Travis since a while ago (https://github.com/KhronosGroup/MoltenVK/pull/63) so asking if it's possible for MoltenVK to add artifacts support to Travis CI? don't know much about Travis CI and if it's free to support artifacts for open source projects like these.. or if other CI systems can support this (AppVeyor for ex.)..
for example wanted the artifacts of recent build for quick testing:
https://travis-ci.com/github/KhronosGroup/MoltenVK/builds/194818878
thanks..

billhollings:
>wanted the artifacts of recent build for quick testing

What artifacts are you looking for...and why?

oscarbg:
a quick update:
*just confirm https://github.com/Gcenx/macOS_Wine_builds include libMoltenVK.dylib for example latest 5.20 in:
Wine Staging.app/Contents/Resources/wine/lib64/libMoltenVK.dylib
so we could test the eventual CI artifact libMoltenVK.dylib by overwriting this file..
*really comparing MoltenVK.dylib shipped by Crossover 20 (which has DXVK support) vs standard shows all needed to fakely expose is:
geometryShader                 = true
shaderCullDistance           = true
as said a custom enviroment variable could do the job of forcely reporting this two caps or not..

oscarbg:
Hi,
altough little unrelated, just seen MetalAngle project has done a new release and includes as an artifact a compressed dylib file:
https://github.com/kakashidinho/metalangle/releases/tag/gles3-0.0.3
perhaps also future MoltenVK releases could have similar artifacts in addition to source code only, or download the Vulkan SDK MacOS release.. I think MetalAngle devs see their project more as an end user upgradeable runtime library and as such their releases include a dylib..
just saying.. no pressure to implement any ofmy wishes.. :-)

PR #1175, opened 2020-12-08
Add GitHub Actions CI

past-due:
- Adds GitHub Actions CI
- CI now builds for macOS, Mac Catalyst, iOS, and tvOS
- Upload build artifacts (Resolves: #1127)
- Cache dependencies for faster CI
   - Once first built, dependencies will be cached for the particular combination of:
      - `runner.os`, Xcode version, target platform, `fetchDependencies` script, `ExternalRevisions` folder contents, `ExternalDependencies.xcodeproj`, `Scripts` folder contents
      - (meaning if any of the above change, the cache is invalidated and dependencies will be built from scratch again before being cached for future runs - i.e. this should "just work" as expected)
- `xcpretty` is now supported for xcodebuild output formatting, if available


past-due:
Note: you will have to merge this before the workflow is run for future PRs.

past-due:
@billhollings This should function as an effective replacement for Travis CI, which seems to be having issues lately. Please let me know if there are any desired changes.

oscarbg:
@past-due this is what I wanted, thanks!!

Gcenx:
@past-due i have a question on this, as I’m not familiar with using the GitHub CI

Would this also attach the build artifacts to GitHub releases section once a release is marked?

past-due:
> @past-due i have a question on this, as I’m not familiar with using the GitHub CI
> 
> Would this also attach the build artifacts to GitHub releases section once a release is marked?

@Gcenx That could be added as another step. (But this workflow does not currently do that.)

There are several GitHub Actions that make doing so easier (examples: [actions/upload-release-asset](https://github.com/actions/upload-release-asset), [softprops/action-gh-release](https://github.com/softprops/action-gh-release)) that could be added to the workflow (or via a separate workflow triggered by the creation of a tag, etc).

However I wasn't sure what the specific build requirements / steps were for releases, or how those might differ from the existing Travis configuration, so I just largely replicated the Travis CI setup for pushes and PRs (with the above-mentioned enhancements).

.github/workflows/CI.yml
```diff
@@ -0,0 +1,95 @@
...
+            scheme: "MoltenVK Package (macOS only)"
+            demos-scheme: "Cube-macOS"
+          - name: "MoltenVK (iOS)"
+            platform: "ios"
+            scheme: "MoltenVK Package (iOS only)"
+            demos-scheme: ""
+          - name: "MoltenVK (tvOS)"
+            platform: "tvos"
+            scheme: "MoltenVK Package (tvOS only)"
+            demos-scheme: ""
```
cdavis5e, line 28: I'd suggest adding one for Catalyst, too, but there doesn't seem to be a `(Mac Catalyst only)` scheme. Someone just hit a build error (#1176) on Catalyst, and we'd like to avoid that in the future.

billhollings: >there doesn't seem to be a (Mac Catalyst only) scheme

Perhaps CI can just run `make`? With the right args, it can build for whichever platforms we want, including Mac Cat.

past-due: I’ll revise how the workflow works, so we can get additional platforms easily. 👍

billhollings: The full `fetchDependencies` build for MoltenVK can take up to half an hour at times. As a result, the Travis CI build was just set up as a smoke test running on macOS only.

We can explore moving to a full CI on all platforms if that is necessary, desirable, and functional.

billhollings:
> > Would this also attach the build artifacts to GitHub releases section once a release is marked?
> 
> @Gcenx That could be added as another step. (But this workflow does not currently do that.)

To combine CI and artifact needs for all platforms, we can run...

    ./fetchDependencies --all
    make

then zip/tar up the resulting `Package/Release/MoltenVK` directory as an artifact for storage.

Gcenx:
Having just libMoltenVK attached to releases would be more then ideal (for for CL tests in general) all other utility’s could be build yourself or grab the VulkanSDK.

past-due:
> > We can explore moving to a full CI on all platforms if that is necessary, desirable, and functional.
> 
> > > Would this also attach the build artifacts to GitHub releases section once a release is marked?
> > 
> > 
> > @Gcenx That could be added as another step. (But this workflow does not currently do that.)
> 
> To combine CI and artifact needs for all supported platforms, we can run...
> 
> ```
> ./fetchDependencies --all
> make
> ```
> 
> then zip/tar up the resulting `Package/Release` directory as an artifact for storage.
> 
> We can probably skip building a Cube demo. And perhaps I'll trim `MoltenVKShaderConverter.xcframework` from the build package, to reduce the size of the artifact. It's unlikely anyone uses it.

@billhollings 
To speed up regular CI (for pushes and PRs), I'm still splitting the builds by platform. This enables building dependencies for each platform in parallel. 

I'm now using the Makefile and have simplified the setup.
As long as the same "platform" (ex. "maccat") is used for both `fetchDependencies` and the `Makefile`, adding another platform to the build matrix is easy.

In regards to attaching build artifacts to a Release, here's what this current iteration will do:
- When pushing a tag, it will rebuild everything and upload build artifacts to the workflow run.
- These can be manually downloaded (from the appropriate run in the "Actions" tab) and uploaded to the GitHub Release.

If you'd like that step to be automated, I'm going to suggest we do it as a separate PR - there are several different choices as to how to accomplish it (and what you want the GitHub Release workflow to look like).

.github/workflows/CI.yml
past-due: Resolved - Mac Catalyst platform added.

fetchDependencies
```diff
@@ -444,6 +446,15 @@ EXT_DEPS=ExternalDependencies
 XC_PROJ="${EXT_DEPS}.xcodeproj"
 XC_DD_PATH="${EXT_DIR}/build"
 
+# Determine if xcpretty is present
+XCPRETTY_PATH=$(command -v xcpretty 2> /dev/null)
+
+# If xcpretty is present, use it to format xcodebuild output
+XC_OUTPUT_FMT_CMD="tee"
```
cdavis5e, line 453: You had the main build write output to a log, but not `fetchDependencies`. Why?

past-due: Handling the various combinations of options in `fetchDependencies` is more problematic (such as parallel builds), and I was trying to minimize changes in a single PR.

That said, I have refactored `fetchDependencies` to handle this better for the "common" case - it'll now use `tee` to output the full dependency build output to a log when using `xcpretty`, and use neither `xcpretty` nor `tee` if `--parallel-build` is specified.

.github/workflows/CI.yml
```diff
@@ -0,0 +1,92 @@
...
+        with:
+          path: |
+            External/build
+            !External/build/Intermediates
+          key: ${{ runner.os }}-${{ steps.prep.outputs.XCODE_VERSION }}-${{ matrix.platform }}-${{ hashFiles('fetchDependencies','ExternalRevisions/**','ExternalDependencies.xcodeproj/**','Scripts/**') }}
+
+      - name: Fetch Dependencies (Use Built Cache)
+        if: steps.cache-dependencies.outputs.cache-hit == 'true'
+        run: |
+          ./fetchDependencies -v --no-parallel-build --none
```
billhollings, line 62: `--no-parallel-build` is the default and not technically needed. Although I understand you want to ensure it.

.github/workflows/CI.yml
```diff
@@ -0,0 +1,92 @@
...
+
+      - name: Fetch Dependencies
+        if: steps.cache-dependencies.outputs.cache-hit != 'true'
+        run: |
+          ./fetchDependencies -v --no-parallel-build --${{ matrix.platform }}
```
billhollings, line 67: Ditto.

fetchDependencies
```diff
@@ -444,6 +452,19 @@ EXT_DEPS=ExternalDependencies
 XC_PROJ="${EXT_DEPS}.xcodeproj"
 XC_DD_PATH="${EXT_DIR}/build"
 
+# Determine if xcpretty is present
+XCPRETTY_PATH=$(command -v xcpretty 2> /dev/null)
```
billhollings, line 456: On my local machine, `fetchDependencies` doesn't get past this line. No error reported...even if I remove `2> /dev/null`.

billhollings (changes requested):
Thanks for taking care of this!

`fetchDependencies` won't run locally. See inline comment.

I see that the repo settings have actions enabled, but the Actions section seems to be uninitialized. Should this PR itself not be triggering an Action?

Also...might as well remove `.travis.yml`. It's currently doing nothing except disappointing us all.

billhollings:
@past-due Thanks for taking the initiative to add this PR!

Khronos is currently exploring moving from Travis to GitHub actions...so this is putting us ahead of the curve!

past-due:
@billhollings:

> `fetchDependencies` won't run locally. See inline comment.

Was able to reproduce - fix pushed. 👍

> I see that the repo settings have actions enabled, but the Actions section of the repo seems to be uninitialized. Shouldn't this PR itself be triggering an Action?

The PR must first be merged into the master / main branch, which will create the first Actions workflow in the actual repo that contains the `on: pull_request` trigger, and then the workflow in *this* repo will run for PRs in the future.

```
When you create a pull request from a forked repository to the base repository, GitHub
sends the pull_request event to the base repository and no pull request events occur
on the forked repository.
```
Source: https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows#pull-request-events-for-forked-repositories

> Also...might as well remove `.travis.yml`. It's currently doing nothing except disappointing us all.

Will also push this change. (Was waiting on approval.)

Re: `--no-parallel-build` - Defaults can always change, so when contributing to another project I always err on the side explicit parameters. I've pushed a commit to remove it. 👍

billhollings:
>The PR must first be merged into the master / main branch, which will create the first Actions workflow in the actual repo that contains the `on: pull_request` trigger, and then the workflow in *this* repo will run for PRs in the future.

```
When you create a pull request from a forked repository to the base repository, GitHub
sends the pull_request event to the base repository and no pull request events occur
on the forked repository.
```

That doc quote.._.that the Action runs on the destination repo, not the source repo_...makes sense. But that in itself doesn't explain why it's not running here. I would have thought it would run whatever workflow the PR provides...not what's currently in the destination repo.

But if what you say is true...that the first PR with an Action triggers them to be enabled within the repo...that's fine.

I'm mostly asking to see how we can test if this actually will run an Action and do it correctly. For instance, the same symptom (not running) might occur if there is say a syntax or logic error in the workflow...or if Actions need to be enabled manually somewhere first.

past-due:
> > The PR must first be merged into the master / main branch, which will create the first Actions workflow in the actual repo that contains the `on: pull_request` trigger, and then the workflow in _this_ repo will run for PRs in the future.
> 
> ```
> When you create a pull request from a forked repository to the base repository, GitHub
> sends the pull_request event to the base repository and no pull request events occur
> on the forked repository.
> ```
> 
> That doc quote.._.that the Action runs on the destination repo, not the source repo_...makes sense. But that in itself doesn't explain why it's not running here. I would have thought it would run whatever workflow the PR provides...not what's currently in the destination repo.
> 
> But if what you say is true...that the first PR with an Action triggers them to be enabled within the repo...that's fine.
> 
> I'm mostly asking to see how we can test if this actually will run an Action and do it correctly. For instance, the same symptom (not running) might occur if there is say a syntax or logic error in the workflow...or if Actions need to be enabled manually somewhere first.

@billhollings

Until this PR is merged, you can look at the branch in my fork that's associated with this PR:
https://github.com/past-due/MoltenVK/tree/github_actions_1

The workflow is getting run *there* because it satisfies the `on: push` trigger.

All the details of the most recent run are here: https://github.com/past-due/MoltenVK/actions/runs/411483678 (You can even download the Build Artifacts if you'd like - the only restriction on those is being logged-in to a GitHub account.)

You could also, for example, push this PR's branch to a branch on this repo, which would similarly trigger the `on: push` trigger.

(Of note: Syntax errors in a workflow file yield an Action run that errors and provides the syntax error details.)

Assuming Actions are, in fact, enabled for this organization / repository (and they appear to be - the tab is visible), merging this PR should kick it all off and replicate exactly what happens in the source branch for this PR.

billhollings (approved):
Very cool. Thanks for providing this!

Gcenx:
@billhollings could the release tag be removed and set again causing the release action to be triggered?

past-due:
> @billhollings could the release tag be removed and set again causing the release action to be triggered?

@Gcenx Just to clarity: this workflow will not (currently) automatically attach the build artifacts to the GitHub Release - to do so requires additional steps in this (or potentially another) workflow file.

If that's something that @billhollings is interested in, let me know and I will try to look into it when I have a chance.

In the meantime, since there were no other code changes since the release tag (besides these GitHub Actions commits), you should be able to use the build artifacts here: https://github.com/KhronosGroup/MoltenVK/actions/runs/411710451

Gcenx:
@past-due sorry I thought that was done.

billhollings:
>this workflow will not (currently) automatically attach the build artifacts to the GitHub Release - to do so requires additional steps in this (or potentially another) workflow file.
>
>If that's something that @billhollings is interested in, let me know and I will try to look into it when I have a chance.

For the most part, official MoltenVK releases are tied to Vulkan SDK releases. Since MoltenVK binaries are already provided in the SDK itself, it looks like this would effectively be redundant.

As I understand the requirements, the real value of the generation of binary artifacts as part of repo CI is to allow devs to access MoltenVK binaries for inter-release builds, without having to build MoltenVK itself offline.

And of course, starting now, the binaries for any release will be available through the PR for that release.

If anyone has additional needs for release binaries that I'm not understanding yet, please let me know.

Gcenx:
I'll provide some reasons for this;

- Artifact won't be code-signed (can be handy)
- Not packaged in a DMG (useful for oxcross based docker workflows)

Other parts of the VulkanSDK can be compiled on lower macOS versions but libMoltenVK requires compiling from the most recent version of macOS along with the latest version of Xcode meaning you could be forced to have multiple systems/installs just to setup an environment. Downloading the entire VulkanSDK just for libMoltenVK would be a waste of bandwidth.

If there were release artifacts all homebrew/macports/osxcross etc would be required to do is grab the needed release artifact aka MoltenVK and grab the needed headers (if needed)

PR #1191, opened 2020-12-30
Don't use barriers in render passes on Apple GPUs.

cdavis5e:
Apple GPUs don't support them, and in fact Metal's validation layer will
complain if you try to use them.

### DIFF
--- a/MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
@@ -1401,7 +1401,6 @@ void MVKPhysicalDevice::initMetalFeatures() {
         _metalFeatures.mslVersionEnum = MTLLanguageVersion2_1;
         _metalFeatures.multisampleArrayTextures = true;
 		_metalFeatures.events = true;
-        _metalFeatures.memoryBarriers = true;
         _metalFeatures.textureBuffers = true;
 		_metalFeatures.quadPermute = true;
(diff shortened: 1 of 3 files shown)


MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
```diff
@@ -1440,6 +1441,9 @@
 			_metalFeatures.maxPerStageDynamicMTLBufferCount = _metalFeatures.maxPerStageBufferCount;
 			_metalFeatures.postDepthCoverage = true;
 			_metalFeatures.renderLinearTextures = true;
+			// Apple devices don't like these.
+			_metalFeatures.memoryBarriers = false;
+			_metalFeatures.textureBarriers = false;
```
billhollings, line 1446: It suppose it might mean more setting lines...but can you flip this around so that all settings are positively set, please?

The pattern on these settings is to start with everything disabled and then selectively enable them when appropriate. None of the other settings are enabled and then disabled again like this. It feels like this could become spaghetti-like and fragile over time.

billhollings (changes requested):
Minor tweak to how the features are enabled/disabled.

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
cdavis5e: Done.

billhollings (approved):
Thanks.

PR #1193, opened 2020-12-30
MVKDeviceMemory: Don't consider Memoryless host-accessible on macOS/tvOS.

cdavis5e:
I missed this when I added support for memoryless on macOS.

### DIFF
--- a/MoltenVK/MoltenVK/GPUObjects/MVKDeviceMemory.h
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKDeviceMemory.h
@@ -52,7 +52,7 @@ public:
 
 	/** Returns whether the memory is accessible from the host. */
     inline bool isMemoryHostAccessible() {
-#if MVK_IOS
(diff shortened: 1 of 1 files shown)


MoltenVK/MoltenVK/GPUObjects/MVKDeviceMemory.h
```diff
@@ -52,7 +52,7 @@ class MVKDeviceMemory : public MVKVulkanAPIDeviceObject {
 
 	/** Returns whether the memory is accessible from the host. */
     inline bool isMemoryHostAccessible() {
-#if MVK_IOS
+#if MVK_IOS_OR_TVOS || MVK_MACOS_APPLE_SILICON
```
billhollings, line 55: Hmmm...I guess at some point we might need something like?

    #define MVK_APPLE_SILICON    (MVK_IOS || MVK_TVOS || MVK_MACOS_APPLE_SILICON)

billhollings (approved):

PR #1194, opened 2020-12-31
MVKRenderPass: Use a non-trivial granularity for TBDR GPUs.

cdavis5e:
This parameter is intended to indicate the optimal granularity for the
render area. For TBDR GPUs, this will be the tile size. IMR GPUs
continue to use 1x1.

Apple GPUs support tile sizes of 16x16, 32x16, and 32x32. But, we can't
read the tile size used for a Metal render pass until a render command
encoder has been created. So for now, hardcode 32x32 for TBDR GPUs.

### DIFF
--- a/MoltenVK/MoltenVK/GPUObjects/MVKRenderPass.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKRenderPass.mm
@@ -767,7 +767,17 @@ MVKRenderPassAttachment::MVKRenderPassAttachment(MVKRenderPass* renderPass,
 #pragma mark -
 #pragma mark MVKRenderPass
 
-VkExtent2D MVKRenderPass::getRenderAreaGranularity() { return { 1, 1 }; }
+VkExtent2D MVKRenderPass::getRenderAreaGranularity() {
+#if MVK_IOS_OR_TVOS || MVK_MACOS_APPLE_SILICON
+    if (_device->_pMetalFeatures->tileBasedDeferredRendering) {
+        // This is the tile area.
+        // FIXME: We really ought to use MTLRenderCommandEncoder.tile{Width,Height}, but that requires
+        // creating a command buffer.
+        return { 32, 32 };
+    }
+#endif
+    return { 1, 1 };
+}
 
 MVKRenderSubpass* MVKRenderPass::getSubpass(uint32_t subpassIndex) { return &_subpasses[subpassIndex]; }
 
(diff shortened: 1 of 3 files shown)


MoltenVK/MoltenVK/GPUObjects/MVKRenderPass.mm
```diff
@@ -767,7 +767,17 @@ static uint32_t getNextViewMaskGroup(uint32_t viewMask, uint32_t* startView, uin
...
+#if MVK_IOS_OR_TVOS || MVK_MACOS_APPLE_SILICON
+    if (_device->_pMetalFeatures->tileBasedDeferredRendering) {
+        // This is the tile area.
+        // FIXME: We really ought to use MTLRenderCommandEncoder.tile{Width,Height}, but that requires
+        // creating a command buffer.
+        return { 32, 32 };
+    }
+#endif
+    return { 1, 1 };
+}
```
billhollings, line 780: Just trying to think of ways to make this not-hardcoded.

Is granularity settable in Vulkan the way it is via `MTLRenderPassDescriptor.tile{Width,Height}`? Or do we just need a way of describing a value set by the system? I couldn't find it as settable within Vulkan...but I may have missed something?

And do we know if the default value of `MTLRenderCommandEncoder.tile{Width,Height}` and/or `MTLRenderPassDescriptor.tile{Width,Height}` are fixed for the GPU, or could they change renderpass by renderpass, based on FB bounds? Wondering if we could create a temp `MTLRenderPassDescriptor` or `MTLRenderCommandEncoder` at `MVKDevice` creation time and read the values?

billhollings (commented):
Question about hardcoding...

MoltenVK/MoltenVK/GPUObjects/MVKRenderPass.mm
cdavis5e: > Is granularity settable in Vulkan the way it is via `MTLRenderPassDescriptor.tile{Width,Height}`? Or do we just need a way of describing a value set by the system? I couldn't find it as settable within Vulkan...but I may have missed something?

I couldn't find one either. So we just need a way of describing the value that was set by Metal.

> And do we know if the default value of `MTLRenderCommandEncoder.tile{Width,Height}` and/or `MTLRenderPassDescriptor.tile{Width,Height}` are fixed for the GPU, or could they change renderpass by renderpass, based on FB bounds?

We don't know. I was afraid it were the latter, so I picked the most conservative value. Maybe we should ask.

PR #1199, opened 2021-01-08
Hide note about a markdown reader if you already use one.

kohver:
I believe this information is unnecessary if you read the document in github for example. Just an idea. Feel free to close the issue if you see any reason to not apply this change.


billhollings:
@kohver 

Thanks for your contribution. Good catch...and interesting fix.

Can you sign the CLA so we can pull this in?

kohver:
Done

billhollings (approved):

PR #1206, opened 2021-01-19
fix incorrect behavior for MVKCmdResolveImage

mbechard:
fix incorrectly changing first resolve layer's src/dst base layer,
as well and out of bounds access to the mtlResolveSlices array


cdavis5e (approved):
Ah, thank you for fixing my fencepost error.

PR #1212, opened 2021-01-22
GitHub Actions: Update Xcode to 12.3

past-due:
To be in-sync with #1211

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -10,7 +10,7 @@ on:
 
 env:
   # See: https://github.com/actions/virtual-environments/tree/main/images/macos
-  XCODE_DEV_PATH: /Applications/Xcode_12.2.app/Contents/Developer
+  XCODE_DEV_PATH: /Applications/Xcode_12.3.app/Contents/Developer
 
 # A workflow run is made up of one or more jobs that can run sequentially or in parallel
 jobs:


.github/workflows/CI.yml
```diff
@@ -10,7 +10,7 @@ on:
 
 env:
   # See: https://github.com/actions/virtual-environments/tree/main/images/macos
-  XCODE_DEV_PATH: /Applications/Xcode_12.2.app/Contents/Developer
+  XCODE_DEV_PATH: /Applications/Xcode_12.3.app/Contents/Developer
```
billhollings, line 14: Ah! Thanks! Good catch!

I'll have to try to remember to keep this up to date as Xcode evolves.

I assume forgetting to do this actually doesn't cause any issues, except in the case where the change in Xcode reflects a change in our use of Apple SDK artifacts (ie. classes, functions, enums introduced in that version of Xcode).

billhollings (approved):

.github/workflows/CI.yml
past-due: For future reference, you'll want to make sure that the GitHub Actions virtual environment image has been updated to include a new version of Xcode.

@billhollings:
At the moment, the `macos-latest` image [maps to `macos-10.15`](https://docs.github.com/en/actions/reference/specifications-for-github-hosted-runners#supported-runners-and-hardware-resources), and that image's list of available Xcode versions (and paths) can be found here: https://github.com/actions/virtual-environments/blob/main/images/macos/macos-10.15-Readme.md#xcode

At some point `macos-latest` will map to `macos-11.0` (the `macos-11.0` image is currently in preview, and has more limited capacity available).

billhollings: Great. Thanks for the education, and for your work on setting Actions up!

PR #1213, opened 2021-01-22
Minor non-behavior-changing updates and clean-up

billhollings:
This PR contains a few code-clean-up updates that should not affect behavior.

- `MVKSmallVector` allow constructor to size with default values.
- Remove obsolete `MVKVector`, which was long ago replaced with `MVKSmallVector`.
- Remove unnecessary concrete implementations of template functions that are used only within a single compilation unit.
- Add strings for all current `VkResult` values.
- Make logging functions globally accessible to ease their use in debugging activities.
- Add and use `MVK_APPLE_SILICON` macro definition.
- Remove a few unused member variables.


billhollings:
@cdavis5e Specifically wanted another pair of eyes on use of `MVK_APPLE_SILICON`.

cdavis5e (approved):

PR #1217, opened 2021-01-25
Advertise support for shaderInt64 feature.

### LINKED ISSUE #1203: Add support for shaderInt64
Now that Metal supports 64-bit integers, please add shaderInt64 support.

billhollings:
Add `MVKDevice::mslVersionIsAtLeast()` to support tests for MSL version.
Also bump MSL support level of standalone `MoltenVKShaderConverter` tool.

Fixes #1203.

### DIFF
--- a/MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
@@ -1575,7 +1575,6 @@ void MVKPhysicalDevice::initFeatures() {
     _features.shaderStorageBufferArrayDynamicIndexing = true;
     _features.shaderClipDistance = true;
     _features.shaderInt16 = true;
-	_features.shaderInt64 = mslVersionIsAtLeast(MTLLanguageVersion2_2);
     _features.multiDrawIndirect = true;
     _features.inheritedQueries = true;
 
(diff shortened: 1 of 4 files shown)


MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
```diff
@@ -1575,6 +1575,7 @@
     _features.shaderStorageBufferArrayDynamicIndexing = true;
     _features.shaderClipDistance = true;
     _features.shaderInt16 = true;
+	_features.shaderInt64 = mslVersionIsAtLeast(MTLLanguageVersion2_2);
```
cdavis5e, line 1578: I think we should only enable this for MSL 2.3. That's the first version that allows loading and storing 64-bit integers without a bitcast. If we want to support this on 2.2, we'll need to have SPIRV-Cross insert those bitcasts, and somehow I don't think H-K is amenable to that.

We may also want to limit this to particular GPU families. According to the Metal Feature Set Tables, only Apple families 3 through 7 support 64-bit integers fully. It might work on Mac families 1 and 2; needs testing.

MoltenVKShaderConverter/MoltenVKShaderConverterTool/OSSupport.mm
```diff
@@ -71,7 +71,11 @@
 #define mslVer(MJ, MN, PT)	mslVersionMajor == MJ && mslVersionMinor == MN && mslVersionPoint == PT
 
 	MTLLanguageVersion mslVerEnum = (MTLLanguageVersion)0;
-	if (mslVer(2, 1, 0)) {
+	if (mslVer(2, 3, 0)) {
+		mslVerEnum = MTLLanguageVersion2_1;
+	} else if (mslVer(2, 2, 0)) {
+		mslVerEnum = MTLLanguageVersion2_1;
```
cdavis5e, line 77: Copy-pasto.

cdavis5e (changes requested):

cdavis5e (approved):

PR #1220, opened 2021-01-26
Make MVKConfiguration access global, ignoring provided VkInstance.

billhollings:
MVKConfiguration access is now global, and the `VkInstance` provided in the
`vkGet/Set/MoltenVKConfigurationMVK()` functions is ignored. This allows these
functions to be provided with a `VkInstance` object that originates from a
different Vulkan layer than MoltenVK, without risking breaking the API.

`MVKConfiguration` extended to cover all MoltenVK environment variables.

Move all environment variable declarations to `MVKEnvironment.h`.
Add `MVKEnvironment.cpp` to define config functions.
Cleanup `.m` files to use `MVKCommonEnvironment.h` instead of `MVKEnvironment.h`.

### DIFF
--- a/Docs/MoltenVK_Runtime_UserGuide.md
+++ b/Docs/MoltenVK_Runtime_UserGuide.md
@@ -373,12 +373,13 @@ where `HEADER_FILE` is one of the following:
   These functions are exposed in this header for your own purposes such as interacting with *Metal* 
   directly, or simply logging data values.
 
->***Note:*** The functions in `vk_mvk_moltenvk.h` are not supported by the *Vulkan SDK Loader and Layers*
- framework. The opaque Vulkan objects used by the functions in `vk_mvk_moltenvk.h` (`VkInstance`, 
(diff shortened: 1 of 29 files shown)


cdavis5e (approved):
Looks good. This should also make sure there's a single point of truth for all MoltenVK-specific configuration. I like that. I remember being confused by the settings for `MTLHeap` or `MTLEvent` not being in that struct.

billhollings:
> Looks good. This should also make sure there's a single point of truth for all MoltenVK-specific configuration. I like that. I remember being confused by the settings for `MTLHeap` or `MTLEvent` not being in that struct.

Yup. It was precisely for the layer-object (`VkInstance` in this case) mismatch issue addressed in this PR that I was migrating away from the `MVKConfiguration` API in favor of environment variables, and was treating `MVKConfiguration` as legacy. 

But env vars have their own issues in that they require external set up (outside of app code). This now brings both approaches together.

PR #1223, opened 2021-01-28
MVKQueryPool: Totally rework the way occlusion queries work.

cdavis5e:
Instead of having Metal directly write to the query pool's internal
storage, we'll have it write to a temp buffer whose lifetime is tied to
the command buffer. The temp buffer's contents are then accumulated to
all queries that were activated.

This last step is particularly important for queries that span multiple
render passes. Since Metal resets the query counter at a render pass
boundary, this means that, up until now, only the last draw counted
toward the query. Data from the others were lost. By using this temp
buffer and accumulating the results to the query storage, the counter
will correctly count draws from all render passes inside the query
bounds.

This will also fix problems using multiple query pools, particularly
with large query pool support on, in a single render pass. Because Metal
requires us to set the visibility results buffer at render pass start
time, we couldn't use multiple query pools inside a single render pass.
Using a single temp buffer bypasses this problem.

Also, don't make queries available to the host unless they became
available to the device first. That way, a query that is immediately
reset during command buffer execution will properly report that the
query is unavailable. This fixes the remaining dEQP-VK.query_pool.*
tests. Fix some bugs that shook out of this.

### DIFF
--- a/MoltenVK/MoltenVK/Commands/MVKCommandBuffer.mm
+++ b/MoltenVK/MoltenVK/Commands/MVKCommandBuffer.mm
@@ -74,7 +74,7 @@ VkResult MVKCommandBuffer::reset(VkCommandBufferResetFlags flags) {
 	_wasExecuted = false;
 	_isExecutingNonConcurrently.clear();
 	_commandCount = 0;
-	_initialVisibilityResultMTLBuffer = nil;		// not retained
+	_needsVisibilityResultMTLBuffer = false;
 	_lastTessellationPipeline = nullptr;
 	_lastMultiviewSubpass = nullptr;
 	setConfigurationResult(VK_NOT_READY);
@@ -198,10 +198,10 @@ MVKCommandBuffer::~MVKCommandBuffer() {
 // found among any of the secondary command buffers, to support the case where a render pass is started in
 // the primary command buffer but the visibility query is started inside one of the secondary command buffers.
 void MVKCommandBuffer::recordExecuteCommands(const MVKArrayRef<MVKCommandBuffer*> secondaryCommandBuffers) {
-	if (_initialVisibilityResultMTLBuffer == nil) {
+	if (!_needsVisibilityResultMTLBuffer) {
 		for (MVKCommandBuffer* cmdBuff : secondaryCommandBuffers) {
-			if (cmdBuff->_initialVisibilityResultMTLBuffer) {
-				_initialVisibilityResultMTLBuffer = cmdBuff->_initialVisibilityResultMTLBuffer;
+			if (cmdBuff->_needsVisibilityResultMTLBuffer) {
+				_needsVisibilityResultMTLBuffer = true;
 				break;
 			}
 		}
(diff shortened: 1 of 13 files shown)


MoltenVK/MoltenVK/Commands/MVKCommandBuffer.mm
```diff
@@ -334,7 +334,21 @@
...
     getSubpass()->populateMTLRenderPassDescriptor(mtlRPDesc, _multiviewPassIndex, _framebuffer, _clearValues.contents(), _isRenderingEntireAttachment, loadOverride);
-    mtlRPDesc.visibilityResultBuffer = _occlusionQueryState.getVisibilityResultMTLBuffer();
+    if (_occlusionQueryState.getNeedsVisibilityResultMTLBuffer()) {
+        if (!_visibilityResultMTLBuffer) {
+            // Unfortunately, the temp buffer mechanism tends to allocate large buffers and return offsets into them.
+            // This won't work with visibility buffers, particularly if the offset is greater than the maximum supported
+            // by the device. So we can't use that.
+            // Use a local variable to make sure it gets copied.
+            id<MTLBuffer> visibilityResultMTLBuffer = [getMTLDevice() newBufferWithLength: _pDeviceMetalFeatures->maxQueryBufferSize options: MTLResourceStorageModePrivate];    // not retained
+            [visibilityResultMTLBuffer setPurgeableState: MTLPurgeableStateVolatile];
```
billhollings, line 344: Ugh! This could slow things down. `MTLBuffers` are really slow to allocate. And I've seen hundreds of Vulkan command buffers per frame. Not all doing occlusion of course...but still.

And the temp buffers also automatically handle the cleanup on command buffer completion, so they're just begging to be used here.

I'm not sure I understand the issue here. What's special about this use case? And what do you mean by an offset that can't be supported on the device?

Maybe (not necessarily on this PR) we can improve temp buffer allocation pooling to deal with whatever this use case needs.

billhollings (commented):
Looks good. I'd like to see if we can do something about the `MTLBuffer` allocation (not necessarily on this PR). See inline comments.

MoltenVK/MoltenVK/Commands/MVKCommandBuffer.mm
cdavis5e: > I'm not sure I understand the issue here. What's special about this use case? And what do you mean by an offset that can't be supported on the device?

From doing tessellation, my experience with the temp buffer mechanism is that it tends to allocate large buffers, and then it carves regions out of them to use. Reading the source confirms that, for small buffers (i.e. 256k or smaller), this is indeed what it does. This is fine for most cases, but it doesn't work for visibility result buffers because the offset passed to -[MTLRenderCommandEncoder setVisibilityResultMode:offset:] cannot be greater than some implementation -defined maximum--formerly 64k less 8 bytes, for later GPUs 256k less 8 bytes. This is why query buffers are limited in size in the first place. If two buffers were taken from the temp buffer pool, the second buffer would have an offset exceeding this maximum, which could cause problems.

billhollings: >the offset passed to -[MTLRenderCommandEncoder setVisibilityResultMode:offset:] cannot be greater than some implementation-defined maximum

Ah. Thanks. Of course. Now I remember. That was the reason for the `supportLargeQueryPools` config setting.

Too bad `MTLRenderPassDescriptor setVisibilityResultBuffer:` doesn't allow for an offset to be set, as Metal often does elsewhere.

Okay. Let's go with this for now. I'll add to my own To-Do list to update the `MTLBuffer` pooling services to support the need for specifying a maximum offset (which might be zero). That will effectively dedicate a `MTLBuffer` per use, for this use-case, but at least they'll be pooled instead of transient.

I'll take care of that To-Do item myself on a separate PR.

PR #1227, opened 2021-01-29
Fix build on Xcode 11.

cdavis5e:
I missed this.

### DIFF
--- a/MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
@@ -1580,7 +1580,11 @@ void MVKPhysicalDevice::initFeatures() {
 #if MVK_TVOS
     _features.textureCompressionETC2 = true;
     _features.textureCompressionASTC_LDR = true;
+#if MVK_XCODE_12
 	_features.shaderInt64 = mslVersionIsAtLeast(MTLLanguageVersion2_3) && supportsMTLGPUFamily(Apple3);
+#else
+	_features.shaderInt64 = false;
+#endif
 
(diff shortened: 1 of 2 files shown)


billhollings (approved):
Thanks for catching. I'll try to remember to attempt builds with Xcode 11.7 as well.

billhollings:
What are the reasons for supporting previous Xcode? And how long for?

In the past we haven't supported legacy Xcode's once the current version is out of beta.

cdavis5e:
> What are the reasons for supporting previous Xcode? And how long for?
> 
> In the past we haven't supported legacy Xcode's once the current version is out of beta.

You don't remember? A bunch of clients requested it--including me. Also, at the time, macOS 11 was still in beta, even though Xcode 12 was out of beta, so the final version of Xcode 12 only had macOS 10.15 SDKs.

I suppose at some point we'll need to move to XC12, particularly since that's the first version that supports building for M1 natively. I'll need to discuss this with my coworkers.

billhollings:
I remember the requests based on builds. But I forget the deets, who needs them, and on what platforms. That's why I'm asking.

There are quite a few outstanding issues with iOS builds in Xcode 11 right now (border color for one, and a couple of others). If we're serious about this, we'll need to fix the iOS breakages too. I'll have a stab at it in a bit.

billhollings:
Looks like we'll need somethings like the following in a few places:

```
#if (MVK_MACOS || (MVK_IOS && MVK_XCODE_12))
#endif
```

billhollings:
I've posted a dev poll as issue #1228. Mostly to see whether we should bother with the changes for iOS builds. 

I'm hoping not, because that `#if (MVK_MACOS || (MVK_IOS && MVK_XCODE_12))` macro is starting to get clumsy and fragile.

PR #1229, opened 2021-02-01
MVKCommandEncoder: Use the temp buffer mechanism for visibility buffers.

cdavis5e:
Add support for "dedicated" temp buffers, where instead of allocating a
big buffer and carving regions out of it, a unique buffer is returned
for each allocation request. This is necessary for visibility buffers,
because the offset passed to `-[MTLRenderCommandEncoder
setVisibilityResultMode:offset:]` cannot exceed an
implementation-defined value, currently 256k less 8 bytes for Mac family
2 on Catalina and up, and on Apple family 7; and 64k less 8 bytes
otherwise.

### DIFF
--- a/MoltenVK/MoltenVK/Commands/MVKCommandBuffer.h
+++ b/MoltenVK/MoltenVK/Commands/MVKCommandBuffer.h
@@ -370,7 +370,7 @@ public:
     void setComputeBytes(id<MTLComputeCommandEncoder> mtlEncoder, const void* bytes, NSUInteger length, uint32_t mtlBuffIndex);
 
     /** Get a temporary MTLBuffer that will be returned to a pool after the command buffer is finished. */
-    const MVKMTLBufferAllocation* getTempMTLBuffer(NSUInteger length);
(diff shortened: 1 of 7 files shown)


billhollings (approved):
Looks good. Thanks for taking care of this one.

And not a moment too soon! As I indicated on the original PR, I put this on my ToDo list and was about to start working on it later today.

Hmmm...which leads me to wonder...how do you want to handle the possibility that eventually we might end up unknowingly working on the same issue simultaneously?

One option I can think of is that we assign ourselves to a repo issue as we start to work on it (possibly creating that repo issue if it doesn't exist, as in this case).

Or would that be overkill for the most part?

What do you think?

cdavis5e:
> One option I can think of is that we assign ourselves to a repo issue as we start to work on it (possibly creating that repo issue if it doesn't exist, as in this case).
> 
> Or would that be overkill for the most part?
> 
> What do you think?

I've actually been thinking about doing that. I agree we do need to coordinate better.

billhollings:
Okay. I agree. Let's start with self-assigning on issues (and creating the issue where needed).

PR #1232, opened 2021-02-01
MVKMTLBufferAllocation: Mark temp buffers as volatile.

### LINKED ISSUE #1231: Temp buffers should be marked volatile
MoltenVK uses a number of temporary buffers to support various features of Vulkan. These buffers are cached in case they are needed again, because creating new resources is expensive. But, this leaves a lot of idle memory tied up, memory which might be put to better use elsewhere.

But, Metal provides a feature to allow the system to reclaim memory from idle resources. By marking the resource with `MTLPurgeableStateVolatile`, Metal can use the resource memory for other purposes; when it does so, it marks the resource `MTLPurgeableStateEmpty`. Then when we need the cached resource, we can mark it `Volatile` again and start using it. We don't really care about their contents after the commands using them are done, so we thankfully don't need to reload them.

cdavis5e:
They are not expected to be useful beyond the commands that use them,
but they take up memory nonetheless. This is exactly the use case
purgeability was designed for. Tell the system that it's OK to reclaim
their memory if necessary.

Doing this every time the buffer is used will cause the purgeable state
to be reset from `MTLPurgeableStateEmpty`, in case the system really did
reclaim their memory.

In accordance with Apple's advice, lock the pages for the buffer when
loading it, so the memory isn't pulled out from under us.

Fixes #1231.

### DIFF
--- a/MoltenVK/MoltenVK/Commands/MVKCmdTransfer.mm
+++ b/MoltenVK/MoltenVK/Commands/MVKCmdTransfer.mm
@@ -28,6 +28,7 @@
 #include "MVKEnvironment.h"
 #include "mvk_datatypes.hpp"
 #include <algorithm>
+#include <sys/mman.h>
 
 
 #pragma mark -
@@ -1615,7 +1616,10 @@ void MVKCmdUpdateBuffer::encode(MVKCommandEncoder* cmdEncoder) {
 
     // Copy data to the source MTLBuffer
     MVKMTLBufferAllocation* srcMTLBufferAlloc = (MVKMTLBufferAllocation*)cmdEncoder->getCommandEncodingPool()->acquireMTLBufferAllocation(_dataSize);
-    memcpy(srcMTLBufferAlloc->getContents(), _srcDataCache.data(), _dataSize);
+    void* pBuffData = srcMTLBufferAlloc->getContents();
+    mlock(pBuffData, _dataSize);
+    memcpy(pBuffData, _srcDataCache.data(), _dataSize);
+    munlock(pBuffData, _dataSize);
 
     [mtlBlitEnc copyFromBuffer: srcMTLBufferAlloc->_mtlBuffer
                   sourceOffset: srcMTLBufferAlloc->_offset
--- a/MoltenVK/MoltenVK/Commands/MVKCommandBuffer.mm
+++ b/MoltenVK/MoltenVK/Commands/MVKCommandBuffer.mm
@@ -26,6 +26,7 @@
 #include "MTLRenderPassDescriptor+MoltenVK.h"
 #include "MVKCmdDraw.h"
 #include "MVKCmdRenderPass.h"
+#include <sys/mman.h>
 
 using namespace std;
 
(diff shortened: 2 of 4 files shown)


MoltenVK/MoltenVK/Utility/MVKObjectPool.h
```diff
@@ -161,7 +161,7 @@ class MVKObjectPool : public MVKBaseObject {
      * contains no objects. This differs from the acquireObject() function, which creates
      * and return a new instance if this pool is empty. This method is not thread-safe.
      */
-    T* nextObject() {
+    virtual T* nextObject() {
```
billhollings, line 164: This function was deliberately not virtualized to allow the core operation of unlinking an object from the head of the list to be preferably inlined for performance efficiency. The polymorphic virtualization for custom types is established in `newObject()`, which is called with much less frequency (if the pool is serving its purpose).

Can we move the invocation of `setPurgeableState:` to `MVKMTLBufferAllocator::acquireMTLBufferRegion()`, after `MVKMTLBufferAllocationPool::acquireObject()`, instead of needing `MVKMTLBufferAllocationPool::nextObject()`?

billhollings (changes requested):
Good catch. This is a great idea.

A couple of questions:

1) What's the contract about when Metal might reclaim the data? Do we know that it will last long enough to make it through the lifetime of the `MTLCommandBuffer`?

2) For a `MVKMTLBufferAllocation` that is not dedicated (per your recent addition), this retention guarantee requirement might also extend across multiple `MTLCommandBuffers` (ie. Hard to say when the entire `MTLBuffer` would be reclaimable).

Plus one minor performance design change request in the inline comments.

cdavis5e:
> A couple of questions:
> 
> 1. What's the contract about when Metal might reclaim the data? Do we know that it will last long enough to make it through the lifetime of the `MTLCommandBuffer`?

I don't know. The docs say it will reclaim the memory without informing the app, implying the app has to check if the state is `Empty` to see if it were reclaimed. I expect, however, that it won't reclaim the memory as long as it's "in use," i.e. referenced by a command buffer and, if untracked, that buffer has `-[MTL*CommandEncoder useResource:usage:]` calls on that resource, or `-[MTL*CommandEncoder useHeap:]` calls if backed by a heap.

> 2. For a `MVKMTLBufferAllocation` that is not dedicated (per your recent addition), this retention guarantee requirement might also extend across multiple `MTLCommandBuffers` (ie. Hard to say when the entire `MTLBuffer` would be reclaimable).

See above.

MoltenVK/MoltenVK/Utility/MVKObjectPool.h
cdavis5e: > This function was deliberately not virtualized to allow the core operation of unlinking an object from the head of the list to be preferably inlined for performance efficiency. The polymorphic virtualization for custom types is established in `newObject()`, which is called with much less frequency (if the pool is serving its purpose).

I was afraid of that.

> 
> Can we move the invocation of `setPurgeableState:` to `MVKMTLBufferAllocator::acquireMTLBufferRegion()`, after `MVKMTLBufferAllocationPool::acquireObject()`, instead of needing `MVKMTLBufferAllocationPool::nextObject()`?

Why didn't I think of that? Done.

billhollings (approved):

billhollings:
> referenced by a command buffer

Ah. Yes. You're probably right. That should cover both cases.

PR #1234, opened 2021-02-02
CI enhancements.

billhollings:
Add `CI-legacy.yml` workflow to build on older Xcode versions.
Update `README.md` to document availability of MoltenVK binaries via CI artifacts.
Update `CI.yml` workflow to use Xcode 12.4.

### DIFF
--- a/.github/workflows/CI-legacy.yml
+++ b/.github/workflows/CI-legacy.yml
@@ -1,4 +1,4 @@
-name: CI
+name: CI-legacy
 
 # Controls when the action will run. 
 on:
@@ -11,8 +11,8 @@ on:
 env:
   # For supported Xcode versions, see: https://github.com/actions/virtual-environments/tree/main/images/macos
   # Then open macos-X.X-Readme.md#xcode
-  # Also see runs-on link for macos-latest below for which macOS platform will be run
-  XCODE_DEV_PATH: /Applications/Xcode_12.4.app/Contents/Developer
+  # Also see runs-on link below for which macOS platform will be run
+  XCODE_DEV_PATH: /Applications/Xcode_11.7.app/Contents/Developer
 
 # A workflow run is made up of one or more jobs that can run sequentially or in parallel
 jobs:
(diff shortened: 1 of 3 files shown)


billhollings:
@past-due   

I don't seem to be able to add you to the Reviewers request list. Don't know why.

Anyway...can you review as well, please?

cdavis5e:
> I don't seem to be able to add you to the Reviewers request list. Don't know why.

Probably because he/she is not a member of KhronosGroup.

past-due:
@billhollings Will take a full look. (From a first glance, there are a few ways this could be done without two separate workflows - unless you prefer having things duplicated / split like that?)

billhollings:
> there are a few ways this could be done without two separate workflows - unless you prefer having things duplicated / split like that

I did think about trying to do this in one workflow (although I am a GH-CI newbie, so didn't look into that in much detail for now), and I'd definitely like to get your insights and suggestions on that.

As you can see from the second file, for legacy, we need to support a different matrix of platforms (which might evolve distinctly in the future), and artifacts should not be created. In addition, I'd like to keep the two workflows as identifiable as separate in the GitHub web UI, so we can enable and disable them separately via the web UI.

Essentially, what is meant by "legacy" in terms of platforms to build and versions of Xcode, will evolve over time. And at times, we may not even need a legacy workflow enabled.

billhollings:
> Probably because he/she is not a member of KhronosGroup.

Hmmm. Possibly. Or maybe they need to be on the repo's official list of collaborators.

@past-due I've added you there now. You should get an invite.

past-due:
> @past-due I've added you there now. You should get an invite.

Accepted. (I believe it should now let you officially request a review.)

billhollings:
>(I believe it should now let you officially request a review

Yup. There you are.

billhollings:
@past-due @cdavis5e Ping for approval.

past-due:
@billhollings I've opened PR #1238 as an example of a different way this can be accomplished.

billhollings:
@past-due Thanks. #1238 is a much more elegant solution. I'm going to merge that PR and close this PR.

PR #1238, opened 2021-02-03
CI enhancements - update to Xcode 12.4, add (legacy) Xcode 11.7 build

past-due:
This is an alternative method of achieving #1234


billhollings:
Thanks for posting this! 

WAY better than my solution! 😉

PR #1239, opened 2021-02-04
Add ability to automatically capture first GPU frame.

billhollings:
Refactor auto GPU capture code to support both device and frame capture.
Add ability to automatically capture first GPU frame by setting
`MVK_CONFIG_AUTO_GPU_CAPTURE_SCOPE` to `2`.
Wrap GPU capture in autorelease pool to handle both cases of capture call
being made in run loop as in an app GUI, or in one-shot app like CTS.


MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
```diff
@@ -3675,6 +3675,62 @@ static uint32_t mvkGetEntryProperty(io_registry_entry_t entry, CFStringRef prope
...
+
+	if (_isCurrentlyAutoGPUCapturing || (mvkGetMVKConfiguration()->autoGPUCaptureScope != autoGPUCaptureScope)) { return; }
+
+	_isCurrentlyAutoGPUCapturing = true;
+
+	@autoreleasepool {
+		MTLCaptureManager *captureMgr = [MTLCaptureManager sharedCaptureManager];
+
+		// Before macOS 10.15 and iOS 13.0, captureDesc will just be nil
+		MTLCaptureDescriptor *captureDesc = [[MTLCaptureDescriptor new] autorelease];
```
cdavis5e, line 3688: Does that actually work? I suppose the class object would be weak-linked and thus `Nil` before 10.15.

billhollings: **_That_** line will work. `MTLCaptureDescriptor` [will be](https://developer.apple.com/library/archive/documentation/MacOSX/Conceptual/BPFrameworks/Concepts/WeakLinking.html#//apple_ref/doc/uid/20002378-107837) `nil` and so `new` and `autorelease` will be fine.

However, you've given me pause to think about this further. I guess it is not definite that the next two lines invoking`captureObject` and `destination` will work. They'd be `NULL` function pointers being called on a `nil` object. I suppose they might break, depending on the order of validation (whether the Obj-C runtime checks the `nil` object or `NULL` function pointer first).

cdavis5e: I think they actually should work. Normally, ObjC property syntax is just sugar for a method call, which in turn gets lowered to a call to `objc_msgSend()` (or variants). So this:
```objc
captureDesc.captureObject = mtlCaptureObject;
```
becomes:
```objc
[captureDesc setCaptureObject: mtlCaptureObject];
```
which in turn becomes:
```c
((void (*)(id, SEL, id))objc_msgSend)(captureDesc, "setCaptureObject:", mtlCaptureObject);
```

Messaging a `nil` object is harmless.

billhollings: Yeah. I agree. 

On my second thought above, I was starting to get worried about what order things were happening inside `objc_msgSend`. If it attempted to lookup the non-existent `SEL` first, I thought there might be a possibility that it might break somehow.

But now on **_third_** thought, `objc_msgSend` uses duck typing based on the `isa` ivar inherent in all Objc objects, and it has no context for looking up the `SEL` if the target is `nil`. So it must test for the `nil` first. [This article](http://blog.zhengdong.me/2013/07/18/a-look-under-the-hood-of-objc-msgsend/#the-runtime) gives a decent verification and explanation of that.

Okay. Leaving it as is now, and pulling it in.

Issue #1240, opened 2021-02-05
maxUniformBufferRange is 0 on Mac Mini M1


BeastLe9enD:
I noticed that the uniform buffer range is 0 on my M1 Mac mini. 
This leads to errors when validation layers are enabled. Without validation layers enabled, its working fine

A screenshot of the VkPhysicalDeviceProperties:
https://i.imgur.com/Eh2C4jb.png 

I used Vulkan SDK 1.2.162.1 version

cdavis5e:
We're setting that field by casting the maximum buffer size to a `uint32_t`. This is a mistake. If the maximum buffer size is 4 GiB or greater, the upper 32 bits will be lost. This value really needs to be clamped to 4 GiB - 1.

cdavis5e:
Does #1242 help? We now have build artifacts from PRs, so you don't need to build it yourself anymore.

BeastLe9enD:
Thanks, its working !:)
https://i.imgur.com/dzAtb5V.png

PR #1242, opened 2021-02-06
MVKPhysicalDevice: Clamp maximum buffer range to 4 GiB - 1.

cdavis5e:
It is possible for the maximum buffer size to be 4 binary gigabytes or
greater. In that case, the upper 32 bits will be lost, and the value of
the `maxUniformBufferRange` and `maxStorageBufferRange` limits would be
too small or even zero. Clamp it to 4 GiB - 1 in that case, since that
is the maximum value of a 32-bit integer.

For #1240.


cdavis5e:
I do wish the C++ standards committee would add versions of `min()` and `max()` with multiple type parameters, because I'm sick of having to cast one of the operands to the type I need.

billhollings (approved):

PR #1243, opened 2021-02-07
Test for null string on MVKConfiguration copy.

billhollings:
When app sets `MVKConfiguration`, if struct not fully populated by app, or if app sets strings to `null`,
copy can fail. Test for this before copying.

Also unrelated trivial tweak to What's New doc.

### DIFF
--- a/Docs/Whats_New.md
+++ b/Docs/Whats_New.md
@@ -32,7 +32,7 @@ Released TBD
 - `Makefile` and `fetchDependencies` support `xcpretty` (if available)
 - Update `VK_MVK_MOLTENVK_SPEC_VERSION` to `30`.
 - Update copyright notices to year 2021.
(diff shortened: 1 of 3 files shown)


cdavis5e (approved):
I take it you were seeing problems from assigning a `NULL` C-string pointer to a `std::string`?

cdavis5e:
> I take it you were seeing problems from assigning a `NULL` C-string pointer to a `std::string`?

Well duh, you said as much in your description! Silly me.

billhollings:
> > I take it you were seeing problems from assigning a `NULL` C-string pointer to a `std::string`?
> 
> Well duh, you said as much in your description! Silly me.

Yeah. With apps that are bound to a previous version of `MVKConfiguration`, without the recently added config attributes, the partially copied `MVKConfiguration` ends up with a `NULL` C-string pointer.

PR #1255, opened 2021-02-10
Fix Metal validation error of duplicate visibility offsets.

### LINKED ISSUE #1247: Invalid occlusion query indexes
Dota2 **_was_** sporadically hitting the following Metal validation:

```
-[MTLDebugRenderCommandEncoder setVisibilityResultMode:offset:]:3322: failed assertion `using the same offset (8) in the visibility buffer twice in the same command encoder is not allowed.'
```

After PR #1223 (presumably), this now seems to be happening every time, which should be helpful in tracking this down.

I'm looking into this.

billhollings:
Don't reset `MVKOcclusionQueryCommandEncoderState::_mtlVisibilityResultOffset`
when ending occlusion query, as subsequent queries need different offsets.

Also, remove `MVKOcclusionQueryCommandEncoderState::_needsVisibilityResultMTLBuffer`,
as it is always set once from `MVKCommandBuffer::_needsVisibilityResultMTLBuffer`.

The current logic around deciding whether a visibility buffer is needed is relatively crude. Currently, it is applied to all render passes in a Vulkan command buffer based on at least one occlusion query being present. A Metal renderpass needs a visibility buffer if a query is started either before it or inside it. For a command buffer with multiple Metal render passes, not all of them may need it. 

Fixing that logic was out of scope for this PR, but I'll add another Issue to track and address that logic.

Unrelated fix, when app setting MVKConfiguration it with a partial copy, start with existing.

Fixes #1247.


cdavis5e (approved):
Thanks for once again cleaning up my mess.

Issue #1253, opened 2021-02-10
Graphics Pipeline creation fails using MSL compiled code


ifiddynine:
I was just testing use of MSL compiled code to avoid SPIRV -> MSL conversion at runtime, but noticed that it's no longer working. The calls to `vkCreateShaderModule` with MSL binary code prepended with the `kMVKMagicNumberMSLCompiledCode` constant succeeds just fine, but we later get:

`[mvk-error] VK_ERROR_INITIALIZATION_FAILED: Failed to get vertex outputs: SPIRV file too small.`

.. when calling vkCreateGraphicsPipelines. This has worked for us in the past, but we weren't ready to integrate this feature by default. However it's broken in the 1.1.1 release. We'd love to switch to precompiled binary code by default in order to validate the MSL compilation in our build step instead of on app launch which has led to a few bugs sneaking through.

billhollings:
Loading MSL code has definitely fallen behind the curve as MoltenVK evolved. 

It is not recommended at this time, until it can be properly addressed.

I'll document that.

cdavis5e:
This is a dupe of #956.

ifiddynine:
Got it, thanks for the clarification!

billhollings:
>I'll document that.

Done. Closing now that documentation is updated, and this is a dup of other issues related to MSL.

PR #1266, opened 2021-02-12
Remove official support for direct MSL shader loading from documentation.

billhollings:
Shader code should be submitted as SPIR-V. Although some simple direct MSL shaders may work,
direct loading of MSL source code or compiled MSL code is not officially supported at this time.
Future versions of MoltenVK may support direct MSL submission again.

Addresses issue #1253

### DIFF
--- a/Docs/MoltenVK_Runtime_UserGuide.md
+++ b/Docs/MoltenVK_Runtime_UserGuide.md
@@ -27,7 +27,6 @@ Table of Contents
 	- [MoltenVK `VK_MVK_moltenvk` Extension](#moltenvk_extension)
 	- [Configuring MoltenVK](#moltenvk_config)
 - [*Metal Shading Language* Shaders](#shaders)
-	- [MoltenVKShaderConverter Shader Converter Tool](#shader_converter_tool)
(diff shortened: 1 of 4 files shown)


cdavis5e (approved):
It's really my fault we have to do this. I wanted to fix this, but I'm so busy with other stuff at the moment that I don't have time. I guess this will do for now.

billhollings:
> It's really my fault we have to do this. I wanted to fix this, but I'm so busy with other stuff at the moment that I don't have time. I guess this will do for now.

Meh. Don't be too hard on yourself. Direct MSL has fallen behind along several dimensions. There have been previous issues raised around how to map Metal resource indexes in a more sophisticated manner. And future Metal argument buffer use will complicate it even further. I expect it will need a fair bit of work to recover effectively and continue to maintain.

I'm also interested in following up on the relatively new pipeline caching options available through Metal. If we can mesh that with Vulkan's pipeline caching, that might provide a more Vulkan-friendly and maintainable approach to improving shader conversion and compiling performance, and mitigate some of the need to directly support MSL.

PR #1273, opened 2021-02-18
Set Metal buffer alignment to 256 on non-Apple Silicon iOS/tvOS simulators.

### LINKED ISSUE #1250: `vkGetBufferMemoryRequirements` reports wrong buffer alignment when running in the iOS simulator
When running a Metal app in the iOS simulator, Apple requires that align constant buffer offsets for render commands to be 256 bytes. (see: https://developer.apple.com/documentation/metal/developing_metal_apps_that_run_in_simulator?language=objc)

Trying to allocate a uniform buffer, `vkGetBufferMemoryRequirements` returns the alignment as 64 bytes. This leads to the following assert from the `metal API validation`: `validateFunctionArguments:3722: failed assertion 'Fragment Function(main0): the offset into the buffer _23 that is bound at buffer index 0 must be a multiple of 256 but was set to 24960.'`

The issue seems to be coming from this file: https://github.com/KhronosGroup/MoltenVK/blob/05066429e29cb5bf11d21cb2bc2b5ebfff94e7ae/MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm where `mtlBufferAlignment` is set to 256 only for `MVK_MACOS`, but there's no check for the iOS simulator.

billhollings:
Derive `MVK_APPLE_SILICON` from target CPU.
Derive `MVK_MACOS_APPLE_SILICON` from target CPU and macOS platform.
Derive `MVK_XCODE_12` from macOS and iOS SDK versions.
Test for simulator on non-Apple GPU using `MVK_OS_SIMULATOR` && !MVK_APPLE_SILICON`.

Fixes issue #1250.

### DIFF
--- a/MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
@@ -1546,13 +1546,9 @@ void MVKPhysicalDevice::initMetalFeatures() {
 #endif
 	}
 
-	// iOS and tvOS adjustments necessary when running in the simulator on non-Apple GPUs.
-	// Apple1 used as baseline for detecting Apple Silicon on macOS because Apple7 not supported for tvOS.
-	// Must run after setting native iOS and tvOS values above.
-#if MVK_OS_SIMULATOR
-	if ( !supportsMTLGPUFamily(Apple1) ) {
-		_metalFeatures.mtlBufferAlignment = 256;
-	}
+// iOS and tvOS adjustments necessary when running in the simulator on non-Apple GPUs.
+#if MVK_OS_SIMULATOR && !MVK_APPLE_SILICON
+	_metalFeatures.mtlBufferAlignment = 256;
 #endif
 }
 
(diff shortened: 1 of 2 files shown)


MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
```diff
@@ -1545,6 +1545,15 @@
 			break;
 #endif
 	}
+
+	// iOS and tvOS adjustments necessary when running in the simulator on non-Apple GPUs.
+	// Apple1 used as baseline for detecting Apple Silicon on macOS because Apple7 not supported for tvOS.
+	// Must run after setting native iOS and tvOS values above.
+#if MVK_OS_SIMULATOR
+	if ( !supportsMTLGPUFamily(Apple1) ) {
```
cdavis5e, line 1553: I normally use `Apple5`, since that's what the Apple Silicon DTKs had, but I guess that wouldn't be supported on tvOS, even on the simulator... would it?

cdavis5e (approved):

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
billhollings: Hmmm. I did some further testing. When running on a simulator, the test for GPU family runs against the device being simulated, not the platform the simulator is running on. So for instance, `Apple5` is not supported on the iPad 7th Gen simulator. Conversely, it IS supported on a different simulator running on macOS without Apple Silicon.

Bottom line, testing for GPU family doesn't test for the platform the simulator is running on.

As a result, I've moved to redefining some of the environment macros in terms of the CPU for which the build is targeted. 

Seems to work now, and be more consistent, but please have a look at the macro logic to see if you can see any issues that I might have missed in the logic path. Or if we are using any of these macros in a manner (rightly or wrongly) that the new logic will break, and which I've not noticed.

cdavis5e (approved):

PR #1275, opened 2021-02-20
MVKMTLBufferAllocation: Support private temp buffers.

cdavis5e:
Some buffers, particularly those used by tessellation, indirect
multiview, or occlusion queries, are never written or read from the
host. Keeping the relevant data in VRAM should improve performance by
reducing the need for the GPU to make expensive accesses over the PCI
bus.

VRAM is a scarce resource on discrete GPUs, but this is part of why we
marked the buffers volatile.

### DIFF
--- a/MoltenVK/MoltenVK/Commands/MVKCommandEncodingPool.mm
+++ b/MoltenVK/MoltenVK/Commands/MVKCommandEncodingPool.mm
@@ -77,9 +77,13 @@ id<MTLDepthStencilState> MVKCommandEncodingPool::getMTLDepthStencilState(bool us
 	MVK_ENC_REZ_ACCESS(_cmdClearDefaultDepthStencilState, newMTLDepthStencilState(useDepth, useStencil));
 }
 
-const MVKMTLBufferAllocation* MVKCommandEncodingPool::acquireMTLBufferAllocation(NSUInteger length, bool isDedicated) {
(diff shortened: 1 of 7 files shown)


MoltenVK/MoltenVK/Commands/MVKCommandEncodingPool.mm
```diff
@@ -77,10 +77,13 @@
 	MVK_ENC_REZ_ACCESS(_cmdClearDefaultDepthStencilState, newMTLDepthStencilState(useDepth, useStencil));
 }
 
-const MVKMTLBufferAllocation* MVKCommandEncodingPool::acquireMTLBufferAllocation(NSUInteger length, bool isDedicated) {
+const MVKMTLBufferAllocation* MVKCommandEncodingPool::acquireMTLBufferAllocation(NSUInteger length, bool isPrivate, bool isDedicated) {
```
billhollings, line 80: To anticipate possible future unexpected behavior confusion, can we handle the `false, true` case (non-private dedicated), in case it ever gets attempted in the future?

For now, I'm okay with just adding an `MVKAssert()`at the top that indicates that `false, true` is not going to do what the caller is expecting.

Another option would be to combine the two flags into an enum, but that is probably overkill.

billhollings (changes requested):
Nice catch and mod!

Just a minor future-proofing change requested.

billhollings (approved):

PR #1281, opened 2021-02-25
Log enhanced command buffer errors in debug mode.

cdavis5e:
If we're on macOS 11 or iOS/tvOS 14, and debug mode (`MVK_DEBUG=1`) were
enabled, then we'll enable enhanced command buffer errors, which give us
more information about which encoder triggered the error. We'll also
print log messages from shader validation.

### DIFF
--- a/MoltenVK/MoltenVK/GPUObjects/MVKQueue.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKQueue.mm
@@ -145,6 +145,26 @@ VkResult MVKQueue::waitIdle() {
 	return mvkWaitForFences(_device, 1, &fence, false);
 }
 
+id<MTLCommandBuffer> MVKQueue::getMTLCommandBuffer(bool retainRefs) {
+#if MVK_XCODE_12
+	if ([_mtlQueue respondsToSelector: @selector(commandBufferWithDescriptor:)]) {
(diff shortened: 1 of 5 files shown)


MoltenVK/MoltenVK/GPUObjects/MVKQueue.mm
```diff
@@ -312,7 +345,33 @@
...
+#endif
 		}
+#if MVK_XCODE_12
+		if (mvkGetMVKConfiguration()->debugMode) {
+			MVKLogInfo("Shader log messages:");
+			for (id<MTLFunctionLog> log in mtlCB.logs) {
+				MVKLogInfo("%s", log.description.UTF8String);
+			}
+		}
+#endif
```
billhollings, line 374: Hmmm...probably lots of under-the-covers Obj-C object creation going on as a result array copying and fast-enumeration. It's got me wondering if Metal makes sure to wrap callbacks in autorelease pools.

Couldn't find any docs on that.

In any case, since this is all debug activity, any leaks are unlikely to affect production, and its probably not worth the performance cost of wrapping every completion handler in an autorelease pool in production.

billhollings (approved):
Thanks for taking care of this. It should prove really helpful!

PR #1283, opened 2021-02-25
WIP: Added support for GPU timestamps

mp-rsx:
This adds proper support for GPU timings for `vkCmdWriteTimestamp`. 

The following changes were made
 - We check for timestamp support in `MVKPhysicalDevice` initialization
 - `MVKTimestampBuffers` is created for each command buffer if timestamps are supported. This object contains/allocates the GPU buffers for the samples.
 - Every encoder creation and `endEncoder` call are recorded in `MVKTimestampBuffers`, recording the time it took to perform the operations in the encoder
 - Every `vkCmdWriteTimestamp` query is associated with a timestamp sample that is last available from `MVKTimestampBuffers` when the query is issued
 - When queries are finalized we resolve the `MVKTimestampBuffers` into CPU memory and convert the GPU timestamps into CPU nanoseconds
 - Additionally, to correlate GPU -> CPU timings we record CPU/GPU timestamps whenever a command buffer is created, and when it finishes execution.

This has been tested on macOS 11.0.1 using Intel HD Graphics 630 and AMD Radeon RSX 570.


cdavis5e:
Looks like you broke the build... everywhere but Xcode 12 on macOS.

MoltenVK/MoltenVK/Commands/MVKCommandBuffer.h
```diff
@@ -172,6 +172,8 @@ class MVKCommandBuffer : public MVKDispatchableVulkanAPIObject,
 	bool _isReusable;
 	bool _supportsConcurrentExecution;
 	bool _wasExecuted;
+	TimestampCorrelationMarker _timestampCorrelationMarker;
+	std::shared_ptr<MVKTimestampBuffers> _timestampBuffers;
```
cdavis5e, line 176: Using the STL smart pointers is actually quite rare in this project. Maybe we should reconsider that.

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
```diff
@@ -1514,6 +1514,32 @@
...
+				isTimestampCounterSupported = true;
+				break;
+			}
+		}
+	}
+	
+	_metalFeatures.timestampCommandSamplingSupported = isTimestampCounterSupported &&
+		[_mtlDevice supportsCounterSampling:MTLCounterSamplingPointAtDrawBoundary] &&
+		[_mtlDevice supportsCounterSampling:MTLCounterSamplingPointAtDispatchBoundary] &&
+		[_mtlDevice supportsCounterSampling:MTLCounterSamplingPointAtBlitBoundary];
```
cdavis5e, line 1542: The calls to `-[MTLDevice supportsCounterSampling:]` need to be in an `#if MVK_XCODE_12` block.

MoltenVK/MoltenVK/GPUObjects/MVKQueryPool.h
```diff
@@ -35,6 +35,58 @@ class MVKCommandEncoder;
 #pragma mark -
 #pragma mark MVKQueryPool
 
+/**
+ * Index of a GPU query and associated timestamp sample (if the query is a timestamp query).
+ */
+struct ActivatedQueryInfo
```
cdavis5e, line 41: This type needs an `MVK` prefix.

MoltenVK/MoltenVK/GPUObjects/MVKQueryPool.h
```diff
@@ -35,6 +35,58 @@ class MVKCommandEncoder;
...
+{
+	uint32_t queryIndex;
+	uint32_t timestampSampleIndex;
+};
+
+/**
+ * CPU and GPU timestamps at two points in time.
+ */
+struct TimestampCorrelationMarker {
```
cdavis5e, line 50: This one, too.

MoltenVK/MoltenVK/GPUObjects/MVKQueryPool.mm
```diff
@@ -24,9 +24,112 @@
...
+	// lapse, then add the CPU start time to match the CPU’s timestamp.
+	double adjGpuTimeStamp = timestampCorrelationMarker.cpuStart + (timestampCorrelationMarker.cpuEnd - timestampCorrelationMarker.cpuStart) * normalizedGpuTime;
+	
+	// Convert from CPU counter to nanoseconds
+	mach_timebase_info_data_t time_info;
+	mach_timebase_info(&time_info);
+
+	return adjGpuTimeStamp * time_info.numer / time_info.denom;
+}
+
```
cdavis5e, line 47: We could probably avoid handling all this complexity ourselves by supporting `VK_EXT_calibrated_timestamps`.

MoltenVK/MoltenVK/GPUObjects/MVKQueryPool.mm
```diff
@@ -24,9 +24,112 @@
...
+constexpr uint32_t MVKTimestampBuffers::kTimestampCountPerBuffer;
+
+MVKTimestampBuffers::MVKTimestampBuffers(MVKDevice* device)
+	:_device(device), _nextTimestampIndex(0), _isResolved(false)
+{
```
cdavis5e, line 52: Braces should go on the same line as the function header.

MoltenVK/MoltenVK/GPUObjects/MVKQueryPool.mm
```diff
@@ -24,9 +24,112 @@
...
+		_bufferPool.push_back(newCounterSampleBuffer);
+	}
+	
+	outSampleBuffer = _bufferPool[destinationBufferIndex];
+	
+	_isResolved = false;
+	const uint32_t queryIndex = _nextTimestampIndex++;
+	return queryIndex % kTimestampCountPerBuffer;
+}
+
```
cdavis5e, line 86: Do we really need a buffer per query? I was thinking we'd do a buffer per pool, and each timestamp query would correspond to a sample location. Vulkan doesn't allow writing a query multiple times anyway.

cdavis5e (changes requested):
I'm sorry, but this change has serious issues. Aside from the build breakages, for one thing, it doesn't handle `vkCmdCopyQueryResultsToBuffer()` for timestamp queries.

MoltenVK/MoltenVK/GPUObjects/MVKQueryPool.h
```diff
@@ -35,6 +35,58 @@ class MVKCommandEncoder;
...
+	MTLTimestamp cpuStart;
+	MTLTimestamp cpuEnd;
+	MTLTimestamp gpuStart;
+	MTLTimestamp gpuEnd;
+};
+
+/**
+ * Manages buffers used for resolving timestamp counter samples.
+ */
+class MVKTimestampBuffers
```
billhollings, line 60: All classes in MoltenVK subclass from `MVKBaseObject`. There is substantial functionality in that class, and its subclass hierarchy.

MoltenVK/MoltenVK/Commands/MVKCommandBuffer.h
billhollings: >Using the STL smart pointers is actually quite rare in this project. Maybe we should reconsider that.

Actually, not rare, but non-existent. 

One of the issues is the use of the `MVKBaseObject` base class to underpin all MoltenVK objects (except POD structs). 

Among other class-hierarchy functionality, that base class is set  up to support a level of memory management. 

Originally this memory management was to include help with Vulkan `VkAllocationCallbacks`, which has yet to be found useful enough to implement. 

But one of the features of this idea that was implemented is the `MVKBaseObject ::destroy()` function, which can be overridden by subclasses with some useful functionality (such as reference counting and deferred deletion). And one of the requirements of **_that_** idea, is that we don't allow classes to be directly deleted via the `delete` operator. 

For better or worse, using STL smart pointers would bypass this management ability.

billhollings (changes requested):
Thanks very much for taking this on and submitting this PR!

@cdavis5e has done a much more complete review, and I've +1'd most of his detailed comments. I've also added a couple of change requests of my own.

MoltenVK/MoltenVK/GPUObjects/MVKQueryPool.mm
mp-rsx: This buffer is not per queries, but for timing samples. The queries themselves are kept as a separate concept that take from the sample buffer as needed.

Doing this a per-query pool basis was my initial approach, but ultimately I had to separate this into different concepts due to the encoder issue mentioned in my other comment.

MoltenVK/MoltenVK/Commands/MVKCmdTransfer.mm
```diff
@@ -519,6 +519,8 @@ static inline MTLSize mvkClampMTLSize(MTLSize size, MTLOrigin origin, MTLSize ma
                     mtlStencilAttDesc.slice = mvkIBR.region.dstSubresource.baseArrayLayer + layIdx;
                 }
                 id<MTLRenderCommandEncoder> mtlRendEnc = [cmdEncoder->_mtlCmdBuffer renderCommandEncoderWithDescriptor: mtlRPD];
+				cmdEncoder->recordTimestamp(mtlRendEnc);
```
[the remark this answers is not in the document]
mp-rsx: That doesn't work because the sampling command is issued on a per-encoder basis. Encoder might end encoding before the timestamp command is requested, or the user might have not even started the encoder if the timestamp is send before doing any other encoder commands.

I tried other approaches, including creating temporary encoders, but that resulted in invalid timings. This only way to catch all the correct timings is to wrap all encoder calls with a recording of the time it started and ended.

If you wanted to optimize this you could do a 'lookahead' and avoid recording the timestamp if there are no `vkCmdWriteTimestamp` commands queued in the command buffer.

mp-rsx:
> I'm sorry, but this change has serious issues. Aside from the build breakages, for one thing, it doesn't handle `vkCmdCopyQueryResultsToBuffer()` for timestamp queries.

How does it need to handle it? Ultimately the only thing I changed regarding queries is the location we retrieve the time from, rather than getting the CPU time we now read the counter sample buffer. I imagined the existing query functionality will work the same.

mp-rsx:
@billhollings Thank you Bill for making this contribution feel appreciated! I am on company time but I will try to make the relevant revisions quickly. The functionality is what we needed internally and I didn't push further than that, any help you can provide if there are any larger changes required would be good.

MoltenVK/MoltenVK/Commands/MVKCommandBuffer.h
mp-rsx: You can provide a custom deleter to `std::shared_pointer` which can call `destroy()`, but I'll see if I can find another alternative.

mp-rsx:
I have made the following changes:
 - Fixed build issues for other platforms
 - Added correct handling for Xcode 11
 - Removed the use of `std::shared_ptr` and replaced it with custom reference tracking
 - Inherited `MVKTimestampBuffers` for `MVKBaseObject` and `MVKDeviceTrackingMixin`
 - Added the missing `MVK` prefixes and resolved other code style issues

MoltenVK/MoltenVK/GPUObjects/MVKQueryPool.mm
mp-rsx: Also note we're not allocating a buffer per query, but rather a buffer per `kTimestampCountPerBuffer` queries.

MoltenVK/MoltenVK/Commands/MVKCommandBuffer.mm
```diff
@@ -551,17 +578,34 @@
     _occlusionQueryState.endMetalRenderPass();
 }
 
+void MVKCommandEncoder::recordTimestamp(id<MTLCommandEncoder> commandEncoder)
```
billhollings, line 581: This parameter declaration results in a build warning because `MTLCommandEncoder` doesn't actually define `sampleCountersInBuffer:atSampleIndex:withBarrier:`, I guess because the `MTLParallelRenderCommandEncoder` subclass doesn't support it.

Replacing `id<MTLCommandEncoder> commandEncoder` with `id commandEncoder` removes the build warning, and is no less safe, since you already check for `respondsToSelector`.

billhollings (changes requested):
Sorry for the delay in getting around to this review.

Thanks for updating this. Making progress. A few small changes that I didn't catch on the first review. Sorry for that. 

Also more discussion questions about the best use of `recordTimestamp()` calls.

MoltenVK/MoltenVK/Commands/MVKCmdTransfer.mm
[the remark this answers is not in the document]
cdavis5e: > I don't see us using `sampleBufferAttachments`, if they exist. Could that be part of the inaccuracy? What does the timestamp represent if no `sampleBufferAttachment` has been configured?

I think that's for `MTLCounterSamplingPointAtStageBoundary`, which this change doesn't support--I assume deliberately so, since that would make this even more complex. The version on Catalina doesn't support that at all. So I assume it isn't necessary.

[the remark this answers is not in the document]
cdavis5e: Does setting `withBarrier: YES` help with the accuracy? According to the Vulkan spec:

> `vkCmdWriteTimestamp` latches the value of the timer *when all previous commands have completed executing* as far as the specified pipeline stage, and writes the timestamp value to memory. [emphasis added]

`withBarrier: YES` inserts a barrier that waits for the previous commands to finish, which is what we want here.

cdavis5e:
> > I'm sorry, but this change has serious issues. Aside from the build breakages, for one thing, it doesn't handle `vkCmdCopyQueryResultsToBuffer()` for timestamp queries.
> 
> How does it need to handle it? Ultimately the only thing I changed regarding queries is the location we retrieve the time from, rather than getting the CPU time we now read the counter sample buffer. I imagined the existing query functionality will work the same.

Upon actually reading the code already there, you are correct, sir. I'm sorry.

What I *really* wanted to say was, we should be able to take advantage of `-[MTLBlitCommandEncoder resolveCounters:inRange:destinationBuffer:destinationOffset:]`, and avoid going through the host. If the results don't use an 8-byte stride, though, we'll need an intermediate buffer and a shader. But we should be able to do that in a later change... unless it's partially responsible for the accuracy problems you were seeing with sampling on each command.

MoltenVK/MoltenVK/Commands/MVKCmdTransfer.mm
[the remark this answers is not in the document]
mp-rsx: You are right that `recordTimestamp()` doesn't handle the case for multiple draw calls within the same render pass. That is an oversight.

As I mentioned creating a temporary encoder did not work for me. I was seeing invalid times (nested timestamp samples reporting higher times than their 'parent' for example). But I have not tried with the barrier option, as the docs state it might negatively impact performance.

mp-rsx:
Thanks for your input. It might be a while until I can get back to this task. I have since added a config option to disable this functionality (disabled by default) so you can perhaps merge it early without fear of it breaking something.

billhollings:
>It might be a while until I can get back to this task

Okay. Thanks for letting us know.

I don't want to officially pull this in until it's ready, because people will start enabling the flag and potentially need support.

I'll mark this as WIP for now. I might pick it up myself if you can't get back to it for a while.

mp-rsx:
I've made some further changes:
 - Moved the timestamp buffer from command buffer to command encoder
 - Added the missing timestamp call when recording a timestamp in the middle of render pass
 - Enabled barrier for samples as it results in more precise timings
 - Added some error handling to prevent issues with potentially invalid GPU data
 - Cleaned up some code as requested in the PR and resolved the conflict

billhollings:
Closing this, because I believe the functionality defined by this PR was included independently in 6ae1745.

If there are areas of this PR that can add to that functionality, we can reopen, or just poach as appropriate.

PR #1284, opened 2021-02-25
Ensure queue submission message logging occurs before submission object is destroyed.

billhollings:
Recent enhancements to shader logging triggered `MTLCommandBuffer` completion handler race condition.

Fixes to `MVKQueueCommandBufferSubmission`:
- Consolidate `MTLCommandBuffer` completion callbacks, and ensure all other activities,
  including logging, which references the object, occurs before `finish()` is called.
- Remove separate flag that indicates whether to track `MTLCommandBuffer` completion times.
- Localize Vulkan error reporting to the queue instead of the device.
- Reduce debug log noise by only logging header msg for shader debug logs if at least one shader log exists.

### DIFF
--- a/MoltenVK/MoltenVK/GPUObjects/MVKQueue.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKQueue.mm
@@ -358,8 +358,12 @@ void MVKQueueCommandBufferSubmission::commitActiveMTLCommandBuffer(bool signalCo
 		}
 #if MVK_XCODE_12
 		if (mvkGetMVKConfiguration()->debugMode) {
-			MVKLogInfo("Shader log messages:");
+			bool isFirstMsg = true;
 			for (id<MTLFunctionLog> log in mtlCB.logs) {
+				if (isFirstMsg) {
+					MVKLogInfo("Shader log messages:");
+					isFirstMsg = false;
+				}
 				MVKLogInfo("%s", log.description.UTF8String);
 			}
 		}
--- a/MoltenVK/MoltenVK/GPUObjects/MVKQueue.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKQueue.mm
@@ -321,11 +321,11 @@ void MVKQueueCommandBufferSubmission::commitActiveMTLCommandBuffer(bool signalCo
 	id<MTLCommandBuffer> mtlCmdBuff = signalCompletion ? getActiveMTLCommandBuffer() : _activeMTLCommandBuffer;
 	_activeMTLCommandBuffer = nil;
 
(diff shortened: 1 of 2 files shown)


MoltenVK/MoltenVK/GPUObjects/MVKQueue.mm
```diff
@@ -315,25 +313,19 @@
...
+	// Use temp var for MTLCommandBuffer commit and release because completion callback
+	// may destroy this instance before this function ends.
+	id<MTLCommandBuffer> mtlCmdBuff = signalCompletion ? getActiveMTLCommandBuffer() : _activeMTLCommandBuffer;
+	_activeMTLCommandBuffer = nil;
 
-	// Use temp vars because callback may destroy this instance before this function ends.
-	MVKDevice* device = _queue->getDevice();
-	id<MTLCommandBuffer> mtlCmdBuff = _activeMTLCommandBuffer;
-	// If command buffer execution fails, log it, and mark the device lost.
+	MVKDevice* mkvDev = _queue->getDevice();
```
cdavis5e, line 324: Typo 'mkvDev'

MoltenVK/MoltenVK/GPUObjects/MVKQueue.mm
```diff
@@ -366,15 +358,22 @@
 		}
 #if MVK_XCODE_12
 		if (mvkGetMVKConfiguration()->debugMode) {
-			MVKLogInfo("Shader log messages:");
+			bool isFirstMsg = true;
 			for (id<MTLFunctionLog> log in mtlCB.logs) {
+				if (isFirstMsg) {
+					MVKLogInfo("Shader log messages:");
```
cdavis5e, line 364: So that's why this was crashing when I didn't use the device to log. I remember changing it to use the device, but I made that change (and the initial change) on another computer, and I excitedly made a patch to copy back in order to submit it *before* I made those changes, so the change I submitted here was still broken.

cdavis5e (approved):
With a minor nit, and me musing on my own folly once again...

MoltenVK/MoltenVK/GPUObjects/MVKQueue.mm
billhollings: Good eye! I think that was already there. But fixed now anyway.

MoltenVK/MoltenVK/GPUObjects/MVKQueue.mm
billhollings: Yeah. `MVKLogInfo()` calls the `reportMessage()` function on the current object, so it needs to be alive.

Issue #1279, opened 2021-02-24
Rely on official include directory of spirv-cross


SpaceIm:
Would it be possible to use official include directory of `spirv-cross`?
https://github.com/KhronosGroup/SPIRV-Cross/blob/621884d70917038caf7509f7b1b3c143807ff43f/CMakeLists.txt#L171
https://github.com/KhronosGroup/SPIRV-Cross/blob/621884d70917038caf7509f7b1b3c143807ff43f/CMakeLists.txt#L194

Basically I had to apply this patch in conan (package manager) recipe of MoltenVK 1.1.1 (we don't rely on fetchDependencies to avoid vendoring and downloads at build time, I had to write a generic CMakeLists from scrach for MoltenVK but that's an other story):

```patch
--- a/MoltenVK/MoltenVK/GPUObjects/MVKPixelFormats.h
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKPixelFormats.h
@@ -22,7 +22,7 @@
 #include "MVKEnvironment.h"
 #include "MVKOSExtensions.h"
 #include "MVKBaseObject.h"
-#include <SPIRV-Cross/spirv_msl.hpp>
+#include <spirv_msl.hpp>
 #include <unordered_map>
 
 #import <Metal/Metal.h>
--- a/MoltenVKShaderConverter/Common/SPIRVSupport.cpp
+++ b/MoltenVKShaderConverter/Common/SPIRVSupport.cpp
@@ -18,7 +18,7 @@
 
 #include "SPIRVSupport.h"
 #include "MVKStrings.h"
-#include <SPIRV-Cross/spirv.hpp>
+#include <spirv.hpp>
 #include <ostream>
 
 #import <CoreFoundation/CFByteOrder.h>
--- a/MoltenVKShaderConverter/MoltenVKShaderConverter/SPIRVReflection.h
+++ b/MoltenVKShaderConverter/MoltenVKShaderConverter/SPIRVReflection.h
@@ -19,10 +19,10 @@
 #ifndef __SPIRVReflection_h_
 #define __SPIRVReflection_h_ 1
 
-#include <SPIRV-Cross/spirv.hpp>
-#include <SPIRV-Cross/spirv_common.hpp>
-#include <SPIRV-Cross/spirv_parser.hpp>
-#include <SPIRV-Cross/spirv_reflect.hpp>
+#include <spirv.hpp>
+#include <spirv_common.hpp>
+#include <spirv_parser.hpp>
[...]

PR #1287, opened 2021-03-01
Remove project qualifiers from references to SPIRV-Cross and glslang header files.

billhollings:
Remove `SPIRV-Cross/` qualifier from include references to SPIRV-Cross header files.
Remove `glslang/` qualifier from include references to glslang header files.
This change allows easier integration with app build scripts.

Fixes issue #1279.

### DIFF
--- a/MoltenVKShaderConverter/MoltenVKShaderConverter.xcodeproj/project.pbxproj
+++ b/MoltenVKShaderConverter/MoltenVKShaderConverter.xcodeproj/project.pbxproj
@@ -625,7 +625,7 @@
 				GCC_WARN_UNUSED_FUNCTION = YES;
 				HEADER_SEARCH_PATHS = (
 					"$(inherited)",
-					"\"$(SRCROOT)\"",
+					"\"$(SRCROOT)/SPIRV-Cross\"",
 					"\"$(SRCROOT)/glslang\"",
 					"\"$(SRCROOT)/glslang/External/spirv-tools/include\"",
 				);
@@ -680,7 +680,7 @@
 				GCC_WARN_UNUSED_FUNCTION = YES;
 				HEADER_SEARCH_PATHS = (
 					"$(inherited)",
-					"\"$(SRCROOT)\"",
+					"\"$(SRCROOT)/SPIRV-Cross\"",
 					"\"$(SRCROOT)/glslang\"",
 					"\"$(SRCROOT)/glslang/External/spirv-tools/include\"",
 				);
--- a/MoltenVKShaderConverter/MoltenVKShaderConverter.xcodeproj/project.pbxproj
+++ b/MoltenVKShaderConverter/MoltenVKShaderConverter.xcodeproj/project.pbxproj
@@ -625,6 +625,7 @@
 				GCC_WARN_UNUSED_FUNCTION = YES;
 				HEADER_SEARCH_PATHS = (
 					"$(inherited)",
+					"\"$(SRCROOT)\"",
 					"\"$(SRCROOT)/SPIRV-Cross\"",
 					"\"$(SRCROOT)/glslang\"",
 					"\"$(SRCROOT)/glslang/External/spirv-tools/include\"",
@@ -680,6 +681,7 @@
 				GCC_WARN_UNUSED_FUNCTION = YES;
 				HEADER_SEARCH_PATHS = (
 					"$(inherited)",
+					"\"$(SRCROOT)\"",
 					"\"$(SRCROOT)/SPIRV-Cross\"",
 					"\"$(SRCROOT)/glslang\"",
 					"\"$(SRCROOT)/glslang/External/spirv-tools/include\"",
--- a/MoltenVKShaderConverter/MoltenVKShaderConverter/GLSLToSPIRVConverter.cpp
+++ b/MoltenVKShaderConverter/MoltenVKShaderConverter/GLSLToSPIRVConverter.cpp
@@ -21,7 +21,7 @@
 #include "SPIRVToMSLConverter.h"
 #include "SPIRVSupport.h"
 #include "MVKStrings.h"
-#include <glslang/SPIRV/GlslangToSpv.h>
+#include <SPIRV/GlslangToSpv.h>
 #include <sstream>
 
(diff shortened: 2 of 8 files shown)


SpaceIm:
@billhollings `glslang` qualifier shouldn't be removed.

glslang installs its public headers under <include install folder>/glslang, and defines <include install folder> as include directory: https://github.com/KhronosGroup/glslang/blob/e04a046ce7c2ce6d37cc7a28802c167d84ffabf3/glslang/CMakeLists.txt#L174

MoltenVKShaderConverter/MoltenVKShaderConverter.xcodeproj/project.pbxproj
```diff
@@ -680,7 +680,7 @@
 				GCC_WARN_UNUSED_FUNCTION = YES;
 				HEADER_SEARCH_PATHS = (
 					"$(inherited)",
-					"\"$(SRCROOT)\"",
+					"\"$(SRCROOT)/SPIRV-Cross\"",
```
SpaceIm, line 683:
```suggestion
					"\"$(SRCROOT)\"",
					"\"$(SRCROOT)/SPIRV-Cross\"",
```

MoltenVKShaderConverter/MoltenVKShaderConverter/GLSLToSPIRVConverter.cpp
```diff
@@ -21,7 +21,7 @@
 #include "SPIRVToMSLConverter.h"
 #include "SPIRVSupport.h"
 #include "MVKStrings.h"
-#include <glslang/SPIRV/GlslangToSpv.h>
+#include <SPIRV/GlslangToSpv.h>
```
SpaceIm, line 24:
```suggestion
#include <glslang/SPIRV/GlslangToSpv.h>
```

SpaceIm (commented):
@billhollings `glslang` qualifier shouldn't be removed.

glslang installs its public headers under `<include install dir>/glslang`, and defines `<include install dir>` as include directory:
https://github.com/KhronosGroup/glslang/blob/e04a046ce7c2ce6d37cc7a28802c167d84ffabf3/glslang/CMakeLists.txt#L174
https://github.com/KhronosGroup/glslang/blob/e04a046ce7c2ce6d37cc7a28802c167d84ffabf3/glslang/CMakeLists.txt#L215-L224

Otherwise, LGTM for `spirv-cross`

MoltenVKShaderConverter/MoltenVKShaderConverter/GLSLToSPIRVConverter.cpp
billhollings: I'm not sure I properly understand the distinction between the two environments now.

For both `glslang` and `SPIRV-Cross`, pulling the Git repo, creating a `mybuild` directory inside it (or somewhere else), and running `cmake` and `make` into it, leaves all the headers relative to the base directory of the respective Git repos. 

So to be consistent, we now point MoltenVK header paths to those base directories: `glslang` and `SPIRV-Cross`, respectively. And MoltenVK `include` statements are now relative to those base directories.

Help me understand what I'm missing about your integrated build environment. How does this work there? Why does the header path for `SPIRV-Cross` point to its Git repo directory, but the header path for `glslang` point to the parent directory of its Git repo directory?

SpaceIm: Build folder layout and official include directories of a lib are orthogonal.

I rely on installation folders and official CMake config files installed with those libs which (should) contain the truth about how to consume these libraries, not build folder which is just an intermediate stuff.

The layout of build subfolders might be completly different from installation layout. Package managers rely on installation layout, and actually they rely on pkg-config/CMake config files information if any (or just a README explaining how to include headers, or just a consensus in the ecosystem).

In my build environment I use a package manager (conan) for my dependencies.
I've created a package for MoltenVK in public repository of conan, which has removed all the vendoring mechanism of MoltenVK because it's a "no go" (or something to remove quickly to avoid side effects like ODR violation, non reproducible builds etc) for a package manager.
Therefore in this context, dependencies of MoltenVK are installed by this package manager, which honors official install layout of each lib. This is why I had to patch includes of spirv-cross in MoltenVK.

The official way to include:
- `glslang` public headers is not `<SPIRV/GlslangToSpv.h>` but `<glslang/SPIRV/GlslangToSpv.h>` (the truth is in CMake config file installed with glslang).
- `spirv-cross` public headers is not `#include <SPIRV-Cross/spirv.hpp>` (this folder doesn't exist actually when you build and install yourself spirv-cross, it's `spirv_cross` AFAIK), not even `#include <spirv_cross/spirv.hpp>` but `#include <spirv.hpp>`.
  So here for example you see that even if they have added a "namespaced folder", authors of spirv-cross have decided that this folder will be useless in a consuming point of view.

Even install layout might be misleading, the truth (include direcrtories, compile flags, dependencies etc) of CMake based projects is in CMake config or pkg-config file installed along headers and binaries.

MoltenVKShaderConverter/MoltenVKShaderConverter.xcodeproj/project.pbxproj
billhollings: Done.

MoltenVKShaderConverter/MoltenVKShaderConverter/GLSLToSPIRVConverter.cpp
billhollings: Done.

cdavis5e:
Well, I was waiting for @SpaceIm, so 🤷‍♂️ .

billhollings:
>Well, I was waiting for @SpaceIm, so 🤷‍♂️ .

Sorry...I was trying to re-request @SpaceIm and ended up taking you off by mistake in my fumbling. The re-request button wasn't taking for them. Maybe because they hadn't been originally requested? Meh.

SpaceIm:
LGTM 👍

cdavis5e (approved):
Formal approval. :)

PR #1290, opened 2021-03-03
Remove LunarG/VulkanSamples as a dependency library.

billhollings:
- LunarG is planning to retire the `LunarG/VulkanSamples` repository.
- Remove dependency to it, and remove the affected `Hologram` and `API-Samples` demo apps.
- Update documents, including directing developers to the `KhronosGroup/Vulkan-Samples` repository for official demo apps.


cdavis5e (changes requested):
You used the `main.m` from the deleted `API-Samples` project for the Cube. Is that correct?

billhollings:
>You used the main.m from the deleted API-Samples project for the Cube. Is that correct?

Wow! You definitely have eagle eyes!

But no. I just copied the Cube directory, whole cloth, up a level. So the two `main.m` in Cube are the same files as they were before.

I think this must be a Git artifact. Since all the `main.m` files were exactly the same per platform, I guess `git add` must have just found the first one that matches to assign to the rename task, and marked the others as deleted. Interesting.

In any case, the files are there, unchanged in content, and the demo builds and runs on all platforms.

cdavis5e (approved):
That figures.

PR #1295, opened 2021-03-04
Add NDEBUG macro to all Release builds to remove assert() calls.

billhollings:
External libraries in particular make liberal use of `assert()` calls, which bypass
error catching, sometimes causing crashes rather than catch errors and moving on.


cdavis5e (approved):
We actually do this ourselves.

billhollings:
>We actually do this ourselves.

By "we", I assume you mean as a MoltenVK build user, as opposed to MoltenVK contributor? 😉

Yeah...embedding it directly in MoltenVK will handle this for the binary users too (although maybe the SDK was already doing the same).

I ran into the issue via CTS crashes.

PR #1297, opened 2021-03-08
Extend MVKConfiguration::advertiseExtensions to optionally specify a few extensions.

billhollings:
Introduce `MVKConfigAdvertiseExtensionBits` enum to specify multiple API config values in
a Vulkan-friendly manner, while automatically documenting the same values for env vars.

### DIFF
--- a/MoltenVK/MoltenVK/API/vk_mvk_moltenvk.h
+++ b/MoltenVK/MoltenVK/API/vk_mvk_moltenvk.h
@@ -58,6 +58,18 @@ typedef unsigned long MTLLanguageVersion;
 #define VK_MVK_MOLTENVK_SPEC_VERSION            31
 #define VK_MVK_MOLTENVK_EXTENSION_NAME          "VK_MVK_moltenvk"
 
+/** Identifies extensions to advertise as part of MoltenVK configuration. */
+typedef enum MVKConfigAdvertiseExtensionBits {
+	MVK_CONFIG_ADVERTISE_EXTENSIONS_ALL                       = 0x00000001,
+	MVK_CONFIG_ADVERTISE_EXTENSIONS_VK_MVK_MOLTENVK           = 0x00000002,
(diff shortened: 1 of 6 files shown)


MoltenVK/MoltenVK/API/vk_mvk_moltenvk.h
```diff
@@ -58,6 +58,18 @@ typedef unsigned long MTLLanguageVersion;
 #define VK_MVK_MOLTENVK_SPEC_VERSION            31
 #define VK_MVK_MOLTENVK_EXTENSION_NAME          "VK_MVK_moltenvk"
 
+/** Identifies extensions to advertise as part of MoltenVK configuration. */
+typedef enum MVKConfigAdvertiseExtensionBits {
+	MVK_CONFIG_ADVERTISE_EXTENSIONS_ALL                       = 0x00000001,
+	MVK_CONFIG_ADVERTISE_EXTENSIONS_VK_MVK_MOLTENVK           = 0x00000002,
+	MVK_CONFIG_ADVERTISE_EXTENSIONS_VK_MVK_IOS_SURFACE        = 0x00000004,
+	MVK_CONFIG_ADVERTISE_EXTENSIONS_VK_MVK_MACOS_SURFACE      = 0x00000008,
+	MVK_CONFIG_ADVERTISE_EXTENSIONS_VK_EXT_METAL_SURFACE      = 0x00000010,
```
cdavis5e, line 67: What about `VK_KHR_swapchain`?

cdavis5e (approved):

MoltenVK/MoltenVK/API/vk_mvk_moltenvk.h
billhollings: Good point. Strict prerequisites are not necessarily enough.

I've redefined the enum in terms of usage families (WSI, Portability), which I think is closer to how the tests and other users would think of enabling them. And now `KHR_swapchain` is advertised when the config asks for WSI extensions.

cdavis5e (approved):

PR #1312, opened 2021-03-13
MVKDescriptor subclasses: Reduce redundant and unused content.

billhollings:
`MVKDescriptor` subclasses: Simplify interface and reduce redundant content.

Remove `MVKDescriptor::setLayout()` and pass `MVKDescriptorSetLayoutBinding`
to `MVKDescriptor::bind(), write(), and read()`.
Remove `MVKSamplerDescriptorMixin::_hasDynamicSampler`.
Remove `MVKInlineUniformBlockDescriptor::_length`.

`MVKDescriptor` subclasses: Reduce unused content.
Don't populate `MVKMTLBufferBinding::size` unless using `mtlBytes` and `isInline`.
Remove `MVKBufferDescriptor::_buffRange`.
Remove `MVKImageDescriptor::_imageLayout`.
Remove `MVKCmdBufferImageCopy::_imageLayout`.

Another prep step compatible with argument buffers.

### DIFF
--- a/MoltenVK/MoltenVK/GPUObjects/MVKDescriptor.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKDescriptor.mm
@@ -161,6 +161,10 @@ void MVKDescriptorSetLayoutBinding::push(MVKCommandEncoder* cmdEncoder,
                 MVKBuffer* buffer = (MVKBuffer*)bufferInfo.buffer;
                 bb.mtlBuffer = buffer->getMTLBuffer();
                 bb.offset = buffer->getMTLBufferOffset() + bufferInfo.offset;
+                if (bufferInfo.range == VK_WHOLE_SIZE)
+                    bb.size = (uint32_t)(buffer->getByteCount() - bb.offset);
+                else
+                    bb.size = (uint32_t)bufferInfo.range;
 
                 for (uint32_t i = kMVKShaderStageVertex; i < kMVKShaderStageMax; i++) {
                     if (_applyToStage[i]) {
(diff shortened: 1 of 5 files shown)


MoltenVK/MoltenVK/GPUObjects/MVKDescriptor.mm
```diff
@@ -161,10 +161,6 @@
                 MVKBuffer* buffer = (MVKBuffer*)bufferInfo.buffer;
                 bb.mtlBuffer = buffer->getMTLBuffer();
                 bb.offset = buffer->getMTLBufferOffset() + bufferInfo.offset;
-                if (bufferInfo.range == VK_WHOLE_SIZE)
-                    bb.size = (uint32_t)(buffer->getByteCount() - bb.offset);
-                else
-                    bb.size = (uint32_t)bufferInfo.range;
```
cdavis5e, line 167: Wait, isn't this used if the shader uses `OpArrayLength`? Admittedly, this is non-obvious from the surrounding code. You'd have to go to `Commands/MVKCommandEncoderState.mm` to see that.

cdavis5e (changes requested):
One possible problem.

MoltenVK/MoltenVK/GPUObjects/MVKDescriptor.mm
billhollings: Ah. Thanks for saving me from myself. I thought I'd traced it through...so good catch! Reverted now.

cdavis5e (approved):

PR #1319, opened 2021-03-15
Fix advertising single-texel alignment capability for texel buffers.

### LINKED ISSUE #1318: Texel buffer alignment validation error
CTS encountered the following Metal validation error:

    validateNewTexture:74: failed assertion `Offset of a buffer-backed texture with pixelFormat(MTLPixelFormatRGBA8Unorm) must be aligned to 16 bytes, found offset(4)'

Looks like the texel buffer alignment we are reporting might not cover everything. Specifically, the validation error message seems to indicate that single-texel alignment, as we report, is not always supported.

Looking into it.

billhollings:
The test for single-texel capability used a mod calculation to compare the linear
alignment requirement reported by Metal to the texel size. Instead, it should use an
inequality comparison, to check if the alignment requirement is LTE the texel size.

Fixes #1318.


cdavis5e (approved):
I don't know what I was thinking with that... Probably something about 3-component formats, which Metal doesn't even support.

billhollings:
> Probably something about 3-component formats, which Metal doesn't even support.

Yeah, that Vulkan 3-byte rule seems somewhat confusing and pedantic, given real-world alignment requirements, and the fact that this response has to cover ALL formats in the system.

PR #1323, opened 2021-03-22
Enhancements and fixes to temporary MTLBuffer allocation.

billhollings:
Check allocations are larger than `MTLBuffer` alignment to avoid Metal errors.
Align returning safely with allocating safely, and change callers to use
`MVKMTLBufferAllocation::returnToPool()` globally, to ensure this match.
Fix use of const on `returnToPool()`;
Move `MVKCommandEncoder::copyToTempMTLBufferAllocation()`
and support setting private and dedicated flags.
Align constructor parameter order on `MVKMTLBufferAllocationPool`
with that of `MVKMTLBufferAllocator`.
Move subclass overrides of `MVKObjectPool::newObject()` to protected access.

### DIFF
--- a/MoltenVK/MoltenVK/Commands/MVKCommandBuffer.h
+++ b/MoltenVK/MoltenVK/Commands/MVKCommandBuffer.h
@@ -373,7 +373,7 @@ public:
     const MVKMTLBufferAllocation* getTempMTLBuffer(NSUInteger length, bool isPrivate = false, bool isDedicated = false);
 
 	/** Copy the bytes to a temporary MTLBuffer that will be returned to a pool after the command buffer is finished. */
-	const MVKMTLBufferAllocation* copyToTempMTLBufferAllocation(const void* bytes, NSUInteger length, bool isPrivate = false, bool isDedicated = false);
+	const MVKMTLBufferAllocation* copyToTempMTLBufferAllocation(const void* bytes, NSUInteger length, bool isDedicated = false);
 
     /** Returns the command encoding pool. */
     MVKCommandEncodingPool* getCommandEncodingPool();
--- a/MoltenVK/MoltenVK/Commands/MVKCommandBuffer.h
+++ b/MoltenVK/MoltenVK/Commands/MVKCommandBuffer.h
@@ -372,6 +372,9 @@ public:
     /** Get a temporary MTLBuffer that will be returned to a pool after the command buffer is finished. */
     const MVKMTLBufferAllocation* getTempMTLBuffer(NSUInteger length, bool isPrivate = false, bool isDedicated = false);
 
+	/** Copy the bytes to a temporary MTLBuffer that will be returned to a pool after the command buffer is finished. */
+	const MVKMTLBufferAllocation* copyToTempMTLBufferAllocation(const void* bytes, NSUInteger length, bool isPrivate = false, bool isDedicated = false);
+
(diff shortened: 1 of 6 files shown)


MoltenVK/MoltenVK/Commands/MVKCommandBuffer.h
```diff
@@ -372,6 +372,9 @@ class MVKCommandEncoder : public MVKBaseDeviceObject {
     /** Get a temporary MTLBuffer that will be returned to a pool after the command buffer is finished. */
     const MVKMTLBufferAllocation* getTempMTLBuffer(NSUInteger length, bool isPrivate = false, bool isDedicated = false);
 
+	/** Copy the bytes to a temporary MTLBuffer that will be returned to a pool after the command buffer is finished. */
+	const MVKMTLBufferAllocation* copyToTempMTLBufferAllocation(const void* bytes, NSUInteger length, bool isPrivate = false, bool isDedicated = false);
```
cdavis5e, line 376: I don't think either of these makes sense here. You wouldn't be able to access a `Private` buffer from the CPU to populate it in the first place, and `isDedicated` is only needed right now for the very particular use case of visibility result buffers. I guess adding `isDedicated` for future-proofing maybe might make sense, if it turns out to be needed for something, but I don't see `isPrivate` ever being useful here.

MoltenVK/MoltenVK/Commands/MVKMTLBufferAllocation.h
```diff
@@ -100,6 +102,7 @@ class MVKMTLBufferAllocationPool : public MVKObjectPool<MVKMTLBufferAllocation>
     MTLStorageMode _mtlStorageMode;
 	MVKSmallVector<id<MTLBuffer>, 64> _mtlBuffers;
     MVKDevice* _device;
+	bool _isThreadSafe;
```
cdavis5e, line 105: This isn't used. I don't see why we need this.

cdavis5e (changes requested):
A few minor quibbles.

MoltenVK/MoltenVK/Commands/MVKCommandBuffer.h
billhollings: Doh! 🤦🏻‍♂️ Brain fart on my part about `isPrivate`. Good catch. I've left the `isDedicated` for future options.

MoltenVK/MoltenVK/Commands/MVKMTLBufferAllocation.h
billhollings: It's used a few lines above there in `returnAllocation()`. It handles the whether to return/return-safely, and is the compliment to the same flag that handles acquire/aquire-safely in `MVKMTLBufferAllocator`

billhollings:
Fixes in place. 

Also, I added some refactoring to use `MVKDeviceTrackingMixin` in the interests of consistent frameworking.

cdavis5e:
I don't see any changes.

MoltenVK/MoltenVK/Commands/MVKMTLBufferAllocation.h
cdavis5e: Oh, sorry. How did I miss that? Normally, I'm more observant than that...

cdavis5e (approved):

PR #1331, opened 2021-04-12
Add MVKConfiguration::resumeLostDevice and MVK_CONFIG_RESUME_LOST_DEVICE env var.

billhollings:
Add `MVKConfiguration::resumeLostDevice` and `MVK_CONFIG_RESUME_LOST_DEVICE` env var,
to allow VkDevice to resume after non-fatal VK_ERROR_DEVICE_LOST error.

When running CTS, when one test triggers a `VK_ERROR_DEVICE_LOST` error condition,
CTS does not create a new `VkDevice` for subsequent tests, which causes all subsequent
tests to fail. This creates an incredible amount of results noise, and makes it
difficult to determine accurate test failure numbers. This is particularly problematic
because `VK_ERROR_DEVICE_LOST` may be triggered sporadically by internal Metal issues,
often making it not traceable to a consistent failure of a particular CTS test.

### DIFF
--- a/MoltenVK/MoltenVK/Utility/MVKEnvironment.h
+++ b/MoltenVK/MoltenVK/Utility/MVKEnvironment.h
@@ -272,3 +272,8 @@ void mvkSetConfig(MVKConfiguration* pMVKConfig);
 #ifndef MVK_CONFIG_ADVERTISE_EXTENSIONS
 #  	define MVK_CONFIG_ADVERTISE_EXTENSIONS    MVK_CONFIG_ADVERTISE_EXTENSIONS_ALL
 #endif
+
+/** Support Metal argument buffers. Disabled by default. */
+#ifndef MVK_CONFIG_RESUME_LOST_DEVICE
+#   define MVK_CONFIG_RESUME_LOST_DEVICE    0
+#endif
--- a/MoltenVK/MoltenVK/Utility/MVKEnvironment.h
+++ b/MoltenVK/MoltenVK/Utility/MVKEnvironment.h
@@ -273,7 +273,7 @@ void mvkSetConfig(MVKConfiguration* pMVKConfig);
 #  	define MVK_CONFIG_ADVERTISE_EXTENSIONS    MVK_CONFIG_ADVERTISE_EXTENSIONS_ALL
 #endif
 
-/** Support Metal argument buffers. Disabled by default. */
+/** Resume MVKDevice VK_ERROR_DEVICE_LOST errors that do not cause MVKPhysicalDevice errors. Disabled by default. */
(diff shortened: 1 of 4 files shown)


MoltenVK/MoltenVK/Utility/MVKEnvironment.h
```diff
@@ -272,3 +272,8 @@ void mvkSetConfig(MVKConfiguration* pMVKConfig);
 #ifndef MVK_CONFIG_ADVERTISE_EXTENSIONS
 #  	define MVK_CONFIG_ADVERTISE_EXTENSIONS    MVK_CONFIG_ADVERTISE_EXTENSIONS_ALL
 #endif
+
+/** Support Metal argument buffers. Disabled by default. */
```
cdavis5e, line 276: Copy-pasto.

cdavis5e (changes requested):
I'm now wondering if I were a bit overzealous mapping all Metal command buffer errors to `VK_ERROR_DEVICE_LOST`.

MoltenVK/MoltenVK/Utility/MVKEnvironment.h
billhollings: Oops. Fixed.

billhollings:
> I'm now wondering if I were a bit overzealous mapping all Metal command buffer errors to `VK_ERROR_DEVICE_LOST`.

Spec is pretty broad about this. As mentioned in my comments, it really leaves everything open to interpretation:

>Typical reasons for device loss will include things like execution timing out (to prevent denial of service), power management events, platform resource management, implementation errors.

>Applications not adhering to valid usage may also result in device loss being reported, however this is not guaranteed. Even if device loss is reported, the system may be in an unrecoverable state, and further usage of the API is still considered invalid.
When this happens, certain commands will return `VK_ERROR_DEVICE_LOST`. After any such event, the logical device is considered lost. It is not possible to reset the logical device to a non-lost state, however the lost state is specific to a logical device (`VkDevice`), and the corresponding physical device (VkPhysicalDevice) may be otherwise unaffected.

TBH...it's not clear to me what the Vulkan concept of re-creating a `VkDevice` without touching the `VkPhysicalDevice` means in the context of `MTLDevice`.

It might make sense to make sense to enable `MVK_CONFIG_RESUME_LOST_DEVICE` by default. Looking at the list of `MTLCommandBufferError` options, most of the remaining ones not covered by your `VkPhysicalDevice` propagation would seem to be ultimately recoverable. Worst case we could tweak the separation between `VkDevice` and `VkPhysicalDevice` errors, in case there is some situation where recreating the `VkDevice` might make sense.

cdavis5e (approved):

PR #1342, opened 2021-04-27
Expose vkGetIOSurfaceMVK() / vkUseIOSurfaceMVK() also under non-ObjC.

kleinerm:
Those functions are also usable from standard C, no need for the caller
to be compiled as Objective-C. Move their definition outside the __OBJC__
guards.

### DIFF
--- a/MoltenVK/MoltenVK/API/vk_mvk_moltenvk.h
+++ b/MoltenVK/MoltenVK/API/vk_mvk_moltenvk.h
@@ -27,10 +27,10 @@ extern "C" {
 #endif	//  __cplusplus
 	
 #include "mvk_vulkan.h"
+#include <IOSurface/IOSurfaceRef.h>
 
 #ifdef __OBJC__
 #import <Metal/Metal.h>
-#import <IOSurface/IOSurfaceRef.h>
 #else
 typedef unsigned long MTLLanguageVersion;
 #endif
@@ -947,14 +947,14 @@ typedef VkResult (VKAPI_PTR *PFN_vkSetMoltenVKConfigurationMVK)(VkInstance ignor
 typedef VkResult (VKAPI_PTR *PFN_vkGetPhysicalDeviceMetalFeaturesMVK)(VkPhysicalDevice physicalDevice, MVKPhysicalDeviceMetalFeatures* pMetalFeatures, size_t* pMetalFeaturesSize);
 typedef VkResult (VKAPI_PTR *PFN_vkGetPerformanceStatisticsMVK)(VkDevice device, MVKPerformanceStatistics* pPerf, size_t* pPerfSize);
 typedef void (VKAPI_PTR *PFN_vkGetVersionStringsMVK)(char* pMoltenVersionStringBuffer, uint32_t moltenVersionStringBufferLength, char* pVulkanVersionStringBuffer, uint32_t vulkanVersionStringBufferLength);
+typedef VkResult (VKAPI_PTR *PFN_vkUseIOSurfaceMVK)(VkImage image, IOSurfaceRef ioSurface);
+typedef void (VKAPI_PTR *PFN_vkGetIOSurfaceMVK)(VkImage image, IOSurfaceRef* pIOSurface);
 
 #ifdef __OBJC__
 typedef void (VKAPI_PTR *PFN_vkGetMTLDeviceMVK)(VkPhysicalDevice physicalDevice, id<MTLDevice>* pMTLDevice);
 typedef VkResult (VKAPI_PTR *PFN_vkSetMTLTextureMVK)(VkImage image, id<MTLTexture> mtlTexture);
 typedef void (VKAPI_PTR *PFN_vkGetMTLTextureMVK)(VkImage image, id<MTLTexture>* pMTLTexture);
 typedef void (VKAPI_PTR *PFN_vkGetMTLBufferMVK)(VkBuffer buffer, id<MTLBuffer>* pMTLBuffer);
-typedef VkResult (VKAPI_PTR *PFN_vkUseIOSurfaceMVK)(VkImage image, IOSurfaceRef ioSurface);
-typedef void (VKAPI_PTR *PFN_vkGetIOSurfaceMVK)(VkImage image, IOSurfaceRef* pIOSurface);
 #endif // __OBJC__
 
 
@@ -1199,6 +1199,8 @@ VKAPI_ATTR void VKAPI_CALL vkGetMTLBufferMVK(
     VkBuffer                                    buffer,
     id<MTLBuffer>*                              pMTLBuffer);
 
+#endif // __OBJC__
+
 /**
  * Indicates that a VkImage should use an IOSurface to underlay the Metal texture.
  *
@@ -1247,8 +1249,6 @@ VKAPI_ATTR VkResult VKAPI_CALL vkUseIOSurfaceMVK(
 VKAPI_ATTR void VKAPI_CALL vkGetIOSurfaceMVK(
     VkImage                                     image,
     IOSurfaceRef*                               pIOSurface);
-
-#endif // __OBJC__
 
 
 #pragma mark -


billhollings:
>Those functions are also usable from standard C

Thanks for the submission. And good catch!

Interesting too. In your environment, are you actually using `IOSurface` without using anything in ObjC?

kleinerm:
My use case is Psychtoolbox-3, a very popular OSS toolkit for neuroscience research. I use Vulkan only as a presentation backend atm., to get access to the HDR/EDR capabilities of Metal and more importantly to the Metal timing/timestamping of Metal, as timing is very important for visual stimulation for vision science.

So Psychtoolbox renders and post-processes all stimulus images via Apples OpenGL, but then i use Vulkan / MoltenVK on top of Metal to present the final image. So essentially this if for OpenGL-Vulkan image sharing.

Let MoltenVK allocate an IOSurface as backing for the MTLTexture of the VkImage interop-image.
vkGetIOSurfaceMVK a handle to the IOSurface, and turn it into a OpenGL GL_TEXTURE_RECTANGLE texture.
Bind to an OpenGL FBO. My OpenGL renderer renders the final image into it, and thereby into the VkImage.
vkCmdBlitImage that image into the swapchain buffer. Present via the VK_GOOGLE_display_timing extension at a specific target time, collect timestamp.

See here for reference for how i use IOSurface:

https://github.com/Psychtoolbox-3/Psychtoolbox-3/blob/c9f1d26d2c51e9cdae19236dc55abeb0f6a67c34/PsychSourceGL/Source/Common/PsychVulkanCore/PsychVulkan.c#L2892

So i don't use Obj-C code for rendering into the IOSurface. That part is pure C + OpenGL.

But ofc. i do have to have one single compilation unit which is ObjC, because i need Cocoa / NSWindow to create a Cocoa window, with a layer-backed content view. The backing CAMetalLayer is passed to MoltenVK as usual, so Metal has something it can present to. As far as i know, those CAMetalLayer backed windows are the only way for Metal - and thereby MoltenVK - to present anything to the screen.

This toolkit is used as plugin for Matlab, GNU/Octave and Python, and due to the peculiarities of their build systems, having Obj-C compilation units in a otherwise pure C/C++ project is a bit of a headache. That's why i wanted to get rid of the need to have more Obj-C just for using those two MoltenVK functions. Atm. i define the function prototypes in my own code, but thanks to your acceptance of the pull request, i can drop that hack again :)

However, this should also simplify future work for Mesa's Gallium zink driver on macOS, which is an FOSS OpenGL-on-top-of-Vulkan implementation. The Mesa developers have already some partial support for macOS + MoltenVK. Good for replacing Apples OpenGL implementation if one needs something less ancient than what Apple ships, or if Apple would rip it out completely. Mesa is also not Obj-C at all at the moment, and tries to import vkGetIOSurfaceMVK / vkUseIOSurfaceMVK. The functions are not used yet, but they cause the build of zink on macOS to fail due to lack of ObjC support in Mesa's build system. I hope this will also help them a little.

billhollings:
Thanks for the comprehensive backstory! Fantastic explanation...and amazing environment! I'm glad to hear it's all hanging together for you.

kleinerm:
This followup is probably more than you wanted to know, and a bit "ranty", so apologies, but maybe it helps somebody else who tries to make sense of timing and timestamping in Metal / Vulkan on macOS, by documenting my quite underwhelming experience so far on Catalina:

The "hanging together" part is unfortunately not quite where i wanted it to be, the jury is still out... The main motivator was to access Metal's timed presentation scheduling, and even more critical for us, the timestamping of when a swapchain image was truly presented onscreen. Iow. the `MTLDrawable presentAtTime: when`, `MTLDrawable:addPresentedHandler`, and `MTLDrawable:presentedTime` api's. These are exposed by MoltenVK's VK_GOOGLE_display_timing extension.

I could only test this on macOS 10.15.7 on AMD and Intel gpu's (MBP Retina 15" 2017), but on that machine + OS combo, Metals timestamping and presentation scheduling proved to be very unreliable and underwhelming in performance, even now that i fixed MVK in that other pull request. As far as dozens of hours of testing and tinkering go, this is not a fault or bug of MoltenVK, but of Metal on Catalina itself. I also downloaded some of Apples "best practices" Metal sample XCode projects, traced their execution via the macOS Performance Tools - the Metal Performance Trace and Game trace templates, to watch application api calls, driver processing, gpu processing of command buffers, when drawable surfaces get scheduled for present, and actually presented in relation to vsync etc. If those went through the compositor/Windowserver or not, DirectToDisplay or not etc. I also used external measurement equipment and procedures of Psychtoolbox to compare Metal reported timestamps against ground truth etc. etc. I hacked Apples "best practices" Metal sample code to add presentAtTime, callbacks etc., to replicate the MoltenVK setup independently. Long story short, all the limtations and bugs i encountered with my Psychtoolbox+Vulkan+MoltenVK+Metal implementation also happened with the independent Apple Metal-only Obj-C sample code. This tells me we are not dealing with MoltenVK bugs or limitations, but macOS Metal brokeness.

Some highlights:

- On AMD, the reported presentedTime randomly goes from reasonable and reasonably accurate (seems to be a timestamp likely taken by the vsync hardware interrupt handler) to completely bogus: Sometimes reporting values that are a dozen milliseconds off wrt. special measurement hardware - and physically impossible. More often it reports a timestamp of zero, which according to Apples docs mean _"The property value is 0.0 if the drawable hasn’t been presented or if its associated frame was dropped."_ - Except this happens at random points in time, for random streaks of presentation (a few frames to a few hundred frames), while looking at the animation on the screen, presentation is just fine, and clearly the drawable has been presented and no frames were dropped at all, in direct contradiction to what Apple docs say a zero timestamp would mean. This happens both in windowed mode, fullscreen mode, and sometimes in DirectToDisplay fullscreen mode. It happens with the application in "steady state" with the animation running and no obvious trigger event, no other apps running, no other GUI activity, nobody touching keyboard or mouse...

- On Intel gpu's the same holds true, except all timestamps are off by a factor of 1000x ! Apparently somebody doesn't understand multiplication and division in Apples Intel driver team, and nobody ever tested this before release of Catalina.

- Sometimes presentation timestamps are not only zero for no reason, but also arrive out of sequence, the presented callback for frame i called after the callback for frame i+1 or i+2, which really throws off software that needs to wait for present completion of frame i before progressing to i+1, i+2. Sometimes this seems to get worse in DirectToDisplay mode or for longer time intervals between presents.

- Things get worse in DirectToDisplay mode. The frequency and severity of above problems. So by default i intentionally choose pixel formats for the the swapchain to prevent DirectToDisplay...

- But that doesn't matter, as DirectToDisplay mode does not seem to have the latency reduction benefit it is supposed to have, at least on Catalina. I see the same one frame extra latency in DtoD as in regular mode, as confirmed by using the Apple performance tools for tracing all application/windowserver/gpu driver/display hardware activity. So if you have an application that needs to wait for frame present completion of frame i before moving on to frame i+1, then that 1 frame extra lag will throttle any animation to half the video refresh rate: max 30 fps on a standard 60 Hz display. Tracing suggests that the WindowServer is still involved in triggering the actual present, even though it is supposed to be bypassed - the purpose of DirectToDisplay. Driver and gpu processing time for the server goes down from milliseconds to a few microseconds, but that doesn't matter because the server still runs on a fixed duty cycle, as if compositing a regular desktop GUI with multiple non-fullscreen application windows, despite there being no other application being on display that would need to be composited, just one DtoD fullscreen exclusive application on a pure single display setup. Things like Retina display vs. standard DPI display, scaled modes versus native modes didn't make any difference.

All taken together, timing and performance is unusable for our very specific purpose due to both performance issues and reliability issues. But as far as i can see, this is not a fault of MoltenVK at all, but of macOS / Metal. And this is so far only confirmed to be broken on Catalina. Also, on macOS 10, and on macOS 11 with AMD/Intel/NVidia gpu's we have clever ways of timing and timestamping just using OpenGL and some custom written kext that does some low-level trickery by directly accessing the gpu hardware. Unfortunately with Apple clamping down on 3rd party kexts, supposedly planning to remove support completely in future macOS versions, and with the introduction of the Apple proprietary M1 SoC, our old tricks either do not work at all (M1's gpu and their OpenGL-on-top-of-Metal implementation), or are expected to fail on Intel machines with a future macOS version.

My hope is that macOS 11 would be less broken and more usable in this area, or maybe at least on the ARM M1 Apple proprietary display hardware, but i currently lack the hardware and even more so the funding and time for further testing and investigation on Big Sur. At least MoltenVK gives us a way to utilize this functionality, in case Apple fixes (or fixed) their system in future macOS releases. And our main recommended target platform isn't macOS anyway, but desktop Linux, where these specific problems don't exist and we have excellent timing and control with the FOSS graphics and display stack since over a decade. I assume at some point one of our users will run this new Vulkan/WSI backend code on their macOS 11 or Apple M1 systems and then i'll see if things are as bad on those setups. So far the feature on our side is less than a month old, so time will tell...

End rant ;-)

PR #1344, opened 2021-04-28
Compliance fixes for VK_GOOGLE_display_timing extension implementation.

kleinerm:
vkGetPastPresentationTimingGOOGLE() does not have return status codes and proper
return values for the pPresentationTimingCount in/out parameter wrt. the specification.

The queuing of returned presentation timing records is also not in agreement with the
spec. This confuses client applications and requires awkward workarounds.

These two commits should fix this.

### DIFF
--- a/MoltenVK/MoltenVK/GPUObjects/MVKSwapchain.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKSwapchain.mm
@@ -432,7 +432,6 @@ VkResult MVKSwapchain::getPastPresentationTiming(uint32_t *pCount, VkPastPresent
 	if (pPresentationTimings == nullptr) {
 		*pCount = _presentHistoryCount;
 	} else {
-		uint32_t index = _presentHistoryHeadIndex;
 		uint32_t countRemaining = std::min(_presentHistoryCount, *pCount);
 		uint32_t outIndex = 0;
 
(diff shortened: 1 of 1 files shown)


MoltenVK/MoltenVK/GPUObjects/MVKSwapchain.mm
```diff
@@ -426,28 +426,37 @@ static inline CIE1931XY VkXYColorEXTToCIE1931XY(VkXYColorEXT xy) {
 VkResult MVKSwapchain::getPastPresentationTiming(uint32_t *pCount, VkPastPresentationTimingGOOGLE *pPresentationTimings) {
 	if (_device->getConfigurationResult() != VK_SUCCESS) { return _device->getConfigurationResult(); }
 
+	VkResult res = VK_SUCCESS;
+
+	// Not allowed, but handle it with some valid error code to avoid crash.
+	if (pCount == nullptr) { return VK_ERROR_OUT_OF_HOST_MEMORY; }
+
```
cdavis5e, line 433: I have a slight preference for not doing error checking if the Vulkan spec says some behavior `must:` or `must: not` happen.

cdavis5e (changes requested):
Looks good. Just one minor quibble.

MoltenVK/MoltenVK/GPUObjects/MVKSwapchain.mm
kleinerm: Thanks for the review. I'll drop that check and update the commit message accordingly.

kleinerm: Done.

cdavis5e (approved):

PR #1345, opened 2021-05-01
Support Xcode 12.5

billhollings:
Xcode 12.5 has added some additional build warnings and a correction to the API for Mac Catalyst.

- Fix build error when querying `MTLDevice` support for BC compression when building for Mac Catalyst. This build error seems to have been exposed in Xcode 12.5 through an API change. Add `mvkSupportsBCTextureCompression()` to manage use of `[MTLDevice supportsBCTextureCompression]` to query for `MTLDevice` support for BC compression across all platforms.
- Add tag names to anonymous structs.
- Suppress switch enum warns in SPIRV-Tools build.
- Support Xcode 12.5 build settings.

@cdavis5e The bulk of this PR are admin fixes that don't change behaviour. But can you check my logic on the refactoring of the use of `supportsBCTextureCompression`. I added the new `mvkSupportsBCTextureCompression()` function because the inline logic of when to call `supportsBCTextureCompression`, and what to do with the results was getting complicated when used inline in two different places.

### DIFF
--- a/MoltenVK/MoltenVK/Commands/MVKMTLResourceBindings.h
+++ b/MoltenVK/MoltenVK/Commands/MVKMTLResourceBindings.h
@@ -29,22 +29,22 @@ class MVKImage;
 
 
 /** Describes a MTLTexture resource binding. */
-typedef struct {
+typedef struct MVKMTLTextureBinding {
     union { id<MTLTexture> mtlTexture = nil; id<MTLTexture> mtlResource; }; // aliases
     uint32_t swizzle = 0;
 	uint16_t index = 0;
     bool isDirty = true;
 } MVKMTLTextureBinding;
 
-/** Describes a MTLSamplerState resource binding. */
-typedef struct {
+/** Describes a MTLSamplerState resource bindi MVKMTLSamplerStateBindingng. */
+typedef struct MVKMTLSamplerStateBinding {
     union { id<MTLSamplerState> mtlSamplerState = nil; id<MTLSamplerState> mtlResource; }; // aliases
     uint16_t index = 0;
     bool isDirty = true;
 } MVKMTLSamplerStateBinding;
 
-/** Describes a MTLBuffer resource binding. */
-typedef struct {
+/** Describes  MVKMTLBufferBindinga MTLBuffer resource binding. */
+typedef struct MVKMTLBufferBinding {
     union { id<MTLBuffer> mtlBuffer = nil; id<MTLBuffer> mtlResource; const void* mtlBytes; }; // aliases
     VkDeviceSize offset = 0;
     uint32_t size = 0;
@@ -54,7 +54,7 @@ typedef struct {
 } MVKMTLBufferBinding;
 
 /** Describes a MTLBuffer resource binding as used for an index buffer. */
-typedef struct {
+typedef struct MVKIndexMTLBufferBinding {
     union { id<MTLBuffer> mtlBuffer = nil; id<MTLBuffer> mtlResource; }; // aliases
     VkDeviceSize offset = 0;
     uint8_t mtlIndexType = 0;		// MTLIndexType
--- a/MoltenVK/MoltenVK/Commands/MVKMTLResourceBindings.h
+++ b/MoltenVK/MoltenVK/Commands/MVKMTLResourceBindings.h
@@ -36,14 +36,14 @@ typedef struct MVKMTLTextureBinding {
     bool isDirty = true;
 } MVKMTLTextureBinding;
 
-/** Describes a MTLSamplerState resource bindi MVKMTLSamplerStateBindingng. */
+/** Describes a MTLSamplerState resource binding. */
 typedef struct MVKMTLSamplerStateBinding {
     union { id<MTLSamplerState> mtlSamplerState = nil; id<MTLSamplerState> mtlResource; }; // aliases
     uint16_t index = 0;
     bool isDirty = true;
 } MVKMTLSamplerStateBinding;
 
-/** Describes  MVKMTLBufferBindinga MTLBuffer resource binding. */
+/** Describes  a MTLBuffer resource binding. */
 typedef struct MVKMTLBufferBinding {
     union { id<MTLBuffer> mtlBuffer = nil; id<MTLBuffer> mtlResource; const void* mtlBytes; }; // aliases
     VkDeviceSize offset = 0;
--- a/MoltenVK/MoltenVK/Commands/MVKMTLResourceBindings.h
+++ b/MoltenVK/MoltenVK/Commands/MVKMTLResourceBindings.h
@@ -43,7 +43,7 @@ typedef struct MVKMTLSamplerStateBinding {
     bool isDirty = true;
 } MVKMTLSamplerStateBinding;
 
-/** Describes  a MTLBuffer resource binding. */
+/** Describes a MTLBuffer resource binding. */
 typedef struct MVKMTLBufferBinding {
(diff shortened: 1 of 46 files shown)


MoltenVK/MoltenVK/Commands/MVKMTLResourceBindings.h
```diff
@@ -29,22 +29,22 @@ class MVKImage;
...
+typedef struct MVKMTLTextureBinding {
     union { id<MTLTexture> mtlTexture = nil; id<MTLTexture> mtlResource; }; // aliases
     uint32_t swizzle = 0;
 	uint16_t index = 0;
     bool isDirty = true;
 } MVKMTLTextureBinding;
 
-/** Describes a MTLSamplerState resource binding. */
-typedef struct {
+/** Describes a MTLSamplerState resource bindi MVKMTLSamplerStateBindingng. */
```
cdavis5e, line 39: This change appears accidental.

MoltenVK/MoltenVK/Commands/MVKMTLResourceBindings.h
```diff
@@ -29,22 +29,22 @@ class MVKImage;
...
+typedef struct MVKMTLSamplerStateBinding {
     union { id<MTLSamplerState> mtlSamplerState = nil; id<MTLSamplerState> mtlResource; }; // aliases
     uint16_t index = 0;
     bool isDirty = true;
 } MVKMTLSamplerStateBinding;
 
-/** Describes a MTLBuffer resource binding. */
-typedef struct {
+/** Describes  MVKMTLBufferBindinga MTLBuffer resource binding. */
```
cdavis5e, line 46: Ditto.

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
```diff
@@ -4171,3 +4165,13 @@ static uint32_t mvkGetEntryProperty(io_registry_entry_t entry, CFStringRef prope
 uint64_t mvkGetRegistryID(id<MTLDevice> mtlDevice) {
 	return [mtlDevice respondsToSelector: @selector(registryID)] ? mtlDevice.registryID : 0;
 }
+
+bool mvkSupportsBCTextureCompression(id<MTLDevice> mtlDevice) {
+#if MVK_MACOS && !MVK_MACCAT
+	if ([mtlDevice respondsToSelector: @selector(supportsBCTextureCompression)]) {
+		return mtlDevice.supportsBCTextureCompression;
+	}
```
cdavis5e, line 4173:
```suggestion
#if MVK_MACOS_APPLE_SILICON
	if ([mtlDevice respondsToSelector: @selector(supportsBCTextureCompression)]) {
		return mtlDevice.supportsBCTextureCompression;
	}
#endif
```
Otherwise, the build fails on Xcode 11.

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
```diff
@@ -4171,3 +4165,13 @@ static uint32_t mvkGetEntryProperty(io_registry_entry_t entry, CFStringRef prope
...
+	return true;
+#endif
+	return false;
```
cdavis5e, line 4176:
```suggestion
#else
	return false;
#endif
```
I'm actually surprised the way you originally had it doesn't warn.

cdavis5e (changes requested):

MoltenVK/MoltenVK/Commands/MVKMTLResourceBindings.h
billhollings: Oh geez. Fat thumbs. Fixed.

MoltenVK/MoltenVK/Commands/MVKMTLResourceBindings.h
billhollings: Similarly fixed.

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
billhollings: Thanks. Yeah. CI caught this too.

I made a slightly different edit. Although in practice for now this only applies to Apple Silicon, I figured I should just address the cause of the build failure, and filter for Xcode 12 and MacCat. But please check my platform logic again.

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
billhollings: Yes. Odd that clang missed it. Good eye! Fixed now.

MoltenVK/MoltenVK/Commands/MVKMTLResourceBindings.h
```diff
@@ -29,22 +29,22 @@ class MVKImage;
...
-typedef struct {
+typedef struct MVKMTLSamplerStateBinding {
     union { id<MTLSamplerState> mtlSamplerState = nil; id<MTLSamplerState> mtlResource; }; // aliases
     uint16_t index = 0;
     bool isDirty = true;
 } MVKMTLSamplerStateBinding;
 
-/** Describes a MTLBuffer resource binding. */
-typedef struct {
+/** Describes  a MTLBuffer resource binding. */
```
cdavis5e, line 46:
```suggestion
/** Describes a MTLBuffer resource binding. */
```

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
cdavis5e: Do BCn textures work on Catalyst?

cdavis5e (changes requested):

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
billhollings: >Do BCn textures work on Catalyst?

I don't know for sure. Not much on it out there and it's an inherently awkward thing to test. An iOS app that deliberately behaved so differently while running on a Mac.

I dithered on where to put Catalyst in this logic, but in the end I accepted that BCn formats are [defined for Catalyst](https://developer.apple.com/documentation/metal/mtlpixelformat/mtlpixelformatbc1_rgba/) for some reason. Since the obvious approach would be to assume an iOS app running on Catalyst wouldn't need them, I figured if Apple has gone to the trouble of declaring otherwise, they are probably supported.

billhollings: > Do BCn textures work on Catalyst?

Actually...in thinking about this further...if `supportsBCTextureCompression` is **_not_** defined for Catalyst, then an Apple Silicon macOS is not able to indicate that it doesn't support BCn. So maybe the lack of `supportsBCTextureCompression` on Catalyst is an indication that BCn are not expected to be supported on Catalyst. 

This is strange given my comment above about those formats existing on Catalyst. But maybe this is just a mixup in platform declarations in the SDK, and the prudent thing to do is assume BCn are not supported on Catalyst.

billhollings: Changed to report no BCn for Catalyst.

MoltenVK/MoltenVK/Commands/MVKMTLResourceBindings.h
billhollings: Fixed.

billhollings:
@cdavis5e Okay. Latest fixes. Hopefully good to go now.

cdavis5e (approved):

PR #1346, opened 2021-05-03
Use SPIRV_CROSS_NAMESPACE define in MVKDescriptor.mm

starbucksDave:
I believe SPIRV_CROSS_NAMESPACE_OVERRIDE should be replaced by SPIRV_CROSS_NAMESPACE defined in SPIRV-Cross spirv_common.hpp. This ensures that SPIRV_CROSS_NAMESPACE_OVERRIDE usage in MVK is optional.

### DIFF
--- a/MoltenVK/MoltenVK/GPUObjects/MVKDescriptor.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKDescriptor.mm
@@ -109,7 +109,7 @@ void mvkPopulateShaderConverterContext(mvk::SPIRVToMSLConversionConfiguration& c
 		mvk::MSLResourceBinding rb;													\
 		auto& rbb = rb.resourceBinding;												\
 		rbb.stage = spvExecModels[stage];											\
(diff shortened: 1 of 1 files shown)


MoltenVK/MoltenVK/GPUObjects/MVKDescriptor.mm
```diff
@@ -109,7 +109,7 @@ void mvkPopulateShaderConverterContext(mvk::SPIRVToMSLConversionConfiguration& c
 		mvk::MSLResourceBinding rb;													\
 		auto& rbb = rb.resourceBinding;												\
 		rbb.stage = spvExecModels[stage];											\
-		rbb.basetype = SPIRV_CROSS_NAMESPACE_OVERRIDE::SPIRType::spvRezType;		\
+		rbb.basetype = SPIRV_CROSS_NAMESPACE::SPIRType::spvRezType;		\
```
billhollings, line 112: Good catch! Thanks!

What an odd copy-paste I made there.

billhollings (approved):
Thanks for the submission! And good catch!

Issue #1336, opened 2021-04-21
-[AGXA10FamilyCommandBuffer logs]: unrecognized selector sent


CrystalBlack:
When I run the Demo project on the iPhone7 Device iOS 12.3.1, the following error occurred

*** Terminating app due to uncaught exception 'NSInvalidArgumentException', reason: '-[AGXA10FamilyCommandBuffer logs]: unrecognized selector sent to instance 0x10310fb50'

billhollings:
Presumably this is happening only with debug mode enabled at runtime?

In any case, this is fixed in PR #. Please retest and confirm, and close this issue if successful.

CrystalBlack:
Thank you.  It has been fixed.

PR #1353, opened 2021-05-05
Fix crash when requesting MTLCommandBuffer logs in runtime debug mode on older OS versions.

billhollings:
Fixes issue #1336.


cdavis5e (approved):
How'd I miss this...?

PR #1357, opened 2021-05-08
Fixes for CTS dEQP-VK.info.device_properties failures.

billhollings:
Adjust `VkPhysicalDeviceLimits` values:
- `minMemoryMapAlignment` raised to MIN of 64 per Vulkan spec.
- `pointSizeRange` upper limit to 511 per Metal.
- `lineWidthGranularity` to 0 per Vulkan spec when wideLines disabled.
- `maxSamplerLodBias` set to 4 to cover setting in shader.

### DIFF
--- a/MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
@@ -2024,7 +2024,7 @@ void MVKPhysicalDevice::initLimits() {
 	_properties.limits.timestampPeriod = mvkGetTimestampPeriod();
 
     _properties.limits.pointSizeRange[0] = 1;
-    _properties.limits.pointSizeRange[1] = 511;
+    _properties.limits.pointSizeRange[1] = 64;
     _properties.limits.pointSizeGranularity = 1;
     _properties.limits.lineWidthRange[0] = 1;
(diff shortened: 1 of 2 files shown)


MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
```diff
@@ -2024,11 +2024,11 @@
 	_properties.limits.timestampPeriod = mvkGetTimestampPeriod();
 
     _properties.limits.pointSizeRange[0] = 1;
-    _properties.limits.pointSizeRange[1] = 64;
+    _properties.limits.pointSizeRange[1] = 511;
```
cdavis5e, line 2027: I lowered this because the actual maximum isn't 511 as documented, but varies from device to device. Changing this causes point rendering tests to fail later on.

cdavis5e (changes requested):

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
billhollings: Thanks. Reverted.

I guess at some point we could enumerate test major devices and set accordingly.

cdavis5e (approved):

PR #1362, opened 2021-05-15
Fix synchronization issue with locking MTLArgumentEncoder.

billhollings:
Move mutex lock for using `MTLArgumentEncoder` to `MVKMTLArgumentEncoder` to be tracked along with the `MTLArgumentEncoder` it guards. Previously, the mutex was tracked in `MVKPipeline`, which did not guard any `MTLArgumentEncoder` that came from an `MVKDescriptorSetLayout`.

### DIFF
--- a/Docs/Whats_New.md
+++ b/Docs/Whats_New.md
@@ -26,6 +26,7 @@ Released TBD
   to find a cached shader, by only considering resources from the current shader stage.
 - Rename `kMVKShaderStageMax` to `kMVKShaderStageCount`.
 - Fix crash when requesting `MTLCommandBuffer` logs in runtime debug mode on older OS versions.
+- Fix synchronization issue with locking `MTLArgumentEncoder` for Metal Argument Buffers.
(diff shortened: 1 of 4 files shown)


cdavis5e (approved):
The change that added this in the first place was really big. No wonder I missed this. This is why, in Wine, we break up our patches to make them smaller and easier to review.

billhollings:
> The change that added this in the first place was really big. No wonder I missed this. This is why, in Wine, we break up our patches to make them smaller and easier to review.

Fair enough. I tried to break some of that large patch up into some precursors where I could...but you're right...it was still too large.

Part of the problem was that, because of the awkward design mis-match between Metal Arg Buffer implementation compared to Vulkan, it required meant I did a number of redesigns, and it was hard to say "I'll definitely need this", and I didn't want to commit a lot of code that would be reverted as it evolved. I'm sure you encountered that with tessellation, which similarly resulted in a kind of large all-or-nothing patch.

But nevertheless, the point is valid, and I'll endeavour to avoid that in any future large enhancements. Perhaps a published branch approach would work. I'm happy to entertain suggestions.

PR #1365, opened 2021-05-20
Fix inconsistent handling of linear attachment decisions on Apple Silicon.

billhollings:
Was causing some CTS attachment clearing tests to fail, or incorrectly report Not Supported, on Apple Silicon.

### DIFF
--- a/MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
@@ -611,7 +611,7 @@ VkResult MVKPhysicalDevice::getImageFormatProperties(VkFormat format,
 					return VK_ERROR_FORMAT_NOT_SUPPORTED;
 				}
 #if !MVK_APPLE_SILICON
-				// - On macOS IMR GPU's, Linear textures may not be used as framebuffer attachments.
+				// - On macOS IMR GPUs, Linear textures may not be used as framebuffer attachments.
 				if (hasAttachmentUsage) { return VK_ERROR_FORMAT_NOT_SUPPORTED; }
 #endif
 				// Linear textures may only have one mip level, layer & sample.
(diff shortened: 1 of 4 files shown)


MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
```diff
@@ -610,8 +610,8 @@
 				if (mvkFmt == kMVKFormatDepthStencil || mvkFmt == kMVKFormatCompressed || isBGRG) {
 					return VK_ERROR_FORMAT_NOT_SUPPORTED;
 				}
-#if MVK_MACOS
-				// - On macOS, Linear textures may not be used as framebuffer attachments.
+#if !MVK_APPLE_SILICON
+				// - On macOS IMR GPU's, Linear textures may not be used as framebuffer attachments.
```
cdavis5e, line 614:
```suggestion
				// - On macOS IMR GPUs, Linear textures may not be used as framebuffer attachments.
```
The apostrophe implies possession.

cdavis5e (changes requested):
What if Intel-architecture MoltenVK finds itself running in Rosetta 2 on an Apple Silicon Mac?

billhollings:
> What if Intel-architecture MoltenVK finds itself running in Rosetta 2 on an Apple Silicon Mac?

Dang! Sorry...because the change was just syntactic...I completely focused on this and didn't see your comment.

Good point. I hadn't considered that situation specifically. 

However...I think this makes sense. As it sits here, Intel MoltenVK under Rosetta would think it was running on an Intel Mac, just like any other Rosetta-driven Intel app, and Apple has taken on the responsibility to make Rosetta handle the silicon differences seamlessly.

I'm thinking if we wander into that territory, we might set up a situation where the app thinks it's on an Intel platform, and its characteristics because it knows no different, and we end up making changes under the covers unexpectedly.

I'm happy to hear counter-arguments.

Based on the result of that discussion, I wonder if we should adopt a consistent policy one way or another about whether an Intel MoltenVK under Rosetta should behave, and modify other internal platform tests accordingly. For instance, if my logic is correct, then perhaps all internal Apple Silicon tests should be based on compile-time arm64 testing instead of runtime Apple7 GPU testing.

galad87:
imho capabilities should depend only on the available gpu, and not on the cpu architecture, I think there was a WWDC session that recommended to optimise for the M1 gpu even when running under rosetta.

PR #1370, opened 2021-05-28
Support the VK_KHR_imageless_framebuffer extension.

f32by:
I'm not sure this is the elegant way but it works in my own
project:)


MoltenVK/MoltenVK/Commands/MVKCmdRenderPass.h
```diff
@@ -50,6 +50,7 @@ class MVKCmdBeginRenderPassBase : public MVKCommand {
 	MVKFramebuffer* _framebuffer;
 	VkRect2D _renderArea;
 	VkSubpassContents _contents;
+	MVKSmallVector<MVKImageView*, 8> _imagelessAttachments;
```
cdavis5e, line 53: This class probably ought to be templated, providing variants with 0, 1, and 8 attachments, to handle the common cases of not using imageless and using it with one attachment.

MoltenVK/MoltenVK/Commands/MVKCommandBuffer.h
```diff
@@ -275,7 +275,8 @@ class MVKCommandEncoder : public MVKBaseDeviceObject {
 						 MVKRenderPass* renderPass,
 						 MVKFramebuffer* framebuffer,
 						 VkRect2D& renderArea,
-						 MVKArrayRef<VkClearValue> clearValues);
+						 MVKArrayRef<VkClearValue> clearValues,
+						 MVKArrayRef<MVKImageView*> imageless_attachments);
```
cdavis5e, line 279:
```suggestion
						 MVKArrayRef<MVKImageView*> imagelessAttachments);
```
Please try to follow the style of the surrounding code.

MoltenVK/MoltenVK/Commands/MVKCommandBuffer.h
```diff
@@ -494,6 +495,7 @@ class MVKCommandEncoder : public MVKBaseDeviceObject {
 	VkRect2D _renderArea;
     MVKActivatedQueries* _pActivatedQueries;
 	MVKSmallVector<VkClearValue, kMVKDefaultAttachmentCount> _clearValues;
+	MVKSmallVector<MVKImageView*, kMVKDefaultAttachmentCount> _imagelessAttachments;
```
cdavis5e, line 498: I feel like there ought to be a better way to handle this. The imageless model, with attachments provided at render pass begin time, corresponds better to the Metal model. Maybe we should see about not keeping the framebuffer object around and just using a vector of attachments, since the purpose of the framebuffer object is to be a container for attachments.

MoltenVK/MoltenVK/Commands/MVKCommandBuffer.mm
```diff
@@ -288,13 +288,17 @@
...
+										MVKArrayRef<MVKImageView*> imagelessAttachments) {
 	_renderPass = renderPass;
 	_framebuffer = framebuffer;
 	_renderArea = renderArea;
 	_isRenderingEntireAttachment = (mvkVkOffset2DsAreEqual(_renderArea.offset, {0,0}) &&
 									mvkVkExtent2DsAreEqual(_renderArea.extent, _framebuffer->getExtent2D()));
 	_clearValues.assign(clearValues.begin(), clearValues.end());
+	for(auto* v : imagelessAttachments) {
+		_imagelessAttachments.push_back(v);
+	}
```
cdavis5e, line 301:
```suggestion
	_imagelessAttachments.assign(imagelessAttachments.begin(), imagelessAttachments.end());
```

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
```diff
@@ -267,6 +267,11 @@
 				inlineUniformBlockFeatures->descriptorBindingInlineUniformBlockUpdateAfterBind = true;
 				break;
 			}
+			case VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_IMAGELESS_FRAMEBUFFER_FEATURES: {
+				auto* imagelessFramebufferFeatures = (VkPhysicalDeviceImagelessFramebufferFeaturesKHR*)next;
+				imagelessFramebufferFeatures->imagelessFramebuffer = true;
+				break;
+			}
```
cdavis5e, line 274: I typically don't add code for this if the feature enable isn't checked elsewhere.

MoltenVK/MoltenVK/GPUObjects/MVKFramebuffer.h
```diff
@@ -57,5 +58,7 @@ class MVKFramebuffer : public MVKVulkanAPIDeviceObject {
 	VkExtent2D _extent;
 	uint32_t _layerCount;
 	MVKSmallVector<MVKImageView*, 4> _attachments;
+	bool _imageless;
+	MVKSmallVector<MVKImageView*, 4> _imagelessAttachments;
```
cdavis5e, line 62: Wait, why do we need another vector for the imageless attachments? This isn't even used.

cdavis5e (changes requested):
Needs a bit of work.

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
cdavis5e: Oops, wrong spot. This is the place you absolutely need to add this, because this is where we report the supported features of the device! Sorry about that. I meant to attach this comment to the other patch hunks.

MoltenVK/MoltenVK/Commands/MVKCommandBuffer.mm
f32by: Hi, the compiler throws an error with new code:
```
MVKCommandBuffer.mm:299:31: Cannot initialize a parameter of type 'size_t' (aka 'unsigned long') with an rvalue of type 'MVKImageView *const *'
```
I think the compiler calls `assign(const size_t new_size, const Type *t)` rather than `assign(InputIterator first, InputIterator last)`.

MoltenVK/MoltenVK/Commands/MVKCmdRenderPass.h
billhollings: > This class probably ought to be templated, providing variants with 0, 1, and 8 attachments, to handle the common cases of not using imageless and using it with one attachment.

Nice catch.

MoltenVK/MoltenVK/Commands/MVKCommandBuffer.mm
billhollings: > I think the compiler calls `assign(const size_t new_size, const Type *t)` rather than `assign(InputIterator first, InputIterator last)`.

Oops...looks like the pointer version of `MVKSmallVectorImpl` doesn't have an iterator-based `assign()` function overload. Odd. 

If you feel up for doing some plumbing, you could add that function (based on the similar `assign()` overload in the non-pointer `MVKSmallVectorImpl` version). But otherwise I'm okay with leaving it as you have it.

billhollings (approved):
Thanks for the contribution!

Nice work. And thanks for working out the two-arg `MVKCmdBeginRenderPass` templating requested by @cdavis5e .

LGTM. If you want to add the additional `assign()` function (see my inline note), please do so. Otherwise, once @cdavis5e approves your updated changes, I'll pull this in.

MoltenVK/MoltenVK/Commands/MVKCommandBuffer.mm
f32by: Sure, I'll add it.

f32by:
Hi,
I added a new assign() which receives iterator in the pointer version of MVKSmallVectorImpl. Any further advice would be great :)

billhollings (approved):
Great! Thanks for adding the additional `MVKSmallVector::assign()` overload.

It all LGTM now.

MoltenVK/MoltenVK/Commands/MVKCmdRenderPass.mm
```diff
@@ -57,25 +58,62 @@
...
+template class MVKCmdBeginRenderPass<2, 0>;
+template class MVKCmdBeginRenderPass<9, 0>;
+
+template class MVKCmdBeginRenderPass<1, 1>;
+template class MVKCmdBeginRenderPass<2, 1>;
+template class MVKCmdBeginRenderPass<9, 1>;
+
+template class MVKCmdBeginRenderPass<1, 8>;
+template class MVKCmdBeginRenderPass<2, 8>;
+template class MVKCmdBeginRenderPass<9, 8>;
```
cdavis5e, line 116: This is rife for combinatorial explosion. But since you only have to supply clear values for all attachments up to the highest-numbered one with a `CLEAR` load op, I don't know what else to do.

You probably should use 9 instead of 8, because the depth/stencil attachment plus max color attachments makes 9 total. And yes, I am aware that I suggested 8 in the first place. :) You may also wish to add a case for two attachments, for the common case of a DS attachment with a single color attachment.

MoltenVK/MoltenVK/Commands/MVKCommandBuffer.mm
```diff
@@ -555,7 +568,7 @@
 	endMetalRenderEncoding();
 
 	_renderPass = nullptr;
-	_framebuffer = nullptr;
+	_attachments.clear();
```
cdavis5e, line 571: Should probably also clear framebuffer-related properties.

cdavis5e (changes requested):
Sorry. Just a couple of minor issues.

f32by:
Added MVKCmdBeginRenderPass<N, 2> and more framebuffer cleanups in MVK.
BTW, if we can move definitions of MVKCmdBeginRenderPass from `MVKCmdRenderPass.mm` to its header, we can use template parameter to let the compiler generate specialization for us. Please refer to https://github.com/f32by/MoltenVK/blob/imageless_framebuffer_way2/MoltenVK/MoltenVK/Commands/MVKCmdRenderPass.h. However, using `std::initializer` in `constexpr` functions requires at least C++ 14.

billhollings:
However, using `std::initializer` in `constexpr` functions requires at least C++14 and MoltenVK is currently compiled under C++11.

Thanks for the note. When MoltenVK started 5 years ago, C++11 was appropriate (C++14 was too new). There have definitely been some good reasons for moving forward from that...probably to C++17 if possible. I'll post a separate issue about that.

cdavis5e (approved):

PR #1373, opened 2021-05-31
Fix crash using memoryless storage for input attachments on Apple Silicon.

billhollings:
Don't allow memoryless storage for `VK_IMAGE_USAGE_INPUT_ATTACHMENT_BIT`.
Memoryless storage for `VK_DESCRIPTOR_TYPE_INPUT_ATTACHMENT` descriptors
requires support for, and use of, shader framebuffer fetch.

Remove compile-time checking for `MVK_APPLE_SILICON` before disabling memoryless storage
for buffers and images, since the requirement exists for all platforms, and memoryless
storage will already otherwise be disabled on other platforms anyway.


billhollings:
Added issue #1374 to mark a larger enhancement to use Metal shader framebuffer fetch.

cdavis5e (approved):

PR #1375, opened 2021-06-01
Begin renderpass fixes

billhollings:
PR #1370 introduced a bug by not clearing attachments when `MVKCmdBeginRenderPass` instances were reused from the command pool.

While fixing that, I also cleaned up some redundancies in the `vkCmdBeginRenderPass()` and `vkCmdBeginRenderPass2KHR()` functions and `MVKCmdBeginRenderPass` constructors.

- Clear attachments in `MVKCmdBeginRenderPass` between invocations. Fixes crash due to incorrect number of attachments.
- Consolidate `vkCmdBeginRenderPass()` and `vkCmdBeginRenderPass2KHR()` handling.
- Derive `MVKCmdBeginRenderPass` attachment count and imageless status once.
 
@f32by please review relative to the recent #1370. I can't request you as a reviewer from the list until you accept the collaboration request.

### DIFF
--- a/Docs/Whats_New.md
+++ b/Docs/Whats_New.md
@@ -18,6 +18,8 @@ MoltenVK 1.1.4
 
 Released TBD
 
+- Add support for extensions:
+	- `VK_KHR_imageless_framebuffer`
 - Make `vkGetPastPresentationTimingGOOGLE()` queuing behavior compliant with Vulkan spec.
 - Expose `vkGetIOSurfaceMVK()` and `vkUseIOSurfaceMVK()` without requiring _Objective-C_.
 - Support _Xcode 12.5_ build settings, build warnings, and SDK change to availability of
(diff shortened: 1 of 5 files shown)


f32by:
Thank you for your invitation!
Apparently your approach is better than mine.

A little advice: we already check if the framebuffer is imageless in `mvkCmdBeginRenderPass()`(https://github.com/billhollings/MoltenVK/blob/begin-renderpass-fixes/MoltenVK/MoltenVK/Vulkan/vulkan.mm#L1904-L1915), I think it's better to also pass the pointer of `VkRenderPassAttachmentBeginInfo ` to `MVKCmdBeginRenderPass::setContent()`(https://github.com/billhollings/MoltenVK/blob/begin-renderpass-fixes/MoltenVK/MoltenVK/Commands/MVKCmdRenderPass.mm#L67-L79) and thus prevent looping twice.

cdavis5e (approved):

billhollings:
>I think it's better to also pass the pointer of `VkRenderPassAttachmentBeginInfo ` to `MVKCmdBeginRenderPass::setContent()`

Good suggestion. I've added a change for that, but slightly modified to pass an `MVKArrayRef` directly into `MVKCmdBeginRenderPass::setContent()`, which completely simplifies that function.

cdavis5e (approved):

PR #1376, opened 2021-06-01
Don't use the "fast" optimization level for external libraries.

js6i:
It implies -ffast-math, which breaks isnan() checks. SPIRV-Cross relies on
isnan() to decide how to print floating point literals.


js6i:
Here's a little example program that I used for testing with the MVK shader tool:
```
               OpCapability Shader
         %65 = OpExtInstImport "GLSL.std.450"
               OpMemoryModel Logical GLSL450
               OpEntryPoint Fragment %1 "main"
               OpExecutionMode %1 OriginUpperLeft
               OpName %1 "main"
          %5 = OpTypeFloat 32
          %3 = OpTypeFunction %5
          %1 = OpFunction %5 None %3
          %4 = OpLabel
        %BAD = OpConstant %5 -0x1.fffffep+128
	       OpReturnValue %BAD
               OpFunctionEnd
```

billhollings:
Interesting. 

Although the edge case may be small, I'm wondering if we should make this configuration change across the board, including with with MoltenVK and MoltenVKShaderConverter?

Since MoltenVK is aiming for standards conformance, perhaps the appropriate behavior is to shun builds "that may break strict standards compliance".

js6i:
That's a good idea, I would go even further and generally stick to O2 unless there's some benefit to using other flags that we can see in benchmarks. In my experience O3 tends to produce unnecessarily bloated code, and Ofast is just scary.

billhollings:
> stick to O2

Looks like you might be right. There are some indications that `O2` mostly equals or outperforms `O3` performance, and of course produces smaller code in less time:

https://discourse.mc-stan.org/t/o2-vs-o3-compiler-optimization-level/3714
https://iitd-plos.github.io/col729/labs/lab0/lab0_submissions/siy187504.pdf

I'll submit a general PR that uses `O2` across the board. Thanks for all your input.

Issue #1256, opened 2021-02-11
Improve tracking of need for Metal visibility buffer in Metal renderpasses


billhollings:
The current logic around deciding whether a visibility buffer is needed is relatively crude. Currently, it is applied to all render passes in a Vulkan command buffer based on at least one occlusion query being present. A Metal renderpass needs a visibility buffer if a query is started either before it or inside it. For a command buffer with multiple Metal render passes, not all of them may need it. 

Recommended logic:
- Track need for visibility buffer in `MVKCmdBeginRenderPass` and `MVKCmdNextSubpass`.
- `MVKCommandBuffer` track when `MVKCmdBeginRenderPass` or `MVKCmdNextSubpass` added and set it's `_needsVisibilityResultMTLBuffer` from `MVKCommandBuffer::_needsVisibilityResultMTLBuffer`.
- Adding, `MVKCmdBeginQuery` sets `MVKCommandBuffer::_needsVisibilityResultMTLBuffer` and if a `MVKCmdBeginRenderPass` or `MVKCmdNextSubpass` is open, set its `_needsVisibilityResultMTLBuffer` too.
sets `MVKCommandBuffer::_needsVisibilityResultMTLBuffer` to false.
- Adding `MVKCmdEndRenderPass` clears open render/sub pass in `MVKCommandBuffer`.- Adding `MVKCmdEndQuery`

billhollings:
I performed some further analysis and experimentation. Additional requirements:

- Vulkan secondary command buffers may contain queries that affect an open begin renderpass command in the primary buffer, and `_needsVisibilityResultMTLBuffer` is set in any of the secondary command buffers.
- `MVKCommandEncoder:: _needsVisibilityResultMTLBuffer` status set by encoding of MVKBeginRenderpass` and `MVKNextSubpass`.
- `MVKCommandEncoder:: _needsVisibilityResultMTLBuffer` status used in `beginMetalRenderPass()` to determine whether visibility buffer should be used.
- Metal renderpass iterations used for Vulkan multi view rendering should not change any flags, but should honor value of `MVKCommandEncoder:: _needsVisibilityResultMTLBuffer`.

billhollings:
Since a full solution appears to be necessarily complex, and the benefits of the improvement are unclear (assumed to be some type of undefined performance improvement), at this point, further work on this is being left until a definite need arises.

PR #1387, opened 2021-06-21
Occlusion query fixes for M1

### LINKED ISSUE #1385: Query pool CTS tests failing on M1
Strangely, for the CTS tests in `vk-default/query-pool.txt`:

- AMD & Intel: Pass: 188 / Fail:  0.
- M1: Pass: 48 / Fail: 140.

The M1 numbers can sometimes change, indicating some form of race condition. Some failing tests can pass occasionally when running in batch, but always fail when run alone.

It seems to be connected to those tests that have multiple render passes with occlusion queries in them. Subsequent renderpass queries seem to be overwriting, or corrupting previous ones, but in ways that have no obvious cause as yet. The writes seem to be occurring to the correct buffer offsets, and appear so in GPU capture, but then the buffer offsets seem to be shifted when reading the contents back to the CPU for test verification.

I'm currently looking into this.

@cdavis5e I'm continuing with this...but if any possible causes pop into your head, feel free to suggest paths for me to research. So far I have eliminated the type of Vulkan wait used before the CPU read, and buffer alignment corruption on the copy from temporary buffer to visibility buffer. Next attempt will be to see if M1 somehow requires `MTLFences` or `MTLEvents` on the `MTLBuffers` between visibility compute encoders (which it shouldn't per docs).

billhollings:
Fixes #1385.

Per discussion in issue, this PR allows a single Metal visibility buffer to be used for all render encoders in a single `MTLCommandBuffer`.

- The first commit in this PR fixes the issue across multiple Metal render encoders in a single Vulkan command buffer.
- The second commit here extends this to an entire Vulkan submit (multiple Vulkan command buffers, one Metal command buffer).
- Third commit cleans up some syntax around tracking query locations in the visibility buffer to make it clearer how the mapping works.

### DIFF
--- a/MoltenVK/MoltenVK/Commands/MVKCommandBuffer.mm
+++ b/MoltenVK/MoltenVK/Commands/MVKCommandBuffer.mm
@@ -350,7 +350,10 @@ void MVKCommandEncoder::beginMetalRenderPass(bool loadOverride) {
 												  _isRenderingEntireAttachment,
 												  loadOverride);
 	if (_cmdBuffer->_needsVisibilityResultMTLBuffer) {
-		mtlRPDesc.visibilityResultBuffer = _pEncodingContext->getVisibilityResultBuffer(this)->_mtlBuffer;
+		if ( !_pEncodingContext->visibilityResultBuffer ) {
+			_pEncodingContext->visibilityResultBuffer = getTempMTLBuffer(_pDeviceMetalFeatures->maxQueryBufferSize, true, true);
+		}
+		mtlRPDesc.visibilityResultBuffer = _pEncodingContext->visibilityResultBuffer->_mtlBuffer;
 	}
 
 	VkExtent2D fbExtent = _framebufferExtent;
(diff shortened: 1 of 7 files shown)


MoltenVK/MoltenVK/Commands/MVKCommandBuffer.mm
```diff
@@ -799,6 +799,33 @@
...
+}
+
+
+#pragma mark -
+#pragma mark MVKCommandEncodingContext
+
+// Increment to the next query slot offset. If we reach the size of the visibility buffer,
+// reset to retrieve and start filling another visibility buffer. This approach may still
+// cause Metal validation errors if the platform does not permit offsets to be reused
+// witin a MTLCommandBuffer, even when a different visibility buffer is used.
```
cdavis5e, line 812:
```suggestion
// within a MTLCommandBuffer, even when a different visibility buffer is used.
```

MoltenVK/MoltenVK/Commands/MVKCommandBuffer.mm
```diff
@@ -799,6 +799,33 @@
...
+// We don't test against the size of the visibility buffer itself, because this call may
+// arrive before the visibiltiy buffer in the case of a query that ends before the renderpass.
```
cdavis5e, line 814:
```suggestion
// arrive before the visibility buffer in the case of a query that ends before the renderpass.
```

cdavis5e (approved):
Just a couple of minor typos.

js6i:
I'm trying to understand what happens when we overflow the buffer. Suppose we only have one slot left in the visibility buffer and we begin a render pass, start a query, draw, end query and start another one, draw again, end, get the results. I'm likely missing something, but:
1) Shouldn't we use a new `MTLRenderPipelineState` after we've ran out of space and made a new buffer? The buffer is baked into it.
2) We only get the newest buffer in `MVKOcclusionQueryCommandEncoderState::endMetalRenderPass`, so shouldn't we also force an accumulation step when we change buffers? Or, better, remember which one was used for each query?
3) If the answer to above is that in the end we only use the new buffer for the whole Metal render pass during which the old buffer overflowed, we still seem to have old (high) offsets in the query array, which might get accidentally reused later?

I'm guessing that the Metal render pass *is* ended somehow in this scenario, or things happen in some different order than I think they do.

billhollings:
>1. Shouldn't we use a new MTLRenderPipelineState after we've ran out of space and made a new buffer? The buffer is baked into it.

Not `MTLRenderPipelineState`. The visibility buffer is baked into `MTLRenderCommandEncoder`, which is created on boundaries of Vulkan subpasses (BeginRenderpass or NewSubpass).

The `incrementMTLVisibilityResultOffset` that triggers the need for a new buffer occurs at the end of an occlusion query. The creation of that new buffer occurs at the beginning of a render pass.

So the situation you describe would be the end of an occlusion query but before the next renderpass. Subsequent queries in that renderpass would overwrite the beginning of the visibility buffer (which may or may not cause a validation error depending on the platform). Subsequent render passes would use the new visibility buffer.

> 2. We only get the newest buffer in MVKOcclusionQueryCommandEncoderState::endMetalRenderPass, so shouldn't we also force an accumulation step when we change buffers? Or, better, remember which one was used for each query?

Yes. Perhaps we should track the visibility buffer in the `OcclusionQueryLocation`. But I think the same situation I described above would occur. The new visibility buffer would not make it into the current renderpass encoder, and queries used in that renderpass would overwrite the beginning of the visibility buffer. 

Or...in both cases, maybe we could detect that the buffer has changed and avoid those overwrites in the current renderpass.

Another simpler and predictable option could be to just use one visibility buffer and permit only up to the maximun numbers of queries per `MTLCommandBuffer`, and just trigger a Vulkan error when it is exceeded.

js6i:
Maybe we could simply restart the encoder (`MVKCommandEncoder::beginMetalRenderPass`?) when we overflow the buffer? We'd accumulate all queries, resume any active ones with the offsets they had assigned and move on - we shouldn't ever reach these old offsets and still need them, as that would mean the render pass limit of queries was exceeded.

billhollings:
In the end I decided to keep this simple until we have a definite need to do further gymnastic workarounds.

I have reverted to just supporting one Metal visibility buffer per `MTLCommandBuffer`, outputting an error log and safely halting visibility tracking within the `MTLCommandBuffer` if the visibility buffer becomes exhausted.

> Maybe we could simply restart the encoder

Hmmm...this sounded fraught with side effects, so I didn't dig into it. It remains a possibility if we identify that we need more flexibility with tracking larger numbers of occlusion queries in a single Vulkan submit.

> we could track the need to allocate a new visibility buffer for a render pass before we start encoding it

So we have #1256 proposing a finer level of tracking the need for a visibility buffer. But again, it appears to be a fairly complex solution to an ill-defined need. Something for the future if needed.

cdavis5e (approved):

js6i (approved):

PR #1389, opened 2021-06-22
Add Scripts/runcts script as a convenience for running Vulkan CTS tests.

billhollings:
Update `.gitignore` to ignore CTS artifacts in `Scripts` directory.

### DIFF
--- a/Scripts/runcts
+++ b/Scripts/runcts
@@ -1,0 +1,124 @@
+#!/bin/bash
+
+# Copyright (c) 2016-2021 The Brenwill Workshop Ltd.
+#
+# runcts - This script simplifies running Vulkan CTS tests on MoltenVK.
+#
+# CTS will save temporary output files to the directory this script is located in. It is 
+# recommended that you copy this file to a location where your testing files will be kept.
+# By default, this should be a directory beside the VK-GL-CTS directory. See the description 
+# of the --cts option below for information about locations.
+#
+# Executing a large case list file of CTS tests can take a long time. It is recommended
+# that you run this using caffeinate, as follows:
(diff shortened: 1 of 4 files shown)


Scripts/runcts
```diff
@@ -0,0 +1,124 @@
...
+         shift 2
+         ;;
+       -p)
+         is_portability="Y"
+         shift 1
+         ;;
+       --portability)
+         is_portability="Y"
+         shift 1
+         ;;
```
cdavis5e, line 67: You could probably consolidate these into one case like this:
```suggestion
       -p|--portability)
         is_portability="Y"
         shift 1
         ;;
```

Scripts/runcts
```diff
@@ -0,0 +1,124 @@
...
+
+if [ "${results_file}" == "" ]; then
+	results_file="${caselist_file}-results.txt"
+fi
+
+
+# -------------- MoltenVK configuration --------------------
+
+if [ "${is_portability}" != "" ]; then
+	export MVK_CONFIG_API_VERSION_TO_ADVERTISE=4194304
```
cdavis5e, line 92: That magic number looks pretty mysterious. You should probably have a comment here explaining it.

cdavis5e (changes requested):
Just a few changes.

Scripts/runcts
billhollings: Done.

Scripts/runcts
billhollings: Done. 

The values are documented in `vk_mvk_moltenvk.h`...and in the documentation in this script file ...but fair enough...doesn't hurt to explain how the values used here as derived.

cdavis5e (approved):

PR #1391, opened 2021-06-24
Fixes #1390 - when MVKRenderSubpass::populateClearAttachments determi…

### LINKED ISSUE #1390: MoltenVK incorrectly clearing stencil when only depth is set to clear [Dota]
I ran into a bug in Dota which is causing the water to render black when setting "Game Screen Render Quality" to less than 100%.  I noticed this when testing the new AMD FidelityFX Super Resolution feature that we added.

The problem is that there is a renderpass that has a renderArea smaller than the attachments and it has Color and Depth clear set to loadOp CLEAR and stencilLoadOp to LOAD.  I've found the bug where MoltenVK is incorrectly clearing stencil in this case, I'll file a PR shortly, but wanted to open a tracking issue.

danginsburg:
…nes whether to clear stencil, it was using the loadOp for the depth.  It needs to look at the loadOp for the stencil which is set independently.

### DIFF
--- a/MoltenVK/MoltenVK/GPUObjects/MVKRenderPass.h
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKRenderPass.h
@@ -201,6 +201,9 @@ public:
     /** Returns whether this attachment should be cleared in the subpass. */
     bool shouldUseClearAttachment(MVKRenderSubpass* subpass);
 
+    /** Returns stencil attachment load op */
+	VkAttachmentLoadOp getAttachmentStencilLoadOp() const;
+
 	/** Constructs an instance for the specified parent renderpass. */
 	MVKRenderPassAttachment(MVKRenderPass* renderPass,
 							const VkAttachmentDescription* pCreateInfo);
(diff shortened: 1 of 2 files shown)


billhollings (approved):
Thanks, Dan!

MoltenVK/MoltenVK/GPUObjects/MVKRenderPass.h
```diff
@@ -201,6 +201,9 @@ class MVKRenderPassAttachment : public MVKBaseObject {
     /** Returns whether this attachment should be cleared in the subpass. */
     bool shouldUseClearAttachment(MVKRenderSubpass* subpass);
 
+    /** Returns stencil attachment load op */
```
cdavis5e, line 204: Doc comments should be complete sentences. Also, this one doesn't have any information that couldn't be inferred from the method name.

cdavis5e (approved):
Looks good. Just one really minor nit.

MoltenVK/MoltenVK/GPUObjects/MVKRenderPass.h
danginsburg: Added clarifying comment that is a complete sentence :)

PR #1395, opened 2021-07-04
Fix incorrect translation of clear color values on Apple Silicon.

billhollings:
The same set of CTS tests either fails or passes on different GPUs
based on whether or not we adjust float clear colors by one ULP.

- Add `MVKFloatRounding` enum.
- Add `MVKPhysicalDeviceMetalFeatures::clearColorFloatRounding`.
- Disable ULP adjustment for clear colors on Apple Silicon.
- For consistency and to simplify bookkeepping, calculate clear color ULP adjustment from bit width of format component. This was not strictly necessary, but is clearer, and could potentially allow us to more easily adjust the ULP offset calcs per GPU if needed (eg. 1/4 ULP offsets, etc).
- Update MoltenVK version to 1.1.5.
- Update VK_MVK_MOLTENVK_SPEC_VERSION to 32.

### DIFF
--- a/MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
@@ -1192,6 +1192,19 @@ void MVKPhysicalDevice::initMetalFeatures() {
 
 	_metalFeatures.maxPerStageStorageTextureCount = 8;
 
+	// GPU-specific features
+	switch (_properties.vendorID) {
+		case kAMDVendorId:
+			_metalFeatures.clearColorFloatRounding = MVK_FLOAT_ROUNDING_DOWN;
+			break;
+		case kAppleVendorId:
+		case kIntelVendorId:
+		case kNVVendorId:
+		default:
+			_metalFeatures.clearColorFloatRounding = MVK_FLOAT_ROUNDING_NEAREST;
+			break;
+	}
+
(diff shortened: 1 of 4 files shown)


MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
```diff
@@ -1376,6 +1378,7 @@
 	_metalFeatures.maxTextureDimension = (16 * KIBI);
 	_metalFeatures.depthSampleCompare = true;
 	_metalFeatures.samplerMirrorClampToEdge = true;
+	_metalFeatures.clearColorFloatRounding = MVK_FLOAT_ROUNDING_DOWN;
```
cdavis5e, line 1381: I actually don't recall seeing this problem with Intel GPUs. I think they use round-to-nearest as well.

cdavis5e (changes requested):
One minor quibble.

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
billhollings: Thanks. I've just tested Intel and looks like you are right. 

I've changed the setting of `_metalFeatures.clearColorFloatRounding` to be driven by GPU `vendorID`.

cdavis5e (approved):

PR #1397, opened 2021-07-07
Fix swizzle of depth and stencil values into RGBA (float4) variable in shaders.

billhollings:
Sampling depth/stencil values into an RGBA value in a shader behaves inconsistency across Metal platforms, resolving to `(v, 0, 0, 1)` (as expected by Vulkan) on some platforms, but to `(v, v, v, v)` on others. This causes several hundred CTS tests to fail.

To get around this, we force an artificial swizzle for D/S textures that are sampled, if it's available on the platform.

- `MVKImageViewPlane` track Vulkan component swizzle instead of just packed swizzle, and modify swizzle when using depth or stencil sampling or reading.
- Refactor `MVKImageView::validateSwizzledMTLPixelFormat()`.
- Move `MVKImageView::validateSwizzledMTLPixelFormat()` to `MVKImageViewPlane::initSwizzledMTLPixelFormat()`, make it non-static, and use instance member content instead of passing all the data as function arguments.
- Use unswizzled `MTLTexture` for attachments.

### DIFF
--- a/Docs/Whats_New.md
+++ b/Docs/Whats_New.md
@@ -19,6 +19,7 @@ MoltenVK 1.1.5
 Released TBD
 
 - Fix incorrect translation of clear color values on Apple Silicon.
+- Fix swizzle of depth and stencil values into RGBA (`float4`) variable in shaders.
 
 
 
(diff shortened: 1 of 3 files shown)


cdavis5e (approved):
I had thought that I couldn't work around this problem that way, so I filed a feedback. But it turns out my test program had a bug that caused it to splat the depth value no matter what, so a swizzle would've done nothing in any case. I guess I can close the feedback since we have a workaround.

PR #1398, opened 2021-07-08
Fix occasional crash when swizzling used but shader swizzling not enabled.

billhollings:
Recent fixes to add depth/stencil swizzling introduced an occasional situational crash. This repairs that, and consolidates tracking of shader swizzling.

`MVKImageViewPlane` distinguish tracking of native and shader swizzling.
Only provide packed swizzle value if shader swizzling is used.
Consolidate and simplify testing for `MVKConfiguration::fullImageViewSwizzle`.

### DIFF
--- a/MoltenVK/MoltenVK/GPUObjects/MVKPipeline.h
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKPipeline.h
@@ -173,9 +173,6 @@ public:
 	/** Returns the current indirect parameter buffer bindings. */
 	const MVKShaderImplicitRezBinding& getIndirectParamsIndex() { return _indirectParamsIndex; }
 
-	/** Returns whether or not full image view swizzling is enabled for this pipeline. */
-	bool fullImageViewSwizzle() const { return _fullImageViewSwizzle; }
-
 	/** Returns whether all internal Metal pipeline states are valid. */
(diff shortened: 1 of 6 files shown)


MoltenVK/MoltenVK/GPUObjects/MVKPipeline.h
```diff
@@ -205,7 +202,6 @@ class MVKPipeline : public MVKVulkanAPIDeviceObject {
 	MVKShaderImplicitRezBinding _indirectParamsIndex;
 	MVKShaderResourceBinding _pushConstantsMTLResourceIndexes;
 	uint32_t _descriptorSetCount;
-	bool _fullImageViewSwizzle;
```
cdavis5e, line 208: I'm worried about this state getting out of sync now that we always go to the configuration, such that we can have a pipeline that was compiled with it on, but bound with it off, so we don't update the swizzle buffer. It's why I put this here in the first place IIRC.

cdavis5e (changes requested):

MoltenVK/MoltenVK/GPUObjects/MVKPipeline.h
billhollings: Ah...shoot. You're right. And that's actually documented in the notes for the config setting.

Looks like I overreached. I'll close this PR and resubmit one with the smaller changes that just fix the crash.

PR #1401, opened 2021-07-10
Support alpha-to-coverage without a color attachment.

billhollings:
If alpha-to-coverage is enabled, we must enable the fragment shader first color output,
even without a color attachment present or in use, so that coverage can be calculated.

@cdavis5e FWIW...in working through this in CTS `dEQP-VK.pipeline.multisample.alpha_to_coverage_no_color_attachment.samples_2.alpha_opaque`, I noticed that enabling all attachments in `mslOptions.enable_frag_output_mask` did not trigger any Metal validation errors, even with no color attachments and `MTLRenderPipelineDescriptor::alphaToCoverageEnabled` disabled. I'm not sure if Apple no longer performs the validation that catalyzed the requirement for `mslOptions.enable_frag_output_mask`, or whether I just didn't encounter the appropriate use case for that validation. Just in case, I've left it intact except for the new `alphaToCoverageEnable` check.


cdavis5e (approved):
I swear I had problems with this... Maybe they did remove the validation.

PR #1404, opened 2021-07-21
Vulkan timestamp query pools use Metal GPU counters when available.

### LINKED ISSUE #520: Timestamp support in MoltenVK
Is there any documentation about the values returned when using vk::CommandBuffer::writeTimestamp()?

On my Linux nvdia machine I get epoch timestamps but on moltenvk I get values that I cannot understand. 

For example: I have a command buffer with 6 timestamps and I took two timestamps with the system clock before and after. The system timestamps make sense but the 6 timestamps are all equal to each other and not in between the cpu timestamps.

### LINKED ISSUE #773: Suggestion for More Accurate GPU Timings for vkCmdWriteTimestamp
Have you considered using MTLCommandBuffer's [GPUStartTime](https://developer.apple.com/documentation/metal/mtlcommandbuffer/1639924-gpustarttime?language=objc) and [GPUEndTime](https://developer.apple.com/documentation/metal/mtlcommandbuffer/1639926-gpuendtime?language=objc) instead of using system time at command buffer submission? These are doubles, in seconds.
Timestamp queries using VK_PIPELINE_STAGE_TOP_OF_PIPE_BIT and VK_PIPELINE_STAGE_BOTTOM_OF_PIPE_BIT seem to correspond to these values.

As a further simplification, I think if you did this, you could multiply the result by one billion to convert it to nanoseconds and change VkPhysicalDeviceLimits::timestampPeriod to one rather than needing to use Mach system time computations.

### LINKED ISSUE #793: Request: Implement Counter Sampling for Timestamp Query Pools
Recently a new primitive got added to Metal that makes it possible to sample hardware counters such as timestamps, which sounds like it would be a really good implementation for time stamp query pools.

Right now timestamp query pools don't really give back any usable info since it's all host timestamps calculated at command buffer boundaries, but using these counter sampling buffers would offer much more reliable timestamps and more granular control.

Is this on the roadmap? If so, when do you think this will be implemented in Molten?

further info:
https://developer.apple.com/documentation/metal/counter_sampling

billhollings:
- Add `MVKPhysicalDeviceMetalFeatures::counterSamplingPoints` to track platform availability of GPU counters.
- `MVKPhysicalDevice` creates and manages `MTLCounterSets` and checks for and enables flags within `MVKPhysicalDeviceMetalFeatures::counterSamplingPoints`.
- Add abstract `MVKGPUCounterQueryPool` class as parent of `MVKTimestampQueryPool` and `MVKPipelineStatisticsQueryPool` concrete classes and refactor access to host and command copy tracking data to allow extraction from `MTLCounterSampleBuffer`.
- `MVKTimestampQueryPool` uses `MTLCounterSampleBuffer` if supported, otherwise reverts to using host data for timestamps.
- `MVKCommandEncoder` encodes Vulkan timestamp commands either as Metal staged or command timestamps, depending on whether the GPU is tile-based or immediate-mode.
- For Metal stage counters, we use a light-weight dummy BLIT encoder to mark timestamp commands executed in the previous Metal encoding pass.
- Add `MVKDevice::getDummyBlitMTLBuffer()` to supply a dummy single-byte buffer that can be used by a stand-alone MTLBlitCommandEncoder as dummy work to mark timestamps.

Fixes #773.
Fixes #793.
Fixes #520.


billhollings:
@cdavis5e Please review if you can. Always good to have a second set of eyes on it...but if you are too busy to right now...let me know and I'll pull it in without the review. We can tweak later if needed. CTS tests pass.

MoltenVK/MoltenVK/GPUObjects/MVKQueryPool.mm
```diff
@@ -357,11 +327,131 @@
...
+		tsDesc.counterSet = mtlCounterSet;
+		tsDesc.storageMode = MTLStorageModeShared;
+		tsDesc.sampleCount = pCreateInfo->queryCount;
+
+		NSError* err = nil;
+		_mtlCounterBuffer = [getMTLDevice() newCounterSampleBufferWithDescriptor: tsDesc error: &err];
+		if (err) {
+			setConfigurationResult(reportError(VK_ERROR_INITIALIZATION_FAILED,
+											   "Could not create MTLCounterSampleBuffer for query pool of type VK_QUERY_TYPE_TIMESTAMP. Reverting to emulated GPU timestamps. (Error code %li): %s",
+											   (long)err.code, err.localizedDescription.UTF8String));
```
cdavis5e, line 352: What happens when we support other types of counter sets, like for pipeline statistics queries?

cdavis5e (changes requested):

MoltenVK/MoltenVK/GPUObjects/MVKQueryPool.mm
```diff
@@ -285,26 +258,25 @@
...
+								length: queryCount * kMVKQuerySlotSizeInBytes
+						  freeWhenDone: false];
 
-    if (shouldOutput64Bit) {
-        *(uint64_t*)pQryData = *pData;
-    } else {
-        *(uint32_t*)pQryData = (uint32_t)(*pData);
-    }
+
+	return [NSData dataWithBytesNoCopy: vizBuff.contents length: vizBuff.length freeWhenDone: false];
```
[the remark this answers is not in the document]
billhollings: Nothing of course. Just leftover code from dev. 

But WTF? Strange that this doesn't trigger an `Unreachable Code` warning.

Thanks for the catch.  Fixed.

MoltenVK/MoltenVK/GPUObjects/MVKQueryPool.mm
billhollings: Thanks. Made it more generic now.

cdavis5e (approved):

PR #1408, opened 2021-07-26
Properly ignore non-null pipeline creation pointers that should be ignored.

billhollings:
See for reference spec description of `VkGraphicsPipelineCreateInfo`, and CTS test
`dEQP-VK.api.pipeline.pipeline_invalid_pointers_unused_structs.graphics`

Struct pointers that should be ignored under certain conditions, are 
permitted to contain non-null garbage according to the spec and tests.

### DIFF
--- a/MoltenVK/MoltenVK/GPUObjects/MVKPipeline.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKPipeline.mm
@@ -1452,7 +1452,7 @@ void MVKGraphicsPipeline::addFragmentOutputToPipeline(MTLRenderPipelineDescripto
 												: mvkMTLPrimitiveTopologyClassFromVkPrimitiveTopology(pCreateInfo->pInputAssemblyState->topology);
 	}
 
-    // Color attachments - must ignore bad pColorBlendState pointer if rasterization is disable or subpass has not color attachments
+    // Color attachments - must ignore bad pColorBlendState pointer if rasterization is disabled or subpass has no color attachments
     uint32_t caCnt = 0;
     if (_isRasterizingColor && pCreateInfo->pColorBlendState) {
         for (uint32_t caIdx = 0; caIdx < pCreateInfo->pColorBlendState->attachmentCount; caIdx++) {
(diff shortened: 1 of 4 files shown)


MoltenVK/MoltenVK/GPUObjects/MVKPipeline.mm
```diff
@@ -1444,9 +1452,9 @@ static MTLVertexFormat mvkAdjustFormatVectorToSize(MTLVertexFormat format, uint3
 												: mvkMTLPrimitiveTopologyClassFromVkPrimitiveTopology(pCreateInfo->pInputAssemblyState->topology);
 	}
 
-    // Color attachments
+    // Color attachments - must ignore bad pColorBlendState pointer if rasterization is disable or subpass has not color attachments
```
cdavis5e, line 1455:
```suggestion
    // Color attachments - must ignore bad pColorBlendState pointer if rasterization is disabled or subpass has no color attachments
```

MoltenVK/MoltenVK/GPUObjects/MVKRenderPass.mm
```diff
@@ -35,6 +35,17 @@
...
+	for (auto& ca : _colorAttachments) {
+		if (ca.attachment != VK_ATTACHMENT_UNUSED) { return true; }
+	}
+	return false;
+}
+
+bool MVKRenderSubpass::hasDepthStencilAttachment() {
+	return _depthStencilAttachment.attachment != VK_ATTACHMENT_UNUSED;
+}
+
```
cdavis5e, line 48: This is simple enough to be inlined.

cdavis5e (changes requested):
Looks good. Just a few minor nits.

MoltenVK/MoltenVK/GPUObjects/MVKPipeline.mm
billhollings: Fixed.

MoltenVK/MoltenVK/GPUObjects/MVKRenderPass.mm
billhollings: Fixed.

cdavis5e (approved):

PR #1415, opened 2021-07-30
Fix pipeline barriers not working inside self-dependent subpasses on Apple GPUs.

billhollings:
On Apple GPUs, `MVKCmdPipelineBarrier` restarts Metal renderpass when encountered inside self-dependent Vulkan subpass where the same attachment acts as both input attachment and render attachment.

Also...`MVKPhysicalDeviceMetalFeatures::memoryBarriers` and `::textureBarriers` are never enabled on Apple Silicon. Remove it as an possibility in `MVKDevice`.

### DIFF
--- a/Docs/Whats_New.md
+++ b/Docs/Whats_New.md
@@ -23,6 +23,7 @@ Released TBD
 - Fix issue with `vkCmdBlitImage()` from compressed textures.
 - Fix incorrect translation of clear color values on Apple Silicon.
 - Fix swizzle of depth and stencil values into RGBA (`float4`) variable in shaders.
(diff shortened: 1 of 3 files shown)


cdavis5e (approved):
Valve is reportedly working on an extension that might be of interest here. Maybe we should keep an eye on it.

billhollings:
>Valve is reportedly working on an extension that might be of interest here. Maybe we should keep an eye on it.

Interesting. What's the nature of that extension, and its connection to this?

cdavis5e:
I can't say much here, because it's still inside Khronos, but you can see it [here](https://gitlab.khronos.org/vulkan/vulkan/-/merge_requests/4469).

PR #1416, opened 2021-08-03
Support resolving attachments with formats that Metal does not natively resolve.

billhollings:
Metal does not support resolving all formats that support MSAA, whereas Vulkan
assumes any MSAA format can be resolved. We fix that by running an optional
post-renderpass compute shader that resolves such textures by simply taking the
first sample as the resolved sample. This works to fix all failing CTS tests,
because such formats are all integer formats, and Vulkan allows an arbitrary
single sample value to be selected.

If we need to resolve, but the Metal format doesn't support it,
cause the Metal renderpass to store the MSAA attachment results.

`MVKRenderSubpass` don't establish Metal resolve attachment textures if format
is not natively resolvable, and encode Metal renderpass store actions accordingly.
`MVKCommandEncodingPool` add `MTLComputePipelineStates` to run simple resolve
compute shaders on attachments that cannot be resolved in Metal renderpass.
Add `MVKRenderSubpass::resolveUnresolvableAttachments()` and call from
`MVKCommandEncoder::endMetalRenderEncoding()`, before subpass index is updated.
Rename `MVKCommandEncodingPool::getClearStateIndex()` to
`getRenderpassLoadStoreStateIndex()` and remove `MVK_MACOS` restriction on
clearing shaders to allow compatibility with resolve shader handling.
`MVKRenderPassAttachment` remove validation of whether a format can be resolved.
`MVKPixelFormats::getMTLTextureUsage()` add read and write usage as appropriate
to allow compute shader to run to resolve formats not natively resolvable.
`MVKPixelFormats` remove obsolete unit test code.
`MVKImageView` clean up access functions and obsolete constructor
use of `MVKConfiguration`.


MoltenVK/MoltenVK/GPUObjects/MVKRenderPass.h
```diff
@@ -130,12 +130,13 @@ class MVKRenderSubpass : public MVKBaseObject {
 							const MVKArrayRef<MVKImageView*>& attachments,
 							bool storeOverride = false);
 
-	/** Constructs an instance for the specified parent renderpass. */
+	/** Resolves any resolve attachments that cannot be handled by native Metal subpass resolve behavior. */
+	void resolveUnresolvableAttachments(MVKCommandEncoder* cmdEncoder, const MVKArrayRef<MVKImageView*>& attachments);
```
cdavis5e, line 134: Since an `MVKArrayRef<T>` is just a pointer and a size, passing it around by reference doesn't buy much. It might actually hurt performance, since cache lines might be evicted to make room for the `MVKArrayRef` in memory. For this reason, LLVM passes its `ArrayRef`s by value. I think we should consider passing this by value instead of by reference in the future.

cdavis5e (approved):
Looks good. Just one note for the future.

MoltenVK/MoltenVK/GPUObjects/MVKRenderPass.h
billhollings: Thanks. Interesting technical detail observation.

No time like the present! I've changed all existing calls passing `MVKArrayRef` to use pass-by-value.

Scripts/runcts
```diff
@@ -99,7 +99,7 @@ export METAL_ERROR_MODE=3
 export METAL_DEBUG_ERROR_MODE=3
 
 # ----- MoltenVK config settings ------
-export MVK_CONFIG_LOG_LEVEL=0
+export MVK_CONFIG_LOG_LEVEL=1
```
cdavis5e, line 102: Looks like an extraneous change slipped in here. I guess it could be useful to see the logs to figure out why a test fails.

cdavis5e (approved):

Scripts/runcts
billhollings: NP. Thanks for noticing. But this was intentional...but unrelated. See last line in second commit.

Just included it because I have it set during all my testing anyway.

Issue #1180, opened 2020-12-14
New VK_ERROR_DEVICE_LOST crash with MoltenVK 1.1.1


richard-lunarg:
I'm getting a crash introduced via MoltenVK 1.1.1 on NVidia GPU's (macOS 10.15.7). The crash does not occur when I select the Intel GPU, and it does not occur with either GPU with MoltenVK 1.1.0

[mvk-error] VK_ERROR_DEVICE_LOST: Command buffer 0x7ffde5d89180 "" execution failed (code 1): Internal Error (IOAF code -536870211)
libc++abi.dylib: terminating with uncaught exception of type std::__1::system_error: mutex lock failed: Invalid argument
12:34:04: The program has unexpectedly finished.

billhollings:
@richard-lunarg Are you able to provide a call stack at the point of crash? You could try running your app in Debug mode from Xcode, or collect the call stack from the console.

richard-lunarg:
@billhollings 

From the crash reports, and probably the most useful bit of info:
Crashed Thread:        4  Dispatch queue: com.Metal.CompletionQueueDispatch

Exception Type:        EXC_CRASH (SIGABRT)
Exception Codes:       0x0000000000000000, 0x0000000000000000
Exception Note:        EXC_CORPSE_NOTIFY

Application Specific Information:
abort() called
terminating with uncaught exception of type std::__1::system_error: mutex lock failed: Invalid argument

Thread 0:: Dispatch queue: com.apple.main-thread
0   libsystem_kernel.dylib        	0x00007fff69933dfa mach_msg_trap + 10
1   libsystem_kernel.dylib        	0x00007fff69934170 mach_msg + 60
2   com.apple.QuartzCore          	0x00007fff3b22b825 CA::CA_BLOCKED_WAITING_FOR_CLIENT_FENCES(unsigned int, unsigned long) + 170
3   com.apple.QuartzCore          	0x00007fff3b20d72a CA::Context::commit_transaction(CA::Transaction*, double) + 1906
4   com.apple.QuartzCore          	0x00007fff3b20bcf0 CA::Transaction::commit() + 644
5   com.apple.AppKit              	0x00007fff2cb4bda1 __62+[CATransaction(NSCATransaction) NS_setFlushesWithDisplayLink]_block_invoke + 266
6   com.apple.AppKit              	0x00007fff2d26b080 ___NSRunLoopObserverCreateWithHandler_block_invoke + 41
[...]

PR #1418, opened 2021-08-10
Prefer MTLEvent for VkSemaphore, except on NVIDIA, prefer emulation.

billhollings:
This was deliberately constructed in two steps in two commits:

1) Prefer `MTLEvent` over `MTLFence`, except on NVIDIA. This still uses `MTLFence` on NVIDIA.

- Prefer MTLEvent over `MTLFence` for `VkSemaphore`, because `MTLEvent` handles sync across `MTLCommandBuffers` and `MTLCommandQueues`, except on NVIDIA GPUs, which have demonstrated trouble with `MTLEvents`, prefer `MTLFence`.
- Add `MVKDevice::VkSemaphoreStyle` enum to consolidate and simplify selection.

This provides behavior similar to existing code, except reverses the priority, but does not fix CTS tests, because `MTLFence` is not suitable for syncing across `MTLCommandBuffers` and `MTLCommandQueues`, so it really doesn't fulfil the Vulkan semaphore contract.

2) Prefer `MTLEvent` and use CPU sync emulation on NVIDIA.

- Disable `MVK_ALLOW_METAL_FENCES` by default.
- Disable use of `MTLEvent` on NVIDIA.
- By default, use `MTLEvent` for `VkSemaphore` everywhere except NVIDIA.
- By default, use CPU synchronization on NVIDIA.

This basically avoids using `MTLFence` at all (unless forced via config), and uses `MTLEvent` everywhere except NVIDIA, and CPU sync emulation with NVIDIA. This fulfills the Vulkan semaphore contract on all GPUs.

Essentially...the second one is where we need to be to satisfy CTS and Vulkan semaphore contract...but I left the option to back out to the first commit functionality if there is some reason that `MTLFences` are preferred over emulation.

This PR reverses and replaces PR #1219, but handles NVIDIA, and adds behavior to cover the discussion at the end of issue #1180, which triggered that PR #1219.

### DIFF
--- a/Docs/Whats_New.md
+++ b/Docs/Whats_New.md
@@ -18,6 +18,7 @@ MoltenVK 1.1.5
 
 Released TBD
 
+- Revert to prefer `MTLEvent` over `MTLFence` for `VkSemaphore`, except on NVIDIA.
 - Vulkan timestamp query pools use Metal GPU counters when available.
 - Support resolving attachments with formats that Metal does not natively resolve.
 - Fix issue where swapchain images were acquired out of order under heavy load.
--- a/Docs/Whats_New.md
+++ b/Docs/Whats_New.md
@@ -18,7 +18,9 @@ MoltenVK 1.1.5
 
 Released TBD
 
-- Revert to prefer `MTLEvent` over `MTLFence` for `VkSemaphore`, except on NVIDIA.
+- Changes to how `VkSemaphore` is supported:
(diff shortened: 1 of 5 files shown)


cdavis5e (approved):

cdavis5e:
Huh... the Mac Catalyst build seems to have hung.

billhollings:
> Huh... the Mac Catalyst build seems to have hung.

Thanks for noting. Meh...it appears to be a Heisenbug. As soon as I opened up the GitHub task Details page and started looking for where it was stuck...it started running again and was completed within 15 seconds. 🤦🏻‍♂️

PR #1426, opened 2021-08-23
Several renderpass clearing and input fixes from CTS

billhollings:
Apologies for the random grouping here. I probably should have done these as three separate PRs. Look at the individual commits for clarity.

1. Support stencil-only partial attachment clearing:

- Rename `MVKRenderPassAttachment::shouldUseClearAttachment()` to `shouldClearAttachment()`, pass whether testing for stencil clearing, and check stencil clearing distinct from depth clearing.
- Refactor `MVKRenderSubpass::populateClearAttachments()` to work with this.
- Remove `MVKRenderPassAttachment::getAttachmentStencilLoadOp()` as obsolete.

2. Fix GPU race condition when clearing a renderpass input attachment on Apple GPUs:

- Apple GPUs do not support rendering/writing to an attachment and then reading from that attachment within a single Metal renderpass. On Apple Silicon, restart the Metal renderpass if an input attachment is cleared inside renderpass.

3. Fix depth/stencil clearing broken during recent fix:
- A recent change for resolving color formats mistakenly broke clearing depth and stencil formats.

### DIFF
--- a/MoltenVK/MoltenVK/Utility/MVKFoundation.h
+++ b/MoltenVK/MoltenVK/Utility/MVKFoundation.h
@@ -63,7 +63,7 @@ typedef struct {
 
 /** Tracks the Vulkan command currently being used. */
 typedef enum : uint8_t {
-    kMVKCommandUseNone,                         /**< No use defined. */
+    kMVKCommandUseNone = 0,                     /**< No use defined. */
 	kMVKCommandUseEndCommandBuffer,             /**< vkEndCommandBuffer (prefilled VkCommandBuffer). */
     kMVKCommandUseQueueSubmit,                  /**< vkQueueSubmit. */
 	kMVKCommandUseAcquireNextImage,             /**< vkAcquireNextImageKHR. */
@@ -73,6 +73,8 @@ typedef enum : uint8_t {
 	kMVKCommandUseInvalidateMappedMemoryRanges, /**< vkInvalidateMappedMemoryRanges. */
     kMVKCommandUseBeginRenderPass,              /**< vkCmdBeginRenderPass. */
     kMVKCommandUseNextSubpass,                  /**< vkCmdNextSubpass. */
+	kMVKCommandUseBeginMultiViewPass,           /**< vkCmdBeginRenderPass. */
+	kMVKCommandUseRestartSubpass,               /**< Restart a subpass because of explicit or implicit barrier. */
     kMVKCommandUsePipelineBarrier,              /**< vkCmdPipelineBarrier. */
     kMVKCommandUseBlitImage,                    /**< vkCmdBlitImage. */
     kMVKCommandUseCopyImage,                    /**< vkCmdCopyImage. */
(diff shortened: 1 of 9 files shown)


MoltenVK/MoltenVK/Utility/MVKFoundation.h
```diff
@@ -73,6 +73,8 @@ typedef enum : uint8_t {
 	kMVKCommandUseInvalidateMappedMemoryRanges, /**< vkInvalidateMappedMemoryRanges. */
     kMVKCommandUseBeginRenderPass,              /**< vkCmdBeginRenderPass. */
     kMVKCommandUseNextSubpass,                  /**< vkCmdNextSubpass. */
+	kMVKCommandUseBeginMultiViewPass,           /**< vkCmdBeginRenderPass. */
```
cdavis5e, line 76: This does not appear to be used yet.

cdavis5e (changes requested):
Looks good, with one small issue.

One of these days, one of us ought to hook up shader framebuffer fetch to input attachments; then we wouldn't need to restart the render pass--at least, not for color attachments.

MoltenVK/MoltenVK/Utility/MVKFoundation.h
billhollings: Done. Good catch. Originally intended to differentiate multi-view from next subpass, but reverted.

billhollings:
>One of these days, one of us ought to hook up shader framebuffer fetch to input attachments; then we wouldn't need to restart the render pass--at least, not for color attachments.

Good plan. Does Metal guarantee `color(m)` as an input will have an implied barrier against previous outputs in the same Metal renderpass? I know that makes sense...and would be consistent with GL framebuffer-fetch behavior...but I haven't seen anything specific about that.

cdavis5e:
> > One of these days, one of us ought to hook up shader framebuffer fetch to input attachments; then we wouldn't need to restart the render pass--at least, not for color attachments.
> 
> Good plan. Does Metal guarantee `color(m)` as an input will have an implied barrier against previous outputs in the same Metal renderpass? I know that makes sense...and would be consistent with GL framebuffer-fetch behavior...but I haven't seen anything specific about that.

My understanding is that framebuffer fetch operates on tile memory; that's why it doesn't require an intermediate store to work on TBDR GPUs.

cdavis5e (approved):

mbarriault:
> Good plan. Does Metal guarantee color(m) as an input will have an implied barrier against previous outputs in the same Metal renderpass? I know that makes sense...and would be consistent with GL framebuffer-fetch behavior...but I haven't seen anything specific about that.

https://developer.apple.com/videos/play/tech-talks/604 About 2:25 in

> Metal guarantees that the results of draws issued before the dispatch are visible when that dispatch is executed. Likewise, Metal guarantees that the results of each dispatch are visible when the next draw or dispatch is executed. This synchronization guarantee allows race-free access to tile memory. No such guarantee is made between draws though.

Tile shader dispatches indeed insert an explicit barrier between draws before and after.

billhollings:
>My understanding is that framebuffer fetch operates on tile memory; that's why it doesn't require an intermediate store to work on TBDR GPUs.

Yeah...but it would still require precognition that a `color(m)` write needs to finish before a `color(m)` read is allowed to occur in a subsequent draw call. 

I can envision that could be ensured as part of the rendering path for the renderpass. But I can also envision a design where only the attachment writes are ensured to be ordered across draw calls.

It would be good to see it in writing. 😉

billhollings:
>No such guarantee is made between draws though.

Huh. They just toss that in as a throw-away comment before moving onto something else. But it has a large impact on framebuffer-fetch coherency across draw calls.

PR #1435, opened 2021-09-08
Several updates for CTS test fixes.

billhollings:
Support maximum point primitive size of 511.
Update to latest SPIRV-Cross version to add support
for `OpSpecConstantOp` ops `OpQuantizeToF16` and `OpSRem`.
Update MoltenVK version to 1.1.6.


MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
```diff
@@ -2065,7 +2065,7 @@
 	_properties.limits.timestampPeriod = mvkGetTimestampPeriod();
 
     _properties.limits.pointSizeRange[0] = 1;
-    _properties.limits.pointSizeRange[1] = 64;
+    _properties.limits.pointSizeRange[1] = 511;
```
cdavis5e, line 2068: The reason I had reduced this is that some cards support a lower limit than the 511 that Apple advertises. I know 511 is the limit for M1, but I think the limit for AMD cards as implemented is 64 (though it could be as large as 8192 based on the reported limits in OpenGL), and that for Intel is 256. I think we're gonna have to have a switch based on the vendor for this...

cdavis5e (changes requested):

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
billhollings: Bah. Looks like this is another victim of Apple's reliance on paper documentation for device capabilities. All GPU's are documented at a 511 point size.

Sorry...I should have checked the commit that set the value to `64` for an explanation, but it slipped my mind that you'd made that change until you just mentioned it here.

The CTS tests `dEQP-VK.rasterization.primitive_size.points.point_size_*` perform the reverse test...they attempt to render a large point, and then fail if it is larger than `pointSizeRange[1]`. 

So, unless we want to be in the business of artificially forcing smaller point sizes at render time,`pointSizeRange[1]` can't be smaller than the native point size limit to pass these tests.

At least the existence of these CTS tests will make it easy to determine the appropriate value for `pointSizeRange[1]` on each GPU. So, I'll set up a GPU-based config for `pointSizeRange[1]` as best I can for the GPU's I can test. Then we can continue to use these CTS tests to determine other GPU's as the are available for testing.

billhollings: Okay. Done and tested on M1, AMD & Intel.

cdavis5e (approved):

Issue #1445, opened 2021-10-03
Invalid SPIR-V binary version 1.5 for target environment SPIR-V 1.3


cfnptr:
I am getting a validation layer error when creating VkShaderModule from compiled shader with glslc --target-env=vulkan1.2. But instance is created with VkApplicationInfo apiVersion = VK_API_VERSION_1_2. Looks like this is a validation layer bug because the shader module is being created and rendering is working correctly.

Vulkan ERROR: Validation Error: [ UNASSIGNED-CoreValidation-Shader-InconsistentSpirv ] Object 0: handle = 0x7fdc9f069a18, type = VK_OBJECT_TYPE_DEVICE; | MessageID = 0x6bbb14 | SPIR-V module not valid: Invalid SPIR-V binary version 1.5 for target environment SPIR-V 1.3 (under Vulkan 1.1 semantics).

Vulkan API: 1.2.189
MacOS 11.6
GPU: Intel Iris Plus Graphics 645

cdavis5e:
We don't support Vulkan 1.2, so you can't use SPIR-V 1.5 yet.

The Vulkan loader is weird here. The *instance* Vulkan version may be 1.2 even as the *driver* Vulkan version is 1.1. That's what's probably happening here.

cfnptr:
> We don't support Vulkan 1.2, so you can't use SPIR-V 1.5 yet.
> 
> The Vulkan loader is weird here. The _instance_ Vulkan version may be 1.2 even as the _driver_ Vulkan version is 1.1. That's what's probably happening here.

Oh, ok, thank you. When do you plan to add 1.2 support?

cdavis5e:
> When do you plan to add 1.2 support?

Unfortunately, it's a ways off, especially since we're busy with trying to fix test conformance with 1.1. Since 1.2 also requires `KHR_buffer_device_address` support, it may never happen.

PR #1446, opened 2021-10-03
Add unsupported VkApplicationInfo->apiVersion warning

cfnptr:
Related to this #1445 issue.

As declared in [spec](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/VkApplicationInfo.html) "Vulkan 1.1 or later must not return VK_ERROR_INCOMPATIBLE_DRIVER for any value of apiVersion". So I decided to check what a Nvidia driver does in this case, it shows validation warning and sets apiVersion to latest available version. I think it would be great to have similar actions in MoltenVK.

### DIFF
--- a/Common/MVKLogging.h
+++ b/Common/MVKLogging.h
@@ -35,9 +35,9 @@ extern "C" {
  * This library adds flexible, non-intrusive logging and assertion capabilities
  * that can be efficiently enabled or disabled via compiler switches.
  *
- * There are four levels of logging: Trace, Info, Error and Debug, and each can be enabled
- * independently via the MVK_LOG_LEVEL_TRACE, MVK_LOG_LEVEL_INFO, MVK_LOG_LEVEL_ERROR and
- * MVK_LOG_LEVEL_DEBUG switches, respectively.
+ * There are five levels of logging: Trace, Info, Warning, Error and Debug, and each can be enabled
+ * independently via the MVK_LOG_LEVEL_TRACE, MVK_LOG_LEVEL_INFO, MVK_LOG_LEVEL_WARNING,
+ * MVK_LOG_LEVEL_ERROR and MVK_LOG_LEVEL_DEBUG switches, respectively.
  *
  * ALL logging can be enabled or disabled via the MVK_LOGGING_ENABLED switch.
  *
@@ -59,12 +59,17 @@ extern "C" {
  *		MVKLogErrorIf(cond, fmt, ...)	- same as MVKLogError if boolean "cond" condition expression evaluates to YES,
  *										  otherwise logs nothing.
  *
+ *      MVKLogWarning(fmt, ...)              - recommended for not immediately harmful errors
+ *                                      - will print if MVK_LOG_LEVEL_WARNING is set on.
+ *      MVKLogWarningIf(cond, fmt, ...)  - same as MVKLogWarning if boolean "cond" condition expression evaluates to YES,
+ *                                        otherwise logs nothing.
+ *
  *		MVKLogInfo(fmt, ...)			- recommended for general, infrequent, information messages
  *										- will print if MVK_LOG_LEVEL_INFO is set on.
  *		MVKLogInfoIf(cond, fmt, ...)	- same as MVKLogInfo if boolean "cond" condition expression evaluates to YES,
  *										  otherwise logs nothing.
  *
- *		MVKLogDebug(fmt, ...)			- recommended for temporary use during debugging
+ *		MVKLogDebug(fmt, ...)		- recommended for temporary use during debugging
  *										- will print if MVK_LOG_LEVEL_DEBUG is set on.
  *		MVKLogDebugIf(cond, fmt, ...)	- same as MVKLogDebug if boolean "cond" condition expression evaluates to YES,
  *										  otherwise logs nothing.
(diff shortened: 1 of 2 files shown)


MoltenVK/MoltenVK/GPUObjects/MVKInstance.mm
```diff
@@ -334,7 +334,17 @@
 	initDebugCallbacks(pCreateInfo);	// Do before any creation activities
 
 	mvkSetOrClear(&_appInfo, pCreateInfo->pApplicationInfo);
-	if (_appInfo.apiVersion == 0) { _appInfo.apiVersion = VK_API_VERSION_1_0; }	// Default
+    
+    if (_appInfo.apiVersion == 0) {
+        _appInfo.apiVersion = VK_API_VERSION_1_0;   // Default
+    }
+    else if (VK_API_VERSION_MAJOR(_appInfo.apiVersion) > MVK_VERSION_MAJOR ||
+             VK_API_VERSION_MINOR(_appInfo.apiVersion) > MVK_VERSION_MINOR) {
```
cdavis5e, line 342: This feels wrong. Probably because it doesn't handle the case where `apiVersion` is nonzero but `VK_API_VERSION_MAJOR(apiVersion)` is zero--even though that shouldn't happen in practice.

It should just be possible to compare `_appInfo.apiVersion` with `MVK_VULKAN_API_VERSION` anyway. That includes the header revision. That way, if an app specifies an `apiVersion` with a greater header revision, the warning will be generated, which would serve as a good reminder to upgrade MoltenVK.

MoltenVK/MoltenVK/GPUObjects/MVKInstance.mm
```diff
@@ -334,7 +334,17 @@
...
+        reportMessage(ASL_LEVEL_WARNING,
+            "Unsupported MoltenVK API version. Assuming VK_API_VERSION_1_1.");
```
cdavis5e, line 344: This should use... wait, there's no `MVKLogWarning()`. Huh. Maybe we should add one.

cdavis5e (changes requested):

MoltenVK/MoltenVK/GPUObjects/MVKInstance.mm
cfnptr: Yeah, I agree with you, we also need to handle incorrect apiVersion. What warning message do you think is better to display?

MoltenVK/MoltenVK/GPUObjects/MVKInstance.mm
cfnptr: I have added it c881389 :)

cfnptr:
I have compiled and tested the changes, now messages are being logged to the log and validation layer.

MoltenVK/MoltenVK/GPUObjects/MVKInstance.mm
```diff
@@ -334,7 +334,13 @@
 	initDebugCallbacks(pCreateInfo);	// Do before any creation activities
 
 	mvkSetOrClear(&_appInfo, pCreateInfo->pApplicationInfo);
-	if (_appInfo.apiVersion == 0) { _appInfo.apiVersion = VK_API_VERSION_1_0; }	// Default
+    
+    if (_appInfo.apiVersion > MVK_VULKAN_API_VERSION) {
```
billhollings, line 338: The spec dictates we should ignore patch version. An app coming in with a higher patch version would fail this test inappropriately. Should we apply `MVK_VULKAN_API_VERSION_CONFORM()` to both sides of this test? 

And the warning message indicates we're going to assume MoltenVK's level. Should we not be correcting `_appInfo` to that value, in this case?

billhollings (changes requested):
Change requested to API level test. Plus some discussion questions about setting the assumed API value, and considering logging config levels.

MoltenVK/MoltenVK/GPUObjects/MVKInstance.mm
cfnptr: > Should we apply MVK_VULKAN_API_VERSION_CONFORM() to both sides of this test?

Yes, that would be more correct.

> Should we not be correcting _appInfo to that value, in this case?

In my opinion, if we are already correcting the smaller value, then the bigger should also be corrected.

Common/MVKLogging.h
```diff
@@ -168,6 +185,7 @@ extern "C" {
 #endif
 
 #define MVKLogErrorImpl(fmt, ...)		reportMessage(ASL_LEVEL_ERR, fmt, ##__VA_ARGS__)
+#define MVKLogWarningImpl(fmt, ...)     reportMessage(ASL_LEVEL_WARNING, fmt, ##__VA_ARGS__)
```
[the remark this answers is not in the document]
cfnptr: > For now, if we're happy not logging warnings when configured for error only

Yeah, I think it would be ok if we don't log the warnings when configured for error only, because Vulkan warnings actually indicates not immediately harmful problems. I hope I haven't added unnecessary problems to you with refactoring the logging system :sweat_smile:.

MoltenVK/MoltenVK/GPUObjects/MVKInstance.mm
billhollings: >In my opinion, if we are already correcting the zero value, then the bigger should also be corrected.

Agreed. Are you able to fix that here?

cfnptr: Of course! :) fae1bb53

PR #1467, opened 2021-11-09
Restore support for BC1_RGB compressed formats.

### LINKED ISSUE #1461: Restore BC1_RGB support
Advertised support for format `BC1_RGB` was removed in PR #1460.

Comment from @cdavis5e at the time:

>Currently, we only use MoltenVK to support D3D10+ graphics, so I guess this is OK for now. At some point though, we'll also want to move to using MoltenVK for D3D9, at which point we'll need the BC1_RGB formats. 

Suggestions in that PR for adding  support back include:

> piggy-backing on existing dynamic decompression options

> Have you considered using a swizzle to support this? If swizzling is supported natively, this should be really cheap.

billhollings:
For Vulkan `BC1_RGB` formats, swizzle alpha of substituted
Metal `BC1_RGBA` to `1.0`, to return value expected by Vulkan.

Fixes #1461 and reverses and improves PR #1460.

### DIFF
--- a/MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
@@ -1489,6 +1489,19 @@ VkResult MVKImageViewPlane::initSwizzledMTLPixelFormat(const VkImageViewCreateIn
 	_componentSwizzle = pCreateInfo->components;
 	VkImageAspectFlags aspectMask = pCreateInfo->subresourceRange.aspectMask;
 
+	// Use swizzle adjustment to bridge some differences between Vulkan and Metal pixel formats.
+	// Do this ahead of other tests and adjustments so that swizzling will be enabled by tests below.
+	switch (pCreateInfo->format) {
+		case VK_FORMAT_BC1_RGB_UNORM_BLOCK:
+		case VK_FORMAT_BC1_RGB_SRGB_BLOCK:
+			// Metal doesn't support BC1_RGB, so force substituted BC1_RGBA alpha to 1.0.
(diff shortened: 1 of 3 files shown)


MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
```diff
@@ -1489,6 +1489,19 @@
 	_componentSwizzle = pCreateInfo->components;
 	VkImageAspectFlags aspectMask = pCreateInfo->subresourceRange.aspectMask;
 
+	// Use swizzle adjustment to bridge some differences between Vulkan and Metal pixel formats.
+	// Do this ahead of other tests and adjustments so that swizzling will be enabled by tests below.
+	switch (pCreateInfo->format) {
+		case VK_FORMAT_BC1_RGB_UNORM_BLOCK:
+		case VK_FORMAT_BC1_RGB_SRGB_BLOCK:
+			// Metal doesn't support BC1_RGB, so force substituted BC1_RGBA alpha to 1.0.
+			_componentSwizzle.a = VK_COMPONENT_SWIZZLE_ONE;
```
cdavis5e, line 1498: What happens if the swizzle's alpha component isn't `IDENTITY` or `A`? For that matter, what should happen if some other swizzle component is set to `A` here?

cdavis5e (changes requested):

MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
billhollings: Ah! Good catch. Thanks.

Updated to deal with both scenarios.

cdavis5e (approved):

PR #1478, opened 2021-11-17
Support building MoltenVK with static Vulkan linkage symbols hidden.

billhollings:
Add build environment variable `MVK_HIDE_VULKAN_SYMBOLS`. to allow MoltenVK to be built with static Vulkan API symbols hidden, to avoid library linking conflicts when bound to a Vulkan Loader that also exports identical symbols.

The default value of `MVK_HIDE_VULKAN_SYMBOLS` is `0`, meaning Vulkan static symbols are exposed by default.

Add `MVK_PUBLIC_VULKAN_SYMBOL` directive to mark each Vulkan call symbols for exporting or hiding.

Update the MoltenVK Xcode project to add the `MVK_HIDE_VULKAN_SYMBOLS` build setting, and set the `ENABLE_TESTABILITY` build setting to `NO`, because it conflicts with stripping symbols.

Update MoltenVK version to 1.1.7.


cdavis5e (approved):
We at CodeWeavers currently use MoltenVK with no native macOS loader. Wine, of course, has its own loader which handles interfacing with the host Vulkan driver.

PR #1489, opened 2021-12-13
Update to latest version of VK_EXT_metal_objects extension

billhollings:
- Support retrieving `MTLTexture` from `VkImageView` and `VkBufferView`.
- Document support for `VK_EXT_metal_objects` in `MoltenVK_Runtime_UserGuide.md`.


billhollings:
It is expected that CI fails, because the Vulkan-Headers do not include the VK_EXT_metal_objects extension yet.

Issue #1487, opened 2021-12-03
MVKRenderSubpass::_mtlDummyTex not released


billhollings:
`MVKRenderSubpass::_mtlDummyTex` not released on object destruction, or before being created again each time `MVKRenderSubpass::populateMTLRenderPassDescriptor()` is called. This is likely creating memory leaks whenever a render pass is begun.

billhollings:
Fixed in PR #1494. Please retest with latest code and close this issue if fixed.

billhollings:
Closing as fixed.

PR #1494, opened 2021-12-25
Fix memory leak of dummy MTLTexture in render subpasses that use no attachments.

billhollings:
Older Metal does not support rendering without subpass attachments. In this case, a dummy attachment with a dummy `MTLTexture` is created whenever the subpass begins, but was not being correctly used.

Move the creation, retaining, and releasing of the dummy MTLTexture to `MVKFramebuffer`, where the extent and layer count is known and can be reused.

Pass framebuffer to `MVKCommandEncoder::beginRenderpass()` and remember current framebuffer in `MVKCommandEncoder`.

Add `getFramebufferExtent()` and `getFramebufferLayerCount()` to `MVKCommandEncoder`.

Pass framebuffer to `MVKRenderSubpass:populateMTLRenderPassDescriptor()` and retrieve dummy `MTLTexture` from framebuffer.

Fixes issue #1487.

### DIFF
--- a/MoltenVK/MoltenVK/GPUObjects/MVKFramebuffer.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKFramebuffer.mm
@@ -17,11 +17,68 @@
  */
 
 #include "MVKFramebuffer.h"
+#include "MVKRenderPass.h"
+
+using namespace std;
 
 
 #pragma mark MVKFramebuffer
 
-#pragma mark Construction
+id<MTLTexture> MVKFramebuffer::getDummyAttachmentMTLTexture(MVKRenderSubpass* subpass, uint32_t passIdx) {
+	if (_mtlDummyTex) { return _mtlDummyTex; }
+
+	// Lock and check again in case another thread has created the texture.
+	lock_guard<mutex> lock(_lock);
+	if (_mtlDummyTex) { return _mtlDummyTex; }
+
+	VkExtent2D fbExtent = getExtent2D();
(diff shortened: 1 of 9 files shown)


MoltenVK/MoltenVK/GPUObjects/MVKFramebuffer.mm
```diff
@@ -17,11 +17,68 @@
...
+#else
+		mtlTexDesc.textureType = MTLTextureType2DArray;
+#endif
+		mtlTexDesc.arrayLength = fbLayerCount;
+	} else if (sampleCount > 1) {
+		mtlTexDesc.textureType = MTLTextureType2DMultisample;
+		mtlTexDesc.sampleCount = sampleCount;
+	}
+#if MVK_IOS
+	if ([_renderPass->getMTLDevice() supportsFeatureSet: MTLFeatureSet_iOS_GPUFamily1_v3]) {
```
cdavis5e, line 67: Copy-pasto here.

cdavis5e (changes requested):
Mostly looks good, but...

MoltenVK/MoltenVK/GPUObjects/MVKFramebuffer.mm
billhollings: Sigh. That's what I get for being rushed and lazy and forgetting to do a full build before submitting.

Fixed now.

cdavis5e (approved):

PR #1497, opened 2021-12-27
On Apple Silicon, set VkPhysicalDeviceProperties::deviceID from GPU capabilities.

### LINKED ISSUE #1486: Can we change the form of VkPhysicalDeviceProperties::deviceID for Apple GPUs?
For Apple GPU's, MoltenVK currently sets `VkPhysicalDeviceProperties::deviceID` in the hex form `0xaMMX`, where `a` indicates an Apple SoC, `MM` is the SoC model number (08, 10...) and `X` is 1 for X version, 0 for other. The original idea was to try to map to the `A8, A10X...A15` nomenclature Apple was using for their SoCs, for transparent identification.

With the introduction of the M1 series of processors under macOS, and now iOS (recent iPad Pro), this naming structure gets confusing. For example, MoltenVK currently reports an M1 macOS as `0xa140`, because it supports the same functionality as the A14 GPU.

Apple distinguishes GPUs based on capabilities using the `MTLGPUFamily` enumeration (`MTLGPUFamilyApple7`, `MTLGPUFamilyApple6`, ...), and these apply to all platforms. For instance, both the `A14` and `M1`GPUs support `MTLGPUFamilyApple7 `.

Some options for moving forward with the `VkPhysicalDeviceProperties ::deviceID` nomenclature:

1. Leave it alone and continue to number just based on historic iOS `AXX` SoC's, even for the M1 series.

2. Introduce a second type of nomenclature for the M1 family (eg. `0xb010`). But this becomes arbitrary, and although we can assume M1 on macOS, it's not clear how we could distinguish M1 from say A14 on iPad Pro, without heroic efforts to extract system info.

3. Scrap the current nomenclature, and replace it with one simply based on the `MTLGPUFamily` enumeration, since that indicates relative GPU functionality, and is consistent on all platforms. For instance, we could create a `deviceID` that is basically just the number of the GPU functionality enumeration (`MTLGPUFamilyApple7` -> `0x7`...or maybe `0xad7` for Apple Device 7).

The third option is preferred because it is trivial to implement, is consistent across all platforms and GPU's, and maps to they way Apple identifies and distinguishes GPU functionality.

Does anyone use `VkPhysicalDeviceProperties::deviceID` in a way that will break if we make this change? Or is there reason to try to perform system query heroics to separately identify individual GPUs beyond Apple's `MTLGPUFamily` functionality designations?

billhollings:
Previously, on Apple Silicon (iOS, tvOS & macOS M1), we tried to guess `deviceID` from GPU parameters, but this is becoming harder as the types of Apple Silicon is growing, and the actual device SoC itself is less relevant that the GPU capabilities. So we now set deviceID from the combination of OS version and GPU type.

Rename `MVKDevice::getHighestMTLFeatureSet()` to `getHighestGPUCapability()`.

Fixes #1486.

### DIFF
--- a/Docs/Whats_New.md
+++ b/Docs/Whats_New.md
@@ -20,6 +20,7 @@ Released TBD
 
 - Support building MoltenVK with static Vulkan linkage symbols hidden.
 - Do not use `MTLEvent` for `VkSemaphore` under *Rosetta2*.
+- On *Apple Silicon (iOS, tvOS & macOS)*, set `VkPhysicalDeviceProperties::deviceID` from GPU capabilities.
 - Support compiling *MSL 2.4* in runtime pipelines and `MoltenVKShaderConverterTool`.
 - Fix issue where *MSL 2.3* only available on *Apple Silicon*, even on *macOS*.
 - Fix memory leak of dummy `MTLTexture` in render subpasses that use no attachments.
--- a/MoltenVK/MoltenVK/GPUObjects/MVKDevice.h
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKDevice.h
@@ -374,7 +374,7 @@ protected:
 	void initCounterSets();
 	MVKArrayRef<MVKQueueFamily*> getQueueFamilies();
 	void initPipelineCacheUUID();
-	uint32_t getHighestMTLFeatureSet();
+	uint32_t getHighestGPUCapability();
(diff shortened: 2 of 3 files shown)


cdavis5e (changes requested):
One tiny change.

cdavis5e (approved):

Issue #1438, opened 2021-09-18
Timestamp issues on MacBook Air M1


SRSaunders:
After a bit of investigation, I have found two things that may be causing problems with timestamps on Apple Silicon:

1. On my MacBook Air M1, using SDK 1.2.189.0, `vkcontext.gpu->props.limits.timestampPeriod = 41.6667`.  Looking at Sascha Willems' Vulkan hardware database (https://vulkan.gpuinfo.org), earlier SDKs showed a `timestampPeriod = 1` for M1 machines.  Since the RBDOOM-3-BFG game uses `timestampPeriod` to calibrate its timing intervals, this would lead to longer than expected times in the stats counter for the game.  This is what I am seeing.  Has something changed regarding `timestampPeriod` in the newer MoltenVK builds?

2. A second issue on the M1 appears to be revealed by calling `vkCmdWriteTimestamp()` back-to-back with no additional commands in between.  There is at least one situation in the RBDOOM-3-BFG game when a rendering stage is skipped for Vulkan (but not for OpenGL), and the following commands are issued:
```
  vkCmdWriteTimestamp( commandBuffer, VK_PIPELINE_STAGE_TOP_OF_PIPE_BIT, queryPool, queryIndex1 );

  <...some trivial cpu logic, but no additional GPU commands when using the Vulkan renderer...>

  vkCmdWriteTimestamp( commandBuffer, VK_PIPELINE_STAGE_BOTTOM_OF_PIPE_BIT, queryPool, queryIndex2 );

  <...and then later on...>

  vkGetQueryPoolResults( vkcontext.device, queryPool, MRB_GPU_TIME, numQueries, results.ByteSize(), results.Ptr(), sizeof( uint64 ), VK_QUERY_RESULT_64_BIT | VK_QUERY_RESULT_WAIT_BIT );
```
For my Intel Mac with Radeon 580 graphics, MoltenVK correctly returns identical timestamps when queried later (i.e. zero time difference).  However, on the M1 this appears to be a problem and the returned timestamps are different, leading to a incorrect non-zero time interval for this situation.

billhollings:
Thanks for submitting this.

[...]

PR #1500, opened 2021-12-28
Improve accuracy of VkPhysicalDeviceLimits::timestampPeriod.

billhollings:
If using GPU counters, on all Apple GPUs lock `timestampPeriod` to 1.0, since Apple GPUs use nanoseconds, and on non-Apple GPUs, dynamically adapt value of `timestampPeriod` by correlating GPU ticks with GPU ticks.

If using CPU sync, set `timestampPeriod` to OS CPU timestamp tick period.

This is a partial fix for issue #1438.

### DIFF
--- a/MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
@@ -1101,6 +1101,25 @@ VkResult MVKPhysicalDevice::getQueueFamilyProperties(uint32_t* pCount,
 	return rslt;
 }
 
+// Don't need to do this for Apple GPUs, where the GPU and CPU timestamps
+// are the same, or if we're not using GPU timestamp counters.
+void MVKPhysicalDevice::startTimestampCorrelation(MTLTimestamp& cpuStart, MTLTimestamp& gpuStart) {
+	if (_properties.vendorID == kAppleVendorId || !_timestampMTLCounterSet) { return; }
+	[_mtlDevice sampleTimestamps: &cpuStart gpuTimestamp: &gpuStart];
+}
+
+// Don't need to do this for Apple GPUs, where the GPU and CPU timestamps
+// are the same, or if we're not using GPU timestamp counters.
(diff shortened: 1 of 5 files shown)


cdavis5e (changes requested):
The type `MTLTimestamp` isn't defined in the Catalina SDK. You'll have to enclose all the uses of it in `MVK_XCODE_12` blocks.

billhollings:
> The type `MTLTimestamp` isn't defined in the Catalina SDK. You'll have to enclose all the uses of it in `MVK_XCODE_12` blocks.

Yeah. Found that out the hard way. Fix in place now. Defined `MTLTimestamp` as `NSUInteger`, per Xcode 11.7 SDK.

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
```diff
@@ -1101,6 +1101,25 @@
 	return rslt;
 }
 
+// Don't need to do this for Apple GPUs, where the GPU and CPU timestamps
+// are the same, or if we're not using GPU timestamp counters.
+void MVKPhysicalDevice::startTimestampCorrelation(MTLTimestamp& cpuStart, MTLTimestamp& gpuStart) {
+	if (_properties.vendorID == kAppleVendorId || !_timestampMTLCounterSet) { return; }
+	[_mtlDevice sampleTimestamps: &cpuStart gpuTimestamp: &gpuStart];
```
cdavis5e, line 1108: You also have to check the OS version before using this method. I don't think it's supported prior to Catalina.

billhollings: > You also have to check the OS version before using this method. I don't think it's supported prior to Catalina.

That should be covered by the `!_timestampMTLCounterSet` guard code. Both `counterSamplingPoints` and `_timestampMTLCounterSet` will be empty prior to 10.15.

cdavis5e: Ah, I see. All right, then.

cdavis5e (approved):

PR #1531, opened 2022-02-24
Don't attempt to store the depth component of a stencil-only renderpass attachment.

billhollings:
For a combined depth-stencil format in an `MVKImageView` attachment with `VK_IMAGE_ASPECT_STENCIL_BIT`, the attachment format may have been swizzled to a stencil-only format. In this case, we want to guard against an attempt
to store the non-existent depth component.

Pass `MVKImageView` attachment to `MVKRenderPassAttachment::encodeStoreAction()` and `MVKRenderPassAttachment::populateMTLRenderPassAttachmentDescriptor()` to check attachment depth format component.

Consolidate calls to `MVKImageView::populateMTLRenderPassAttachmentDescriptor()` by calling it from within `MVKRenderPassAttachment::populateMTLRenderPassAttachmentDescriptor()`.

### DIFF
--- a/MoltenVK/MoltenVK/GPUObjects/MVKRenderPass.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKRenderPass.mm
@@ -237,13 +237,8 @@ void MVKRenderSubpass::populateMTLRenderPassDescriptor(MTLRenderPassDescriptor* 
 
             // Configure the color attachment
             MVKRenderPassAttachment* clrMVKRPAtt = &_renderPass->_attachments[clrRPAttIdx];
-			attachments[clrRPAttIdx]->populateMTLRenderPassAttachmentDescriptor(mtlColorAttDesc);
-			bool isMemorylessAttachment = false;
-#if MVK_APPLE_SILICON
-			isMemorylessAttachment = attachments[clrRPAttIdx]->getMTLTexture(0).storageMode == MTLStorageModeMemoryless;
-#endif
-			if (clrMVKRPAtt->populateMTLRenderPassAttachmentDescriptor(mtlColorAttDesc, this,
-                                                                       isRenderingEntireAttachment, isMemorylessAttachment,
+			if (clrMVKRPAtt->populateMTLRenderPassAttachmentDescriptor(mtlColorAttDesc, this, attachments[clrRPAttIdx],
+                                                                       isRenderingEntireAttachment,
                                                                        hasResolveAttachment, canResolveFormat,
 																	   false, loadOverride)) {
 				mtlColorAttDesc.clearColor = pixFmts->getMTLClearColor(clearValues[clrRPAttIdx], clrMVKRPAtt->getFormat());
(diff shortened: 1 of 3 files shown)


MoltenVK/MoltenVK/GPUObjects/MVKRenderPass.mm
```diff
@@ -637,28 +622,46 @@ static uint32_t getNextViewMaskGroup(uint32_t viewMask, uint32_t* startView, uin
...
+	MVKPixelFormats* pixFmts = _renderPass->getPixelFormats();
 	bool isDepthFormat = pixFmts->isDepthFormat(mtlFmt);
 	bool isStencilFormat = pixFmts->isStencilFormat(mtlFmt);
 	bool isColorFormat = !(isDepthFormat || isStencilFormat);
 
+	bool isMemorylessAttachment = false;
+#if MVK_APPLE_SILICON
+	isMemorylessAttachment = attachment->getMTLTexture().storageMode == MTLStorageModeMemoryless;
+#endif
+	MTLStoreAction storeAction = getMTLStoreAction(subpass, isRenderingEntireAttachment, isMemorylessAttachment, hasResolveAttachment, canResolveFormat, isStencil, storeOverride);
```
cdavis5e, line 663: Wait, where's the part where we set the depth store action to `DontCare`?

In fact, if the DS image only has a stencil component, why are we telling Metal it's a depth attachment at all?

cdavis5e (changes requested):

MoltenVK/MoltenVK/GPUObjects/MVKRenderPass.mm
billhollings: > Wait, where's the part where we set the depth store action to DontCare?

We don't. On a stencil-only `MVKImageView` attachment, `isDepthFormat` is `false`, and `setDepthStoreAction:` will not be called at all, leaving the store action as `Unknown`

The Metal error was not a validation error, but an `EXC_BAD_ACCESS` crash on the call to `setDepthStoreAction:`, so I thought it would be best to avoid calling it altogether, in case `setDepthStoreAction:` is blindly stomping on something that was not there. Metal seems happy to leave the store action at `Unknown` if there is not a component to do something with.

>In fact, if the DS image only has a stencil component, why are we telling Metal it's a depth attachment at all?

The _image_ has both components (as does the renderpass attachment). The _image view_ only has a stencil component, based on `VK_IMAGE_ASPECT_STENCIL_BIT`. The problem was that we were previously testing the renderpass attachment format, not the image view format.

billhollings: I _could_ change this to extract a `DontCare` and call `setDepthStoreAction:` with it, which would be consistent with the related behavior in `populateMTLRenderPassAttachmentDescriptor()`, assuming `setDepthStoreAction:` didn't choke on the call.

cdavis5e: > > In fact, if the DS image only has a stencil component, why are we telling Metal it's a depth attachment at all?
> 
> The _image_ has both components (as does the renderpass attachment). The _image view_ only has a stencil component, based on `VK_IMAGE_ASPECT_STENCIL_BIT`. The problem was that we were previously testing the renderpass attachment format, not the image view format.

Ah, OK.

cdavis5e: > I _could_ change this to extract a `DontCare` and call `setDepthStoreAction:` with it, which would be consistent with the related behavior in `populateMTLRenderPassAttachmentDescriptor()`, assuming `setDepthStoreAction:` didn't choke on the call.

No, it's fine.

cdavis5e (approved):

Issue #1538, opened 2022-03-06
vkCmdPipelineBarrier no longer works


tellowkrinkle:
Attempting to use `vkCmdPipelineBarrier` does not result in any barriers being issued to Metal

Broken by 18642002ce68fa3071885afbaf4fa42e6ea17ded

Specifically the change on `MVKDevice.mm:1450`, which makes it so `_metalFeatures.textureBarriers = true;` is only set on macOS 10.15 and lower

billhollings:
Fixed in PR #1540. Please retest and close this issue if it fixes the problem you encountered.

tellowkrinkle:
Can confirm it fixes the issue

PR #1540, opened 2022-03-08
Reinstate memory barriers on non-Apple GPUs.

billhollings:
Ensure non-Apple GPU's enable memory barriers.
A previous commit inadvertently disabled GPU memory barriers.

Change tests for memory barriers to runtime test for Apple GPU, instead of
build-time test for Apple Silicon, to accommodate running on Rosetta2, and
refactor tests for Apple Silicon and OS version on some macOS GPU feature settings.

Fixes issue #1538.

### DIFF
--- a/MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
@@ -1478,38 +1478,42 @@ void MVKPhysicalDevice::initMetalFeatures() {
 	}
 
 #if MVK_XCODE_12
-	if ( mvkOSVersionIsAtLeast(10.16) ) {
+	if ( mvkOSVersionIsAtLeast(11.0) ) {
 		_metalFeatures.mslVersionEnum = MTLLanguageVersion2_3;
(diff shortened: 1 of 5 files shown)


MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
```diff
@@ -1478,38 +1478,42 @@
...
+		_metalFeatures.maxPerStageDynamicMTLBufferCount = _metalFeatures.maxPerStageBufferCount;
+		_metalFeatures.postDepthCoverage = true;
+		_metalFeatures.renderLinearTextures = true;
+		_metalFeatures.tileBasedDeferredRendering = true;
+		if (supportsMTLGPUFamily(Apple6)) {
+			_metalFeatures.astcHDRTextures = true;
+		}
+		if (supportsMTLGPUFamily(Apple7)) {
+			_metalFeatures.maxQueryBufferSize = (256 * KIBI);
+		}
```
cdavis5e, line 1505: This needs to be inside an `MVK_XCODE_12` block.

cdavis5e (changes requested):

Issue #1504, opened 2022-01-11
Wrong VkClearColorValue color values when using sRGB swapchain/textures


saschasc:
In the method `MVKPixelFormats::getMTLClearColor` there is a fix to increment the clear value by half the ULP. This results in errors when using linear color values in combination with sRGB swapchain/textures.

https://github.com/KhronosGroup/MoltenVK/blob/f99384da9630717684e2838c70f13cdc1bae15bf/MoltenVK/MoltenVK/GPUObjects/MVKPixelFormats.mm#L472

Excerpt from the [Vulkan specs](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/VkClearColorValue.html):

> Floating point values are automatically converted to the format of the image, with the clear value being treated as linear if the image is sRGB.

So what does that mean when using a linear swapchain fromat like `VK_FORMAT_B8G8R8A8_SRGB` with a color space `VK_COLOR_SPACE_SRGB_NONLINEAR_KHR`?  Assume we want to have a clear color value of sRGB(11 / 255, 22 / 255, 33 / 255). To use this color we have to convert it to linear as the specs states. To accomplish this the well known formula can be used (e.g. [found in chapter 7.7.7 of the Metal Shading Language](https://developer.apple.com/metal/Metal-Shading-Language-Specification.pdf#//apple_ref/doc/uid/TP40014364?language=objc)).

1. Desired clear color in sRGB color space: sRGB(11 / 255, 22 / 255, 33 / 255) = sRGB(0.043, 0.086, 0.123)
2. Manual conversion for `VkClearColorValue`: linear(0.0033, 0.0080, 0.0152)
3. bug introduced in `MVKPixelFormats::getMTLClearColor`by incrementing the clear color by half the ULP: linear(0.0053, 0.0099, 0.0172)
4. Conversion back to sRGB colorspace done by Vulkan: sRGB(16 / 255, 25 / 255, 35 / 255)

In my opinion, one should simply pass on the SRGB values without changing them. Any other opinions on this? @billhollings @cdavis5e?

The image below shows a triangle rendered with sRGB(11 / 255, 22 / 255, 33 / 255) and the background with the measured sRGB value of sRGB(16 / 255, 25 / 255, 35 / 255).

![image](https://user-images.githubusercontent.com/84325903/149034722-857b945a-3851-48e0-8148-18f74cd8659c.png)

billhollings:
[...]

PR #1545, opened 2022-03-11
Avoid adjusting SRGB clear color values by half-ULP.

billhollings:
For GPUs that round float clear colors down, a half-ULP adjustment is performed on normalized formats. But this adjustment should not be performed on sRGB formats, which Vulkan requires to be treated as linear, with the value managed by the app.

Fixes issue #1504.


cdavis5e (approved):
I remember being worried about this when I first wrote this.

PR #1560, opened 2022-04-05
Added optimized support for VK_COMMAND_BUFFER_USAGE_ONE_TIME_SUBMIT_BIT when MVK_CONFIG_PREFILL_METAL_COMMAND_BUFFERS is used

BeastLe9enD:
When a command buffer uses ```VK_COMMAND_BUFFER_USAGE_ONE_TIME_SUBMIT_BIT``` and ```MVK_CONFIG_PREFILL_METAL_COMMAND_BUFFERS``` is enabled, we can directly encode the ```MVKCommand```s into the ```MTLCommandEncoder``` instead of first storing it inside an array. This should reduce memory consumption in some cases where many commands are recorded.


billhollings (changes requested):
LGTM so far. Just need some robustness around the transient object lifetimes, per my suggested change.

Also need to figure out user Lorenz. Maybe add GitHub user...or just squash all the commits into one under @BeastLe9enD?

PR #1563, opened 2022-04-08
Fix crash in vkGetMTLCommandQueueMVK()

bangnoise:
Calls to `vkGetMTLCommandQueueMVK()` returned garbage because `MVKQueue*` cannot be cast from `VkQueue`, use `MVKQueue::getMVKQueue()` instead.


billhollings:
Thanks for finding this, and submitting the fix!

PR #1564, opened 2022-04-08
Fix leak of CoreFoundation objects during calls to vkUseIOSurfaceMVK()

bangnoise:
A small leak occurs if no existing IOSurface is provided to vkUseIOSurfaceMVK() because CoreFoundation objects returned from functions with Create in their name must be released with CFRelease()


billhollings:
Thanks for finding this, and submitting the fix!

Issue #1552, opened 2022-03-28
Add support for VK_EXT_sample_locations extension


billhollings:

cdavis5e:
I've actually wanted to do this for a long time. It's something we need for `vkd3d` in order to support programmable sample positions in D3D12 apps. I've had this on the back burner, though, because of an apparent lack of demand.

billhollings:
Starting work on this now.

billhollings:
Implemented in #1566.

PR #1566, opened 2022-04-08
Add support for VK_EXT_sample_locations extension.

billhollings:
Supports only setting custom sample locations in subpasses via `vkBeginRenderpass`. Does not support setting custom sample locations via `vkCmdBindPipeline` or `vkCmdSetSampleLocationsEXT`, although collects that info for possible future enhancements.

- `MVKPhysicalDevice` track platform support and respond to property queries.
- `MVKCmdBeginRenderPassBase` collect subpass custom sample locations.
- `MVKPipeline` support dynamic state values beyond 31.
- `MVKPipeline` collect custom sample locations.
- Add `MVKCmdSetSampleLocations` to support `vkCmdSetSampleLocations` to collect dynamic custom sample locations.
- `MVKCommandEncoder` support collecting custom sample positions from subpass and dynamic, and set into `MTLRenderPassDescriptor` for each Metal render pass.
- `MVKArrayRef` add assignment operator.
- Add `MVKPhysicalDeviceMetalFeatures::programmableSamplePositions`.
- Update `VK_MVK_MOLTENVK_SPEC_VERSION` to version `34`.
- `MVKCommandBuffer.h` remove obsolete comment documentation.
- Update `Whats_New.md`.

---------------------------
The spec includes the following VU wordings:

>VUID-vkCmdBindPipeline-variableSampleLocations-01525 If VkPhysicalDeviceSampleLocationsPropertiesEXT::variableSampleLocations is VK_FALSE, and pipeline is a graphics pipeline created with a VkPipelineSampleLocationsStateCreateInfoEXT structure having its sampleLocationsEnable member set to VK_TRUE but without VK_DYNAMIC_STATE_SAMPLE_LOCATIONS_EXT enabled then the current render pass instance must have been begun by specifying a VkRenderPassSampleLocationsBeginInfoEXT structure whose pPostSubpassSampleLocations member contains an element with a subpassIndex matching the current subpass index and the sampleLocationsInfo member of that element must match the sampleLocationsInfo specified in VkPipelineSampleLocationsStateCreateInfoEXT when the pipeline was created

>VUID-vkCmdSetSampleLocationsEXT-variableSampleLocations-01530
If VkPhysicalDeviceSampleLocationsPropertiesEXT::variableSampleLocations is VK_FALSE then the current render pass must have been begun by specifying a VkRenderPassSampleLocationsBeginInfoEXT structure whose pPostSubpassSampleLocations member contains an element with a subpassIndex matching the current subpass index and the sampleLocationsInfo member of that element must match the sample locations state pointed to by pSampleLocationsInfo

I have taken these to mean that if the app is using custom sample locations, and the implementation returns `variableSampleLocations = VK_FALSE`, the app **_must always_** set the custom sample locations as part of the `vkCmdBeginRenderPass` command, even if they are also being set via `vkCmdBindPipeline` and `vkCmdSetSampleLocationsEXT`.

However, CTS does not follow this interpretation, and many tests assume the custom sample locations can be set only via `vkCmdBindPipeline` and `vkCmdSetSampleLocationsEXT`, even if `variableSampleLocations == VK_FALSE`. But from the spec wording above, I take this to be an error in CTS.

This is relevant because Metal **_only_** supports setting custom sample locations when the render subpass is encoded, not when the `vkCmdBindPipeline` or  `vkCmdSetSampleLocationsEXT` command is encoded (which can often occur after begin renderpass or next subpass is encoded).
I have submitted a [clarification request](https://gitlab.khronos.org/vulkan/vulkan/-/issues/3067) to the Vulkan WG, and a filed an [issue](https://gitlab.khronos.org/Tracker/vk-gl-cts/-/issues/3623) with CTS.

If the Vulkan WG response come back that my assumptions are not accurate, we will need to enhance this feature to support using custom sample locations set in the pipeline or dynamically. To do so, we might have to defer creating a Metal renderpass until the first draw command, or we might have to track the renderpass and next subpass command objects in the `MVKCommandBuffer` to be able to update it as later pipeline binding or dynamic location setting commands are added.

@cdavis5e Let me know your thoughts on this as you perform the review.

Implements #1552.


MoltenVK/MoltenVK/Commands/MVKCmdRenderPass.mm
```diff
@@ -61,13 +85,23 @@
...
+	
 	cmdEncoder->beginRenderpass(this,
 								_contents,
 								_renderPass,
 								_framebuffer,
 								_renderArea,
 								_clearValues.contents(),
-								_attachments.contents());
+								_attachments.contents(),
+								MVKArrayRef<MVKArrayRef<MTLSamplePosition>>(spSPRefs, spSPCnt));
```
cdavis5e, line 104: There should really be an `mvkMakeArrayRef()` function. Then we wouldn't need to repeat the template arguments.

MoltenVK/MoltenVK/GPUObjects/MVKPipeline.h
```diff
@@ -323,6 +327,8 @@ class MVKGraphicsPipeline : public MVKPipeline {
 
 	MVKSmallVector<VkViewport, kMVKCachedViewportScissorCount> _viewports;
 	MVKSmallVector<VkRect2D, kMVKCachedViewportScissorCount> _scissors;
+	MVKSmallVector<VkDynamicState> _dynamicState;
```
cdavis5e, line 330: I wonder if this wouldn't be better suited to a set-like data structure (e.g. `std::set`).

cdavis5e (approved):
Looks mostly good, but I have a few minor quibbles...

MoltenVK/MoltenVK/GPUObjects/MVKPipeline.h
billhollings: Yeah. I flipped a coin on this one. The number of dynamic states on any pipeline is so small I figured the retrieval overhead would be small compared to the memory and insertion overhead of a set.

MoltenVK/MoltenVK/Commands/MVKCmdRenderPass.mm
billhollings: Ah. Good catch and suggestion. Fix applied...but it turned out the `MVKArrayRef` constructor itself was sufficient to take care of this.

cdavis5e (approved):
Huh. I guess they allowed deduction of class template arguments on constructor expressions in C++17 (14?).

Issue #1536, opened 2022-03-02
Extension request for VK_KHR_dynamic_rendering


edrickhong:
As the title suggests, is it possible to get this extension implemented in Moltenvk?

billhollings:
Thanks for the request. At a first look, this should be possible.

billhollings:
Starting work on this now.

billhollings:
Implemented in PR #1575.

PR #1575, opened 2022-04-19
Add support for VK_KHR_dynamic_rendering extension

billhollings:
- `MVKDevice` track enabling `VK_KHR_dynamic_rendering` extension features.
- Add `MVKCmdBeginRendering` and `MVKCmdEndRendering` command objects and associated command pools.
- Add `MVKCommandEncoder::beginRendering()` which dynamically instantiates temporary `MVKRenderpass` and (imageless) `MVKFramebuffer` objects to handle equivalent renderpass operations.
- Add support functions to create `MVKRenderpass` and (imageless) `MVKFramebuffer` objects from `VkRenderingInfo`.
- `MVKCommandEncodingContext` tracks and manages lifespan of transient renderpass and framebuffer, potentially across multiple `MVKCommandEncoders`.
- `MVKRenderpass` track `VkRenderingFlags`.
- When resuming a suspended dynamic renderpass, reuse the renderpass and framebuffer cached in `MVKCommandEncodingContext`, otherwise create them.
- `MVKRenderSubpass` create and track an internal `VkPipelineRenderingCreateInfo`.
- `MVKRenderPassAttachment` support `LoadOp` and `StoreOp` behavior for dynamic rendering suspend and resume.
- Refactor `viewMask` processing to permit operations on `viewMask` outside of `MVKGraphicsPipeline` without requiring the presence of an `MVKSubpass`.
- `MVKGraphicsPipeline` extract `VkPipelineRenderingCreateInfo` from `MVKRenderSubpass` if available, or `VkGraphicsPipelineCreateInfo::pNext` if not.
- `MVKCommandBuffer` track `VkCommandBufferInheritanceRenderingInfo` (currently unused).
- Exclude attachment from shader if clearing format is not specified.
- (unrelated) `vulkan.mm` move location of `VK_EXT_sample_locations` functions for layout readability consistency.

Implements enhancement request #1536.

Passes all 1739 `dEQP-VK.dynamic_rendering.*` CTS tests.


MoltenVK/MoltenVK/GPUObjects/MVKRenderPass.mm
```diff
@@ -894,6 +864,246 @@ static uint32_t getNextViewMaskGroup(uint32_t viewMask, uint32_t* startView, uin
...
+		depthResolveMode = pRenderingInfo->pDepthAttachment->resolveMode;
+		attRef.layout = pRenderingInfo->pDepthAttachment->imageLayout;
+		rslvLayout = pRenderingInfo->pDepthAttachment->resolveImageLayout;
+	}
+	if (pRenderingInfo->pStencilAttachment && pRenderingInfo->pStencilAttachment->imageView) {
+		attRef.aspectMask |= VK_IMAGE_ASPECT_STENCIL_BIT;
+		stencilResolveMode = pRenderingInfo->pStencilAttachment->resolveMode;
+		attRef.layout = pRenderingInfo->pStencilAttachment->imageLayout;
+		rslvLayout = pRenderingInfo->pStencilAttachment->resolveImageLayout;
+	}
```
cdavis5e, line 977: Metal actually supports separate depth/stencil rendering. I wonder if we couldn't implement this the other way around--that is, render passes implemented in terms of dynamic rendering.

cdavis5e (approved):

MoltenVK/MoltenVK/Commands/MVKCommandBuffer.mm
```diff
@@ -30,6 +30,32 @@
 
 using namespace std;
 
+#pragma mark -
+#pragma mark MVKCommandEncodingContext
+
+// Sets the rendering objects, releasing the old objects, and retaining the new objects.
+// Retaining the new is performed first, in case the old and new are the same object.
+// With dynamic rendering. the objects are transient and only lives as long as the
```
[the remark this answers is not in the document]
billhollings: Thanks for the catch.

MoltenVK/MoltenVK/GPUObjects/MVKRenderPass.mm
```diff
@@ -600,11 +557,13 @@ static uint32_t getNextViewMaskGroup(uint32_t viewMask, uint32_t* startView, uin
...
 #endif
+	bool isResuming = mvkIsAnyFlagEnabled(_renderPass->getRenderingFlags(), VK_RENDERING_RESUMING_BIT);
 
 	// Only allow clearing of entire attachment if we're actually
 	// rendering to the entire attachment AND we're in the first subpass.
+	// If the renderpass was suspended, and is now being resumed, load the contents.
 	MTLLoadAction mtlLA;
-	if (loadOverride || !isRenderingEntireAttachment || !isFirstUseOfAttachment(subpass)) {
+	if (loadOverride || isResuming || !isRenderingEntireAttachment || !isFirstUseOfAttachment(subpass)) {
 		mtlLA = MTLLoadActionLoad;
```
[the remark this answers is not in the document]
billhollings: Yeah...it basically behaves like the other forced load/store overrides in these logic flows. 

Sadly ironic...because the point of the Vulkan suspend/resume renderpass flags is to do exactly the opposite...and NOT perform load/store activities. But I couldn't think of any other way of doing this within the architecture of Metal.

If you have any suggestions, let me know.

MoltenVK/MoltenVK/GPUObjects/MVKRenderPass.mm
billhollings: I did look into creating and populating some kind of mid-level objects that both dynamic rendering and fixed render passes could use. It would require moving the plumbing around populating Metal RP descriptors to those object constructs, and for now it seemed like they'd end up looking very much like the existing MVKRP object constructs that it didn't seem to be worth it, or that it would save much during dynamic behavior.

But I'm open for discussion about a v2 of all this that might optimize things better.

billhollings: I'm not sure if you're tying those changes to D/S here. But I figured we would address separate D/S as part of implementing `VK_KHR_separate_depth_stencil_layouts`, which is an extension that needs to be addressed on the road to 1.2.

Issue #1577, opened 2022-04-25
Function prototype mismatch vkSetMoltenVKConfigurationMVK


smanhaeve:
The type of the second argument to `vkSetMoltenVKConfigurationMVK` is missing a `const` qualifier in the function typedef.

Compare the following two lines:

https://github.com/KhronosGroup/MoltenVK/blob/1236d2f5350062327a2428f80dee615a2e67304f/MoltenVK/MoltenVK/API/vk_mvk_moltenvk.h#L996

https://github.com/KhronosGroup/MoltenVK/blob/1236d2f5350062327a2428f80dee615a2e67304f/MoltenVK/MoltenVK/API/vk_mvk_moltenvk.h#L1089-L1092

billhollings:
Fixed in PR #1578.

PR #1578, opened 2022-04-25
Fix vkSetMoltenVKConfigurationMVK function typedef

smanhaeve:
The type of the second argument to `vkSetMoltenVKConfigurationMVK` is missing a `const` qualifier in the function typedef.

Compare the following two lines:

https://github.com/KhronosGroup/MoltenVK/blob/1236d2f5350062327a2428f80dee615a2e67304f/MoltenVK/MoltenVK/API/vk_mvk_moltenvk.h#L996

https://github.com/KhronosGroup/MoltenVK/blob/1236d2f5350062327a2428f80dee615a2e67304f/MoltenVK/MoltenVK/API/vk_mvk_moltenvk.h#L1089-L1092

Fixes issue https://github.com/KhronosGroup/MoltenVK/issues/1577


billhollings:
Nice catch, and thanks for submitting the fix!

Fixes #1577.

PR #1587, opened 2022-05-05
Add support for VK_KHR_separate_depth_stencil_layouts extension.

billhollings:
Add `VK_KHR_separate_depth_stencil_layouts` and enable its features.
Metal generally ignores image layouts, so nothing further needed.
Update `Whats_New.md` and `MoltenVK_Runtime_UserGuide.md` documents with recently-added extensions.

Passes almost all supported separate_layouts CTS tests, with the few remaining tests failing due to issues with unrelated capabilities.

Unrelated cleanup:
- `MVKDevice` enable extensions before features.
- `MVKDevice` reorder ivar declarations to reduce memory layout gaps, and define default values.
- Rename `VkSemaphoreStyle` to `MVKSemaphoreStyle` to remove potential conflicts with Vulkan `Vk` name space.

### DIFF
--- a/Docs/MoltenVK_Runtime_UserGuide.md
+++ b/Docs/MoltenVK_Runtime_UserGuide.md
@@ -270,6 +270,7 @@ In addition to core *Vulkan* functionality, **MoltenVK**  also supports the foll
 - `VK_KHR_device_group`
 - `VK_KHR_device_group_creation`
 - `VK_KHR_driver_properties`
+- `VK_KHR_dynamic_rendering`
 - `VK_KHR_get_memory_requirements2`
 - `VK_KHR_get_physical_device_properties2`
 - `VK_KHR_get_surface_capabilities2`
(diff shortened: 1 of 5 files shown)


cdavis5e:
I know it's a bit late, but...

In D3D12, combined depth/stencil formats aren't packed but are [explicitly planar](https://microsoft.github.io/DirectX-Specs/d3d/PlanarDepthStencilDDISpec.html). I suspect that this is also true of many Vulkan implementations, particularly when this extension and `VK_EXT_separate_stencil_usage` are used. I wonder, then, if we shouldn't model this explicitly--that is, when both extensions are enabled and the different aspects have different usages and/or aspects, use separate `MTLTexture`s for depth and stencil.

billhollings:
> I know it's a bit late, but...

Sorry. I didn't explicitly ask you to review my recent PR's, as I have been doing in the past, because I didn't want you to feel pressure to do it on your own time. Let me know. I can always request a review and if you don't get to it, pull it in anyway.

> I wonder, then, if we shouldn't model this explicitly--that is, when both extensions are enabled and the different aspects have different usages and/or aspects, use separate MTLTextures for depth and stencil.

Interesting idea. In pulling together those two extensions, I was expecting to have to work magic around things like separately write-enabling parameters for depth or stencil under certain conditions, but I never actually encountered it in CTS. The only CTS tests that are failing are some layout tests that [break on discarding fragments](https://stackoverflow.com/questions/72131365/using-metal-discard-fragment-to-discard-individual-samples-in-an-msaa-attachme) and are unrelated to the actual layout or usage behavior. So both extensions effectively turned out to be almost no-ops under Metal.

It might be that CTS just hasn't covered the right scenarios yet, but it would definitely help to have some concrete use cases that make planar D/S required, or at least attractive.

Issue #1592, opened 2022-05-11
Regression in MoltenVK 1.1.9 affecting DXVK games


FleetAdmiralButter:
It appears that release 1.1.9 has some significant regressions when running Final Fantasy XIV under Wine compared to 1.1.8:

![Screen Shot 2022-05-11 at 9 53 05 pm](https://user-images.githubusercontent.com/37036084/167843943-3bd773a1-ccc9-4c9d-b284-e4de1bb452f2.png)

The game is pretty much rendered unplayable with this bug - all the colours appear incorrectly and it will eventually lock up when loading into a character.

seathasky:
Also experienced this on Final Fantasy XIV with 1.1.9.

FleetAdmiralButter:
Reverting https://github.com/KhronosGroup/MoltenVK/commit/fada8e08c7a9868003992a087474e23b8c096142 seems to resolve the issue in both FFXIV as well as other games running under Wine/DXVK.

rcaridade145:
Isn't this fixed by the corrections made to fix  https://github.com/KhronosGroup/MoltenVK/issues/1569 ?

billhollings:
> Isn't this fixed by the corrections made to fix #1569 ?

Unfortunately, the 1.1.9 release was cut before that #1573 regression fix went in.

Can someone please test this issue against the latest MoltenVK version, or any version later than 6b3ef7b51a1793cdc268b5d5d9d96c5f8606e822, please?

ZoeyCluff:
Completely broken in new and exciting ways!
[CleanShot 2022-05-11 at 16 13 19](https://user-images.githubusercontent.com/2716914/167948827-cd1dd608-74fa-4297-9f45-f778bd26cf14.png)
![image](https://user-images.githubusercontent.com/2716914/167948995-da854332-698b-4554-8813-b5e123f63020.png)

Gcenx:
The following env options are being set
```
MVK_ALLOW_METAL_FENCES=1
MVK_CONFIG_RESUME_LOST_DEVICE=1
MVK_CONFIG_FULL_IMAGE_VIEW_SWIZZLE=1
```

Along with this copy of [DXVK-macOS](https://github.com/Gcenx/DXVK-macOS/releases/download/v1.10.1/dxvk-macOS-async-1.10.1.tar.gz) (and yes this contains the hacks from DXVK-CX21)

ZoeyCluff:
videos:
[...]

Issue #1542, opened 2022-03-09
Flickering in highly lit areas on Mac in FFXIV


tyrexis26:
Flickering is an issue in areas of high light, such as Mist and Limsa Lominsa in FFXIV.  Adjustments/support may be needed for occlusion culling.

marzent:
Seems to be fine with 1.1.1 and lower and happens with 1.1.2 and onwards...

It's probably related to 55d206e058bb17f2c6fab718c939119e1f1a30f3 and 8e8edbadb10672c1d175f52b422536b33b69ca2d since changing occlusion settings in-game makes the flickering actually go away for a few seconds after which it gradually returns

marzent:
If it would help I could also quickly record a video of it happening

cdavis5e:
The irony is that I initially wrote those patches trying to *fix* occlusion problems in FFXIV... /me sighs

tyrexis26:
Your work is very much appreciated - without it, I doubt I would be playing on my Mac.  Sometimes the flickering gives me a headache, though, and I'm not really sure what to do about it.

tyrexis26:
(sorry about closing it, hit the wrong button)

FleetAdmiralButter:
Just backing up @marzent's point about occlusion:

When standing in the Crystal Tower, the flickering is especially bad when Occlusion Culling is enabled. Disabling Culling does improve it in this area:

https://user-images.githubusercontent.com/37036084/158099752-2b77d4c4-d9f1-454a-9431-5acd455743a8.mp4

However, even with Culling disabled, the flickering will still occur in those areas depending on light source:

https://user-images.githubusercontent.com/37036084/158099813-745c6e0c-f86b-415a-b593-4ca3eceee75a.mp4

marzent:
For me on a 6900XT I actually barely notice any changes between turning Occlusion Culling on and off, however the flickering is also faster (looks like it is tied to framerate though)

seathasky:
Can confirm, flickering happens on bot my RX 570 Mac system and 5700XT Mac in FF14 in all versions past 1.1.2 The only version it doesnt, is CX's mvk 1.1.0, has no source code.

Ceolona:
Adding a “me too” to the list. M1 MBA.  My experience seems to have it affected by Occlusion Culling, Glare, and camera angle.

jakeaw:
I'm having the same issue.  M1 Max

seathasky:
[...]

PR #1594, opened 2022-05-17
Reducing redundant state changes

js6i:
Here's a couple of patches that in my tests can reduce Metal command buffer encoding time pretty significantly in some cases, by avoiding encoding things redundantly.
There's another one (not included) to allocate MVKCommands into a linear buffer, which speeds things up a little and reduces freeing them to a trivial "count = 0;" (which otherwise takes some time of the order of 1ms from what I've seen), but there's the complication of a handful commands which actually do things in their destructor, which I haven't resolved yet.

### DIFF
--- a/MoltenVK/MoltenVK/Commands/MVKCommandEncoderState.h
+++ b/MoltenVK/MoltenVK/Commands/MVKCommandEncoderState.h
@@ -373,7 +373,7 @@ protected:
     // Template function that marks both the vector and all binding elements in the vector as dirty.
     template<class T>
     void markDirty(T& bindings, bool& bindingsDirtyFlag) {
-        for (auto& b : bindings) { b.isDirty = true; }
+        for (auto& b : bindings) { b.markDirty(); }
         bindingsDirtyFlag = true;
     }
 
@@ -384,18 +384,18 @@ protected:
 
         if ( !b.mtlResource ) { return; }
 
-        T db = b;   // Copy that can be marked dirty
         MVKCommandEncoderState::markDirty();
         bindingsDirtyFlag = true;
-        db.isDirty = true;
 
         for (auto iter = bindings.begin(), end = bindings.end(); iter != end; ++iter) {
-            if( iter->index == db.index ) {
-                *iter = db;
+            if (iter->index == b.index) {
+                iter->update(b);
                 return;
             }
         }
-        bindings.push_back(db);
+
+        bindings.push_back(b);
+        bindings.back().markDirty();
     }
 
 	// For texture bindings, we also keep track of whether any bindings need a texture swizzle
--- a/MoltenVK/MoltenVK/Commands/MVKMTLResourceBindings.h
+++ b/MoltenVK/MoltenVK/Commands/MVKMTLResourceBindings.h
@@ -34,6 +34,16 @@ typedef struct MVKMTLTextureBinding {
     uint32_t swizzle = 0;
 	uint16_t index = 0;
     bool isDirty = true;
+
+    inline void markDirty() { isDirty = true; }
+
+    inline void update(const MVKMTLTextureBinding &other) {
+        if (mtlTexture != other.mtlTexture || swizzle != other.swizzle) {
+            mtlTexture = other.mtlTexture;
+            swizzle = other.swizzle;
+            markDirty();
+        }
+    }
 } MVKMTLTextureBinding;
 
 /** Describes a MTLSamplerState resource binding. */
@@ -41,6 +51,15 @@ typedef struct MVKMTLSamplerStateBinding {
     union { id<MTLSamplerState> mtlSamplerState = nil; id<MTLSamplerState> mtlResource; }; // aliases
     uint16_t index = 0;
     bool isDirty = true;
+
+    inline void markDirty() { isDirty = true; }
+
+    inline void update(const MVKMTLSamplerStateBinding &other) {
+        if (mtlSamplerState != other.mtlSamplerState) {
+            mtlSamplerState = other.mtlSamplerState;
+            markDirty();
+        }
+    }
 } MVKMTLSamplerStateBinding;
 
 /** Describes a MTLBuffer resource binding. */
@@ -49,8 +68,27 @@ typedef struct MVKMTLBufferBinding {
     VkDeviceSize offset = 0;
     uint32_t size = 0;
 	uint16_t index = 0;
+    bool justOffset = false;
     bool isDirty = true;
     bool isInline = false;
+
+    inline void markDirty() { justOffset = false; isDirty = true; }
+
+    inline void update(const MVKMTLBufferBinding &other) {
+        if (mtlBuffer != other.mtlBuffer || size != other.size || isInline != other.isInline) {
+            mtlBuffer = other.mtlBuffer;
+            size = other.size;
+            isInline = other.isInline;
+            offset = other.offset;
+
+            justOffset = false;
+            isDirty = true;
+        } else if (offset != other.offset) {
+            offset = other.offset;
+            justOffset = !isDirty || justOffset;
+            isDirty = true;
+        }
+    }
 } MVKMTLBufferBinding;
 
 /** Describes a MTLBuffer resource binding as used for an index buffer. */
--- a/MoltenVK/MoltenVK/Commands/MVKCommandEncoderState.mm
+++ b/MoltenVK/MoltenVK/Commands/MVKCommandEncoderState.mm
@@ -752,6 +752,10 @@ void MVKGraphicsResourcesCommandEncoderState::encodeImpl(uint32_t stage) {
                                                            b.mtlBytes,
                                                            b.size,
                                                            b.index);
+                           else if (b.justOffset)
+                               [cmdEncoder->getMTLComputeEncoder(kMVKCommandUseTessellationVertexTessCtl)
+                                            setBufferOffset: b.offset
+                                            atIndex: b.index];
                            else
                                [cmdEncoder->getMTLComputeEncoder(kMVKCommandUseTessellationVertexTessCtl) setBuffer: b.mtlBuffer
                                                                                                              offset: b.offset
@@ -785,9 +789,14 @@ void MVKGraphicsResourcesCommandEncoderState::encodeImpl(uint32_t stage) {
                                                               b.size,
                                                               b.index);
                                } else {
-                                   [cmdEncoder->_mtlRenderEncoder setVertexBuffer: b.mtlBuffer
-                                                                           offset: b.offset
-                                                                          atIndex: b.index];
+                                   if (b.justOffset) {
+                                       [cmdEncoder->_mtlRenderEncoder setVertexBufferOffset: b.offset
+                                                                                    atIndex: b.index];
+                                   } else {
+                                       [cmdEncoder->_mtlRenderEncoder setVertexBuffer: b.mtlBuffer
+                                                                               offset: b.offset
+                                                                              atIndex: b.index];
+                                   }
 
                                    // Add any translated vertex bindings for this binding
                                    auto xltdVtxBindings = pipeline->getTranslatedVertexBindings();
@@ -828,6 +837,9 @@ void MVKGraphicsResourcesCommandEncoderState::encodeImpl(uint32_t stage) {
                                                            b.mtlBytes,
                                                            b.size,
                                                            b.index);
+                           else if (b.justOffset)
+                               [cmdEncoder->getMTLComputeEncoder(kMVKCommandUseTessellationVertexTessCtl) setBufferOffset: b.offset
+                                                                                                                  atIndex: b.index];
                            else
                                [cmdEncoder->getMTLComputeEncoder(kMVKCommandUseTessellationVertexTessCtl) setBuffer: b.mtlBuffer
                                                                                                              offset: b.offset
@@ -858,6 +870,9 @@ void MVKGraphicsResourcesCommandEncoderState::encodeImpl(uint32_t stage) {
                                                           b.mtlBytes,
                                                           b.size,
                                                           b.index);
+                           else if (b.justOffset)
+                               [cmdEncoder->_mtlRenderEncoder setVertexBufferOffset: b.offset
+                                                                            atIndex: b.index];
                            else
                                [cmdEncoder->_mtlRenderEncoder setVertexBuffer: b.mtlBuffer
                                                                        offset: b.offset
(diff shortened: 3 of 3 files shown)


cdavis5e (changes requested):
I was wondering about this myself. "Why are we emitting all these redundant state changes?" Hopefully this will help with the performance problems other people have been seeing.

Just one small problem... (Besides the fact that #1595 changed the same part of the code.)

cdavis5e:
@ZoeyCluff Does this change help with the performance problems you've been seeing? (You may have trouble applying it because of #1595.)

MoltenVK/MoltenVK/Commands/MVKCommandEncoderState.h
```diff
@@ -387,11 +388,12 @@ class MVKResourcesCommandEncoderState : public MVKCommandEncoderState {
...
+        db.markDirty();
 
         for (auto iter = bindings.begin(), end = bindings.end(); iter != end; ++iter) {
             if( iter->index == db.index ) {
-                *iter = db;
+                if (!areEqual(&db, iter))
+                    *iter = db;
                 return;
             }
         }
```
billhollings, line 399: Per the comment below by @cdavis5e, regarding side effects in `areEqual()`, I'm wondering if it might be a  good idea to merge this logic into something like an `updateIfReplaced(b)` function on the binding structs, and call `areEqual()` and `markDirty()`, plus handle the additional specialized buffer logic for `justOffset` within that `updateIfReplaced(b)` function. Then we'd end up with something here like:

```
for (auto iter = bindings.begin(), end = bindings.end(); iter != end; ++iter) {
    if (iter->updateIfReplaced(b))
        return;
}
```

with no need for the `db` copy, because the `updateIfReplaced()` function in `*iter` would call it's own `markDirty()`. And I've assumed `areEqual() could presumably include the `index == b.index` check too.

MoltenVK/MoltenVK/Commands/MVKMTLResourceBindings.h
```diff
@@ -34,33 +34,64 @@ typedef struct MVKMTLTextureBinding {
...
+    inline void markDirty() { justOffset = false; isDirty = true; }
 } MVKMTLBufferBinding;
 
+static inline
+bool areEqual(MVKMTLBufferBinding *a, MVKMTLBufferBinding *b) {
+    if (a->mtlBuffer == b->mtlBuffer && a->size == b->size && a->isInline == b->isInline) {
+        if (a->offset == b->offset)
+            return true;
+        if (!a->isInline && !b->isDirty)
+            a->justOffset = true;
```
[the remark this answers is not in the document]
billhollings: Agreed. I've added some thoughts above about refactoring this part (plus other logic) into a possible `updateIfReplaced()` function.

billhollings (changes requested):
Thanks so much for submitting this! It is definitely a helpful performance improvement!

Need changes to align with #1595, and I've included a suggestion for refactoring, adding to @cdavis5e 's thoughts.

js6i:
Thanks, I changed the isEqual() procedure to update(), which makes more sense. Also changed the justOffset logic a little bit - setting the flag doesn't have to be predicated on the binding not being inline, as that is checked during encoding. Additionally, if the old binding already was justOffset dirty, we can preserve that (pretty much an edge case, but hey).

FleetAdmiralButter:
Hi @js6i,

Apologies for butting in - I gave this PR a spin with Final Fantasy XIV but unfortunately, it completely breaks lighting/shadows in the game:

https://www.youtube.com/watch?v=WTcn3lbWeOY
https://www.youtube.com/watch?v=bpM36Q_CMDk

I'll try this out with another DXVK game shortly and see if it exhibits the same issue.

js6i:
> Hi @js6i,
> 
> Apologies for butting in - I gave this PR a spin with Final Fantasy XIV but unfortunately, it completely breaks lighting/shadows in the game:
> 
> https://www.youtube.com/watch?v=WTcn3lbWeOY https://www.youtube.com/watch?v=bpM36Q_CMDk
> 
> I'll try this out with another DXVK game shortly and see if it exhibits the same issue.

Interesting, I tested it with FFXIV, but using wined3d and not DXVK. Note that I did a second force push after the first one, where I got the update backwards, but I don't see how that could make things worse. I'll try with DXVK. Thanks!

FleetAdmiralButter:
> Interesting, I tested it with FFXIV, but using wined3d and not DXVK. Note that I did a second force push after the first one, where I got the update backwards, but I don't see how that could make things worse. I'll try with DXVK. Thanks!

Thanks for looking into this @js6i!

Just tested with a couple NieR games I have in CrossOver and they don't seem to exhibit the same lighting issue fortunately:

image
image

MoltenVK/MoltenVK/Commands/MVKCommandEncoderState.h
```diff
@@ -384,17 +385,18 @@ class MVKResourcesCommandEncoderState : public MVKCommandEncoderState {
...
+            if (iter->index != b.index) continue;
+
+            iter->update(b);
+            return;
         }
+
+        T db = b;   // Copy that can be marked dirty
+        db.markDirty();
         bindings.push_back(db);
     }
```
billhollings, line 401: Minor nit. Could also skip need for `db` copy with:

```
bindings.push_back(b);
bindings.back().markDirty();
```

MoltenVK/MoltenVK/Commands/MVKMTLResourceBindings.h
```diff
@@ -49,8 +68,27 @@ typedef struct MVKMTLBufferBinding {
...
+            mtlBuffer = other.mtlBuffer;
+            size = other.size;
+            isInline = other.isInline;
+            offset = other.offset;
+
+            justOffset = false;
+            isDirty = true;
+        } else if (offset != other.offset) {
+            offset = other.offset;
+            justOffset = !isDirty || justOffset;
```
billhollings, line 88: TBH...I'm having trouble following the logic of when `justOffset` will be set to `true`. Here we seem to be setting it to `true` only if I am not already dirty or `justOffset` was previously set to `true` somehow. 

I may just be missing something in the flow, but it's not clear, in the same way that the values of `isDirty` and `justOffset` are clearly set on each invocation everywhere else.

billhollings (changes requested):
Couple of new questions and comments.

billhollings:
Thanks to everyone for proactively testing this PR!

We'll need some form of verification about whether the possible regression is avoided, before pulling it in.

MoltenVK/MoltenVK/Commands/MVKCommandEncoderState.h
```diff
@@ -384,17 +385,18 @@ class MVKResourcesCommandEncoderState : public MVKCommandEncoderState {
...
         MVKCommandEncoderState::markDirty();
         bindingsDirtyFlag = true;
-        db.isDirty = true;
 
         for (auto iter = bindings.begin(), end = bindings.end(); iter != end; ++iter) {
-            if( iter->index == db.index ) {
-                *iter = db;
-                return;
-            }
+            if (iter->index != b.index) continue;
```
billhollings, line 392: Why is this flow not the more direct positive form:

```
if (iter->index == b.index) {
    iter->update(b);
    return;
}
```

MoltenVK/MoltenVK/Commands/MVKMTLResourceBindings.h
js6i: My thinking is that if the binding was updated previously, we can't set it to justOffset here, unless the previous update also only changed the offset. If we did, we wouldn't set the changed buffer from the first update when encoding state.

MoltenVK/MoltenVK/Commands/MVKCommandEncoderState.h
js6i: I guess I slightly more like it my way, but I'll change to this next time I update the branch.

Gcenx:
@js6i Fleet isn’t using the SE client but the third-party client [XIV on Mac](https://www.xivmac.com/) that’s the same client that Bill used to fix #1592

That uses MoltenVK-CX21-v1.1.3 by default or MoltenVK-v1.1.8 when the “modern” checkbox is checked, it bundles [dxvk-macOS-async-1.10.1](https://github.com/Gcenx/DXVK-macOS/releases/download/v1.10.1/dxvk-macOS-async-1.10.1.tar.gz)

The flickering lighting is best seen in Crystarium  (dosal gate)

My DXVK-macOS packages work with upstream wine64 with stock MoltenVK no problem just use the env options I’d provided in my prior comment.

js6i:
Okay, I reproduced the issue with flickering shadows. Presumably the lights was the same problem. The depth-stencil state wasn't getting marked dirty in its beginMetalRenderPass override in the cases handled by the parent method. That should be fixed now.

italomandara:
Tested on CX21 with the latest dxvk from gcenx and the mentioned flags turned on.
Titanfall2: no glitches, fps bump from 80-90 to 110-120
Rocket league: no glitches (but the old missing labels bug) fps bump from 90 to 120 fps
Elite dangerous odyssey: no glitches, shader pre-compilation is faster(?) no noticeable in-game fps boost
Guilty gear Xrd: no glitches,, no noticeable fps boost.
looks good so far.
Maybe this is not that much but i wanted to help somehow.

Gcenx:
@italomandara the Rocket League nameplate issue is lightly due to the current lack of transform-feedback/geometry shaders support.

I’d asked @seathasky to test this and they’d confirmed the shadows regression was indeed resolved, however the lighting bug still remains that’s been the case for any version except MoltenVK shipped in CX20.

Gcenx:
@js6i not sure if you’ve seen my comment.

The shadows regression was resolved, the lighting wasn’t resolved. From my perspective it’s not worth holding their back from getting merged just for the FFXIV lightly problem.

js6i:
> @js6i not sure if you’ve seen my comment.
> 
> The shadows regression was resolved, the lighting wasn’t resolved. From my perspective it’s not worth holding their back from getting merged just for the FFXIV lightly problem.

Wait, I thought you meant that the lighting problem was introduced earlier. If that's not the case, and it is in fact caused by changes in this PR, it should definitely be fixed before merging. In such case I'd ask for directions to this Crystorium place, or somewhere else where I'm going to see it easily.

Gcenx:
@js6i this PR didn’t cause the lighting problem that was prior behavior, the shadows problem was caused by and fix in this PR.

The lighting problem has been around since 1.1.2, as 1.1.1 was the last version without that problem see #1542

FleetAdmiralButter:
Hey @js6i,

Thanks for looking into this! I grabbed the latest build from Actions but it's unfortunately still exhibiting the same behaviour:

image

FleetAdmiralButter:
Gave this a try on my M1 Pro and got a similar result:

image

seathasky:
Tested with RX 5700 XT on MacOS Big Sur gives this result: 
image

seathasky:
Suprising this issue doesnt affect Polaris based card (RX500 Series), but does affect M1 and Navi based AMD card. 

One thing to note, on Polaris (500 series), the regressions and stuttering are still very apparent and unplayable with this build.

seathasky:
Here is video footage of 5700 XT on Big Sur:

https://user-images.githubusercontent.com/17598024/170817906-0f5475b1-05d1-4088-8189-faf33b60cb5b.mp4

seathasky:
Here is footage of **(Polaris) RX570** with same build:

https://user-images.githubusercontent.com/17598024/170818080-853855a6-7577-4b28-bd1f-d1461227bb70.mp4

seathasky:
> Here is footage of **(Polaris) RX570** with same build:
> 2022-05-28-03-54-31.1.mp4
> 
> Stutter regression remains there but lighting seems fine compared to Navi.

I need to point something out very important here that I forgot to mention in my initial post about AMD 500 series. 

@billhollings There has yet to be a release of MoltenVK from your page that has worked properly for 500 series. 

Every single release has introduced stuttering to these 500 series cards. Jeremy over at Codeweavers provided the us, the "XIV on Mac" launcher team distribution of one of there CX 21 MVK dylibs that solved this issue entirely. 

It's actually the release we currently use as our "stable" release that ships with the XIV on Mac launcher. I just didnt want to confuse anyone.

Gcenx:
@billhollings I'm bugging them on discord now

Gcenx:
> Are there lighting/shadow issues that are being introduced with this PR, or is it safe to pull in, relative to existing MoltenVK behavior?

The only thing to look into having in upstream is explained by Chip https://github.com/KhronosGroup/MoltenVK/issues/1597

seathasky:
We can ignore the 500 series issue for now. One thing I will say about it though is Codeweavers did provide the source. We weren't able to achieve the same results as their compiled dylib after compiling their source. Maybe they sent the wrong source? Something to do with xcode? Who knows.

**On to this PR.** 

I've done some extensive testing in FF14 today with my 5700xt and the latest pr. Hours of testing actually. The PR is extremely stable, but it still has flickering in certain places. It's interesting too. In the Dossal gate, you can see it strobes faster than usual. 


Short video:
https://www.youtube.com/watch?v=IpD-y6Ouyks

js6i:
@billhollings I don't reproduce any more issues with AMD Radeon Pro 5500M. I want to test it on M1 too, given the possibility of something hardware dependent, but I couldn't get it to run right away and was busy with some other work. Let me ping you once I manage that.

FleetAdmiralButter:
Can confirm the latest commit fixes the lighting issues for me as well, thanks @js6il.

js6i:
> Can confirm the latest commit fixes the lighting issues for me as well, thanks @js6il.

That's confusing, because I didn't change anything since your last report!

Anyways, @billhollings, this time running the game on M1 worked and I didn't see any lighting issues there either. I found a bug in `MVKDepthBiasCommandEncoderState::setDepthBias` that is fixed in the latest version, however that wasn't causing any visible trouble. If anyone's still seeing regressions caused by these changes, let me know, but I couldn't reproduce any - sorry!

Gcenx:
@js6i your changed had introduced a regression with shadows but you’ve already resolved that.

The other thing mentioned here was #1542 that’s a rather long standing bug. Your changes have no made any difference to that issue.

<br>

I hope this helps clear things up for you.

billhollings (approved):

billhollings:
@cdavis5e It's been a bit since your review. All good now from your perspective?

cdavis5e (approved):

PR #1595, opened 2022-05-20
Fix vertex buffer binding counts when establishing implicit buffers binding indexes.

billhollings:
Fix vertex buffer binding counts when establishing implicit buffers binding indexes.

This patch fixes a regression caused by 1b6b8bc9 when the implicit buffers were moved to top of Metal buffer index range.

Fix vertex buffer binding counts when binding indexes are not consecutive, or when additional synthetic buffer bindings are required to accommodate vertex attributes that are outside the vertex buffer stride values. Take into consideration that the app may bind more vertex attribute buffers than a pipeline may consume, and don't allow these to overwrite any implicit buffers sent to the shader.

`MVKResourcesCommandEncoderState::encodeBinding()` clear binding dirty flag before calling lambda function to allow function operation to possibly override and leave binding marked dirty.

Set tessellation vertex buffer indexes based on platform maximum vertex stage  buffer count, instead of hardcoding them, to preemptively avoid conflict with implicit buffers should the platform counts ever change.

Fixes issue #1592.

### DIFF
--- a/Docs/Whats_New.md
+++ b/Docs/Whats_New.md
@@ -30,6 +30,7 @@ Released TBD
   used by a subsequent pipeline that does not use push constants.
 - Fix error on some Apple GPU's where a `vkCmdTimestampQuery()` after a renderpass was 
   writing timestamp before renderpass activity was complete.
+- Fix regression error in vertex buffer binding counts when establishing implicit buffers binding indexes.
 - Work around zombie memory bug in Intel Iris Plus Graphics driver when repeatedly retrieving GPU counter sets.
 - Update to latest SPIRV-Cross:
 	- MSL: Emit interface block members of array length 1 as arrays instead of scalars.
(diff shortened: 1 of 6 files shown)


FleetAdmiralButter:
Hi @billhollings thanks for this!

This has eliminated the corruption on the title screen but it's still present once you load into the game:

[Screen Recording 2022-05-20 at 11.45.59 pm.mov](https://youtu.be/uZzj2fN3WGE)

billhollings:
> once you load into the game

Thanks for reporting. Looking at this now.

I'm not really familiar with the game. Is there a way to bypass the world login queues and long intro cutscenes to get to areas that are experiencing the problem? Otherwise, it make debugging painfully slow.

I'm testing with XIV on Mac.

cdavis5e (approved):

billhollings:
Based on [feedback](https://github.com/KhronosGroup/MoltenVK/issues/1592#issuecomment-1132946969) from the original version of this PR, I've updated this PR to properly handle the app possibly binding extra vertex attribute buffers that are not used by the current pipeline, but overlap the buffer binding range used by the implicit buffers needed by the current pipeline.

ZoeyCluff:
Nope, worse this time.
![CleanShot 2022-05-22 at 15 19 52]
CleanShot 2022-05-22 at 15 20 13
(https://user-images.githubusercontent.com/2716914/169714235-40697b63-5ddb-4c74-a28c-a851f9851555.png)

https://nextcloud.zoeyrae.dev/s/zGFKGkrS8ExxaCy

cdavis5e (approved):

cdavis5e (changes requested):
Well, that didn't work...

cdavis5e (approved):
OK, glad they fixed the problems on their end.

Gcenx:
> OK, glad they fixed the problems on their end.

We’ve never seen the cache causing a problem between MoltenVK versions before so that was a new one.

PR #1603, opened 2022-05-30
Fix reorder-ctor warnings

scandit-opm:
When building, I get all these warnings regarding the ordering of the member variable initialization being different to the order of declaration. This pull request fixes some of them (for the files we are building).


billhollings:
Thanks for catching this, and submitting the fix.

Because they don't include the `-Wreorder` warning, I assume you were able to catch this because you're not using the Xcode projects to build MoltenVK? 

Once I pull this in, I'll add `-Wreorder` to the Xcode projects, so we can keep constructors aligned for all build environments in real-time in the future.

billhollings:
> Thanks for merging so quickly. Yes, we are using cmake.

You're welcome. Thanks for the fix.

I've also added PR #1604 to add the `-Wreorder` warning in the Xcode project...so that any future changes to member orders will be detected immediately by Xcode builds.

PR #1620, opened 2022-06-10
Add support for VK_EXT_metal_objects extension.

billhollings:
- Add support for `VK_EXT_metal_objects` extension.
- Update Vulkan-Headers version to include new `VK_EXT_metal_objects` extension.
- Update MoltenVK version to 1.1.11.

PR #1621, opened 2022-06-12
Add support for VK_EXT_metal_objects extension.

billhollings:
- Update Vulkan-Headers version to include new `VK_EXT_metal_objects extension`.
- Update MoltenVK version to 1.1.11.
- Update What's New document.

Squashes and replaces PR #1620.


MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
```diff
@@ -986,6 +993,35 @@
...
+			case VK_STRUCTURE_TYPE_EXPORT_METAL_OBJECT_CREATE_INFO_EXT:
+				pExportInfo = (VkExportMetalObjectCreateInfoEXT*)next;
+				break;
+			default:
+				break;
+		}
+	}
+
+	// If we're expecting to export an IOSurface, and weren't give one,
+	// base this image on a new IOSurface that matches its configuration.
```
cdavis5e, line 1021:
```suggestion
	// If we're expecting to export an IOSurface, and weren't given one,
	// base this image on a new IOSurface that matches its configuration.
```

cdavis5e (approved):
I still don't see why we don't support `VK_KHR_external_memory`-style usage here. All I know is, this will probably make using this in Wine (in order to support `VK_KHR_external_memory_win32`, `VK_KHR_external_fence_win32`, and `VK_KHR_external_semaphore_win32`, which is needed to support some games) more complicated.

MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
billhollings: We will at some point...maybe driven by the Android emulator requirements.

The use cases here were different, really about making the existing functionality, including accessing resources, but also including `MTLCommandQueue` and `MTLDevice`, Vulkan friendly.

PR #1632, opened 2022-07-06
Support Xcode 14, macOS 13, and iOS/tvOS 16

### LINKED ISSUE #1627: Remove obsolete over-retention, as a result of Apple fix
_According to feedback from Apple for Metal 3:_

Under some circumstances, Xcode's frame capture layer was retaining Capture Scopes for a long time, leading to Command Queue not being recycled. You implemented a workaround to force command queues being recycled here :

```
MVKGPUCaptureScope::MVKGPUCaptureScope(MVKQueue* mvkQueue)
{
    // ...
    _mtlCaptureScope = [[MTLCaptureManager sharedCaptureManager] newCaptureScopeWithCommandQueue: _mtlQueue];
    // ...
    while (_mtlCaptureScope.retainCount > 1)
    {
        [_mtlCaptureScope release];
    }
    // ...
}
```

We changed that behavior of Capture Scopes - and now this workaround will not be needed anymore - The fix would look like this :

```
MVKGPUCaptureScope::MVKGPUCaptureScope(MVKQueue* mvkQueue)
{
    // ...
    _mtlCaptureScope = [[MTLCaptureManager sharedCaptureManager] newCaptureScopeWithCommandQueue: _mtlQueue];
    // ...
    if (@available(macOS 13.0, iOS 16.0, tvOS 16.0, *))
    { }
    else
    {
        while (_mtlCaptureScope.retainCount > 1)
        {
            [_mtlCaptureScope release];
        }
        // ...
    }
}
```

billhollings:
> _**Note:** For review, first commit contains the meaningful changes. Second commit contains the standard Xcode build setting update noise._

- Update minimum Xcode deployment targets to macOS 10.13, iOS 11, and tvOS 11, to avoid Xcode build warnings.
- Add support for `MTLLanguageVersion3_0` enumeration.
- Build efficiencies:
  - Build scripts `create_dylib.sh` and `gen_moltenvk_rev_hdr.sh` only run if build dependencies require it.
  - Packaging and `copy_to_staging.sh` scripts are too complex to define dependencies, and are fast, so configured to run every time, to avoid build warning.
- Replace use of deprecated `sprintf()` with  `snprintf()`.
- Replace use of deprecated `kIOMasterPortDefault` with `kIOMainPortDefault`.
- Support old-style GPU debug capture only if building for earlier minimum deployment targets, to avoid deprecation warning.
- Update minimum Xcode deployment targets of Cube demo to macOS 10.14, iOS 12, and tvOS 12, to avoid Xcode build warning regarding `MTLSharedEvent`.
- Update `README.md` document regarding minimum Xcode deployment targets.

Fix Xcode projects to support Xcode 11.7:
- Add dummy output files to each Xcode Run Script build phase that doesn't already have dependencies set, to avoid setting `alwaysOutOfDate` flag (by disabling _Based on dependency analysis_ UI flag), which forces the Xcode project to a version that can't be read by Xcode 11.7.

Xcode 14 build fixes from code review and further testing:
- Replace use of deprecated kIOMasterPortDefault with MACH_PORT_NULL.
- Convert an inline VLA with constant length array.
- Add input files to all build phases to trigger dependencies when inputs change (packages were not being updated correctly on incremental builds).
- Add packaging scripts to per-platform ExternalDependencies targets.

Fixes #1627.

### DIFF
--- a/MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
@@ -1354,6 +1354,11 @@ void MVKPhysicalDevice::initMetalFeatures() {
 		_metalFeatures.mslVersionEnum = MTLLanguageVersion2_4;
 	}
 #endif
+#if MVK_XCODE_14
+	if ( mvkOSVersionIsAtLeast(16.0) ) {
+		_metalFeatures.mslVersionEnum = MTLLanguageVersion3_0;
+	}
+#endif
 
 #endif
 
@@ -1463,6 +1468,11 @@ void MVKPhysicalDevice::initMetalFeatures() {
 		_metalFeatures.mslVersionEnum = MTLLanguageVersion2_4;
 	}
 #endif
+#if MVK_XCODE_14
+	if ( mvkOSVersionIsAtLeast(16.0) ) {
+		_metalFeatures.mslVersionEnum = MTLLanguageVersion3_0;
+	}
+#endif
 
 #endif
 
@@ -1541,6 +1551,11 @@ void MVKPhysicalDevice::initMetalFeatures() {
 #if MVK_XCODE_13
 	if ( mvkOSVersionIsAtLeast(12.0) ) {
 		_metalFeatures.mslVersionEnum = MTLLanguageVersion2_4;
+	}
+#endif
+#if MVK_XCODE_14
+	if ( mvkOSVersionIsAtLeast(13.0) ) {
+		_metalFeatures.mslVersionEnum = MTLLanguageVersion3_0;
 	}
 #endif
 
@@ -1643,6 +1658,11 @@ void MVKPhysicalDevice::initMetalFeatures() {
 	_metalFeatures.mslVersion = SPIRV_CROSS_NAMESPACE::CompilerMSL::Options::make_msl_version(maj, min);
 
 	switch (_metalFeatures.mslVersionEnum) {
+#if MVK_XCODE_14
+		case MTLLanguageVersion3_0:
+			setMSLVersion(3, 0);
+			break;
+#endif
 #if MVK_XCODE_13
 		case MTLLanguageVersion2_4:
 			setMSLVersion(2, 4);
@@ -1668,7 +1688,7 @@ void MVKPhysicalDevice::initMetalFeatures() {
 		case MTLLanguageVersion1_1:
 			setMSLVersion(1, 1);
 			break;
-#if MVK_IOS_OR_TVOS
+#if MVK_IOS_OR_TVOS || MVK_XCODE_14
 		case MTLLanguageVersion1_0:
 			setMSLVersion(1, 0);
 			break;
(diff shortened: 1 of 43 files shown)


MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
```diff
@@ -2309,13 +2329,26 @@ static uint32_t mvkGetEntryProperty(io_registry_entry_t entry, CFStringRef prope
...
+#if MVK_XCODE_13
+#	if MVK_MACCAT
+	const mach_port_t kIOMainPortDefaultMVK = mvkOSVersionIsAtLeast(12.0) ? kIOMainPortDefault : MACH_PORT_NULL;
+#	else
+	const mach_port_t kIOMainPortDefaultMVK = mvkOSVersionIsAtLeast(12.0) ? kIOMainPortDefault : kIOMasterPortDefault;
+#	endif
+#else
+	const mach_port_t kIOMainPortDefaultMVK = kIOMasterPortDefault;
+#endif
+
```
cdavis5e, line 2344: Why don't we just use `MACH_PORT_NULL` everywhere? That's been the behavior of IOKit since long before Metal was introduced--since at least 10.4, in fact, and I think maybe even earlier.

MoltenVK/MoltenVK/GPUObjects/MVKQueue.mm
```diff
@@ -218,8 +218,9 @@
 
 void MVKQueue::initName() {
 	const char* fmt = "MoltenVKQueue-%d-%d-%.1f";
-	char name[256];
-	sprintf(name, fmt, _queueFamily->getIndex(), _index, _priority);
+	size_t nameLen = 256;
```
cdavis5e, line 221: This should probably be a `const`; otherwise, you're needlessly creating a VLA. Or, you could pass `sizeof(name)/sizeof(char)` to `snprintf(3)`.

MoltenVK/MoltenVK/Utility/MVKBaseObject.mm
```diff
@@ -129,9 +129,10 @@
 
 	// Prepend the error code to the format string
 	const char* vkRsltName = mvkVkResultName(vkErr);
-	char fmtStr[strlen(vkRsltName) + strlen(format) + 4];
-	sprintf(fmtStr, "%s: %s", vkRsltName, format);
-    
+	size_t rsltLen = strlen(vkRsltName) + strlen(format) + 4;
+	char fmtStr[rsltLen];
+	snprintf(fmtStr, rsltLen, "%s: %s", vkRsltName, format);
+
```
cdavis5e, line 135: Well, I guess this one was already a VLA...

cdavis5e (changes requested):
Just a few small nits...

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
billhollings: Mostly I was just timid about changing more than the minimum required, based on old Apple code comments.

But okay...updated now.

MoltenVK/MoltenVK/GPUObjects/MVKQueue.mm
billhollings: Thanks. Good catch. Fixed now.

cdavis5e (approved):
@js6i Heads up--the default minimum version has been bumped. You may wish to adjust CrossOver's build scripts accordingly.

billhollings:
> @js6i Heads up--the default minimum version has been bumped. You may wish to adjust CrossOver's build scripts accordingly.

It's a default, based on avoiding build warnings in Xcode 14. Looks like Apple only supports 5 years worth of back versions. The MoltenVK Xcode projects can still be taken down to macOS 11 / iOS 9, before building, if needed.

PR #1639, opened 2022-07-13
MVKDescriptorPool: Move _hasPooledDescriptors to the top to ensure it's initialized first.

js6i:
Creating an empty MVKDescriptorSet will read this member and tip off undefined behavior sanitizer. It shouldn't actually cause any bugs, but with UB you never know.

### DIFF
--- a/MoltenVK/MoltenVK/GPUObjects/MVKDescriptorSet.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKDescriptorSet.mm
@@ -707,10 +707,10 @@ static size_t getInlineBlockPoolSize(const VkDescriptorPoolCreateInfo* pCreateIn
 }
 
 // Although poolDescriptors is derived from MVKConfiguration, it is passed in here to ensure all components of this instance see a SVOT for this value.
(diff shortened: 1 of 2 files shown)


MoltenVK/MoltenVK/GPUObjects/MVKDescriptorSet.mm
```diff
@@ -707,10 +707,10 @@ static size_t getInlineBlockPoolSize(const VkDescriptorPoolCreateInfo* pCreateIn
 }
 
 // Although poolDescriptors is derived from MVKConfiguration, it is passed in here to ensure all components of this instance see a SVOT for this value.
-// Alternate might have been to force _hasPooledDescriptors to be set first by changing member declaration order in class declaration.
 MVKDescriptorPool::MVKDescriptorPool(MVKDevice* device, const VkDescriptorPoolCreateInfo* pCreateInfo, bool poolDescriptors) :
```
billhollings, line 710: As this comment indicates, if we're forcing `_hasPooledDescriptors` to be set first now, there is a bunch of simplifications we can do here.

However, that's a bit out of scope for this PR, so I'll take care of that in a separate PR. Tracking that now in #1640.

billhollings (approved):
Thanks for catching this.

PR #1666, opened 2022-07-28
Update copy_lib_to_staging.sh

MennoVink:
Allow for building on removable volumes with a space in their name, ie "/Volumes/T7 Touch/"


cdavis5e (approved):

billhollings (approved):
Good catch. Thanks for the fix!

Issue #1591, opened 2022-05-10
Cube example doesn't update contentsScale for different monitor scales


mbechard:
If you have a hiDPI (contentsScale=2) and regular monitor (contentsScale=1) connected to your computer, when dragging the window between the monitors the contentsScale isn't updated, so the swap chain isn't the optimal resolution for the current display the window is on.
Interestingly, even if the window is originally opened on the regular monitor, the first call to makeBackingLayer gets a viewScale == 2 from the View. I'm not sure if that behaviour is dependable, so if somehow the View returned a scale == 1 on first open, then dragging the window onto the scale==2 monitor could result in a very ugly output.

billhollings:
PR #1668 fixes this. Please retest and close this issue if it fixes the problem you encountered. 

The contents scale is now logged for easy verification.

PR #1668, opened 2022-07-28
Update macOS Cube demo to demonstrate optimizing swapchain across multiple screens.

billhollings:
- `DemoView` implements `NSViewLayerContentScaleDelegate` protocol to update `[CAMetalLayer contentsScale]` property when moved between screens.
- Log `contentsScale` value during swapchain creation.
- Remove a few unnecessary inline declarations.

Fixes issue #1591.

### DIFF
--- a/Demos/Cube/macOS/DemoViewController.m
+++ b/Demos/Cube/macOS/DemoViewController.m
@@ -88,4 +88,16 @@ static CVReturn DisplayLinkCallback(CVDisplayLinkRef displayLink,
 	return layer;
 }
 
+/**
+ * If this view moves to a screen that has a different resolution scale (eg. Standard <=> Retina),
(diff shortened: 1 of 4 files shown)


cdavis5e (approved):
Do we want to send this upstream?

billhollings:
> Do we want to send this upstream?

What do you mean by "send it upstream"?

cdavis5e:
> > Do we want to send this upstream?
> 
> What do you mean by "send it upstream"?

I thought the cube demo came from LunarG. I guess this one is a fork?

billhollings:
> I thought the cube demo came from LunarG. I guess this one is a fork?

Ah. The `cube.c` file itself comes from the `Vulkan-Tools` dependency, and if I'd needed a change to that file, I'd have to change it upstream there. But all the rest of the Cube demo (ie. all the Apple-SDK bits: Xcode projects, `DemoViewController`, etc) originate here.

PR #1674, opened 2022-08-05
MVKDevice consolidate enabling device feature tracking.

billhollings:
- Add `MVKDeviceFeatureStructs.def` file to describe each device feature struct, and define, populate, and manage `MVKDevice` device feature tracking iVars using the list in this file.
- Reorder switch cases in `MVKPhysicalDevice::getFeatures()` to match the list in `MVKDeviceFeatureStructs.def`.
- Where possible, remove extension suffixes from device feature structs and associated `VkStructureType` values.
- Remove const qualifier from device feature enabling structs.

This supports automatic enablement tracking for all device features, ensures such enablement tracking will evolve as new device features are added, and better sets up enabling device features in Vulkan 1.2 and beyond.

### DIFF
--- a/MoltenVK/MoltenVK/GPUObjects/MVKDeviceFeatureStructs.def
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKDeviceFeatureStructs.def
@@ -1,0 +1,71 @@
+/*
+ * MVKDeviceFeatureStructs.def
+ *
+ * Copyright (c) 2015-2022 The Brenwill Workshop Ltd. (http://www.brenwill.com)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * 
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ * 
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+// To use this file, define the macro MVK_DEVICE_FEATURE and optionally
+// MVK_DEVICE_FEATURE_EXTN if needed, then #include this file.
+
+// To add a new device feature struct, simply add a line below.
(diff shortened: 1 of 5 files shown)


MoltenVK/MoltenVK/GPUObjects/MVKDeviceFeatureStructs.def
```diff
@@ -0,0 +1,72 @@
...
+#endif
+#ifndef MVK_DEVICE_FEATURE_EXTN
+#define MVK_DEVICE_FEATURE_EXTN(structName, enumName, extnSfx, flagCount) MVK_DEVICE_FEATURE(structName, enumName, flagCount)
+#endif
+
+// INTEL is a common macro defined elsewhere in the system.
+// We need to undefine it so that it doesn't expand when used in a macro below.
+// Remember the system level value so that we can restore it when we're done.
+#define MVK_INTEL_OLD INTEL
+#undef INTEL
```
cdavis5e, line 37: This and line 70 don't do what you think they do.

cdavis5e (changes requested):

MoltenVK/MoltenVK/GPUObjects/MVKDeviceFeatureStructs.def
billhollings: Ah. Okay. It works for my situation. But where have I gone wrong? What does it actually do?

The issue is that the macros are expanding the `INTEL` in line 68, which is supposed to be used intact as a symbol to append to the struct and `sType` enum. 

Happy to consider suggestions for other options for appending the `INTEL` symbol. Feel free to suggest.

cdavis5e: > But where have I gone wrong? What does it actually do?

It won't save and restore the old definition of `INTEL`.

cdavis5e: Since we're (ironically) not concerned about portability to other platforms, it occurs to me you can use the `push_macro` pragma to do what you want.

billhollings: Ah! A fine suggestion. And thanks for the catch in the first place. Fixed now.

cdavis5e (approved):

PR #1675, opened 2022-08-06
Support Vulkan 1.2 device feature and property structs.

billhollings:
In preparation for Vulkan 1.2, support the following device feature and property structures, and use them to populate the corresponding device feature and property values originally supplied by Vulkan extensions, to ensure a single source of truth for these values:

- `VkPhysicalDeviceVulkan11Features`
- `VkPhysicalDeviceVulkan11Properties`
- `VkPhysicalDeviceVulkan12Features`
- `VkPhysicalDeviceVulkan12Properties`

Disable `VkPhysicalDeviceVulkan12Features::drawIndirectCount` and `VkPhysicalDeviceVulkan12Features::samplerFilterMinmax`, to indicate that Vulkan 1.2 support will not include extensions `VK_KHR_draw_indirect_count` and `VK_EXT_sampler_filter_minmax`, respectively.

Support enabling device features during `VkDevice` creation using `VkPhysicalDeviceVulkan11Features` and `VkPhysicalDeviceVulkan12Features`.

### DIFF
--- a/MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
@@ -93,57 +93,143 @@ void MVKPhysicalDevice::getFeatures(VkPhysicalDeviceFeatures* features) {
 }
 
 void MVKPhysicalDevice::getFeatures(VkPhysicalDeviceFeatures2* features) {
+
+	// Create a SSOT for these Vulkan 1.1 features, which can be queried via two mechanisms here.
+	VkPhysicalDeviceVulkan11Features supportedFeats11 = {
+		.sType = VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_VULKAN_1_1_FEATURES,
+		.pNext = nullptr,
+		.storageBuffer16BitAccess = true,
+		.uniformAndStorageBuffer16BitAccess = true,
+		.storagePushConstant16 = true,
+		.storageInputOutput16 = true,
+		.multiview = true,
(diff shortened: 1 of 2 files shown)


MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
```diff
@@ -93,57 +93,143 @@
...
+		.imagelessFramebuffer = true,
+		.uniformBufferStandardLayout = true,
+		.shaderSubgroupExtendedTypes = _metalFeatures.simdPermute || _metalFeatures.quadPermute,
+		.separateDepthStencilLayouts = true,
+		.hostQueryReset = true,
+		.timelineSemaphore = true,
+		.bufferDeviceAddress = mvkOSVersionIsAtLeast(12.05, 16.0),
+		.bufferDeviceAddressCaptureReplay = false,
+		.bufferDeviceAddressMultiDevice = false,
+		.vulkanMemoryModel = false,									// VK_KHR_vulkan_memory_model
```
cdavis5e, line 160: FYI: Sooner or later, we're gonna have to figure out how to support this. Vulkan memory model support is required for Vulkan 1.3.

cdavis5e (approved):

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
billhollings: Agreed. Setting these `false` is a precursor for now. We will definitely dig into each of the marked extensions further leading up to a Vulkan 1.2 release (and beyond).

PR #1678, opened 2022-08-11
MVKCommandBuffer: Optimize management of command storage.

js6i:
Note: it's a large-ish change and probably would require some more clean up. For example, I think it no longer makes sense to have a separate MVKCmdRenderPassBase. Still, I'm submitting for overall review and suggestions.

The main motivation behind this commit is to reduce the time needed to free or
reset a command buffer. With the current scheme, we need to loop over all
recorded commands and return them to their pools. This can take a significant
amount of time in games, of the order of a millisecond.

With this change, instead of pooling individual commands by their type, we
instead keep chunks of memory in each command buffer, from which we bump
allocate memory for the commands to be recorded. MVKCommandVector is introduced
to use the same memory pool for dynamic arrays of data associated with the
commands.

Pointers to commands that actually have non-trivial destructors (i.e. they keep
references to other objects) are stored separately, so that we can call their
destructors when reseting the command buffer. All the memory simply gets
discarded by setting the current chunk's index and offset to zero.

As a side note, specific applications might opt to not retain pipeline layouts
in bind/push descriptor commands (introduced by 30113fc1235ff51a125dd972504af536c15c9cba)
if it's known not to be needed, to further reduce time spent resetting command buffers.


MoltenVK/MoltenVK/Utility/MVKSmallVector.h
```diff
@@ -59,9 +59,10 @@ using MVKSmallVector = std::vector<T>;
 template<typename Type, typename Allocator = mvk_smallvector_allocator<Type, 0>>
 class MVKSmallVectorImpl
 {
-  Allocator  alc;
-  
+
 public:
+  Allocator  alc;
+
```
cdavis5e, line 65: I wonder if you shouldn't add a constructor to `MVKSmallVector<T>` and its base to initialize the allocator. Then you wouldn't have to reach into the allocator in order to set the command buffer for the bump allocator.

cdavis5e (changes requested):
So much boilerplate... I'm amazed you had the patience to slog through all those code changes!

billhollings:
Thanks for submitting this. I'll review this next week, after the SDK release this weekend.

italomandara:
@js6i is this based on the latest **KhronosGroup:master** ? I'd like to run some benchmarks and comparisons against that, currently I noticed that some shaders that are really slow to compile on some games now compile much faster (but still slow-ish). 
Nevertheless, the average framerate is lower, compared to the latest **KhronosGroup:master** which has some performance improvements commits merged in yesterday.

italomandara:
Bad news: compiled _libMoltenVK.dylib_   from this branch via `xcodebuild build -project MoltenVKPackaging.xcodeproj -scheme "MoltenVK Package (macOS only)" -configuration "Debug" -arch "x86_64"`, even though it works fine for most games, it causes one particular game to freeze randomly. I wasn't able to trace the error i followed the instructions to enable moltenvk logging but it doesn't work with this build if you want more info I'll try to help but i need to understand why logging doesn't work on crossover.

js6i:
> Bad news: compiled _libMoltenVK.dylib_ from this branch via `xcodebuild build -project MoltenVKPackaging.xcodeproj -scheme "MoltenVK Package (macOS only)" -configuration "Debug" -arch "x86_64"`, even though it works fine for most games, it causes one particular game to freeze randomly. I wasn't able to trace the error i followed the instructions to enable moltenvk logging but it doesn't work with this build. I'd love to help but i need to understand why logging doesn't work on crossover. (any help appreciated)

What is the game? You can save logs in CrossOver with the "run with options" dialog, which I think should include MoltenVK errors by default, but you can also set MVK_CONFIG_LOG_LEVEL=3 environment variable. It may be useful to also set METAL_DEVICE_WRAPPER_TYPE=1 to enable Metal API validation.

italomandara:
> > Bad news: compiled _libMoltenVK.dylib_ from this branch via `xcodebuild build -project MoltenVKPackaging.xcodeproj -scheme "MoltenVK Package (macOS only)" -configuration "Debug" -arch "x86_64"`, even though it works fine for most games, it causes one particular game to freeze randomly. I wasn't able to trace the error i followed the instructions to enable moltenvk logging but it doesn't work with this build. I'd love to help but i need to understand why logging doesn't work on crossover. (any help appreciated)
> 
> What is the game? You can save logs in CrossOver with the "run with options" dialog, which I think should include MoltenVK errors by default, but you can also set MVK_CONFIG_LOG_LEVEL=3 environment variable. It may be useful to also set METAL_DEVICE_WRAPPER_TYPE=1 to enable Metal API validation.

elite dangerous: odyssey is the game, i used to get mvk logs even without setting MVK_CONFIG_LOG_LEVEL before, now with this build [mkv-info] and [mv-error] are completely missing, all i can see is dxvk logs i'll try again setting METAL_DEVICE_WRAPPER_TYPE=1 and MVK_CONFIG_LOG_LEVEL=3 instead of 2 to see if that changes it but i think my debug build is broken somehow? by "run with options" do you mean the bottle context menu "run command..." that's what i usually do to save the logs and it used to work with CX stock MVK builds.

italomandara:
@js6i it worked, i was able to enable the logs, thanks

js6i:
@italomandara Do you use any workarounds to launch this game with CrossOver? I'm getting an error saying "Unhandled Exception: mono-io-layer-error" on startup.

italomandara:
> @italomandara Do you use any workarounds to launch this game with CrossOver? I'm getting an error saying "Unhandled Exception: mono-io-layer-error" on startup.

not using particular workarounds that i don't use with other games too i have:
`"MVK_CONFIG_RESUME_LOST_DEVICE" = "1"` and `"MVK_ALLOW_METAL_FENCES" = "1"` set
I'm using a newer version of DXVK not the stock one, but this works fine with other MVK builds.

Other MVK builds included the latest master work just fine. 
I also rebased your branch with the latest master to make sure it wasn't some other issue that have been fixed recently. same issue.

phush0:
This WIP also crash Guild Wars 2. The easiest way to reproduce is fast travel. Upon appear on new spot, loading textures and assets will make game crash or freeze entirely.

js6i:
@italomandara, how to trigger this freeze? Can I expect to hit it by playing for a few minutes through the tutorial?

phush0:
you have to unlock fast travel point and travel from somewhere to it that's it for me triggers every time

italomandara:
@js6i the freeze is random in Elite dangerous odyssey, usually happens at the main menu, but sometimes while starting playing, I'm using M1Max, can't confirm you get the same on x86 machines and discrete GPUs

phush0:
I am also with m1 max

js6i:
@italomandara I think I'll need more info about your setup. On my M1 Air Elite Dangerous seems to work fine with this change. Which version of CrossOver are you using? Do you use the version of DXVK that ships with it? Did you change any potentially important registry settings or environment variables? Esync? High res/retina mode? Fullscreen, windowed fullscreen, other graphics settings in game? Unfortunately I don't have direct access to M1 Max specifically..

js6i:
I didn't reproduce the freeze in GW2 either, but I spotted a possible problem. This PR is now updated and also rebased against current master. Feel free to retest..
@italomandara @phush0

phush0:
yes now it will not crash, also seems that helps on fps in group events. I have played with patch like hour - hour and a half and traveled several times in very populated zones where before will hang and is ok

MoltenVK/MoltenVK/Utility/MVKSmallVector.h
js6i: Maybe, to me that seems to add complication for no clear reason, thus I'm leaving it for others to decide. Are you thinking doing something of the sort:
`_bindings = MVKCommandVector(MVKCommandVectorAllocator<BlahBlah>(cmdBuff));`
in `setContent` functions, or passing the command buffer to all MVKCommand constructors and putting
`_bindings(MVKCommandVector(MVKCommandVectorAllocator<BlahBlah>(cmdBuff)))`
in the inializer list, or something else?

italomandara:
> @italomandara I think I'll need more info about your setup. On my M1 Air Elite Dangerous seems to work fine with this change. Which version of CrossOver are you using? Do you use the version of DXVK that ships with it? Did you change any potentially important registry settings or environment variables? Esync? High res/retina mode? Fullscreen, windowed fullscreen, other graphics settings in game? Unfortunately I don't have direct access to M1 Max specifically..

@js6i I do use some tweaks, but they're practically the same i usually use and EDO doesn't crash with those, I'll send you a configuration after testing if it still crashes after your rebase.

italomandara:
When I tested it all went well, i turned on async in dxvk but never forced MTLEvents, so this might narrow down the issue to MTLEvents?

phush0:
It happens with single queue also when async is enabled

js6i:
Rebased on current master and some minor clean ups.

@phush0, are you getting any error messages from DXVK or MoltenVK? What about Metal runtime? (run with METAL_DEVICE_WRAPPER_TYPE=1) Isn't dxvk-async just compiling shaders in a thread? That in principle shouldn't care about how commands are stored in buffers.

phush0:
Will check and will report back. Basically display just hangs. Actually async will skip draw at first and introduce it later as it is compiled in background, so it is more a hack than solution. May be this skip will lead to crash I am not entirely sure.

Gcenx:
I’d applied the first version of this on 1.11.1 and haven’t seen any issue with Witcher3.

Tested in the training grounds where you’d usually get a lockup when hit (need to have MVK_CONFIG_RESUME_LOST_DEVICE=1 set) completed it without problem.

Gcenx:
@js6i any reason to still consider this WIP?

italomandara:
I tested this on my entire game library and it works

phush0:
with latest version I don't have hangs also

oscarbg:
sadly seems will miss the next Vulkan SDK feature cutoff (https://github.com/KhronosGroup/MoltenVK/issues/1736)
still interesting if can be merged soon after, seeing it works now..

Gcenx:
@oscarbg agreed but we don’t know if there’s a reason it’s still marked as WIP

Once the next release is marked I’ll provide the usual stock package along with another that includes this PR and a coupe of other interesting patches/hacks.

oscarbg:
@Gcenx nice.. it's one of these additional patches the UE4 black screen fix: https://github.com/nastys/MoltenVK/releases/tag/ue4-workaround-1? 
I just opened an issue (https://github.com/KhronosGroup/MoltenVK/issues/1741) asking if this workaround makes sense to be integrated into MoltenVK and enabled via some env variable given it fixes lots of games on MacOS..

Gcenx:
@oscarbg yes the UE4hack is included along with zeroinit

@nastys I know you said you were still working on tracking down the root cause of this.

MoltenVK/MoltenVK/Utility/MVKSmallVector.h
aerofly: @js6i , I highly discourage this change to MVKSmallVector. The allocator should not be exposed at all and the overall design wasn't ment to be used with a specific constructor as well. The allocator type is a pure template argument.
I think the problem is, that MVKCommandVectorAllocator is already bad by design, it should not have the member cmdBuffer and it should not call mvkPushCommandMemory directly.
There are multiple ways of redesigning this, but since I am not really familiar with MoltenVK, I can't suggest a good solution right now, since I am afraid I will brake other things.
But please do not change MVKSmallVector with the above change, it will introduce all kind of issues in the future.

js6i: @aerofly, I don't see what's the argument you're making, other than saying it's bad. Setting the allocator with accompanying data for an expandable array is a very simple thing to do, and I don't understand how that's going to be a source of issues.

aerofly: Sorry, I didn't mean to be rude here. What I tried to say is, that exposing the allocator and actually accessing allocator elements from another class is basically mixing separate concepts. The allocator as well as the actual vector should not be intertwined like this.

I have added MVKSmallVector to MoltenVK to have a std::vector compatible container that allows the easy use of pre allocated stack elements to be used, to avoid new/delete's.
The way it is used now would not be STL compatible.

Now for this special case, I recommend to write a special class that encapsulates everything that is intended here.

js6i: Regarding the first point, I still don't get it. If you want to use some allocator for an array or whatever, you gotta set it and its things. I don't see how that's mixing concepts. If I put the pointer to command buffer directly into the vector class, then I'd agree it would be mixing unrelated things.

I don't know if we need STL compatibility for anything. If we do, that's probably a valid argument.

That said, I also don't mind having dedicated code for this.

phush0:
I still have hangs in Guild Wars 2 even with the last patch. This is what I have in log as errors 

```
[mvk-error] VK_ERROR_INVALID_SHADER_NV: Unable to convert SPIR-V to MSL:
MSL conversion error: Vertex attribute type mismatch between host and shader


[mvk-error] VK_ERROR_INVALID_SHADER_NV: Vertex shader function could not be compiled into pipeline. See previous logged error.
err:   DxvkGraphicsPipeline: Failed to compile pipeline
err:     vs  : VS_7628e2c6c3fd5510b370e8166966c9f45137cad7
err:     fs  : FS_6dbf94ee014a451d8ba108e51e15e38b602ce07a
err:     attr 0 : location 0, binding 0, format VK_FORMAT_R32G32B32_SFLOAT, offset 0
err:     attr 1 : location 1, binding 0, format VK_FORMAT_R32G32B32_SFLOAT, offset 12
err:     attr 2 : location 2, binding 0, format VK_FORMAT_R8G8B8A8_UINT, offset 0
err:     attr 3 : location 3, binding 0, format VK_FORMAT_R16G16_SFLOAT, offset 48
err:     binding 0 : binding 0, stride 52, rate 0, divisor 0
[mvk-error] VK_ERROR_INVALID_SHADER_NV: Unable to convert SPIR-V to MSL:
MSL conversion error: Vertex attribute type mismatch between host and shader


[mvk-error] VK_ERROR_INVALID_SHADER_NV: Vertex shader function could not be compiled into pipeline. See previous logged error.
err:   DxvkGraphicsPipeline: Failed to compile pipeline
err:     vs  : VS_f89fed9e2085bf1ab8b74645384b6177cf16fa67
err:     fs  : FS_847fb2fbf55b3d81b0d05c9786d0576f485e7be9
err:     attr 0 : location 0, binding 0, format VK_FORMAT_R32G32B32_SFLOAT, offset 0
err:     attr 1 : location 1, binding 0, format VK_FORMAT_R32G32B32_SFLOAT, offset 12
err:     attr 2 : location 2, binding 0, format VK_FORMAT_R8G8B8A8_UINT, offset 0
err:     attr 3 : location 3, binding 0, format VK_FORMAT_R32G32_SFLOAT, offset 24
err:     attr 4 : location 7, binding 1, format VK_FORMAT_R32G32B32A32_SFLOAT, offset 48
err:     attr 5 : location 6, binding 1, format VK_FORMAT_R32G32B32A32_SFLOAT, offset 32
err:     attr 6 : location 5, binding 1, format VK_FORMAT_R32G32B32A32_SFLOAT, offset 16
err:     attr 7 : location 4, binding 1, format VK_FORMAT_R32G32B32A32_SFLOAT, offset 0
err:     binding 0 : binding 0, stride 32, rate 0, divisor 0
err:     binding 1 : binding 1, stride 64, rate 1, divisor 1
[mvk-error] VK_ERROR_INVALID_SHADER_NV: Unable to convert SPIR-V to MSL:
MSL conversion error: Vertex attribute type mismatch between host and shader
```

after several such hangs of the app I reach the point where whole laptop freeze and reboots itself

Gcenx:
If your using CrossOver-22.0.1 did you use the shipped DXVK or replace it with a third-party build?

phush0:
I am using latest build from your page

Gcenx:
> I am using latest build from your page and yes it is CrossOver-22.0.1

The latest DXVK-macOS build would be 1.10.3 that includes DXVK-CX hacks (that are still relevant) along with DXVK-async that’s only auto enabled on select titles.

phush0:
I don't use async for guild wars. Easiest way to reproduce e port to big city from somewhere. Loading objects will lead to hang.

italomandara:
That seems a dxvk error not a moltenvk one...

phush0:
this don't change the fact that Guild Wars hangs with this pr and not without it

phush0:
```
-[MTLDebugBlitCommandEncoder dealloc]:89: failed assertion `Command encoder released without endEncoding.'
0174:02fc:trace:seh:dispatch_exception code=c000001d flags=0 addr=00007FF8004E0D5D ip=00007FF8004E0D5D tid=02fc
code=c000001d flags=0 addr=0x689d768a ip=689d768a tid=02fc
```

This is what I found when game hang

Gcenx:
@phush0 I noticed at the end of the log `[mvk-info] Compiling Metal shader with FastMath enabled.` have you tried disabling FastMath?, on discord nas had shown some titles having graphical issues with this enabled so maybe that's a contributing factor.

phush0:
> @phush0 I noticed at the end of the log `[mvk-info] Compiling Metal shader with FastMath enabled.` have you tried disabling FastMath?, on discord nas had shown some titles having graphical issues with this enabled so maybe that's a contributing factor.

yes no change same assert and hang

alphavector:
It works for Foxhole game

Gcenx:
@js6i could you please rebase this

marzent:
Took the liberty of rebasing here https://github.com/marzent/MoltenVK/tree/cmdbuffer-storage2 if anyone needs it

italomandara:
@js6i this is not from your code changes but maybe your changes revealed some existing issues
in MVKCmdTransfer.mm lines 652 and 864 i can see ```[mtlRendEnc popDebugGroup]``` comes before ```[mtlRendEnc endEncoding]``` 
now if the error is ```failed assertion 'Command encoder released without endEncoding.'``` could it be because ```[mtlRendEnc endEncoding]``` should be called before releasing the command encoder instead?
Forgive me if I'm mistaken.

cdavis5e:
> could it be because `[mtlRendEnc endEncoding]` should be called before releasing the command encoder instead?

No. `[mtlRendEnc popDebugGroup]` doesn't release the command encoder. It undoes the effect of a prior `-pushDebugGroup:`.

italomandara:
> > could it be because `[mtlRendEnc endEncoding]` should be called before releasing the command encoder instead?
> 
> No. `[mtlRendEnc popDebugGroup]` doesn't release the command encoder. It undoes the effect of a prior `-pushDebugGroup:`.

right it's even done in the same order in the apple docs, sorry.

Gcenx:
@js6i is there and issues you’ve encountered when using this that’s caused it to be still marked as WIP?

Gcenx:
@js6i could you rebase this PR to master, also is there any reason this is sitting in limbo?

js6i:
I'll most likely rebase on a newer version soon, the problem is that I haven't been able to reproduce the issue @phush0 was getting, and understandably @billhollings is not eager to merge regressive changes. Perhaps it's something that will go away on its own after rebasing.

Gcenx:
@js6i here is an easy example of a game crashing quickly with this applied.

Jedi: Fallen Order, fails to load a save without the game will load the save though the game will eventually crash out when remembering an ability (first one being wall running) when using DXVK not the case when using WineD3D Vulkan though it’s unplayable with that. The issue is probably with spirv-cross.

Though the above may have been resolved via https://github.com/KhronosGroup/MoltenVK/commit/75b4d26fbcccc610a26a929da7a2983c8b7d943c as something fails to compile causing the pipeline to get locked up.

I suspect this PR was just uncovering these issues much sooner than without.

phush0:
Just rebase will test again

PR #1680, opened 2022-08-12
Update dependency libraries to match Vulkan SDK 1.3.224.

billhollings:
- Update dependency libraries to match Vulkan SDK 1.3.224.
- Update What's New document.


spnda (commented):
Is `Templates/spirv-tools/build.zip` supposed to be part of this commit?

Also, should mention the requirement of Metal 3/macOS 13/iOS 16?
https://github.com/billhollings/MoltenVK/blob/a87e2235434bed6c8c5e942c72307cc12303a1b9/Docs/Whats_New.md?plain=1#L23

billhollings:
> Is `Templates/spirv-tools/build.zip` supposed to be part of this commit?

Yes. It's built from `SPIRV-Tools` and cached in `Templates/spirv-tools/build.zip` when we update the version of `SPIRV-Tools`.

> Also, should mention the requirement of Metal 3/macOS 13/iOS 16? https://github.com/billhollings/MoltenVK/blob/a87e2235434bed6c8c5e942c72307cc12303a1b9/Docs/Whats_New.md?plain=1#L23

Based on feedback from Apple, `[MTLBuffer gpuAddress]` doesn't require Metal 3. It requires Tier 2 argument buffer support:

https://github.com/KhronosGroup/MoltenVK/blob/a87e2235434bed6c8c5e942c72307cc12303a1b9/MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm#L3026-L3030

PR #1681, opened 2022-08-15
Add MVK_USE_CEREAL build option to avoid use of Cereal library.

scandit-opm:
We cannot use Cereal due to legal reasons. However, we are fine with not having the pipeline cache serialization feature.

### DIFF
--- a/MoltenVK/MoltenVK/GPUObjects/MVKPipeline.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKPipeline.mm
@@ -25,7 +25,11 @@
 #include "MTLRenderPipelineDescriptor+MoltenVK.h"
 #include "mvk_datatypes.hpp"
 
-#ifndef MVK_EXCLUDE_CEREAL
+#ifndef MVK_USE_CEREAL
+#define MVK_USE_CEREAL (1)
+#endif
+
+#if MVK_USE_CEREAL
 #include <cereal/archives/binary.hpp>
 #include <cereal/types/string.hpp>
 #include <cereal/types/vector.hpp>
@@ -1973,7 +1977,7 @@ MVKShaderLibraryCache* MVKPipelineCache::getShaderLibraryCache(MVKShaderModuleKe
 
 #pragma mark Streaming pipeline cache to and from offline memory
 
-#ifndef MVK_EXCLUDE_CEREAL
+#if MVK_USE_CEREAL
 static uint32_t kDataHeaderSize = (sizeof(uint32_t) * 4) + VK_UUID_SIZE;
 #endif
 
@@ -2009,7 +2013,7 @@ protected:
 // returns the number of bytes required to serialize the contents of this pipeline cache.
 // This is the compliment of the readData() function. The two must be kept aligned.
 VkResult MVKPipelineCache::writeData(size_t* pDataSize, void* pData) {
-#ifndef MVK_EXCLUDE_CEREAL
+#if MVK_USE_CEREAL
 	lock_guard<mutex> lock(_shaderCacheLock);
 
 	try {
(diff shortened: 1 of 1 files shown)


MoltenVK/MoltenVK/GPUObjects/MVKPipeline.mm
```diff
@@ -25,9 +25,11 @@
 #include "MTLRenderPipelineDescriptor+MoltenVK.h"
 #include "mvk_datatypes.hpp"
 
+#ifndef MVK_EXCLUDE_CEREAL
```
billhollings, line 28: Please modify this to align with the conventions for definitions of other `MVK_` build switches. Something like:

```
#ifndef MVK_USE_CEREAL
#	define MVK_USE_CEREAL	1
#endif
```

Then when using:

```
#if MVK_USE_CEREAL
...
[#else
...]
#endif
```

MoltenVK/MoltenVK/GPUObjects/MVKPipeline.mm
```diff
@@ -2037,11 +2042,15 @@ static MTLVertexFormat mvkAdjustFormatVectorToSize(MTLVertexFormat format, uint3
 		*pDataSize = 0;
 		return reportError(VK_INCOMPLETE, "Error writing pipeline cache data: %s", ex.what());
 	}
+#else
+	MVKAssert(false, "Pipeline cache serialization is unavailable. To enable pipeline cache serialization, build MoltenVK without the MVK_EXCLUDE_CEREAL build setting.");
```
billhollings, line 2046: Not sure that `MVKAssert` is appropriate here, as it will crash the app in a Debug build. It is probably better to just simply set `*pDataSize = 0;` and report the result as incomplete with an appropriate error message, as you are doing. 

And do make sure you set `*pDataSize = 0;`, to indicate nothing was exported, for those cases where the assert doesn't cause a crash.

MoltenVK/MoltenVK/GPUObjects/MVKPipeline.mm
```diff
@@ -2077,11 +2086,15 @@ static MTLVertexFormat mvkAdjustFormatVectorToSize(MTLVertexFormat format, uint3
 	// Mark the end of the archive
 	cacheEntryType = MVKPipelineCacheEntryTypeEOF;
 	writer(cacheEntryType);
+#else
+	MVKAssert(false, "Pipeline cache serialization is unavailable. To enable pipeline cache serialization, build MoltenVK without the MVK_EXCLUDE_CEREAL build setting.");
```
billhollings, line 2090: `MVKAssert` is okay here, because it would be a MoltenVK code error to call this function when CEREAL is not available.

MoltenVK/MoltenVK/GPUObjects/MVKPipeline.mm
```diff
@@ -2151,6 +2164,9 @@ static MTLVertexFormat mvkAdjustFormatVectorToSize(MTLVertexFormat format, uint3
 	} catch (cereal::Exception& ex) {
 		setConfigurationResult(reportError(VK_SUCCESS, "Error reading pipeline cache data: %s", ex.what()));
 	}
+#else
+	MVKAssert(false, "Pipeline cache serialization is unavailable. To enable pipeline cache serialization, build MoltenVK without the MVK_EXCLUDE_CEREAL build setting.");
```
billhollings, line 2168: Similarly, I don't think `MVKAssert` is appropriate here. Probably better to return an error result:

	setConfigurationResult(reportError(VK_ERROR_INITIALIZATION_FAILED, ...

While you're at it...can you change the the existing return of `VK_SUCCESS` to `VK_ERROR_INITIALIZATION_FAILED`, as well, please, as that existing return value looks to be a mistake.

billhollings (changes requested):
Thanks for the submission. I've suggested some form and behavior tweaks.

Out of curiosity, and so we can assess the use of Cereal, what about the Cereal license do you have legal trouble with? Is it the lack of patent protection that is available in Apache 2.0?

MoltenVK/MoltenVK/GPUObjects/MVKPipeline.mm
scandit-opm: Ok, done.

MoltenVK/MoltenVK/GPUObjects/MVKPipeline.mm
scandit-opm: Ok, done.

MoltenVK/MoltenVK/GPUObjects/MVKPipeline.mm
scandit-opm: Ok.

MoltenVK/MoltenVK/GPUObjects/MVKPipeline.mm
scandit-opm: Ok, done.

scandit-opm:
> Thanks for the submission. I've suggested some form and behavior tweaks.
> 
> Out of curiosity, and so we can assess the use of Cereal, what about the Cereal license do you have legal trouble with? Is it the lack of patent protection that is available in Apache 2.0?

I am not into all the details, but AFAIK cereal is BSD-3 and not Apache 2.0, right ? I think that was the issue.

billhollings:
Thanks for making the requested changes. LGTM now.

> I am not into all the details, but AFAIK cereal is BSD-3 and not Apache 2.0, right ? I think that was the issue.

I was curious, because BSD-3 is one of the most open, and least restrictive licenses, and specifically, is more open and less restrictive than Apache 2.0. So I'm curious what your legal team was concerned about.

billhollings (approved):

scandit-opm:
Thanks for the fast acceptance.

If I have understood our legal department correctly, then BSD-3 requires attribution, i.e. our customers would be required to state that they are using Cereal. We want to avoid that.

PR #1683, opened 2022-08-16
Fix retention of MVKSwapchain for future drawable presentations.

### LINKED ISSUE #1679: MVKPresentableSwapchainImage::presentCAMetalDrawable occasionally triggers observer detach error
MoltenVK attaches an observer to the view holding the `CAMetalLayer` to detect the situation where the layer is swapped for a different layer in the view.

There is an unpredictable crash bug associated with incorrectly releasing the observer that sometimes occurs during long CTS runs. This error occurs once every 2 or 3 full CTS runs, occurs late in the run, but at unpredictable times and in seemingly random tests.

```
2022-08-09 14:53:40.627 deqp-vk[40203:34550812] *** Terminating app due to uncaught exception 'NSRangeException', reason: 'Cannot remove an observer <MVKBlockObserver 0x60000de4dac0> for the key path "layer" from <AGXG13GFamilyTexture 0x15080d140> because it is not registered as an observer.'
*** First throw call stack:
(
	0   CoreFoundation                      0x00000001bd65d1a8 __exceptionPreprocess + 240
	1   libobjc.A.dylib                     0x00000001bd3a7e04 objc_exception_throw + 60
	2   Foundation                          0x00000001be4b2c94 -[NSObject(NSKeyValueObserverRegistration) _removeObserver:forProperty:] + 676
	3   Foundation                          0x00000001be4b299c -[NSObject(NSKeyValueObserverRegistration) removeObserver:forKeyPath:] + 144
	4   Foundation                          0x00000001be4b2624 -[NSObject(NSKeyValueObserverRegistration) removeObserver:forKeyPath:context:] + 204
	5   libMoltenVK.dylib                   0x00000001174470f4 -[MVKBlockObserver stopObserving] + 36
	6   libMoltenVK.dylib                   0x0000000117447020 -[MVKBlockObserver dealloc] + 24
	7   libMoltenVK.dylib                   0x00000001174fab24 _ZN12MVKSwapchainD2Ev + 64
	8   libMoltenVK.dylib                   0x00000001174fab98 _ZN12MVKSwapchainD0Ev + 12
	9   libMoltenVK.dylib                   0x000000011748768c _ZN28MVKPresentableSwapchainImageD2Ev + 140
	10  libMoltenVK.dylib                   0x00000001174876d8 _ZN28MVKPresentableSwapchainImageD0Ev + 12
	11  libMoltenVK.dylib                   0x0000000117487474 ___ZN28MVKPresentableSwapchainImage22presentCAMetalDrawableEPU26objcproto15CAMetalDrawable11objc_object20MVKPresentTimingInfo_block_invoke + 120
	12  QuartzCore                          0x00000001c464ffcc -[CAMetalDrawable didPresentAtTime:] + 172
	13  QuartzCore                          0x00000001c4460b98 _ZL31layer_private_cleanup_callbacksP20_CAMetalLayerPrivate + 264
	14  QuartzCore                          0x00000001c4460434 _ZL25layer_private_flush_asyncP20_CAMetalLayerPrivate + 228
	15  QuartzCore                          0x00000001c464fa44 _ZL31layer_private_flush_backbuffersP20_CAMetalLayerPrivate + 40
	16  QuartzCore                          0x00000001c448cf70 _ZN2CA6Render17post_notificationENS0_16NotificationNameEPNS0_6ObjectEPvb + 964
	17  libdispatch.dylib                   0x00000001bd34c1b4 _dispatch_client_callout + 20
	18  libdispatch.dylib                   0x00000001bd34f670 _dispatch_continuation_pop + 500
	19  libdispatch.dylib                   0x00000001bd3628e0 _dispatch_source_invoke + 1596
	20  libdispatch.dylib                   0x00000001bd35e658 _dispatch_kevent_worker_thread + 1292
	21  libsystem_pthread.dylib             0x00000001bd50c398 _pthread_wqthread + 344
	22  libsystem_pthread.dylib             0x00000001bd50b080 start_wqthread + 8
)
libc++abi: terminating with uncaught exception of type NSException
```

billhollings:
When presenting drawables in the future, the `MVKPresentableSwapchainImage`and `MVKSwapchain` were both being retained by the drawable-presented callback, so that the presentation timing info can be recorded on the swapchain. Unfortunately, in the case where the presentation timing is set far enough into the future (I'm looking at you CTS), the swapchain, and even the `CAMetalLayer's` view may be destroyed, causing occasional bad access crashes.

- `MVKSwapchainImage` don't retain() the swapchain, and move clearing the swapchain from the destructor to destroy(), so the `MVKSwapchain` is not necessarily retained by the `MVKSwapchainImage`, for a drawable being presented in the future.
- `MVKSwapchainImage` add a lock around clearing swapchain and accessing it from callbacks.
- Add lockable `releaseLayer()` function in both `MVKSwapchain` and `MVKSurface`, which is called from both layer observer and destructor, to handle race conditions better.
- `MVKSwapchain::initCAMetalLayer()` call `MVKSurface::getCAMetalLayer()` only once.
- Update MoltenVK version to 1.1.12.
- Update What's New document.

Fix #1679.

### DIFF
--- a/MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
@@ -1190,12 +1190,16 @@ MVKSwapchainImage::MVKSwapchainImage(MVKDevice* device,
 									 uint32_t swapchainIndex) : MVKImage(device, pCreateInfo) {
 	_swapchain = swapchain;
 	_swapchainIndex = swapchainIndex;
-
-	_swapchain->retain();
 }
 
-MVKSwapchainImage::~MVKSwapchainImage() {
-	if (_swapchain) { _swapchain->release(); }
+void MVKSwapchainImage::detachSwapchain() {
+	lock_guard<mutex> lock(_swapchainLock);
+	_swapchain = nullptr;
+}
+
+void MVKSwapchainImage::destroy() {
+	detachSwapchain();
+	MVKImage::destroy();
 }
 
 
@@ -1308,10 +1312,19 @@ void MVKPresentableSwapchainImage::presentCAMetalDrawable(id<MTLCommandBuffer> m
 
 	_swapchain->willPresentSurface(getMTLTexture(0), mtlCmdBuff);
 
-	// Get current drawable now. Don't retrieve in handler, because a new drawable might be acquired by then.
+	// According to Apple, it is more performant to call MTLDrawable present from within a
+	// MTLCommandBuffer scheduled-handler than it is to call MTLCommandBuffer presentDrawable:.
+	// But get current drawable now, intead of in handler, because a new drawable might be acquired by then.
+	// Attach present handler before presenting to avoid race condition.
 	id<CAMetalDrawable> mtlDrwbl = getCAMetalDrawable();
 	[mtlCmdBuff addScheduledHandler: ^(id<MTLCommandBuffer> mcb) {
-		presentCAMetalDrawable(mtlDrwbl, presentTimingInfo);
+		if (presentTimingInfo.hasPresentTime) {
+			// Convert from nsecs to seconds for Metal
+			addPresentedHandler(mtlDrwbl, presentTimingInfo);
+			[mtlDrwbl presentAtTime: (double)presentTimingInfo.desiredPresentTime * 1.0e-9];
+		} else {
+			[mtlDrwbl present];
+		}
 	}];
 
 	MVKSwapchainSignaler signaler;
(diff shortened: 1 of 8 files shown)


MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
```diff
@@ -1190,12 +1190,16 @@
...
-	_swapchain->retain();
+void MVKSwapchainImage::detachSwapchain() {
+	lock_guard<mutex> lock(_swapchainLock);
+	_swapchain = nullptr;
 }
 
-MVKSwapchainImage::~MVKSwapchainImage() {
-	if (_swapchain) { _swapchain->release(); }
+void MVKSwapchainImage::destroy() {
+	detachSwapchain();
```
cdavis5e, line 1201: This is the only call site for `detachSwapchain()`. Is that intentional? The code in `presentCAMetalDrawable()` implies that it was supposed to be called elsewhere, e.g. when the swapchain is destroyed. Otherwise, I fail to see how the swapchain pointer becomes `nullptr` after the swapchain is destroyed.

cdavis5e (changes requested):

MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
billhollings: For each swap chain image, `MVKSwapchain::destroy()` calls `MVKPresentableSwapchainImage::destroy()` (via `_device->destroyPresentableSwapchainImage()`).

So the app request to destroy the swapchain triggers the `detachSwapchain()` and breaks the link back to the swapchain from the image. The swapchain itself is not retained by the images, so it is free to deallocate at that time.

In the end, the drawable-presented callback block will hold onto the swapchain image, but not the swapchain itself once it has been destroyed by the app. When that callback block executes, it will find the swapchain gone, and won't record the timings.

cdavis5e: > For each swap chain image, `MVKSwapchain::destroy()` calls `MVKPresentableSwapchainImage::destroy()` (via `_device->destroyPresentableSwapchainImage()`).
> ...
> In the end, the drawable-presented callback block will hold onto the swapchain image, but not the swapchain itself once it has been destroyed by the app. When that callback block executes, it will find the swapchain gone, and won't record the timings.

All right, I see now.

cdavis5e (approved):

PR #1686, opened 2022-08-17
Remove KHR suffix typo from vkGetDeviceMemoryOpaqueCaptureAddress promotion to Vulkan 1.2.

billhollings:
Fixes typo from earlier promotion of extension calls to Vulkan 1.2.


spnda (approved):
oops

billhollings:
> I hope it's the only one though.

Yup. Looks like it was the only one. That typo crashed a CTS run. Fixing it allowed the run to complete.

PR #1690, opened 2022-08-18
Fix crash in vkCreateSwapchainKHR() on macOS 10.14 and earlier

bangnoise:
NSScreen's `localizedName` property is only available on macOS 10.15+, so check before use.

### DIFF
--- a/MoltenVK/MoltenVK/GPUObjects/MVKSwapchain.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKSwapchain.mm
@@ -402,10 +402,12 @@ void MVKSwapchain::initSurfaceImages(const VkSwapchainCreateInfoKHR* pCreateInfo
 		_presentableImages.push_back(_device->createPresentableSwapchainImage(&imgInfo, this, imgIdx, NULL));
 	}
 
-#if MVK_MACOS && !MVK_MACCAT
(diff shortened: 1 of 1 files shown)


cdavis5e (approved):

MoltenVK/MoltenVK/GPUObjects/MVKSwapchain.mm
```diff
@@ -402,10 +402,12 @@ static inline CIE1931XY VkXYColorEXTToCIE1931XY(VkXYColorEXT xy) {
 		_presentableImages.push_back(_device->createPresentableSwapchainImage(&imgInfo, this, imgIdx, NULL));
 	}
 
-#if MVK_MACOS && !MVK_MACCAT
-	NSString* screenName = _mtlLayer.screenMVK.localizedName;
-#else
 	NSString* screenName = @"Main Screen";
+#if MVK_MACOS && !MVK_MACCAT
+	if ([_mtlLayer.screenMVK respondsToSelector:@selector(localizedName)])
+	{
```
billhollings, line 408:
```suggestion
	if ([_mtlLayer.screenMVK respondsToSelector:@selector(localizedName)]) {
```

Style nit.

billhollings (changes requested):
Good catch, and thanks for the fix. Just a minor style consistency nit fix to add.

billhollings (approved):

PR #1691, opened 2022-08-18
Add support for the VK_KHR_shader_float_controls.

billhollings:
- Add support for `VK_KHR_shader_float_controls` extension.
- Enable `shaderSignedZeroInfNanPreserveFloat16` and `shaderSignedZeroInfNanPreserveFloat32`. Other float control properties are not settable in Metal.
- `MVKShaderModule` debug log whether compiling with fast math enabled.
- Update `MoltenVK_Runtime_UserGuide.md` and `Whats_New.md` documents.
- `runcts` script enable `MVK_CONFIG_FAST_MATH_ENABLED` by default.

### DIFF
--- a/Docs/MoltenVK_Runtime_UserGuide.md
+++ b/Docs/MoltenVK_Runtime_UserGuide.md
@@ -289,6 +289,7 @@ In addition to core *Vulkan* functionality, **MoltenVK**  also supports the foll
 - `VK_KHR_sampler_ycbcr_conversion`
 - `VK_KHR_separate_depth_stencil_layouts`
 - `VK_KHR_shader_draw_parameters`
+- `VK_KHR_shader_float_controls`
 - `VK_KHR_shader_float16_int8`
 - `VK_KHR_shader_subgroup_extended_types` *(requires Metal 2.1 on Mac or Metal 2.2 and Apple family 4 on iOS)*
 - `VK_KHR_storage_buffer_storage_class`
--- a/Docs/Whats_New.md
+++ b/Docs/Whats_New.md
@@ -18,6 +18,8 @@ MoltenVK 1.1.12
 
 Released TBD
 
+- Add support for extensions:
+	- `VK_KHR_shader_float_controls`
 - Fix occassional crash from retention of `MVKSwapchain` for future drawable presentations.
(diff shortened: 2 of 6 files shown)


cdavis5e:
But we don't even support `OpPtr[Not]Equal` and `OpPtrDiff` yet...

cdavis5e (changes requested):
I don't think we should expose `VK_KHR_spirv_1_4` until the new operations it adds are supported in SPIRV-Cross.

billhollings:
> I don't think we should expose `VK_KHR_spirv_1_4` until the new operations it adds are supported in SPIRV-Cross.

Yeah. A fair point. I was focused on that extension being associated with `VK_KHR_shader_float_controls`. But you are correct that if we advertise `VK_KHR_spirv_1_4` or move to 1.2, we should make a decent attempt to support SPIR-V 1.4 and 1.5.

I've reverted that commit now.

spnda:
Are `shaderSignedZeroInfNanPreserveFloat16` and `shaderSignedZeroInfNanPreserveFloat32` already supported in SPIRV-Cross? If not it also seems like none of the features are actually required for the extension by the spec, or am I reading that wrong?

billhollings:
> Are `shaderSignedZeroInfNanPreserveFloat16` and `shaderSignedZeroInfNanPreserveFloat32` already supported in SPIRV-Cross? I don't see a update of the SPIRV-Cross checkout commit. If not it also seems like none of the features are actually required for the extension by the spec, or am I reading that wrong?

Yes. They are already supported by SPIRV-Cross, in the sense that we were already retrieving the info from SPIRV-Cross, and making compilation use of it here:

https://github.com/KhronosGroup/MoltenVK/blob/d447ea79400f0250d5e16cb669be18406240d885/MoltenVKShaderConverter/MoltenVKShaderConverter/SPIRVToMSLConverter.cpp#L492

billhollings:
BTW...for `shaderSignedZeroInfNanPreserveFloat16` and `shaderSignedZeroInfNanPreserveFloat32`, of the 1,503 new CTS tests those two flags expose, we pass 1,493 out of the box. We'll deal with the 10 failures during future conformance cleanups.

cdavis5e (approved):

PR #1693, opened 2022-08-21
Replace MTLFence semaphores with forcing a single queue

### LINKED ISSUE #796: Issues using MTLFence for VkSemaphore
Hello,

From what I understand, it seems the scopes for MTLFence and VkSemaphore are radically different.

The vulkan specs states that semaphore are intended for queue<->queue synchronization:
- signaling a semaphore will include any command submitted in the queue before the signal operation as its source scope.
- waiting on a semaphore will block all commands submitted in the queue after the wait operation

However, metal specs states that:
- MTLFence is "An object that can capture, track, and manage resource dependencies across command encoders."
- "Updates the given fence to capture all GPU work enqueued by the command encoder up to this specific point."
- "Prevents further GPU commands from being enqueued by the command encoder until the given fence is reached."

So a MTLFence's scope only includes commands in the encoder where it's being signaled/waited upon.

This means that, if we have 3 encoders A, B and C, and a MTLFence signaled by A waited by B, we have the guarantee that B will execute after A is completed. But C is free to execute in parallel to both A and B (whereas a VkSemaphore would guarantee that both B and C would wait on A to complete).

Moreover, the current pattern to inject an empty blit encoder to wait or signal a semaphore means that signaling and waiting scope are empty, and are not actually synchronizing anything.

I'm assuming though that it has been tested and found to work on Mac at least. So I'd assume that Mac's GPU doesn't have (for now) the technical capability to do fine-grained encoder only scopes, and result in a MTLFence behaving like a MTLEvent. But this completely breaks on iOS (using iPhone XS 13.2.3).

By un-allowing VkSemaphore to use the MTLFence implementation, and reverting to the use of MTLEvent, the synchronization works as expected.

MTLFence however is appropriate to handle subpass synchronization inside a vk render pass, as it exactly matches VkSubpassDependency's semantic (assuming each subpass have its encoder).

tellowkrinkle:
Closes #796

The previous MTLFence code placed MTLFences on extra blit encoders, however MTLFences only prevent other commands on their encoder from being reordered, not the entire command buffer, meaning they were doing nothing at all.

On the other hand, MTLFences only actually are useful if you're using `MTLHazardTrackingModeUntracked`.  Resources with `MTLHazardTrackingModeTracked` act like they're paired with a MTLFence that gets updated for every write and waited for every read, guaranteeing total order.  We currently don't use `MTLHazardTrackingModeUntracked`, so the only thing MTLFences *could* have done is make things slower.

(If we do use `MTLHazardTrackingModeUntracked` in the future, we can either start using `MTLFence`s, or enable it only when using `MTLEvent`s for synchronization.  The latter being much easier as adapting the MTLFence API to Vulkan's semaphore API is going to be incredibly difficult.)

This addresses the only other issue MTLFence semaphores had, their lack of cross-queue communication ability, by only exposing a single queue when MTLEvents aren't available.  Since the implementation no longer uses MTLFences and requires a single queue, I've renamed the associated runtime flags.  I've also enabled it by default, since it should no longer be able to violate the Vulkan spec.

Hopefully this way we can avoid performance the problems caused by emulated semaphores on Nvidia and Rosetta


MoltenVK/MoltenVK/Utility/MVKEnvironment.h
```diff
@@ -231,16 +231,16 @@ void mvkSetConfig(const MVKConfiguration& mvkConfig);
...
+ * Allow the use of a single queue or MTLEvent for VkSemaphore synchronization behaviour.
  * By default:
  *   - MVK_ALLOW_METAL_EVENTS is enabled
- *   - MVK_ALLOW_METAL_FENCES is disabled
+ *   - MVK_ALLOW_SINGLE_QUEUE_SEMAPHORE is enabled
  * */
 #ifndef MVK_ALLOW_METAL_EVENTS
 #   define MVK_ALLOW_METAL_EVENTS    1
 #endif
 #ifndef MVK_ALLOW_METAL_FENCES
```
italomandara, line 242: Shouldn't this be MVK_ALLOW_SINGLE_QUEUE_SEMAPHORE?

italomandara, line 242: Shouldn't his be #ifndef MVK_ALLOW_SINGLE_QUEUE_SEMAPHORE?

tellowkrinkle: It should

billhollings:
Sorry. I haven't had a chance to review this in detail. And I want to run it against CTS as part of that. I'll look at it as soon as I can.

italomandara:
Don't know if this is useful but it works just fine on many games i've been using this build on M1 for a while.

billhollings:
Thanks for the submission. 

LGTM, and I've run it against CTS and it seems to be fine (it does cause a large number of multi-queue tests to be recorded as Not Supported now...which is to be expected).

I do have concerns about breaking backwards compatibility in the renaming of both the `MVKConfiguration` struct member and environment variable. I also want to align this with expectations for resolving issue #1695.

With that in mind, I've pulled this into a separate branch, so I can make those changes, and I'll pull it into master once that's completed.

PR #1704, opened 2022-08-26
Support automatically enable Metal argument buffers when VK_EXT_descriptor_indexing extension is enabled.

### LINKED ISSUE #1689: Automatically enable Metal argument buffers when VK_EXT_descriptor_indexing is enabled on a device
Since `MVKConfiguration::useMetalArgumentBuffers` (`MVK_CONFIG_USE_METAL_ARGUMENT_BUFFERS`) is disabled by default, Metal argument buffers are not automatically used.

But since argument buffers are essential when the `VK_EXT_descriptor_indexing` is used, it would be useful to enable Metal argument buffers automatically in an `MVKDevice`, when the `VK_EXT_descriptor_indexing` extension is enabled by the app.

Create a flag in `MVKDevice`, and set it based on the combination of `MVKConfiguration::useMetalArgumentBuffers` and the enabling of the `VK_EXT_descriptor_indexing` extension, or enabling `VkPhysicalDeviceVulkan12Features::descriptorIndexing`.

Enable `VkPhysicalDeviceVulkan12Features::descriptorIndexing` for feature query.

billhollings:
- Change `MVKConfiguration::useMetalArgumentBuffers` to enum instead of boolean and default to enabling Metal argument buffers for VK_EXT_descriptor_indexing.
- **_However, leave Metal argument buffers disabled by default until they are improved._**
- Enable `VkPhysicalDeviceVulkan12Features::descriptorIndexing`.
- Add `MVKPhysicalDeviceVulkan12FeaturesNoExt` to track and enable Vulkan 1.2 features not part of any prior extensions absorbed by Vulkan 1.2.
- Update `VK_MVK_MOLTENVK_SPEC_VERSION` to version `36`.
- `MVKDeviceTrackingMixin` remove unnecessary inline declarations.
- `MoltenShaderConveter` tool support Metal Argument Buffers and MSL 3.0.

Fixes #1689.

### DIFF
--- a/MoltenVK/MoltenVK/GPUObjects/MVKDevice.h
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKDevice.h
@@ -90,6 +90,17 @@ typedef NSUInteger MTLTimestamp;
 #pragma mark -
 #pragma mark MVKPhysicalDevice
 
+/** VkPhysicalDeviceVulkan12Features entries that didn't originate in a prior extension. */
+typedef struct MVKPhysicalDeviceVulkan12FeaturesNoExt {
+	VkBool32 samplerMirrorClampToEdge;
+	VkBool32 drawIndirectCount;
+	VkBool32 descriptorIndexing;
+	VkBool32 samplerFilterMinmax;
+	VkBool32 shaderOutputViewportIndex;
+	VkBool32 shaderOutputLayer;
+	VkBool32 subgroupBroadcastDynamicId;
+} MVKPhysicalDeviceVulkan12FeaturesNoExt;
+
 /** Represents a Vulkan physical GPU device. */
 class MVKPhysicalDevice : public MVKDispatchableVulkanAPIObject {
 
@@ -333,8 +344,10 @@ public:
 	/** Returns whether the MSL version is supported on this device. */
 	bool mslVersionIsAtLeast(MTLLanguageVersion minVer) { return _metalFeatures.mslVersionEnum >= minVer; }
 
-	/** Returns whether this device is using Metal argument buffers. */
-	bool isUsingMetalArgumentBuffers() const  { return _metalFeatures.argumentBuffers && mvkConfig().useMetalArgumentBuffers; };
+	/** Returns whether this physical device supports Metal argument buffers. */
+	bool supportsMetalArgumentBuffers() const  {
+		return _metalFeatures.argumentBuffers && mvkConfig().useMetalArgumentBuffers != MVK_CONFIG_USE_METAL_ARGUMENT_BUFFERS_NEVER;
+	};
 
 	/**
 	 * Returns the start timestamps of a timestamp correlation.
@@ -406,6 +419,7 @@ protected:
 	MVKInstance* _mvkInstance;
 	const MVKExtensionList _supportedExtensions;
 	VkPhysicalDeviceFeatures _features;
+	MVKPhysicalDeviceVulkan12FeaturesNoExt _vulkan12FeaturesNoExt;
 	MVKPhysicalDeviceMetalFeatures _metalFeatures;
 	VkPhysicalDeviceProperties _properties;
 	VkPhysicalDeviceTexelBufferAlignmentPropertiesEXT _texelBuffAlignProperties;
@@ -694,6 +708,9 @@ public:
 	/** Returns the underlying Metal device. */
 	inline id<MTLDevice> getMTLDevice() { return _physicalDevice->getMTLDevice(); }
 
+	/** Returns whether this device is using Metal argument buffers. */
+	bool isUsingMetalArgumentBuffers() { return _isUsingMetalArgumentBuffers; };
+
 	/**
 	 * Returns an autoreleased options object to be used when compiling MSL shaders.
 	 * The useFastMath parameter is and-combined with MVKConfiguration::fastMathEnabled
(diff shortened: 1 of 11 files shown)


MoltenVK/MoltenVK/GPUObjects/MVKDevice.h
```diff
@@ -786,6 +803,9 @@ class MVKDevice : public MVKDispatchableVulkanAPIObject {
 	VkPhysicalDevice##structName##Features##extnSfx _enabled##structName##Features;
 #include "MVKDeviceFeatureStructs.def"
 
+	/** Vulkan 1.2 features that did originate in a prior extension available and enabled. */
```
cdavis5e, line 806:
```suggestion
	/** Vulkan 1.2 features that did not originate in a prior extension available and enabled. */
```

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
```diff
@@ -1986,9 +1986,16 @@
...
 
-    if (_metalFeatures.indirectDrawing && _metalFeatures.baseVertexInstanceDrawing) {
-        _features.drawIndirectFirstInstance = true;
-    }
+	_features.drawIndirectFirstInstance = _metalFeatures.indirectDrawing && _metalFeatures.baseVertexInstanceDrawing;
+
+	mvkClear(&_vulkan12FeaturesNoExt);	// Start with everything cleared
+	_vulkan12FeaturesNoExt.samplerMirrorClampToEdge = _metalFeatures.samplerMirrorClampToEdge;
+	_vulkan12FeaturesNoExt.drawIndirectCount = false;
+	_vulkan12FeaturesNoExt.descriptorIndexing = true;
```
cdavis5e, line 1994: It should now be possible to set this properly.

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
```diff
@@ -1986,9 +1986,16 @@
...
+	_vulkan12FeaturesNoExt.samplerFilterMinmax = false;
+	_vulkan12FeaturesNoExt.shaderOutputViewportIndex = true;
```
cdavis5e, line 1996: This should probably be gated on `multiViewport`.

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
```diff
@@ -1986,9 +1986,16 @@
...
+	_vulkan12FeaturesNoExt.shaderOutputLayer = true;
```
cdavis5e, line 1997: This should require layered rendering.

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
```diff
@@ -1986,9 +1986,16 @@
...
+	_vulkan12FeaturesNoExt.subgroupBroadcastDynamicId = true;
```
cdavis5e, line 1998: This should require SIMD-group support.

cdavis5e (changes requested):

MoltenVK/MoltenVK/GPUObjects/MVKDevice.h
billhollings: Good catch. Fixed.

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
billhollings: What do you mean by this suggestion? 

`VK_EXT_descriptor_indexing` is reported as always supported, so this aligns with that.

We don't restrict descriptor indexing support to Metal arg buffer support. Are you thinking both `VK_EXT_descriptor_indexing` and `descriptorIndexing` support should be tested against something else?

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
billhollings: Good catch. Fixed.

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
billhollings: Good catch. Fixed.

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
billhollings: Good catch. Fixed.

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
cdavis5e: I was referring to the comment formerly on line 128, stating that this should only be set if `_metalFeatures.arrayOfTextures && _metalFeatures.arrayOfSamplers && shaderStorageBufferArrayNonUniformIndexing` are all true.

cdavis5e (approved):

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
billhollings: Hmmm. Okay. I removed those tests as obsolete...but does it make sense that descriptor indexing (`VK_EXT_descriptor_indexing` and `descriptorIndexing`) only be permitted if texture, sampler, and storage buffer arrays are available after all? Would it be better to put those tests back in place?

PR #1715, opened 2022-09-13
MVKPipeline: Add builtins that are read but not written to tessellation pipelines.

cdavis5e:
It is always legal in Vulkan to read a builtin, particularly `BuiltInPosition`, even if it weren't written by the previous stage. The CTS tests that this scenario works in the driver.

Update SPIRV-Cross to pull in a change required for this.

Fixes 8 CTS tests under `dEQP-VK.pipeline.*.no_position`. (Eight other tests worked solely by accident without this change.)


cdavis5e:
cc @gpx1000

billhollings (approved):
LGTM. Elegant solution!

PR #1716, opened 2022-09-16
Updated readme to reflect 1.2 support as announced

lukebrowell:
Updated readme to reflect 1.2 support as announced

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,7 +69,7 @@ document in the `Docs` directory.
 Introduction to MoltenVK
 ------------------------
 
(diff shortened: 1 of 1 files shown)


lukebrowell:
Might be worth dropping the version number from the readme or including a Semver version range to negate need for updating version numbers in the readme.

billhollings:
> Updated readme to reflect 1.2 support as announced

MoltenVK does not quite yet support Vulkan 1.2 The docs, and API, will be updated once it does.

What announcement are you referring to?

Issue #1717, opened 2022-09-17
Build of shared MoltenVK 1.1.11 fails with undefined refence to `_vkGetBufferDeviceAddressEXT` if `MVK_HIDE_VULKAN_SYMBOLS=1`


SpaceIm:
If I try to build shared MoltenVK 1.1.11 and to hide vulkan symbols, link fails while creating MoltenVK.dylib:

```
Undefined symbols for architecture x86_64:
  "_vkGetBufferDeviceAddressEXT", referenced from:
      MVKInstance::initProcAddrs() in MVKInstance.mm.o
```

There was no issue in 1.1.10

Seems to come from https://github.com/KhronosGroup/MoltenVK/pull/1638 and the fact that `MVK_PUBLIC_VULKAN_ALIAS` as created in https://github.com/KhronosGroup/MoltenVK/pull/1490 is no-op if `MVK_HIDE_VULKAN_SYMBOLS=1`.

Here is a fix:

```patch
--- a/Common/MVKCommonEnvironment.h
+++ b/Common/MVKCommonEnvironment.h
@@ -108,6 +108,9 @@ extern "C" {
 /** Directive to make a public alias of another symbol. */
 #define MVK_PUBLIC_ALIAS(ALIAS, TARGET)   asm(".globl _" #ALIAS "\n\t_" #ALIAS " = _" #TARGET)
 
+/** Directive to make a private extern alias of another symbol. */
+#define MVK_PRIVATE_EXTERN_ALIAS(ALIAS, TARGET)   asm(".private_extern _" #ALIAS "\n\t_" #ALIAS " = _" #TARGET)
+
 /**
  * Directives to hide public symbols from the Vulkan API, to avoid library linking
  * conflicts when bound to a Vulkan Loader that also exports identical symbols.
@@ -117,7 +120,7 @@ extern "C" {
 #endif
 #if MVK_HIDE_VULKAN_SYMBOLS
 #	define MVK_PUBLIC_VULKAN_SYMBOL
-#	define MVK_PUBLIC_VULKAN_ALIAS(ALIAS, TARGET)
+#	define MVK_PUBLIC_VULKAN_ALIAS(ALIAS, TARGET)	MVK_PRIVATE_EXTERN_ALIAS(ALIAS, TARGET)
 #else
 #	define MVK_PUBLIC_VULKAN_SYMBOL		MVK_PUBLIC_SYMBOL
[...]

PR #1718, opened 2022-09-18
Fix `MVK_PUBLIC_VULKAN_ALIAS` when vulkan symbols are hidden

SpaceIm:
closes https://github.com/KhronosGroup/MoltenVK/issues/1717

With this fix, alias symbols are properly defined with external linkage but hidden in export table of shared lib (same behavior than `visibility("hidden")`).


billhollings:
Thanks for the fix submission, and for highlighting the issue in #1717.

Since the v1.1.11 release, there has been a transition of the `vkGetBufferDeviceAddressEXT` function as we move towards Vulkan 1.2. Those changes removed the problem you were encountering, but introduced a different issue, in that the function wasn't being mapped in the pointer functions.

PR #1719 fixes that issue (and your original issue), in a manner that is compatible with the promoted Vulkan 1.2 functions.

Closing this as being replaced by #1719.

PR #1723, opened 2022-09-23
Makefile: forward to xcode any variable declared on the command line

rasky:
Makefile currently special-case MVK_HIDE_VULKAN_SYMBOLS as the only variable that can be customised on the command line. There are many others that are useful though, such as MVK_LOG_*. This commit makes the Makefile more generic and allows to forward to xcode whatever variable was declared on the command line.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,12 +14,11 @@ else
 	OUTPUT_FMT_CMD = -quiet
 endif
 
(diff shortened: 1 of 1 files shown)


billhollings:
LGTM. Thanks. Nice enhancement.

PR #1724, opened 2022-09-23
Advertise support for extension VK_KHR_spirv_1_4.

billhollings:
- Advertise support for extension `VK_KHR_spirv_1_4`.
- `MoltenVKShaderConverter` automatically map bindings when converting GLSL.
- `MoltenVKShaderConverter` improvements to diagnostic logging.
- Update `Whats_New.md` document.

Support for _SPIR-V 1.4_ features added in [recent updates](https://github.com/KhronosGroup/SPIRV-Cross/issues/2004) to `SPRIV-Cross`.

### DIFF
--- a/MoltenVKShaderConverter/MoltenVKShaderConverter/SPIRVSupport.cpp
+++ b/MoltenVKShaderConverter/MoltenVKShaderConverter/SPIRVSupport.cpp
@@ -20,6 +20,7 @@
 #include "MVKStrings.h"
 #include <spirv.hpp>
 #include <ostream>
+#include <sstream>
 
 #import <CoreFoundation/CFByteOrder.h>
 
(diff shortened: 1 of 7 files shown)


MoltenVKShaderConverter/MoltenVKShaderConverter/SPIRVSupport.cpp
```diff
@@ -97,16 +98,25 @@ void mvk::logSPIRV(vector<uint32_t>& spirv, string& spvLog) {
 	uint32_t options = (SPV_BINARY_TO_TEXT_OPTION_INDENT);
 	spv_text text;
 	spv_diagnostic diagnostic = nullptr;
-	spv_context context = spvContextCreate(SPV_ENV_VULKAN_1_0);
+	spv_context context = spvContextCreate(SPV_ENV_VULKAN_1_2);
```
cdavis5e, line 101: Doesn't that allow SPIR-V 1.5? I guess it doesn't really matter much at this point, given that this function just logs the SPIR-V it was given.

cdavis5e (approved):

MoltenVKShaderConverter/MoltenVKShaderConverter/SPIRVSupport.cpp
spnda: According to https://github.com/KhronosGroup/SPIRV-Cross/issues/2004 SPIRV-Cross now also supports 1.5, so this should be perfectly fine.

billhollings: Yes. That particular line change was primarily about allowing `spirv-tools` to log v1.4 (and v1.5) SPIR-V (I guess I could have identified that in the commit note).

And also yes. Both SPIR-V 1.4 and 1.5 are effectively supported by SPIRV-Cross now. There is no specific Vulkan extension to be identified to indicate SPIR-V 1.5 support in MoltenVK, otherwise this PR would have included that.

Next stop Vulkan 1.2!

Issue #1567, opened 2022-04-09
Vulkan 1.2 requirements


billhollings:
[Appendix D of the Vulkan spec](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/html/vkspec.html#versions-1.2) summarizes the extension list below, and adds additional Vulkan 1.2 enhancement requirements.

The following extensions are required for Vulkan 1.2 but are **_not yet_** supported by MoltenVK:

- [VK_KHR_buffer_device_address](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/html/vkspec.html#VK_KHR_buffer_device_address)
- [VK_KHR_draw_indirect_count](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/html/vkspec.html#VK_KHR_draw_indirect_count)
- [VK_KHR_separate_depth_stencil_layouts](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/html/vkspec.html#VK_KHR_separate_depth_stencil_layouts)
- [VK_KHR_shader_atomic_int64](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/html/vkspec.html#VK_KHR_shader_atomic_int64)
- [VK_KHR_shader_float_controls](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/html/vkspec.html#VK_KHR_shader_float_controls)
- [VK_KHR_spirv_1_4](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/html/vkspec.html#VK_KHR_spirv_1_4)
- [VK_KHR_vulkan_memory_model](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/html/vkspec.html#VK_KHR_vulkan_memory_model)
- [VK_EXT_sampler_filter_minmax](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/html/vkspec.html#VK_EXT_sampler_filter_minmax)
- [VK_EXT_separate_stencil_usage](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/html/vkspec.html#VK_EXT_separate_stencil_usage)

The following extensions are required for Vulkan 1.2 and **_are_** supported by MoltenVK:

- [~VK_KHR_8bit_storage~](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/html/vkspec.html#VK_KHR_8bit_storage)
[...]

PR #1726, opened 2022-09-26
Add support for Vulkan 1.2.

billhollings:
- Set `MVK_VULKAN_API_VERSION` from `VK_API_VERSION_1_2`.
- Update MoltenVK version to `1.2.0`.
- Remove several previously overlooked extension suffixes in use of promoted Vulkan struct and enum values.
- Update documentation.

Implements #1567.

Well...this should have everything we need, but I have a nagging feeling there might be some kind of _unknown-unknown_ that's been missed. Please identify anything that might have been overlooked in the transition to 1.2.

### DIFF
--- a/MoltenVK/MoltenVK/GPUObjects/MVKInstance.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKInstance.mm
@@ -402,18 +402,18 @@ void MVKInstance::initDebugCallbacks(const VkInstanceCreateInfo* pCreateInfo) {
 #define ADD_INST_ENTRY_POINT(func)						ADD_ENTRY_POINT(func, VK_API_VERSION_1_0, nullptr, nullptr, false)
 #define ADD_DVC_ENTRY_POINT(func)						ADD_ENTRY_POINT(func, VK_API_VERSION_1_0, nullptr, nullptr, true)
 
+// Add a core function.
 #define ADD_INST_1_1_ENTRY_POINT(func)					ADD_ENTRY_POINT(func, VK_API_VERSION_1_1, nullptr, nullptr, false)
 #define ADD_INST_1_3_ENTRY_POINT(func)					ADD_ENTRY_POINT(func, VK_API_VERSION_1_3, nullptr, nullptr, false)
 #define ADD_DVC_1_1_ENTRY_POINT(func)					ADD_ENTRY_POINT(func, VK_API_VERSION_1_1, nullptr, nullptr, true)
 #define ADD_DVC_1_2_ENTRY_POINT(func)					ADD_ENTRY_POINT(func, VK_API_VERSION_1_2, nullptr, nullptr, true)
 #define ADD_DVC_1_3_ENTRY_POINT(func)					ADD_ENTRY_POINT(func, VK_API_VERSION_1_3, nullptr, nullptr, true)
 
-// Adds both the 1.1 function and the promoted extension function.
+// Add both the promoted core function and the extension function.
 #define ADD_INST_1_1_PROMOTED_ENTRY_POINT(func, EXT)	\
 	ADD_INST_1_1_ENTRY_POINT(func);	\
 	ADD_ENTRY_POINT_MAP(func##KHR, func, 0, VK_##EXT##_EXTENSION_NAME, nullptr, false)
 
-// Adds both the 1.1 function and the promoted extension function.
 #define ADD_DVC_1_1_PROMOTED_ENTRY_POINT(func, EXT)	\
 	ADD_DVC_1_1_ENTRY_POINT(func);	\
 	ADD_ENTRY_POINT_MAP(func##KHR, func, 0, VK_##EXT##_EXTENSION_NAME, nullptr, true)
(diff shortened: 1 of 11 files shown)


spnda (changes requested):
There's a few more values where the EXT or KHR suffix should be removed, which I have committed here (perhaps cherry-pick it):
https://github.com/spnda/MoltenVK/commit/d25623fd6107f48fc1e675aeb73f0fc13dcf0ea0

billhollings:
> There's a few more places where the EXT or KHR suffix should be removed, which I have committed here (perhaps cherry-pick it):
> [spnda@6148dc7](https://github.com/spnda/MoltenVK/commit/6148dc7e89decce657234c53974491ccd85badaf) (fixed commit)

Done. Thanks for the addition.

MoltenVK/MoltenVK/GPUObjects/MVKInstance.mm
```diff
@@ -402,18 +402,18 @@
 #define ADD_INST_ENTRY_POINT(func)						ADD_ENTRY_POINT(func, VK_API_VERSION_1_0, nullptr, nullptr, false)
 #define ADD_DVC_ENTRY_POINT(func)						ADD_ENTRY_POINT(func, VK_API_VERSION_1_0, nullptr, nullptr, true)
 
+// Add a core function.
 #define ADD_INST_1_1_ENTRY_POINT(func)					ADD_ENTRY_POINT(func, VK_API_VERSION_1_1, nullptr, nullptr, false)
```
cdavis5e, line 406: Should this be renamed?

MoltenVK/MoltenVK/GPUObjects/MVKInstance.mm
```diff
@@ -402,18 +402,18 @@
...
 #define ADD_INST_1_3_ENTRY_POINT(func)					ADD_ENTRY_POINT(func, VK_API_VERSION_1_3, nullptr, nullptr, false)
 #define ADD_DVC_1_1_ENTRY_POINT(func)					ADD_ENTRY_POINT(func, VK_API_VERSION_1_1, nullptr, nullptr, true)
 #define ADD_DVC_1_2_ENTRY_POINT(func)					ADD_ENTRY_POINT(func, VK_API_VERSION_1_2, nullptr, nullptr, true)
 #define ADD_DVC_1_3_ENTRY_POINT(func)					ADD_ENTRY_POINT(func, VK_API_VERSION_1_3, nullptr, nullptr, true)
 
-// Adds both the 1.1 function and the promoted extension function.
+// Add both the promoted core function and the extension function.
 #define ADD_INST_1_1_PROMOTED_ENTRY_POINT(func, EXT)	\
```
cdavis5e, line 413: This too.

MoltenVK/MoltenVK/GPUObjects/MVKInstance.mm
```diff
@@ -402,18 +402,18 @@
...
 	ADD_INST_1_1_ENTRY_POINT(func);	\
 	ADD_ENTRY_POINT_MAP(func##KHR, func, 0, VK_##EXT##_EXTENSION_NAME, nullptr, false)
 
-// Adds both the 1.1 function and the promoted extension function.
 #define ADD_DVC_1_1_PROMOTED_ENTRY_POINT(func, EXT)	\
```
cdavis5e, line 417: Ditto.

cdavis5e (approved):

MoltenVK/MoltenVK/GPUObjects/MVKInstance.mm
billhollings: What are you thinking these should be renamed to? And why just the 1.1 versions?

spnda (approved):

MoltenVK/MoltenVK/GPUObjects/MVKInstance.mm
cdavis5e: Well, I assume we are now using them for 1.2 entry points as well.

billhollings: Yes. The macros set the minimum API version needed to enable the function. So those should stay at 1.1 in terms of operational content.

Are you indicating that you think something will break here if the app tries to access one of those functions after setting the `VkApplicationInfo::apiVersion` to `VK_API_VERSION_1_2`?

Or are you just suggesting that we just rename the macro to something like `ADD_INST_1_1_AND_1_2_ENTRY_POINT` to make it clear that those functions are available for both API versions?

cdavis5e: Don't  `VK_KHR_buffer_device_address` and `VK_KHR_create_renderpass2` add new entry points?

billhollings: Yes. They're already in there as `ADD_DVC_1_2_PROMOTED_ENTRY_POINT` entries.

cdavis5e: Oh wait, I'm being silly. This is for new functions that were *not* part of an existing extension. Got it.

cdavis5e (approved):
Approving again in case you missed it the first time.

billhollings:
> Approving again in case you missed it the first time.

Thanks for confirming. I did see your first approval, but wanted to make sure all questions and concerns were resolved before pulling in, in case I missed something.

PR #1730, opened 2022-09-29
Makefile: fix environment variable definitions

rasky:
My previous PR was wrong because of a wrong test did before submission. The syntax `A=B` on the `xcodebuild` command line only defines a Xcode build setting, but that is not what we are looking for: we need preprocess macros. To do so, we need to use the `GCC_PREPROCESSOR_DEFINITIONS` build setting, and set the variables there. We also use `${inherited}` because that build setting is already used in the Xcode project and we want to append to it rather than overwrite.


cdavis5e:
It worked for `MVK_HIDE_VULKAN_SYMBOLS` because that actually *is* an Xcode build setting--MoltenVK's Xcode project automatically converts it into a preprocessor definition. I wonder, then, if more build settings should be added for the other `#define` switches...

rasky:
Yes I saw that. With this PR, it works for all macros *without* having a build setting, and the one for `MVK_HIDE_VULKAN_SYMBOLS` could be removed afterwards.

I tried with the logging settings and it works. In general, it can be tested with `make MVK_WHATEVER=1` and then adding a `#ifndef / #error` in some header file.

billhollings:
Thanks for the update fix. Makes sense.

> the one for `MVK_HIDE_VULKAN_SYMBOLS` could be removed afterwards

Might as well remove it in this PR, otherwise it will get forgotten. Please make that change before I pull it in. Thanks.

billhollings (changes requested):
Thanks for the update fix. Makes sense.

> the one for `MVK_HIDE_VULKAN_SYMBOLS` could be removed afterwards

Might as well remove it in this PR, otherwise it will get forgotten. Please make that change before I pull it in. Thanks.

rasky:
@billhollings done. It seems to work but I would appreciate a double check

billhollings (approved):
Thanks for updating. LGTM now.

billhollings:
BTW...I've also added PR #1737, to fix a couple of breakages in the use of `MVK_HIDE_VULKAN_SYMBOLS=1` that were unrelated to this PR, but were exposed while I was testing it.

PR #1735, opened 2022-10-05
Fix prefer MTLEvent semaphore fallback

marzent:
The current behaviour does not align with the description here
https://github.com/KhronosGroup/MoltenVK/blob/c2e9f1d876584f664aef19744332737e78bf8f93/MoltenVK/MoltenVK/API/vk_mvk_moltenvk.h#L595

This makes it so the single queue semaphore style is used as fallback in the `MVK_CONFIG_VK_SEMAPHORE_SUPPORT_STYLE_METAL_EVENTS_WHERE_SAFE` case for Nvidia and Apple Silicon instead of using CPU callbacks.


billhollings:
Thanks for catching this, and for the submission. It triggered me to review the semaphore selection flow and fallback logic, and usage documentation. Rather than ask you to make a number of changes to this PR, it was easier to replace it with PR #1738.

PR #1738 makes `MVK_CONFIG_VK_SEMAPHORE_SUPPORT_STYLE_SINGLE_QUEUE` the default in all cases where `MTLEvents` are unusable, provides legacy consistency with `semaphoreUseMTLEvent` and `semaphoreUseMTLFence`, and clarifies documentation.

Issue #1736, opened 2022-10-09
Vulkan SDK release feature cut-off Friday Oct 14


billhollings:
LunarG is planning the next Vulkan SDK release.

A MoltenVK SDK release for this Vulkan SDK will be made Monday Oct 17.

MoltenVK PRs approved by Friday Oct 14 will be included in this release.

@cdavis5e @spnda @HansKristian-Work @richard-lunarg @KarenGhavam-lunarG

Gcenx:
Could we have #1735 merged since that resolves a regression before the cutoff.

billhollings:
> Could we have #1735 merged since that resolves a regression before the cutoff.

Yup. Definitely planning to bring that change in before the cutoff.

billhollings:
> Yup. Definitely planning to bring that change in before the cutoff.

PR #1735 replaced by PR #1738, which has been merged.

PR #1738, opened 2022-10-10
Vulkan semaphore fallback to using single queue if `MTLEvents` unusable.

billhollings:
- If `MTLEvents` are unusable, fallback to single queue.
- Adjust values of `MVKVkSemaphoreSupportStyle` enumeration.
- For legacy compatibility, if legacy `semaphoreUseMTLEvent` and `semaphoreUseMTLFence` are both disabled, use CPU callback emulation.
- Update and expand related documentation in `vk_mvk_moltenvk.h`.

This PR is an expansion of, and replaces, PR #1735, based on a suggestion from @marzent in that PR.

This PR makes `MVK_CONFIG_VK_SEMAPHORE_SUPPORT_STYLE_SINGLE_QUEUE` the default in all cases where `MTLEvents` are unusable, provides legacy consistency with `semaphoreUseMTLEvent` and `semaphoreUseMTLFence`, and clarifies documentation.

@marzent @Gcenx Please review relative to #1735 & regression fix request in issue #1736.

### DIFF
--- a/MoltenVK/MoltenVK/API/vk_mvk_moltenvk.h
+++ b/MoltenVK/MoltenVK/API/vk_mvk_moltenvk.h
@@ -106,10 +106,10 @@ typedef enum MVKUseMetalArgumentBuffers {
 
 /** Identifies the Metal functionality used to support Vulkan semaphore functionality (VkSemaphore). */
 typedef enum MVKVkSemaphoreSupportStyle {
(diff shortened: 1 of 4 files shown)


Gcenx:
Seems good to me, testing on Witcher3 (my usual testing title)

cdavis5e (approved):

marzent (approved):

Issue #1729, opened 2022-09-29
`MTLRenderPassDescriptor`s leaking for `vkCmdBeginRenderPass` when recording from a worker thread with `prefillMetalCommandBuffers`


James2022-rgb:
`MTLRenderPassDescriptor` objects allocated for `vkCmdBeginRenderPass` seem to leak if it's recorded from a worker thread with `prefillMetalCommandBuffers` enabled, presumably because its allocation and usage is not wrapped inside an autoreleasepool block,
resulting in gradually growing memory usage every frame.

I'm not knowledgeable in iOS/Cocoa ecosystem but I'm assuming the reason I don't see these leaks when not using `prefillMetalCommandBuffers` is because there's a default autoreleasepool block for the main thread,
which seems to suggest that these leaks would occur anyway if we did the `vkQueueSubmit` from a worker thread.

`MTL(Render|Blit)PassSampleBufferAttachmentDescriptor` objects also seem to leak if we're using timestamp queries, but I haven't investigated any deeper yet.

billhollings:
This should be fixed in PR #1744. Please retest with that PR and close this issue if it fixes the issue you encountered.

@James2022-rgb
I've tried testing the leaks as best I can, but I've had trouble finding an app that has sufficient complexity **_and_** is configured with sufficient permissions to be  opened in the Xcode Instruments Leaks tool or command line leaks utility. I know this is last minute...but any testing you can do by end of day Monday Oct 17 would help get this into the SDK release due then.

Issue #1732, opened 2022-10-04
Memory leak when not using @autoreleasepool


MennoVink:
I have recently updated to 1.3.244. I am now experiencing a huge memory leak, about 100GB in 1 minute. I am using `prefillMetalCommandBuffers` and `specializedQueueFamilies`.
I noticed that if i place an @autoreleasepool around my frame then the leak is gone. Trying to figure out which call caused the issue i ended up at the vkCmdBeginRenderPass function. If i wrap just that call with an @autoreleasepool then i get an error that the Metal debug command encoder was released before any commands were recorded to it. Shouldn't this encoder be retained by the command buffer until i free that buffer?

I started digging into MVK a bit but i've got a few questions.

In `id<MTLBlitCommandEncoder> MVKCommandEncoder::getMTLBlitEncoder(MVKCommandUse cmdUse)`, why is the blit encoder not retained? If i place an @autoreleasepool around that then that pool errors out with the same issue. This already errors out way before i start rendering though, from the first `vkCmdCopyBufferToImage` call i'm making.

Similar question for `void MVKCommandEncoder::beginMetalRenderPass(MVKCommandUse cmdUse)` The render encoder isn't being retained either?

Finally in `void MVKCommandEncoder::encodeTimestampStageCounterSamples()` i'm seeing an autorelease but without it being in a pool?

MennoVink:
![Screenshot 2022-10-04 at 18 40 09](https://user-images.githubusercontent.com/26071817/193877557-270f4534-2966-4d00-a24b-f94048512beb.png)
This is what instruments is showing for metal resource events. Does this mean that the render pass commands are retaining the images? That could cause a lot of memory to leak since i am destroying images but creating new ones each frame.

billhollings:
This should be fixed in PR #1744. Please retest with that PR and close this issue if it fixes the issue you encountered.

@MennoVink
I've tried testing the leaks as best I can, but I've had trouble finding an app that has sufficient complexity **_and_** is configured with sufficient permissions to be  opened in the Xcode Instruments Leaks tool or command line leaks utility. I know this is last minute...but any testing you can do by end of day Monday Oct 17 would help get this into the SDK release due then.

PR #1744, opened 2022-10-16
Fix memory leaks when configured for prefilling Metal command buffers.

billhollings:
- Add `MVKPrefillMetalCommandBuffersStyle` enumeration to enable prefilling, and configure memory recovery options.
- Redefine `MVKConfiguration::prefillMetalCommandBuffers` as a value from `MVKPrefillMetalCommandBuffersStyle` instead of a simple boolean. Set the values of `MVKPrefillMetalCommandBuffersStyle` and the default value of `MVKConfiguration::prefillMetalCommandBuffers` to be compatible with legacy use of this setting as a simple enable/disable boolean value.
- `MVKCommandEncoder` track prefill style, and support option to maintain a Metal autorelease pool that is created on `vkBeginCommandBuffer()` and drained on `vkEndCommandBuffer()`, when prefiling a command buffer from a single thread, and an option to wrap each command addition in a separate `@autoreleasepool{}` scope to create and drain an autorelease pool at every command addition to the Vulkan command buffer, when prefilling a single command buffer from multiple threads.
- `MVKCommandEncoder` retain Metal encoders when prefilling, because prefilling may span multiple autorelease pools.
- ` MVKCommandEncoder` clean up memory leak of `MTLBlitPassDescriptor` during prefilling.
- `MVKCommandEncoder` rearrange declarations of `MVKCommandUse` member variables to tighten up class memory use (unrelated).
- `MVKCommandBuffer` remove unnecessary inline declarations (unrelated).
- Change MoltenVK logging to use stdout instead of stderr (unrelated).

This should fix issues #1729 and #1732.

@MennoVink @James2022-rgb @BeastLe9enD 
I've tried testing the leaks as best I can, but I've had trouble finding an app that has sufficient complexity **_and_** is configured with sufficient permissions to be  opened in the Xcode Instruments Leaks tool. Last minute...but any testing you can do by end of day Monday Oct 17 would help get this into the SDK release due then.


MoltenVK/MoltenVK/API/vk_mvk_moltenvk.h
```diff
@@ -206,16 +215,26 @@ typedef struct {
...
+	 * to the Vulkan queue. This allows a single Metal command buffer to be used for all of the
+	 * Vulkan command buffers in a queue submission. The Metal command buffer is filled on the
+	 * thread that processes the command queue submission.
+	 *
+	 * If set to any value other than MVK_CONFIG_PREFILL_METAL_COMMAND_BUFFERS_STYLE_NO_PREFILL,
+	 * where possible, a Metal command buffer will be created and filled when each Vulkan
+	 * command buffer is filled. For applications that parallelize the filling of Vulkan
 	 * commmand buffers across multiple threads, this allows the Metal command buffers to also
 	 * be filled on the same parallel thread. Because each command buffer is filled separately,
 	 * this requires that each Vulkan command buffer requires a dedicated Metal command buffer.
```
cdavis5e, line 229:
```suggestion
	 * this requires that each Vulkan command buffer have a dedicated Metal command buffer.
```

MoltenVK/MoltenVK/Commands/MVKCommandBuffer.mm
```diff
@@ -466,6 +466,16 @@
...
+// Retain encoders when prefilling, because prefilling may span multiple autorelease pools.
+#define retainIfPrefilling(mtlEncIVar)            \
+	if (_prefillStyle) { [mtlEncIVar retain]; }
+
+#define endMetalEncoding(mtlEncIVar)              \
+	[mtlEncIVar endEncoding];                     \
+	if (_prefillStyle) { [mtlEncIVar release]; }  \
+	mtlEncIVar = nil
+
+
```
cdavis5e, line 478: I don't like these macros, but I don't have any better suggestions. Member functions, maybe?

MoltenVK/MoltenVK/Utility/MVKBaseObject.mm
```diff
@@ -100,7 +100,7 @@
 	va_end(origArgs);
 
 	// Log the message to the standard error stream
-	if (shouldLog) { fprintf(stderr, "[%s] %s\n", getReportingLevelString(logLevel), pMessage); }
+	if (shouldLog) { fprintf(stdout, "[%s] %s\n", getReportingLevelString(logLevel), pMessage); }
```
cdavis5e, line 103: This seems wrong, and the old way (logging to `stderr`) right. Typically in Unix programming, actual program output goes to `stdout` while diagnostics go to `stderr`. Also, Wine logs to `stderr`; changing ours will make capturing MoltenVK logs alongside Wine logs harder.

cdavis5e (changes requested):

James2022-rgb:
Thanks!
`MVK_CONFIG_PREFILL_METAL_COMMAND_BUFFERS_STYLE_MULTI_THREAD_AUTORELEASE` is the only prefill option I'm able to test right now,
(unfortunately  not all of our `VkCommandBuffer` uses suit `SINGLE_THREAD_AUTORELEASE`)
but I just confirmed that the leaks I was seeing of `MTLRenderPassColorAttachmentDescriptorInternal` and the like is now gone,
and the memory usage is stable.

MennoVink:
I did some runs with my triangle example. I've left it running for a minute and see what kind of memory footprint it has (in MBs). For the frame-wide @autorelease i have an @autorelease inside my display link callback. In this example i'm running that encompasses all command buffers being recorded. (ie not including async uploads)

|   | no @autorelease |  frame-wide @autorelease  | 
|---|---|---|
| MVK_CONFIG_PREFILL_METAL_COMMAND_BUFFERS_STYLE_NO_PREFILL | 80.5  | 67.1  | 
| MVK_CONFIG_PREFILL_METAL_COMMAND_BUFFERS_STYLE_SINGLE_THREAD_AUTORELEASE | 341.4  | 73.1  |
| MVK_CONFIG_PREFILL_METAL_COMMAND_BUFFERS_STYLE_MULTI_THREAD_AUTORELEASE |  342.7 | 66.8  |
|  MVK_CONFIG_PREFILL_METAL_COMMAND_BUFFERS_STYLE_NO_AUTORELEASE | 348.0  | 73.4  |

So, unfortunately, for my scenario none of the options work. I'd be using the STYLE_SINGLE_THREAD_AUTORELEASE as i'm using the prefilling to spread out the transcribing costs over multiple threads.

In Instruments i'm seeing a lot of MTLRenderPassColorAttachmentDescriptor's being created and being live. This is with single threaded without frame-wide @autorelease:
Screenshot 2022-10-17 at 13 44 36
There are also quite a few of these:
Screenshot 2022-10-17 at 13 47 03

You mentioned you cant test easily. I'm linking mvk statically so i gues sending you a binary of the example wouldn't help much?

billhollings:
@MennoVink 

> i'm seeing a lot of MTLRenderPassColorAttachmentDescriptor's being created

That's very helpful. Thanks.

> You mentioned you cant test easily. I'm linking mvk statically so i gues sending you a binary of the example wouldn't help much?

Unfortunately, yeah...Instruments is sandboxed to heck these days, so binaries from elsewhere seem to be effectively blocked (although I think it might partially be a bug with recent Xcode versions). I wasn't able to use Instruments on any of the Vulkan Samples. Do you have a buildable version of your triangle example, so I can run it through Instruments? If not, please send the binary anyway, and I'll give it a shot with Instruments here. If you want to email either rather than post them publicly, please send it to support@brenwill.com.

MoltenVK/MoltenVK/API/vk_mvk_moltenvk.h
billhollings: Good catch. Fixed.

MoltenVK/MoltenVK/Commands/MVKCommandBuffer.mm
billhollings: Replaced with template member functions.

MoltenVK/MoltenVK/Utility/MVKBaseObject.mm
billhollings: Yeah. Makes sense. Thanks for the convention background.

I had changed it because for some reason `stderr` wasn't being logged from background threads, but `stdout` was. But not sure why, or whether that was an appropriate solution, so I've reverted it to match convention.

billhollings:
Okay. I think I've finally got this right. 🤞🏻🤞🏻

Thanks to @MennoVink for providing me with a small app that made testing and profiling this much easier! I appreciate your assistance!

I've restored the old deferred `MTLCommandBuffer` encoding on `vkEndCommandBuffer()`, and redefined the prefiling options around immediate vs deferred options, and with or without autorelease pools. All options seem to hold steady on memory now.

Also included all of @cdavis5e's review suggestions.

@MennoVink @James2022-rgb Again...last minute request...but if you can verify with testing right away, I'd appreciate it.

MennoVink:
Works like a charm, thanks for investigating and fixing 👍 

Just some nitpicking:
Both IMMEDIATE_ENCODING and IMMEDIATE_ENCODING_NO_AUTORELEASE mention they have the smallest memory footprint. I think for the NO_AUTORELEASE variant this is least true because it depends on where the application has the autoreleasepool, which is likely around the entire frame or entire command buffer recording. If it's around the recording then i'm thinking it'd have the same footprint as DEFERRED_ENCODING. If it's around the frame then it has a  largerfootprint. If it's around each command though then it's not the fastest anymore, similar to IMMEDIATE_ENCODING.

In the readme's contribution guidelines there is a mention that contributions should not rely on autoreleasepools. Obviously this is for contributors and not users, but the way it's worded now made me believe as a user i dont need autoreleasepools anywhere. Maybe this mention should be expanded to only allow relying on the user's autoreleasepools when the user has provided an option (which is off by default)?

Considering maintenance reasons, is it necessary to offer and thus keep maintaining these four options instead of just on/off? What is the intended use case for immediate encoding options? Maybe it's just for profiling granularity? And if you do want to keep options, what about the other options? Currently you have these:
|   | transcribe on command | transcribe on EndCommandBuffer | transcribe on Submit
|---|---|---|---|
| @autoreleasepool in MVK | IMMEDIATE_ENCODING | DEFERRED_ENCODING | NO_PREFILL
| @autoreleasepool in app | IMMEDIATE_ENCODING_NO_AUTORELEASE | _1_ | _2_
1. Probably never needed as deferred encoding with @autoreleasepool in mvk is only done for performance reasons i think?
2. Some apps might want this if they have autorelease around the entire frame?

Personally i think just having NO_PREFILL and DEFERRED_ENCODING options would be sufficient.

billhollings:
> Works like a charm, thanks for investigating and fixing 👍

Wonderful! Thanks for testing so quickly!

> Just some nitpicking: Both IMMEDIATE_ENCODING and IMMEDIATE_ENCODING_NO_AUTORELEASE mention they have the smallest memory footprint. 

Thanks. I've fixed this in the comment documentation now.

> I think for the NO_AUTORELEASE variant this is least true because it depends on where the application has the autoreleasepool, which is likely around the entire frame or entire command buffer recording. If it's around the recording then i'm thinking it'd have the same footprint as DEFERRED_ENCODING.

The immediate modes do not hold MoltenVK command objects in the MoltenVK command buffer, so there should be significant savings for either immediate option.

> In the readme's contribution guidelines there is a mention that contributions should not rely on autoreleasepools. Obviously this is for contributors and not users, but the way it's worded now made me believe as a user i dont need autoreleasepools anywhere. Maybe this mention should be expanded to only allow relying on the user's autoreleasepools when the user has provided an option (which is off by default)?

Thanks. I've fleshed out the `README.md` document to clarify that this is meant for MoltenVK contributors.

> Personally i think just having NO_PREFILL and DEFERRED_ENCODING options would be sufficient.

That's what we had originally, but PR #1561 was added to try to lower the memory footprint of caching MoltenVK command objects in the MoltenVK command buffer. Ultimately we're trying to give the app opportunities to balance their time vs memory options.

PR #1755, opened 2022-11-01
MVKPipeline: Stop using vertex-style input for tessellation evaluatio…

cdavis5e:
…n shaders.

This has caused us nothing but trouble. The code to build up the vertex descriptor is fragile; we can rip that out now.

Also, make sure to positively identify per-patch blocks as per-patch. For those, the individual members have the `Patch` decoration.

Update SPIRV-Cross to pull in the changes needed for this.

Fixes 66 tests in the CTS.


billhollings (commented):
Nice simplification.

I am able to run `./runcts -p <cts-path>/vk-default.txt`, and get the expected reduction of about 60 CTS failure.

However, when I try running without the `-p` option to evaluate with 1.2 + extns, I get a very odd result. The test run bogs down in `dEQP-VK.synchronization.op.multi_queue.*` to the point of taking 5-6 seconds per test (effectively infinite time for the test run). This does not happen with `-p` probably because timeline semaphores are `Not Supported`, it doesn't happen with 1.2 + extns in the current code base without this patch. And it doesn't happen if I run the individual tests one-by-one with this patch.

Very strange. I can't see any reason why this patch should be causing this. And maybe it's not. Investigating as best I can.

cdavis5e:
> Very strange. I can't see any reason why this patch should be causing this. And maybe it's not. I hate these kind of Heisenbugs. Investigating as best I can.

Yeah, this has the feel of a Heisenbug. I suspect it's some kind of race condition--not doing this work changed all the timings and exposed it.

cdavis5e:
Were you able to figure this out? Do you need an extra set of eyes or two?

billhollings:
> Were you able to figure this out? Do you need an extra set of eyes or two?

Still working on it, but making progress. 

The PR code invokes orders of magnitude more SPIRVCross conversions and Metal shader compiles that existing code. For example, on the first 263 tests in the CTS `dEQP-VK.synchronization.*` category:

```
                   Existing        PR
SPIRV-Cross           63          1,347
Metal Compiles      92,102      1,188,034
```

The additional SPIRVCross invocations seems to be arising because `MVKShaderLibraryCache::findShaderLibrary()` can't find an existing converted shader, which I think is somehow tied to the new `shader_patch_input_buffer_index` alternating between `29` and `20` between the tess stage and fragment stage. Still investigating this.

The egregious number of Metal compiles (in either case) is substantially due to CTS's use of pipeline caches. In MoltenVK, current pipeline caching only caches the MSL code. So when loading from cache, it doesn't need to do a SPRIV-Cross conversion, but it does need to compile the Metal shaders. We should definitely get proper Metal Binary Archives implemented, to avoid pipeline caches from having to recompile Metal shaders during loading.

MoltenVK/MoltenVK/GPUObjects/MVKPipeline.mm
```diff
@@ -2348,7 +2244,8 @@ void serialize(Archive & archive, CompilerMSL::Options& opt) {
 				opt.ios_use_simdgroup_functions,
 				opt.emulate_subgroups,
 				opt.vertex_index_type,
-				opt.force_sample_rate_shading);
+				opt.force_sample_rate_shading,
+				opt.raw_buffer_tese_input);
 	}
```
billhollings, line 2249: Looks like this Archive struct is missing the new `shader_patch_input_buffer_index` member. This is causing the pipeline cache values being read in to never match what was stored to them whenever this member is set to anything other than the default value. This affects both the tess stage and the following frag stage, since they use the same `SPIRVToMSLConversionConfiguration` struct.

Can you add `shader_patch_input_buffer_index` in the appropriate place in Archive definition, between these two existing entries, please?

    opt.shader_index_buffer_index,
    opt.shader_patch_input_buffer_index,     //  <--new
    opt.shader_input_wg_index,

billhollings (changes requested):
Got it. See code note for fix.

MoltenVK/MoltenVK/GPUObjects/MVKPipeline.mm
cdavis5e: /me smacks head

D'oh! Fixed.

cdavis5e:
Long term, though, we need to 1) reduce time spent compiling shaders, and 2) figure out the underlying race condition this exposed.

MoltenVK/MoltenVK/GPUObjects/MVKPipeline.mm
```diff
@@ -2311,6 +2207,7 @@ void serialize(Archive & archive, CompilerMSL::Options& opt) {
 				opt.view_mask_buffer_index,
 				opt.dynamic_offsets_buffer_index,
 				opt.shader_input_buffer_index,
+				opt.shader_patch_input_buffer_index,
```
billhollings, line 2210: Not trying to be too pedantic here, but can you move this down one line to just before `opt.shader_input_wg_index`, so that it's location aligns with its location in the definition of the struct in `spirv_msl.hpp`, please? It makes it much easier to visually check for required struct members in the Archive definition if they appear in the same order as the struct member definition, wherever possible.

billhollings (changes requested):
Thanks for adding the Archive member. One minor nit about the location of it. _(Ignore this. It is a dup of the review below where I actually added the suggested change. I don't see any way to now delete this earlier review)_.

MoltenVK/MoltenVK/GPUObjects/MVKPipeline.mm
```diff
@@ -2311,6 +2207,7 @@ void serialize(Archive & archive, CompilerMSL::Options& opt) {
...
 				opt.shader_index_buffer_index,
 				opt.shader_input_wg_index,
```
billhollings, line 2212: Not trying to be too pedantic here, but can you move this down one line to just before `opt.shader_input_wg_index`, so that it's location aligns with its location in the definition of the struct in `spirv_msl.hpp`, please? It makes it much easier to visually check for required struct members in the Archive definition if they appear in the same order as the struct member definition, wherever possible.

```suggestion
				opt.shader_index_buffer_index,
				opt.shader_patch_input_buffer_index,
				opt.shader_input_wg_index,
```

billhollings (changes requested):
Thanks for adding the Archive member. One minor nit about the location of it.

billhollings:
> Long term, though, we need to 1) reduce time spent compiling shaders, and 2) figure out the underlying race condition/Heisenbug this exposed.

Agreed. As I mention above I think Metal Binary Archives might help with (1). I've added issue #1765 to track adding them.

billhollings (approved):
Thanks. LGTM now.

PR #1769, opened 2022-11-11
Only wait on emulated semaphores once...

### LINKED ISSUE #1764: Emulated semaphores freeze application
Using MVK 1.3.231 with this configuration causes my application to freeze due to MVK reverting to emulated semaphores:
```
	config.prefillMetalCommandBuffers = MVK_CONFIG_PREFILL_METAL_COMMAND_BUFFERS_STYLE_DEFERRED_ENCODING;
	config.semaphoreSupportStyle = MVK_CONFIG_VK_SEMAPHORE_SUPPORT_STYLE_CALLBACK;
```

What's happening is that during `MVKQueueCommandBufferSubmission::execute` it's calling `commitActiveMTLCommandBuffer` more than once. This is a problem because the first time it's called it will wait for semaphores without a command buffer. This resets the `MVKSemaphoreEmulated`'s internal `_blocker`. Then the second time it wants to commit (this time the prefilled command buffer) it wants to wait on the semaphore again, but the `_blocker` has already been reset so this freezes.

The first commit happens from `MVKCommandBuffer::submit` under the `_prefilledMTLCmdBuffer` branch. The second commit happens after unwinding the stack back to `MVKQueueCommandBufferSubmission::execute` as last call there.

I can circumvent this issue by tracking whether or not the nil wait has already been done and then bypassing it, but that's probably not the best solution so any ideas how to solve this?

MennoVink:
… to prevent freezing when using prefilled command buffers.

fix #1764

### DIFF
--- a/MoltenVK/MoltenVK/GPUObjects/MVKQueue.h
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKQueue.h
@@ -220,6 +220,7 @@ protected:
 	MVKFence* _fence;
 	id<MTLCommandBuffer> _activeMTLCommandBuffer;
 	MVKCommandUse _commandUse;
+	bool _emulatedWaitDone; //Used to track if we've already waited for emulated semaphores.
 };
 
 
--- a/MoltenVK/MoltenVK/GPUObjects/MVKQueue.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKQueue.mm
@@ -359,7 +359,17 @@ static const char* mvkStringFromErrorState(MTLCommandEncoderErrorState errState)
 void MVKQueueCommandBufferSubmission::commitActiveMTLCommandBuffer(bool signalCompletion) {
 
 	// If using inline semaphore waiting, do so now.
-	for (auto& ws : _waitSemaphores) { ws.first->encodeWait(nil, ws.second); }
+	// When prefilled command buffers are used, multiple commits will happen because native semaphore
+	// waits need to be committed before the prefilled command buffer is committed. Since semaphores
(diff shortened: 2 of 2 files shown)


billhollings:
@MennoVink Thanks for submitting this. 

You originally marked this with a _Draft_ prefix. I've changed that to _WIP_, as our convention for marking it as still under development.

I'm assuming you marked it that way because you are still testing the changes, or expect to make further changes before it should be reviewed and merged?

When you are happy with this PR, remove the _WIP_ title prefix, and request a review from myself (and maybe @cdavis5e too), or just post a comment that you think it's ready. Thanks.

MennoVink:
Thanks for the rename, i assumed the gitlab convention, but looking at the other MRs it should've been WIP indeed.

I've marked it as wip for testing purposes. If i forcefully disable MTL events then this change fixes it on my intel MBP, but i dont have a nvidia mac so i've sent that part off to our testers. Most likely i'll get a response early this week.

MennoVink:
@billhollings @cdavis5e As per github rules i'm not allowed to fill in the reviewers. Only owners or collaborators with write access can.

Our testers reported it works on nvidia as well so i'm good for this to be merged. I think the comment in the code explains enough but let me know if you have any questions.

One thing i didn't mention in the comment is that i expect this to also happen with multiple submits in a single batch. I didn't test this though, but i find it hard to believe that none of the users or tests do this so not sure what's going on there.

cdavis5e (approved):
I'd wait for @billhollings to approve as well, but FWIW, LGTM.

billhollings (approved):
LGTM. Thanks for the submission!

Issue #1760, opened 2022-11-06
Vulkan CTS test crash because of invalid dimensions for PVRTC texture


spnda:
This happens with MoltenVK from master and VK-GL-CTS from master on macOS 13.1 with a M1. Looks like MoltenVK reports a lot of image formats as supported in the `vkGetPhysicalDeviceImageFormatProperties` call even though they're not supported. Though the reason for the crash, namely Metal asserting because PVRTC_RGBA_4BPP did not have a compatible size.

From the [CTS source](https://github.com/KhronosGroup/VK-GL-CTS/blob/main/external/vulkancts/modules/vulkan/api/vktApiMemoryRequirementInvarianceTests.cpp#L512) it seems like the test generates a random image width and height, ranging somewhere between 2 and 18. It does not look like the Vulkan spec makes a requirement for the size of a PVRTC image dimensions needing to be a power of 2 or be a perfect square. The Metal docs also don't note any requirement either, though it seems like Apple GPUs have these implicit requirements in their PVRTC implementation.

```
Test case 'dEQP-VK.api.invariance.random'..
[mvk-error] VK_ERROR_FORMAT_NOT_SUPPORTED: VkFormat VK_FORMAT_G10X6B10X6G10X6R10X6_422_UNORM_4PACK16 is not supported on this device.
[mvk-error] VK_ERROR_FORMAT_NOT_SUPPORTED: VkFormat VK_FORMAT_G10X6B10X6G10X6R10X6_422_UNORM_4PACK16 is not supported on this device.
[mvk-error] VK_ERROR_FORMAT_NOT_SUPPORTED: VkFormat VK_FORMAT_G10X6B10X6G10X6R10X6_422_UNORM_4PACK16 is not supported on this device.
[mvk-error] VK_ERROR_FORMAT_NOT_SUPPORTED: VkFormat VK_FORMAT_G10X6B10X6G10X6R10X6_422_UNORM_4PACK16 is not supported on this device.
[mvk-error] VK_ERROR_FORMAT_NOT_SUPPORTED: VkFormat VK_FORMAT_G10X6B10X6G10X6R10X6_422_UNORM_4PACK16 is not supported on this device.
[mvk-error] VK_ERROR_FORMAT_NOT_SUPPORTED: VkFormat VK_FORMAT_G10X6B10X6G10X6R10X6_422_UNORM_4PACK16 is not supported on this device.
[...]

PR #1780, opened 2022-11-19
Fix mistaken YCBCR format support indication.

billhollings:
- `MVKVkFormatDesc::chromaSubsamplingPlaneCount` must be `> 1` to be considered supported.

- Fixes mistaken `VK_ERROR_FORMAT_NOT_SUPPORTED ` errors for some `YCBCR` formats found in issue #1760.

- Has the same `YCBCR` CTS test results as before this patch.

@Lichtso Please also review.


Lichtso (commented):
LGTM

Issue #1746, opened 2022-10-19
MoltenVK 1.2.0 performance regression on AMD gpu


Qotonana:
Hi all

First, congrats on this project to all involved.
Doing some testing with CrossOver 22 and new MoltenVK (unsupported off course :)) and noticed that the new MoltenVK-1.2.0 has terrible performance regression on Intel Mac with AMD Radeon card.

Uses libMoltenVK dylibs from @Gcenx releases (just repackaged) from
https://github.com/The-Wineskin-Project/MoltenVK/releases/
Tested on Monterey 12.6. Cleared the DXVK state cache before every test. 

Also replaced DXVK with macOS-async-1.10.3 from
https://github.com/Gcenx/DXVK-macOS/releases

**Sniper Contracts MoltenVK macos-1.1.11**
https://github.com/The-Wineskin-Project/MoltenVK/releases/download/v1.1.11/macos-1.1.11.tar.xz
![Sniper Contracts macos-1 1 11](https://user-images.githubusercontent.com/102525166/196627138-f892d7e3-c3d9-4ea4-8634-efdf58b07aee.png)

**Sniper Contracts MoltenVK macOS-1.2.0**
https://github.com/The-Wineskin-Project/MoltenVK/releases/download/v1.2.0/macos-1.2.0.tar.xz
![Sniper Contracts macos-1 2 0](https://user-images.githubusercontent.com/102525166/196627227-5e253f3f-8a02-4dec-90e1-8da9b9826c06.png)

**Automobilista 2 DemoAVX MoltenVK macos-1.1.11**
![Automobilista 2 DemoAVX 2560x1440 DXVK113async macos-1 1 11](https://user-images.githubusercontent.com/102525166/196627306-fee5513c-d723-4576-bfa7-51d32f9e8391.jpg)

**Automobilista 2 DemoAVX MoltenVK macos-1.2.0**
![Automobilista 2 DemoAVX 2560x1440 DXVK113async macos-1 2 0-beta](https://user-images.githubusercontent.com/102525166/196627377-a3258d14-5cf9-440e-88c4-38e750f6e86a.jpg)

Tried also to add various env variables and getting the same results
"MVK_ALLOW_METAL_FENCES" = "1"
"MVK_ALLOW_METAL_EVENTS" = "1"
[...]

PR #1793, opened 2022-12-07
Fix app performance regression triggered by the previous introduction of VK_KHR_shader_float_controls.

billhollings:
The introduction of `shaderSignedZeroInfNanPreserveFloat32` was causing Metal fast math to be aggressively disabled by the use of `SignedZeroInfNanPreserve` in shaders in some apps, causing a regression in shader performance in those apps.

- Change `MVKConfiguration::fastMathEnabled` from `bool` to enumeration to distinguish between always using fast math or allowing shaders to selectively disable it.
- Default value to fast math always.

Fixes issue #1746.

### DIFF
--- a/Docs/Whats_New.md
+++ b/Docs/Whats_New.md
@@ -20,6 +20,7 @@ Released TBD
 
 - Fix crash on descriptor update with out-of-bounds descriptor count data.
 - Fix Metal buffer index binding overrides for push constants and attachment clearing.
+- Fix app performance regression triggered by the previous introduction of `VK_KHR_shader_float_controls`.
 - Work around `MTLCounterSet` crash on additional Intel Iris Plus Graphics devices.
 - Fix mistaken YCBCR format support indication.
 - Document new linkage model used by *Xcode 14* and later, and how to link **MoltenVK**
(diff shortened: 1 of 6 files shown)


cdavis5e (approved):

cdavis5e:
It occurs to me that what we really want is a toggle for infinity/NaN preservation, not one to disable all floating-point optimization.

billhollings:
> It occurs to me that what we really want is a toggle for infinity/NaN preservation, not one to disable all floating-point optimization.

I assume you mean we should request from Apple to have finer optimization tuning in Metal? Yes. That is something that would be useful as a mapping for similar control in Vulkan.

Or do you mean we should have such a flag in MoltenVK? Currently in Metal there is no way to enable only `SignedZeroInfNanPreserve` by itself, so we have to disable all of fast math. That was the problem in #1746. The apps blindly enable `SignedZeroInfNanPreserve` because it was now being advertised as available, without realizing that doing so was going to dramatically reduce shader performance, because of the lack of tuning granularity in Metal.

cdavis5e:
> I assume you mean we should request from Apple to have finer optimization tuning in Metal?

Well of course. This (`VK_KHR_shader_float_controls`) *is* the Vulkan control for infinity/NaN preservation, that we can't implement properly without destroying performance.

Issue #1650, opened 2022-07-16
FFXIV fails to run on Mac with > 100GB RAM


cbackas42:
This is a fairly weird issue, so please bear with me.  Final Fantasy XIV exhibits the following behavior:

- Upon startup, it tries to allocate a very large surface from Metal. I don't know the full details of this surface (or if it's a single one or multiple), but I do know that it's 16Kx16K with 2048 Layers - the total allocation ends up being somewhere around 80GB of RAM.
- On *most* Macs this allocation fails, having insufficient RAM available to fulfill the request. MVK reports the command buffer failure and things continue.
- On Macs with sufficient RAM (Studio Ultra 128GB for instance) the allocation *succeeds*, and the game proceeds to run some vertex shader on this surface. Said shader takes > 12ms to process a single vertex, so the firmware restarts the GPU.

This doesn't seem like it's MVK's fault per-se, but we're looking for solutions to work around it. Maybe a build setting or ENV setting to limit the max size of surfaces -  the thinking is if we can induce the failure we see from lower-RAM machines the game would work on the higher-RAM machines, even if the defect is in the game itself.

Can you suggest places in the MVK code base I could target to try out hacks? I have actual hardware to test on.

billhollings:
I have to say...this is pretty bizarre. 😲

Aside from the fundamental question of why the heck this is happening in a game...from my calculations, the math doesn't seem to make sense. A surface of 16K x 16k x 2k is 512 Gigapixels. You don't indicate which surface format you're using, but any format has got to be way beyond even your 128GB memory availability too.

Typically, the surface itself is created from an existing `CAMetalLayer` (or more accurately one per swapchain image). So this allocation will be happening outside MoltenVK anyway. Plus it's hard to see how that could be a 3D texture.

Can you get a handle on where this initial giant surface is coming from? Surely it's a bug in the game itself, or something?

cbackas42:
Yup. Pretty bizarre is about right!

I agree the math doesn't make a ton of sense. All I know from debugging the shader timeout was that those are the dimensions of the surface it was operating on at the time, and also that I can see ~80GB of RAM become "Wired" by the metal driver around this time. 

I'm not super familiar with how MVK works, in this case it's DXVK->MVK so I assumed all the "Metal" allocations would happen at the MVK layer. You're saying it may in fact be DXVK creating the surface?

[...]

PR #1797, opened 2022-12-12
Fix excessive Metal tile memory preallocation when rendering without attachments.

billhollings:
Metal uses `MTLRenderPassDescriptor` properties `renderTargetWidth`, `renderTargetHeight`, and `renderTargetArrayLength` to preallocate tile memory storage on machines using tiled rendering. This memory preallocation is not necessary if we are not rendering to attachments, and some apps actively define extremely oversized framebuffers when they know they are not rendering to actual attachments, making this internal tile memory allocation even more wasteful, occasionally to the point of triggering OOM crashes.

Update MoltenVK version to 1.2.2.

- Fixes issue #1650.

### DIFF
--- a/Docs/Whats_New.md
+++ b/Docs/Whats_New.md
@@ -13,6 +13,15 @@ Copyright (c) 2015-2022 [The Brenwill Workshop Ltd.](http://www.brenwill.com)
 
 
 
+MoltenVK 1.2.2
+--------------
+
+Released TBD
+
+- Fix excessive Metal tile memory preallocation when rendering without attachments.
+
+
+
 MoltenVK 1.2.1
 --------------
 
@@ -51,6 +60,7 @@ Released 2022/12/08
   - MSL: Don't dereference forwarded copies of `OpVariable` pointers.
   - MSL: Refactor member reference in terms of one boolean.
   - Fix MSL Access Chain.
+
 
 
 MoltenVK 1.2.0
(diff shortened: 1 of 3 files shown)


cdavis5e:
But wait, then how does it know how many fragment threads to spawn? Does this pass Metal validation?

billhollings:
> But wait, then how does it know how many fragment threads to spawn? Does this pass Metal validation?

Yes, it passes Metal validation. I do not see Metal validation errors on Apple Silicon when `renderTargetWidth`, `renderTargetHeight`, and `renderTargetArrayLength` are left unset. And things work correctly. So Metal is figuring out something at actual render time, possibly improvising based on the viewport, etc.

I haven't tested what this might do to performance. It's possible that providing advanced warning to Metal by setting those properties may allow it to better organize and optimize the tile cache and rendering thread groups.

js6i:
Do we know that it actually does spawn the threads based on the viewport? I'm somewhat worried that it just defaults to 0 in this case and successfully doesn't do anything.

billhollings:
@js6i @cdavis5e 

Uggh. You are of course both correct. I realize the testing I did always involved a render target (which is I guess where Metal picks up the frag shader thread count from, and not from the viewport). Thanks for going to the trouble to create a little test app to show me where I went wrong.

I'm not sure how to handle this over-allocation issue identified in #1650 then. We can't be guaranteed to have the correct viewport size (or any viewport size) at the time the Metal render encoder is created. The only option might be to defer creating the Metal render encoder until the first draw (which kinda sorta happens with tess pipelines already).

billhollings:
@js6i @cdavis5e

I've pushed PR #1802 to replace this PR, and fix the issue in #1650 in a way that always sets the Metal render area to something: the frame buffer render area, or the viewport, if there are no attachments. Since the viewport is not guaranteed until the draw call, we now defer creating the Metal renderpass until then (or until needed by other renderpass operations that involve drawing, like clearing attachments).

PR #1802, opened 2022-12-22
WIP: Fix missing render area when rendering without attachments.

billhollings:
- Rewrite logic for setting `MTLRenderPassDescriptor` properties `renderTargetWidth`, `renderTargetHeight`, and `renderTargetArrayLength` to reduce render area to viewport area when rendering without attachments, and ensure that some kind of render area is always defined.
- Since viewport area is possibly not available until drawing, defer calling `MVKCommandEncoder::beginMetalRenderPass()` until the first `finalizeDrawState()` of the renderpass.
- Add `MVKCommandEncoder::enqueueNextMetalRenderPass()` to end current render encoding, but does not start a new render pass, and call it wherever `beginMetalRenderPass()` was being called.
- Add `MVKCommandEncoder::ensureMetalRenderPass()` to call `beginMetalRenderPass()` if it hasn't been invoked since `enqueueNextMetalRenderPass()` was last called.
- Add convenience functions `mvkMakeVkRect()`, `mvkVkRectFromVkViewport()`, `mvkVkRect2DIntersection()`, and `mvkVkRect2DUnion()`.
- Set label for render area clear different from clear attachments.
- Update MoltenVKPackaging.xcodeproj build settings to Xcode 14.1 (unrelated).

- Rewinds and replaces PR #1797.
- Fixes issue #1650.

### DIFF
--- a/MoltenVK/MoltenVK/Commands/MVKCommandBuffer.mm
+++ b/MoltenVK/MoltenVK/Commands/MVKCommandBuffer.mm
@@ -472,13 +472,13 @@ void MVKCommandEncoder::setSubpass(MVKCommand* subpassCmd,
 							   (_device->_pMetalFeatures->multisampleLayeredRendering ||
 							    (getSubpass()->getSampleCount() == VK_SAMPLE_COUNT_1_BIT)));
 
-	beginMetalRenderPass(_renderSubpassIndex == 0 ? kMVKCommandUseBeginRenderPass : kMVKCommandUseNextSubpass);
+	enqueueNextMetalRenderPass(_renderSubpassIndex == 0 ? kMVKCommandUseBeginRenderPass : kMVKCommandUseNextSubpass);
 }
 
 void MVKCommandEncoder::beginNextMultiviewPass() {
 	encodeStoreActions();
 	_multiviewPassIndex++;
-	beginMetalRenderPass(kMVKCommandUseNextSubpass);
+	enqueueNextMetalRenderPass(kMVKCommandUseNextSubpass);
 }
 
 uint32_t MVKCommandEncoder::getMultiviewPassIndex() { return _multiviewPassIndex; }
@@ -501,13 +501,22 @@ void MVKCommandEncoder::endMetalEncoding(T& mtlEnc) {
 	mtlEnc = nil;
 }
 
+void MVKCommandEncoder::enqueueNextMetalRenderPass(MVKCommandUse cmdUse) {
+	endMetalRenderEncoding();
+	_mtlRenderEncoderUse = cmdUse;
+}
+
+id<MTLRenderCommandEncoder> MVKCommandEncoder::ensureMetalRenderPass() {
+	if (_mtlRenderEncoderUse && !_mtlRenderEncoder) { beginMetalRenderPass(); }
+	return _mtlRenderEncoder;
+}
 
 // Creates _mtlRenderEncoder and marks cached render state as dirty so it will be set into the _mtlRenderEncoder.
-void MVKCommandEncoder::beginMetalRenderPass(MVKCommandUse cmdUse) {
+void MVKCommandEncoder::beginMetalRenderPass() {
 
     endCurrentMetalEncoding();
 
-	bool isRestart = cmdUse == kMVKCommandUseRestartSubpass;
+	bool isRestart = _mtlRenderEncoderUse == kMVKCommandUseRestartSubpass;
     MTLRenderPassDescriptor* mtlRPDesc = [MTLRenderPassDescriptor renderPassDescriptor];
 	getSubpass()->populateMTLRenderPassDescriptor(mtlRPDesc,
 												  _multiviewPassIndex,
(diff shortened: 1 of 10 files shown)


MoltenVK/MoltenVK/Commands/MVKCommandEncoderState.mm
```diff
@@ -156,6 +169,19 @@
...
+	if ( !usingScissors.empty() ) {
+		rslt = usingScissors[0];
+		size_t sCnt = usingScissors.size();
+		for (uint32_t sIdx = 1; sIdx < sCnt; sIdx++) {
+			rslt = mvkVkRect2DUnion(rslt, usingScissors[sIdx]);
+		}
+	}
+	return rslt;
+}
+
```
cdavis5e, line 184: Is this used for anything?

MoltenVK/MoltenVK/Utility/MVKFoundation.h
```diff
@@ -285,6 +285,53 @@ static inline VkOffset3D mvkVkOffset3DDifference(VkOffset3D minuend, VkOffset3D
...
+
+	VkRect2D rslt = {};
+	if (rightEdge > leftEdge && topEdge > bottomEdge) {
+		rslt.offset.x = leftEdge;
+		rslt.offset.y = bottomEdge;
+		rslt.extent.width = rightEdge - leftEdge;
+		rslt.extent.height = topEdge - bottomEdge;
+	}
+	return rslt;
+}
```
cdavis5e, line 318: This isn't used for anything, either. I'd prefer not to have unused code, because it's more work to maintain.

cdavis5e (changes requested):

js6i:
One bit that seems to be missing is that when the application changes the viewport, we might use a stale setting?

More broadly, I'm not sure that it's MoltenVK that should be working around the issue. It's an unfortunate quirk that you can't preset a huge render area, but also something that applications should be able to avoid.

MoltenVK/MoltenVK/Commands/MVKCommandBuffer.mm
```diff
@@ -523,43 +532,42 @@
...
-        }
+		// As above, to avoid overallocation, do not use layering if there are no attachments.
+		// And Metal does not support layered render passes with a mix of 2D & 3D targets.
+		// For a multiview pass, the framebuffer layer count will be 1, so we need to use
+		// the view count for this multiview pass. Otherwise, use the framebuffer layer count.
+		if (hasAttachments && !(found3D && found2D)) {
+			mtlRPDesc.renderTargetArrayLengthMVK = (getSubpass()->isMultiview()
+													? getSubpass()->getViewCountInMetalPass(_multiviewPassIndex)
+													: getFramebufferLayerCount());
+		}
```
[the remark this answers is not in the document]
billhollings: Fair enough. 

What kind of test would we use to determine that?

Before I started messing with all this, originally, this test was:

        // Metal does not allow layered render passes where some RTs are 3D and others are 2D.
        if ( !(found3D && found2D) || renderTargetArrayLength > 1 ) {

This test seemed odd to me, because the the comment indicates we can't handle a mix of 2D & 3D attachments, but then we override that if the attachments have more than one layer. I don't understand the connection between the two conditions. They seem to be orthogonal. That's why I removed the layer override, to clarify that we don't set render layer count if we don't have any attachments, or if the attachment mix is unsupported.

At the same time, in issue #1650 which catalyzed this fix attempt, when not using attachments, FFXIV was setting the FB layer count to 2048, which triggers a large unnecessary tile memory allocation if we accept it blindly.

One of the DXVK hacks around this overallocation issue, was to say if we have no attachments, then set the layer count to 1, and I was following that thinking with `hasAttachments`.

With all that in mind, how do we figure out whether the app _really_ does want the extra layers?

MoltenVK/MoltenVK/Commands/MVKCommandEncoderState.mm
billhollings: It's a companion to the almost identical `MVKViewportCommandEncoderState::getUnionVkRect()`, but it's not used right now.

TBH...during this, I had experimented with using either (or both), viewport or scissor rect as the mechanism for trimming the render area down to what is actually being rendered, and eventually settled on viewport. 

Scissor seemed like a possible alternative because it could trim it down further if scissor is small. But then I realized scissor doesn't really have an impact on tiles, since the pixel probably has to exist in a tile in order for the pixel to be tested by the scissor test. And on top of that, scissors are often left sized to the entire render area anyway, so it wouldn't help much.

Anyway...I left it in since it is similar to `MVKViewportCommandEncoderState::getUnionVkRect()`, thinking it might be useful at some point. But it can certainly come out if that's never going to be a possibility.

MoltenVK/MoltenVK/Utility/MVKFoundation.h
billhollings: I take your point.

Like above, I just added this in as a companion to `mvkVkRect2DUnion()`. `MVKFoundation.h` is a library of misc helper functions, not all of which are used at any code point, and I figured finding the intersection of rectangles might come in handy someday.

billhollings:
@js6i 

> One bit that seems to be missing is that when the application changes the viewport, we might use a stale setting?

Potential late viewport setting is the reason I deferred creating a Metal renderpass until the first draw call. And there is definitely a danger that the viewport could move or be upsized, which would mean tiles were not available, which will trigger a Metal validation warning. To get around this, when no attachments are present, we could use a change in viewport to trigger a Metal renderpass restart to reestablish the render area based on the new viewport. 

BTW...the danger is even worse if we let the app (or emulator) take care of this before the renderpass begin, because it could then also submit a viewport that is not compatible with the viewport it used to set the render area.

> More broadly, I'm not sure that it's MoltenVK that should be working around the issue. It's an unfortunate quirk that you can't preset a huge render area, but also something that applications should be able to avoid.

I'd love to be able to dump this. It is awkward. But it is a disappointment that other Vulkan drivers on other tile-based platforms seem to somehow cope with synthetically-huge render areas and layer counts, without drastically over-allocating tile storage.

Or are they? Maybe all the platforms encountered in other drivers from games that are doing this, are being run on IM drivers that don't preallocate tile storage.

@cdavis5e @js6i  

If we do decide to push this back to the games and Vulkan emulators, formally, that would just remove the `hasAttachments` check here. But is there still value in retaining the code changes I've made to defer creating the Metal renderpass until the draw call is being finalized? Or is that behavior unnecessarily complex, fragile, or a source of possible performance or lag issues?

cdavis5e:
> But is there still value in retaining the code changes I've made to defer creating the Metal renderpass until the draw call is being finalized? Or is that behavior unnecessarily complex, fragile, or a source of possible performance or lag issues?

I believe the latter.

billhollings:
>More broadly, I'm not sure that it's MoltenVK that should be working around the issue. It's an unfortunate quirk that you can't preset a huge render area, but also something that applications should be able to avoid.

> > But is there still value in retaining the code changes I've made to defer creating the Metal renderpass until the draw call is being finalized? Or is that behavior unnecessarily complex, fragile, or a source of possible performance or lag issues?
> 
> I believe the latter.

Given the openness of the discussion, I'm going to mark this as WIP, until some of the discussion points above are addressed better, and we see whether this code path is actually needed to solve the problem it is meant to solve.

PR #1823, opened 2023-01-14
Set more accurate values for some VkPhysicalDeviceLimits members.

### LINKED ISSUE #1822: Set correct values for maxClipDistances and maxCombinedClipAndCullDistances
- According to Apple engineers, `maxClipDistances` should be `8`.

- `maxCombinedClipAndCullDistances` should be the number of distance slots that can be used either as clip and cull distances. In most cases, a GPU shares the same slots for either clip and cull distances, and so should be the max of the two individual values.

billhollings:
- Update `maxClipDistances` and `maxCombinedClipAndCullDistances` to more accurate values.
- Update `maxDrawIndexedIndexValue` to acknowledge automatic primitive restart.

Fixes #1822.

### DIFF
--- a/MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
@@ -2239,6 +2239,11 @@ void MVKPhysicalDevice::initLimits() {
 	_properties.limits.maxDescriptorSetStorageImages = (_properties.limits.maxPerStageDescriptorStorageImages * 5);
 	_properties.limits.maxDescriptorSetInputAttachments = (_properties.limits.maxPerStageDescriptorInputAttachments * 5);
 
+	_properties.limits.maxClipDistances = 8;	// Per Apple engineers.
+	_properties.limits.maxCullDistances = 0;	// unsupported
+	_properties.limits.maxCombinedClipAndCullDistances = max(_properties.limits.maxClipDistances,
+															 _properties.limits.maxCullDistances);  // If supported, these consume the same slots.
+
 	// Whether handled as a real texture buffer or a 2D texture, this value is likely nowhere near the size of a buffer,
 	// needs to fit in 32 bits, and some apps (I'm looking at you, CTS), assume it is low when doing 32-bit math.
 	_properties.limits.maxTexelBufferElements = _properties.limits.maxImageDimension2D * (4 * KIBI);
@@ -2486,7 +2491,8 @@ void MVKPhysicalDevice::initLimits() {
     _properties.limits.minTexelGatherOffset = _properties.limits.minTexelOffset;
     _properties.limits.maxTexelGatherOffset = _properties.limits.maxTexelOffset;
 
-    // Features with no specific limits - default to unlimited int values
+
+    // Features with no specific limits - default to effectively unlimited int values
 
     _properties.limits.maxMemoryAllocationCount = kMVKUndefinedLargeUInt32;
 	_properties.limits.maxSamplerAllocationCount = getMaxSamplerCount();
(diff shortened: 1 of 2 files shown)


cdavis5e (approved):

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
```diff
@@ -2239,6 +2239,11 @@
 	_properties.limits.maxDescriptorSetStorageImages = (_properties.limits.maxPerStageDescriptorStorageImages * 5);
 	_properties.limits.maxDescriptorSetInputAttachments = (_properties.limits.maxPerStageDescriptorInputAttachments * 5);
 
+	_properties.limits.maxClipDistances = 8;	// Per Apple engineers.
+	_properties.limits.maxCullDistances = 0;	// unsupported
+	_properties.limits.maxCombinedClipAndCullDistances = max(_properties.limits.maxClipDistances,
```
lexaknyazev, line 2244: Since Metal's clip distance slots cannot be used for cull distances, shouldn't the combined number be zero?

billhollings: > Since Metal's clip distance slots cannot be used for cull distances, shouldn't the combined number be zero?

Hmmm. I don't think so. 

I interpret "Combined" to mean the number of slots that can be used in total between the two clip and cull functionality, implying some kind of overlap between the two if that number is not a simple sum of the two. This contrasts with an interpretation that it means the number that can be optionally used for either clip or cull, which I think is what you're describing.

There are these notes in the spec that lead me to this conclusion:

`maxCombinedClipAndCullDistances` is the maximum combined number of clip and cull distances that can be used in a single shader stage. The sum of the sizes of any pair of arrays declared with the `ClipDistance` and `CullDistance` built-in decoration used by a single shader stage in a shader module must be less than or equal to this limit.

`VUID-VkPipelineShaderStageCreateInfo-maxCombinedClipAndCullDistances-00710`
If the identified entry point includes any variables in its interface that are declared with the `ClipDistance` or `CullDistance` `BuiltIn` decoration, those variables must not have array sizes which sum to more than `VkPhysicalDeviceLimits::maxCombinedClipAndCullDistances`.

The VU will fail if `maxCombinedClipAndCullDistances` is zero.

lexaknyazev: Historically, GL supported only clip distances and did not have any combined limit. Later, `GL_ARB_cull_distance` introduced both the new built-in output and the combined limit, so the exact semantic of the latter was not defined for the case with no cull distance support.

The Vulkan spec language mentions "pair of arrays", which makes little sense when only one of two types is supported. On the other hand, VU is clear but, in my opinion, the spec should be the authoritative source.

billhollings: > VU is clear but, in my opinion, the spec should be the authoritative source.

I interpret the Vulkan spec description of `maxCombinedClipAndCullDistances`, which I included above, as covering the same concept as the VU.

Is that not your interpretation as well? Or are you looking at a different area of the Vulkan spec that describes the intention of `maxCombinedClipAndCullDistances` differently?

lexaknyazev: I see VU as being more strict than the spec description, see the internal issue 3325. Let's refine the spec language first.

billhollings: > Or are you looking at a different area of the Vulkan spec that describes the intention of `maxCombinedClipAndCullDistances` differently?

No, I'm just interpreting the description and VU that I quoted above. They both consistently indicate that the _**sum**_ of consumed clip and cull distance counts cannot be greater than `maxCombinedClipAndCullDistances`.

So, in our case, with `maxClipDistances = 8` and `maxCullDistances = 0`, a user can consume up to 8 clip distances and `0` cull distances, for a sum of up to `8`. This value must not be greater than `maxCombinedClipAndCullDistances`, meaning that `maxCombinedClipAndCullDistances` must be at least `8` in order for the rule to be satisfied in all cases.

I assume (but do not know for sure) the reason for needing 3 distinct limits is because some GPU's might physically share clip and cull distance slots, and others might not. 

For a GPU that has a 8 separate slots for each, we might see:

```
maxClipDistances = 8
maxCullDistances = 8
maxCullDistances = 16
```

And for a GPU that shares 8 slots between clip and cull operations, we might see:

```
maxClipDistances = 8
maxCullDistances = 8
maxCullDistances = 8
```

lexaknyazev: As shown in the quoted "Limit Requirements" spec section above, the combined limit is tied to the `shaderCullDistance` feature, thus it should be set to zero when the feature is unsupported. Other parts of the spec are not so specific.

billhollings: > the combined limit is tied to the `shaderCullDistance` feature, thus it should be set to zero when the feature is unsupported

Ah. Okay. I see what you're saying regarding the `shaderCullDistance ` feature flag. 

I agree there is a conflict in the wording within the spec. Hopefully your spec MR will get some clarity around this. For the time being though, I believe the VU will get triggered if we set `maxCombinedClipAndCullDistances` to zero, so let's leave it the way it is until the spec wording is clarified.

PR #1834, opened 2023-02-04
Work around problems with explicit LoD with arrayed depth images on A…

cdavis5e:
…pple Silicon.

Use an explicit gradient to make it sample the correct level.

Update SPIRV-Cross to pull in the change needed for this.

Fixes two tests in the Vulkan CTS.


billhollings (approved):
LGTM.

Issue #1791, opened 2022-12-01
vkGetPhysicalDeviceSurfaceCapabilitiesKHR reports incorrect surface extent on iOS


jswillard:
I found that checking the extent of the surface does not match the actual width and height of the underlying CAMetalLayer. I am testing on 12.9inch iPad Pro which has a 2732x2048 resolution display. The CAMetalLayer will report the drawableSize as 2732x2048 but I get a surface extent from Vulkan of 2731x2048.

I believe the culprit is this function in CAMetalLayer+MoltenVK.m
```
-(CGSize) naturalDrawableSizeMVK {
    CGSize drawSize = self.bounds.size;
    CGFloat scaleFactor = self.contentsScale;
    drawSize.width = trunc(drawSize.width * scaleFactor);
    drawSize.height = trunc(drawSize.height * scaleFactor);
    return drawSize;
}
```
The bounds of the layer are 1590x1192 with a contentsScale of 1.7181208053691275. 

I tried replacing trunc() with round() which worked great on my 12.9inch iPad but broke the 11inch iPad version.

I can't think of a great way to solve this as per Apple's docs `bounds * contentsScale` is the expected way to get the extent of the surface. If you set `drawableSize = bounds * contentsScale` then Apple will do some rounding behind the scenes to the correct value, it's not clear what rounding they are doing except that drawableSize will be the correct value (2732x2048).

> By default, a layer creates textures sized to match its content—that is, this property’s value is the layer’s [bounds](https://developer.apple.com/documentation/quartzcore/calayer/1410915-bounds?language=objc) size multiplied by its [contentsScale](https://developer.apple.com/documentation/quartzcore/calayer/1410746-contentsscale?language=objc) factor.

For now, I am just creating swapchains of 2731x2048 so that the image extent and the surface extent match and then letting the system scale appropriately. It would be great to have a proper solution for this, thanks!

billhollings:
Could be adjusted along with fixes for https://github.com/KhronosGroup/MoltenVK/issues/1753#issuecomment-1338004413.

billhollings:
PR #1835 should fix this issue. Please retest and close this issue if the problem you encountered has been resolved.

jswillard:
[...]

Issue #1753, opened 2022-10-28
vkAcquireNextImageKHR returns VK_SUCCESS even when window is resized


johnpayne-dev:
Happens on both v1.1.11 and v1.2.0 for me, I don't remember this always being an issue so I'll try and find a regression.

When I fullscreen the window VK_SUBOPTIMAL_KHR is returned, it's just resizing that continues to return VK_SUCCESS.

timprepscius:
I am also seeing that the behavior for vkAcquireNextImageKHR has changed.
I believe it changed before 1.1.4.

In the cube demo for 1.1.11
The vkAcquireNextImageKHR isn't the driver for recreation, but the

```
    } else if (err == VK_SUBOPTIMAL_KHR) {
        // SUBOPTIMAL could be due to a resize
        VkSurfaceCapabilitiesKHR surfCapabilities;
        err = demo->fpGetPhysicalDeviceSurfaceCapabilitiesKHR(demo->gpu, demo->surface, &surfCapabilities);
        assert(!err);
        if (surfCapabilities.currentExtent.width != (uint32_t)demo->width ||
            surfCapabilities.currentExtent.height != (uint32_t)demo->height) {
            demo_resize(demo);
        }
    }
```

In my code, I also never get a VK_SUBOPTIMAL_KHR from the present frame :-/
I'm not sure what this means.

billhollings:
MoltenVK returns `VK_SUBOPTIMAL_KHR` if it detects that the `CAMetalLayer` underlying the `NSView/UIView` has changed dimensions (resized) or content scale (eg. moved to a different physical display).

However, as a Vulkan driver, MoltenVK does not manage resizing the view. It is the responsibility of the app to configure the view to be resizable. It's possible that your app does not support the view being resized to different dimensions (as opposed to simply being stretched).

timprepscius:
Yeah.  Before I could rely on VK_SUBOPTIMAL_KHR or some other Vulkan exception occurring (I forget exactly), which I would use to trigger rebuilding the surface.

Now, for whatever reason, they no longer occur (after I updated MoltenVK), and I now rely on the apple:
    func mtkView(_ view: MTKView, drawableSizeWillChange size: CGSize)

i'm not exactly sure if this is completely in line with what you wrote or not.

billhollings:
[...]

Issue #1803, opened 2022-12-23
VK_EXT_swapchain_maintenance1, VK_EXT_surface_maintenance1 support


egorov-k:
There are some gaps in vulkan WSI spec. The spec doesn't explain clearly when it's safe to destroy a swapchain and destroy a semaphore passed to VkPresentInfo.waitSemaphores. VK_EXT_swapchain_maintenance1, VK_EXT_surface_maintenance1 fixes these gaps.

billhollings:
PR #1835 implements this request.

PR #1835, opened 2023-02-04
Add support for VK_EXT_swapchain_maintenance1 and VK_EXT_surface_maintenance1

billhollings:
This PR is an overhaul of how swapchain image sizing is handled, implements two recent maintenance extensions, and fixes a couple of outstanding surface resizing bugs.

- Support querying scaling capabilities and present mode capabilities when querying surface capabilities.
- Rename `MVKPresentTimingInfo` to `MVKImagePresentInfo` and add present mode and fence to support dynamic present mode changes and fence signalling.
- `MVKPresentableSwapchainImage` remove static functions from class declaration.
- `MVKSwapchain` support releasing swapchain images on command.
- `MVKSwapchain` support configuring with scaling and gravity info, apply it to `CAMetalLayer`, and do not return `VK_SUBOPTIMAL_KHR` if swapchain was configured with scaling info.
- Rename `MVKSwapchain::acquireNextImageKHR()` to `acquireNextImage()`.
- `CAMetalLayer naturalDrawableSizeMVK()` compute precise drawable size.
- `CAMetalLayer` remove obsolete and unused `updatedDrawableSizeMVK()` method.
- Rename `MVKConfiguration::swapchainMagFilterUseNearest` to `swapchainMinMagFilterUseNearest` to apply `CAMetalLayer` size filtering to both magnification and minification, and rename corresponding  env var `MVK_CONFIG_SWAPCHAIN_MAG_FILTER_USE_NEAREST` to `MVK_CONFIG_SWAPCHAIN_MIN_MAG_FILTER_USE_NEAREST`.
- Detect when size of surface has changed under the covers.
- Change rounding of surface size provided by Metal from truncation to rounding-with-half-to-even.
- Remove `MVKLogSizeOf()` logging macro as redundant to `mvkPrintSizeOf()` macro.
- Update `VK_MVK_MOLTENVK_SPEC_VERSION` to version `37`.
- Queue submissions retain wait semaphores until `MTLCommandBuffer` finishes (unrelated but exposed during CTS testing).
- Add additional Vulkan error strings (unrelated).

Passes all CTS WSI macOS tests, including new tests for `VK_EXT_swapchain_maintenance1` and `VK_EXT_surface_maintenance1`.

- Fixes enhancement #1803.
- Fixes issue #1791.
- Fixes issue #1753.


MoltenVK/MoltenVK/Utility/MVKFoundation.h
```diff
@@ -339,11 +339,13 @@ static inline bool mvkVkComponentMappingsMatch(VkComponentMapping cm1, VkCompone
...
 
 
-#pragma mark -
-#pragma mark Template functions
-
 #pragma mark Math
 
+/** Rounds the value to nearest integer using half-to-even rounding. */
+static inline double mvkRoundHalfToEven(const double val) {
+	return val - std::remainder(val, 1.0);	// remainder() uses half-to-even rounding
```
cdavis5e, line 346: I guess, unlike in MSL, `rint(3)` isn't guaranteed to perform nearest-even rounding, even when `fegetround(3)` returns `FE_TONEAREST`.

cdavis5e (changes requested):

MoltenVK/MoltenVK/Utility/MVKFoundation.h
```diff
@@ -339,11 +339,13 @@ static inline bool mvkVkComponentMappingsMatch(VkComponentMapping cm1, VkCompone
 #define mvkPrintSizeOf(type)    printf("Size of " #type " is %lu.\n", sizeof(type))
 
 
-#pragma mark -
-#pragma mark Template functions
-
 #pragma mark Math
 
+/** Rounds the value to nearest integer using half-to-even rounding. */
+static inline double mvkRoundHalfToEven(const double val) {
```
cdavis5e, line 345: Oh, I forgot to mention: this could probably be a `constexpr` function.

MoltenVK/MoltenVK/GPUObjects/MVKImage.h
```diff
@@ -447,33 +449,29 @@ class MVKPresentableSwapchainImage : public MVKSwapchainImage {
 #pragma mark Metal
 
 	/** Presents the contained drawable to the OS. */
-	void presentCAMetalDrawable(id<MTLCommandBuffer> mtlCmdBuff,
-								MVKPresentTimingInfo presentTimingInfo);
+	void presentCAMetalDrawable(id<MTLCommandBuffer> mtlCmdBuff, MVKImagePresentInfo presentInfo);
```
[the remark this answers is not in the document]
billhollings: Good catch. Fixed.

MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
```diff
@@ -1404,13 +1392,39 @@
...
+		_availabilitySignalers.clear();
+		_availability.isAvailable = true;
+	}
+}
+
 
 #pragma mark Construction
 
+// The deferImgMemAlloc parameter is ignored, because image memory allocation is provided by a MTLDrawable,
+// which is retrieved lazily, and hence is already as deferred (or as deferred as we can make it).
```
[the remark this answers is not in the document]
billhollings: I added it based on the extension features, but later realized it's not needed. I've removed it now, but retained a descriptive comment.

MoltenVK/MoltenVK/GPUObjects/MVKSurface.mm
```diff
@@ -42,8 +42,6 @@
 					   const Vk_PLATFORM_SurfaceCreateInfoMVK* pCreateInfo,
 					   const VkAllocationCallbacks* pAllocator) : _mvkInstance(mvkInstance) {
 
-//	MVKLogInfo("%s(): This function is obsolete. Consider using the vkCreateMetalSurfaceEXT() function from the VK_EXT_metal_surface extension instead.", STR(vkCreate_PLATFORM_SurfaceMVK));
-
```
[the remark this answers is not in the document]
billhollings: Yeah. That would be good in the long run.

MoltenVK/MoltenVK/GPUObjects/MVKSwapchain.h
```diff
@@ -110,30 +122,35 @@ class MVKSwapchain : public MVKVulkanAPIDeviceObject {
...
     void willPresentSurface(id<MTLTexture> mtlTexture, id<MTLCommandBuffer> mtlCmdBuff);
     void renderWatermark(id<MTLTexture> mtlTexture, id<MTLCommandBuffer> mtlCmdBuff);
     void markFrameInterval();
-	void recordPresentTime(MVKPresentTimingInfo presentTimingInfo, uint64_t actualPresentTime = 0);
+	void recordPresentTime(MVKImagePresentInfo presentInfo, uint64_t actualPresentTime = 0);
 
-	CAMetalLayer* _mtlLayer;
-    MVKWatermark* _licenseWatermark;
+	CAMetalLayer* _mtlLayer = nil;
+    MVKWatermark* _licenseWatermark = nil;
```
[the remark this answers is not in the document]
billhollings: Fixed.

MoltenVK/MoltenVK/Utility/MVKFoundation.h
billhollings: Good idea. I've added a separate commit to this PR to fix this, and to also similarly change a bunch of other inlines that could possibly be resolved at compile-time.

cdavis5e: Huh. It looks like Xcode 11 and 12 are too old to recognize `mvkRoundHalfToEven()` as `constexpr`.

cdavis5e: Also, if the functions aren't `static` and don't wind up inlined, won't they appear in the public symbol table? I guess it doesn't matter much, since the ODR requires that every instance of such a function have the same definition, and they all have prefixes which makes a name collision unlikely anyway.

cdavis5e: > Huh. It looks like Xcode 11 and 12 are too old to recognize `mvkRoundHalfToEven()` as `constexpr`.

Almost certainly because `std::remainder()` isn't `constexpr` in these versions.

billhollings: > Also, if the functions aren't static and don't wind up inlined, won't they appear in the public symbol table?

Ah. Thanks for catching that.

Are you thinking we should use `static constexpr` instead of `constexpr`, to ensure they are always inlined? If so...I guess there's no reason not to. 

In current practice it shouldn't make a difference. The `MVKFoundation.h` header is not meant for external consumption, and none of its functions are marked with `MVK_PUBLIC_SYMBOL`, so they don't survive the `GCC_SYMBOLS_PRIVATE_EXTERN=1` we use.

However, I guess the code itself shouldn't depend on how we build, and so we should be deliberate about marking things static when they should be. I'll add the `static` part back.

billhollings: > Almost certainly because `std::remainder()` isn't `constexpr` in these versions.

Hmmm...`std::remainder()` isn't `constexpr` [until](https://en.cppreference.com/w/cpp/numeric/math/remainder) C++23. Strange that Xcode 14 doesn't complain. I've reverted to `static inline`.

cdavis5e (approved):

PR #1836, opened 2023-02-05
WIP: enable availability of EXT_shader_atomic_float 

rcoreilly:
also adds docs for how to replace Vulkan SDK .dylib while avoiding need to reboot.

This goes together with https://github.com/KhronosGroup/SPIRV-Cross/pull/2105 to enable the opcodes.

See discussion here: https://github.com/KhronosGroup/MoltenVK/discussions/1616#discussioncomment-4868519


spnda (changes requested):
1. This does not cover extension features, which are required to be set because otherwise the extension is not allowed to be advertised as per the spec (To an application it currently looks like nothing is supported). Look at the massive switches in MVKDevice.mm. Its a matter of just a few little lines to properly enable all the float32 features as described in [this struct](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VkPhysicalDeviceShaderAtomicFloatFeaturesEXT.html).
2. The changes to the user guide should really be in a separate PR, as they're completely irrelevant to the actual change proposed here.
3. When the associated SPIRV-Cross PR is merged, update the git revision in ExternalRevisions/SPIRV-Cross accordingly.

billhollings:
Since the functionality in this PR depends on an update to SPIRV-Cross, we have to wait for the PR there to be merged, and then the SPIRV-Cross merge commit value set into the MoltenVK `ExternalRevisions/SPIRV-Cross_repo_revision` file, and included in this PR.

phush0:
SPIRV-Cross PR is accepted

rcoreilly:
when I run ./fetchDependencies I get this error:

```
SPIRV-Cross repo: https://github.com/KhronosGroup/SPIRV-Cross.git
SPIRV-Cross revision: 3550a54ae01b295c40ce972d951b420b388b9401
fatal: reference is not a tree: 3550a54ae01b295c40ce972d951b420b388b9401
```

That SHA is the one of the merge, in the main branch.  Not sure what the issue is.

billhollings:
> when I run ./fetchDependencies I get this error:...

I downloaded this PR and didn't have any trouble running `fetchDependencies`. Perhaps clear away the `External\SPRIV-Cross` directory first?

Docs/MoltenVK_Runtime_UserGuide.md
```diff
@@ -219,6 +220,27 @@ To link **MoltenVK** to your application as a dynamic library (`.dylib`), follow
...
+
+If you do not do the remove first, your application will terminate immediately with a singularly unhelpful `Killed: 9` message.
+
+2. Do *not* copy the `MoltenVK_icd.json` file from the newly-built package to `/usr/local/share/vulkan/icd.d` -- it will not work and will result in errors about not being able to initialize the instance.  The one installed by Vulkan SDK uses a relative path to specify the location of the `.dylib`, whereas the one in the package specifies it in the same directory.
+
+3. The default config for command-line build has verbose logging info turned on -- if you want it to be like the original, use this command for building:
+
+```bash
+$ make macos MVK_CONFIG_LOG_LEVEL=1
+```
```
billhollings, line 243: Please text-wrap this content in a similar manner to the rest of the content in this document, so it's also readable in a regular text editor. Open the document in Xcode or a text editor to check the wrapping.

Docs/MoltenVK_Runtime_UserGuide.md
```diff
@@ -219,6 +220,27 @@ To link **MoltenVK** to your application as a dynamic library (`.dylib`), follow
       - `MoltenVK/dylib/tvOS/libMoltenVK.dylib` *(tvOS)* 
      
 
+<a name="install_vksdk"></a>
+### Install *MoltenVK* replacing the Vulkan SDK .dylib
```
billhollings, line 224: Please clarify by using `libMoltenVK.dylib` instead of just `.dylib`.

Docs/MoltenVK_Runtime_UserGuide.md
billhollings: Your instructions are great for devs just looking to build and use MoltenVK from the repo. Thanks for adding them.

But for future reference if you are developing and testing changes to MoltenVK, I rename the `libMoltenVK.dylib` file in `/usr/local/lib/` and then create a symlink there to `<path-to-MoltenVK>/Package/Latest/MoltenVK/dylib/macOS/libMoltenVK.dylib`. That way any rebuild of MoltenVK is automatically available. I just have to remember to do it each time I download and install a new SDK.

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
```diff
@@ -401,6 +401,11 @@
 				shaderIntFuncsFeatures->shaderIntegerFunctions2 = true;
 				break;
 			}
+			case VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_SHADER_ATOMIC_FLOAT_FEATURES_EXT: {
+				auto* atomicFloatFeatures = (VkPhysicalDeviceShaderAtomicFloatFeaturesEXT*)next;
```
billhollings, line 405: We can't assume the app has cleared the `VkPhysicalDeviceShaderAtomicFloatFeaturesEXT` before submitting it. Please either explicitly set the values of all the flags, or add the following line after this line, and before setting the `shaderBufferFloat32AtomicAdd` flag:

`mvkClear(atomicFloatFeatures);`

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
```diff
@@ -401,6 +401,11 @@
...
+				atomicFloatFeatures->shaderBufferFloat32AtomicAdd = _metalFeatures.mslVersion >= 030000;
```
billhollings, line 406: What about the other atomic float capabilities? From the Metal docs, I would expect that all four buffer 32-bit  float capabilities are available:

 ```
shaderBufferFloat32Atomics
shaderBufferFloat32AtomicAdd
shaderSharedFloat32Atomics
shaderSharedFloat32AtomicAdd
```

Metal supports supports the load, store, and exchange of `shaderBufferFloat32Atomics`. And the `Shared` capabilities seem to be covered by Metal's `threadgroup A* object` function parameter types.

Am I missing something? Or did you exclude them due to time constraints relative to your requirements?

**Review by billhollings — CHANGES_REQUESTED** [review]
Thanks for your submission! 

I've requested some changes, and asked a couple of questions inline.

We'll also need to run the appropriate CTS to verify `VK_EXT_shader_atomic_float` conformance. I can take care of that once this PR is further along.

File: MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm (hunk 4 above)
Comment (rcoreilly) [reply MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm]: Ok I went ahead and enabled those other flags, and addressed the other comments above in the latest push.  Thanks!

File: Docs/Whats_New.md [hunk 5]
```diff
@@ -22,7 +22,7 @@ Released TBA
   was sometimes incorrectly disabled due to a Metal driver bug.
 - Work around problems with using explicit LoD with arrayed depth images
   on Apple Silicon.
-
+- Added support for `VK_EXT_shader_atomic_float` *(requires Metal 3.0)*.
```
billhollings, line 25:
```suggestion

```

Oops. I can't merge because it looks like your dev pull from this repo is not current, and it's causing a conflict in `Whats_New.md`.

The simplest thing to do is for you to just remove the change I requested to `Whats_New.md`, and I'll add `VK_EXT_shader_atomic_float` into that document when I clean it up before the upcoming Vulkan SDK release.

billhollings (changes requested):
Okay. LGTM now. And CTS test results are reasonable.

One last thing...I can't merge because it looks like your dev pull from this repo is not current, and it's causing a conflict in `Whats_New.md`.

The simplest thing to do is for you to just remove the change I requested to `Whats_New.md`. I'll add `VK_EXT_shader_atomic_float` into that document when I clean it up before the upcoming Vulkan SDK release.

billhollings (approved):

PR #1842, opened 2023-02-22
Fixes to dynamic rendering and occlusion queries.

billhollings:
This PR fixes some issues with dynamic rendering and occlusion queries, and modifies the some Metal debug labelling to make it easier to distinguish attachment clearing activities within a GPU capture.

- Fix crash when `VkCommandBufferInheritanceInfo::renderPass` is `VK_NULL_HANDLE`.
- Do not clear attachments when dynamic rendering is resumed.
- Allow ending dynamic rendering to trigger next multiview pass if needed.
- Move deciding to begin next multiview pass to `MVKCommandEncoder`.
- Fix premature caching of occlusion query results during tessellation rendering. Tessellation ends Metal renderpass for compute control and eval stages. Wait until end of Metal renderpass after rasterization stage.
- `vkCmdCopyQueryPoolResults()`: Fix loss of queries when query count is not a multiple of GPU threadgroup execution width.
- Disable occlusion recording while clearing attachments or render area.
- `MVKCmdClearAttachments` improve labelling of `MTLDebugGroup` to better distinguish clearing renderpass render area from `vkCmdClearAttachments()` in an Xcode GPU capture (unrelated but helpful during debugging).
- `MVKCmdClearAttachments` re-order member variables to optimize memory requirements (unrelated).
- `MVKCommandBuffer` remove unused renderpass tracking functions (unrelated).

### DIFF
--- a/MoltenVK/MoltenVK/Commands/MVKCommandBuffer.h
+++ b/MoltenVK/MoltenVK/Commands/MVKCommandBuffer.h
@@ -107,6 +107,12 @@ public:
     /** Returns whether this command buffer can be submitted to a queue more than once. */
     bool getIsReusable() { return _isReusable; }
 
+	/**
+	 * If this is a secondary command buffer, returns the number of views inherited
+	 * from the primary command buffer. If this is a primary command buffer, returns 1.
+	 */
+	uint32_t getViewCount() const;
+
     /**
      * Metal requires that a visibility buffer is established when a render pass is created, 
      * but Vulkan permits it to be set during a render pass. When the first occlusion query
@@ -129,24 +135,6 @@ public:
 
 	/** The most recent recorded tessellation pipeline */
 	MVKCmdBindPipeline* _lastTessellationPipeline;
-
-
-#pragma mark Multiview render pass command management
-
-	/** Update the last recorded multiview render pass */
-	void recordBeginRenderPass(MVKCmdBeginRenderPassBase* mvkBeginRenderPass);
-
-	/** Update the last recorded multiview subpass */
-	void recordNextSubpass();
-
-	/** Forget the last recorded multiview render pass */
-	void recordEndRenderPass();
-
-	/** The most recent recorded multiview render subpass */
-	MVKRenderSubpass* _lastMultiviewSubpass;
-
-	/** Returns the currently active multiview render subpass, even for secondary command buffers */
-	MVKRenderSubpass* getLastMultiviewSubpass();
 
 
 #pragma mark Construction
@@ -188,7 +176,7 @@ protected:
 	MVKSmallVector<VkFormat, kMVKDefaultAttachmentCount> _colorAttachmentFormats;
 	MVKCommandPool* _commandPool;
 	VkCommandBufferInheritanceInfo _secondaryInheritanceInfo;
-	VkCommandBufferInheritanceRenderingInfo _inerhitanceRenderingInfo;
+	VkCommandBufferInheritanceRenderingInfo _secondaryInerhitanceRenderingInfo;
 	id<MTLCommandBuffer> _prefilledMTLCmdBuffer = nil;
     MVKCommandEncodingContext* _immediateCmdEncodingContext = nullptr;
     MVKCommandEncoder* _immediateCmdEncoder = nullptr;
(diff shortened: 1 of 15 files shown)


MoltenVK/MoltenVK/GPUObjects/MVKQueryPool.mm
```diff
@@ -183,9 +185,16 @@
...
 		cmdEncoder->setComputeBytes(mtlComputeCmdEnc, _availability.data(), _availability.size() * sizeof(Status), 5);
 		_availabilityLock.unlock();
+
 		// Run one thread per query. Try to fill up a subgroup.
-		[mtlComputeCmdEnc dispatchThreadgroups: MTLSizeMake(max(queryCount / mtlCopyResultsState.threadExecutionWidth, NSUInteger(1)), 1, 1)
-						  threadsPerThreadgroup: MTLSizeMake(min(NSUInteger(queryCount), mtlCopyResultsState.threadExecutionWidth), 1, 1)];
+		NSUInteger threadCount = NSUInteger(queryCount);
+		NSUInteger threadExecutionWidth = mtlCopyResultsState.threadExecutionWidth;
+		NSUInteger tgWidth = min(threadCount, threadExecutionWidth);
+		NSUInteger tgCount = threadCount / threadExecutionWidth;
```
cdavis5e, line 193: What happens when there are fewer queries than will fill up an SIMD-group?

MoltenVK/MoltenVK/GPUObjects/MVKQueryPool.mm
```diff
@@ -281,12 +290,9 @@
 }
 
 void MVKOcclusionQueryPool::beginQueryAddedTo(uint32_t query, MVKCommandBuffer* cmdBuffer) {
+	// In multiview passes, one query is used for each view.
+	NSUInteger queryCount = cmdBuffer->getViewCount();
```
cdavis5e, line 294: Wait, what happens when we're *not* in a secondary command buffer?

cdavis5e (changes requested):
I had noticed some of these problems myself, while I was trying out a potential fix for some of the CTS tests.

MoltenVK/MoltenVK/Commands/MVKCommandBuffer.h
```diff
@@ -188,7 +176,7 @@ class MVKCommandBuffer : public MVKDispatchableVulkanAPIObject,
 	MVKSmallVector<VkFormat, kMVKDefaultAttachmentCount> _colorAttachmentFormats;
 	MVKCommandPool* _commandPool;
 	VkCommandBufferInheritanceInfo _secondaryInheritanceInfo;
-	VkCommandBufferInheritanceRenderingInfo _inerhitanceRenderingInfo;
+	VkCommandBufferInheritanceRenderingInfo _secondaryInerhitanceRenderingInfo;
```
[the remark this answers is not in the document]
billhollings: Thanks. Good eyes. Fixed.

MoltenVK/MoltenVK/GPUObjects/MVKQueryPool.mm
billhollings: Is your concern about passing the right values to `dispatchThreadgroups:threadsPerThreadgroup:`?

If so, I think this

    NSUInteger threadCount = NSUInteger(queryCount);
    NSUInteger tgWidth = min(threadCount, threadExecutionWidth);

combined with  the guard code in the shader

    if (query >= numQueries) { return; }

 should cover that?

Or are you concerned that there is something deeper in the hardware that will run into trouble?

cdavis5e: > Is your concern about passing the right values to `dispatchThreadgroups:threadsPerThreadgroup:`?

Yes.

What I'm trying to say is, if `queryCount` is less than `threadExecutionWidth`, then the division will come out zero, which isn't what we want. This division probably needs to round up.

billhollings: > This division probably needs to round up.

Ah. Yes. The line following it does that.

cdavis5e: How did I not notice that...?

Ah well. Never mind, then.

MoltenVK/MoltenVK/GPUObjects/MVKQueryPool.mm
billhollings: It would default to 1. This was the same as existing behavior, since the existing `_lastMultiviewSubpass` was never set (`recordBeginRenderPass` was never called).

The `queryCount` value here is only used as a sanity check on multipass queries that approach the end of the visibility buffer, so I toyed with simply removing all this code.

However, in the end, I added a `MVKCurrentSubpassInfo` helper in `MVKCommandBuffer` to track the current renderpass, subpass, or dynamic rendering, to derive a `viewMask` for the command corresponding to the currently added subpass, replacing and expanding on what `_lastMultiviewSubpass` was supposed to do.

So, even though it's just doing an edge-case validation test, we've got some plumbing to track other subpass command info if we need to.

cdavis5e (approved):

Issue #1837, opened 2023-02-09
Occlusion query Metal validation issue


danginsburg:
I'm using MoltenVK on macOS 12.5.1 AMD Firepro D500 built to commit dc2dad84d61b68c0244fdb8220bfd0218570cd46

I have some fairly simple code that issues a begin/end occlusion query in a query pool with 1024 queries allocated.  When the command buffer is submitted to the queue, it gives this validation error on commit:

-[MTLDebugCommandBuffer lockPurgeableObjects]:2103: failed assertion `MTLResource 0x60000a8109a0 (label: (null)), referenced in cmd buffer 0x7fe6953f4600 (label: vkQueueSubmit CommandBuffer on Queue 0-0) is in volatile or empty purgeable state at commit'

I've narrowed it down to the visibilityResultBuffer which is allocated with getTempMTLBuffer().

If I remove the calls to setPurgeableState in MVKMTLBufferAllocationPool::acquireAllocationUnlocked() and MVKMTLBufferAllocationPool::returnAllocationUnlocked() the problem goes away.

I'm unable to provide a repro, although as I said it's really a pretty simple single begin/end on an occlusion query.

cdavis5e:
Huh. So I guess you can't use a purgeable buffer as a visibility buffer.

This should be easy to fix. We just need to set the purgeable state to `MTLPurgeableStateNonVolatile` before the command buffer is committed, and then after the query results are copied, we can set it back to `Volatile`.

danginsburg:
> This should be easy to fix. We just need to set the purgeable state to MTLPurgeableStateNonVolatile before the command buffer is committed, and then after the query results are copied, we can set it back to Volatile.

My impression is that this is what the code is currently doing, but perhaps I don't understand.  getTempMTLBuffer sets the state to NonVolatile and then when it is returned to the pool it gets set to Volatile and that happens on a completion handler.

billhollings:
> This should be easy to fix.

@cdavis5e Are you able to investigate a fix for this, or should I look into it.

billhollings:
I believe I've got a handle on this. I'll push a PR soon.

billhollings:
This should be fixed by PR #1843.

PR #1843, opened 2023-02-23
Use a different visibility buffer for each MTLCommandBuffer in a queue submit.

billhollings:
Fixes issue #1837.


danginsburg:
I've confirmed this fixes issue #1837 for me.

cdavis5e (approved):

PR #1851, opened 2023-03-16
Fixed invalid local variable reuse.

js6i:


MoltenVK/MoltenVK/GPUObjects/MVKPipeline.mm
```diff
@@ -1240,8 +1240,8 @@ static MTLVertexFormat mvkAdjustFormatVectorToSize(MTLVertexFormat format, uint3
     // Vertex buffer divisors (step rates)
     std::unordered_set<uint32_t> zeroDivisorBindings;
     if (pVertexInputDivisorState) {
-        vbCnt = pVertexInputDivisorState->vertexBindingDivisorCount;
-        for (uint32_t i = 0; i < vbCnt; i++) {
+        uint32_t vbdCnt = pVertexInputDivisorState->vertexBindingDivisorCount;
+        for (uint32_t i = 0; i < vbdCnt; i++) {
```
billhollings, line 1244: By "invalid", I assume you don't mean invalid in a C++ sense, and mean that reusing and changing the `vbCnt` variable here gives incorrect results when `vbCnt` is used later in the function to fill attributes, etc?

That code is almost 5 years old now. Surprising that it hasn't been encountered before!

@cdavis5e Was your original intention to deliberately alter the value of `vbCnt` when divisors are present?

billhollings (commented):
See my comment below. Just want to get @cdavis5e to approve this too.

MoltenVK/MoltenVK/GPUObjects/MVKPipeline.mm
js6i: Yeah, I meant it in the sense that the changed value is invalid when it's used later in the function. I'm also surprised that it didn't manifest itself sooner. For me it bumped into the "buffer has no stride" assertion, because the loop below didn't fix zero stride for a constant step rate attribute.

cdavis5e: > @cdavis5e Was your original intention to deliberately alter the value of `vbCnt` when divisors are present?

No; in fact, I had assumed they would be the same value. Now, reading the [relevant portion of the Vulkan spec](https://registry.khronos.org/vulkan/specs/1.3-extensions/html/chap22.html#fxvertex-attribute_divisor), it's clear that they don't even have to be!

cdavis5e (approved):

Issue #794, opened 2019-12-16
Request: Support for VK_EXT_external_memory_host


starbucksDave:
Having dug around the Metal spec a bit, I discovered that it may be possible to implement the VK_EXT_external_memory_host extension in Molten.

This extension essentially allows you to create a piece of memory that points to some arbitrary pointer in system RAM. This is useful for things such as having multiple GPUs read/write to the same buffer that is backed by "imported" or "external" memory on the host, without needing explicit uploads or downloads.

Is this something that is on the roadmap for Molten?

Further reading
https://developer.apple.com/documentation/metal/mtldevice/1433382-makebuffer

starbucksDave:
By the way, I think it is also possible to add similar support for device memory, through these constructs:

https://developer.apple.com/documentation/metal/mtlbuffer/2967416-remotestoragebuffer
https://developer.apple.com/documentation/metal/mtlbuffer/2967415-makeremotebufferview

Though adding support for this might be more complex..

Wunkolo:
Would like to +1 this as well. I use it over on Linux/Windows to avoid having to make redundant copies of memory into a staging buffer.

billhollings:
This enhancement is implement in PR #1854.

PR #1854, opened 2023-03-20
Add support for VK_EXT_external_memory_host extension.

billhollings:
- To expose host memory page size, add `mvkGetHostMemoryPageSize()` and `MVKPhysicalDeviceMetalFeatures::hostMemoryPageSize`.
- `MVKPhysicalDevice::getProperties()` sort EXT property structs alphabetically (unrelated).
- `MVKExtensions.def` fix alignment in OS versions (unrelated).

Implements #794.


MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
```diff
@@ -3469,6 +3483,27 @@ static uint32_t mvkGetEntryProperty(io_registry_entry_t entry, CFStringRef prope
...
+	if (pMemHostPtrProps) {
+		switch (handleType) {
+			case VK_EXTERNAL_MEMORY_HANDLE_TYPE_HOST_ALLOCATION_BIT_EXT:
+				pMemHostPtrProps->memoryTypeBits = _physicalDevice->getHostVisibleMemoryTypes();
+				break;
+			case VK_EXTERNAL_MEMORY_HANDLE_TYPE_HOST_MAPPED_FOREIGN_MEMORY_BIT_EXT:
+				// Vulkan requires only host-visible memory types here, but Metal
+				// only supports Private storage for cross-device buffer sharing.
+				pMemHostPtrProps->memoryTypeBits = 0;
+				break;
```
cdavis5e, line 3498: AFAICT, the docs for `-[MTLDevice newBufferWithBytesNoCopy:length:options:deallocator:]` only state that the memory allocation must belong to a single VM region. I don't see how passing the `-[MTLBuffer contents]` of a foreign device's buffer violates that constraint. What happens when you do that?

Also, I don't recall Metal allowing the direct sharing of `MTLBuffer` objects at all. All of the resource-sharing mechanisms I know of (`MTLSharedTextureHandle`, `IOSurface`) only publicly support `MTLTexture`s. (I know about the secret `MTLDevice` methods for creating `MTLBuffer`s from `IOSurface`s, but they're obviously not supported for public use.) The one thing I think you might be thinking of, I believe is intended to be used within a peer group (i.e. a Vulkan device group).

cdavis5e (changes requested):
Looks good. I just have one objection...

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
billhollings: > I don't see how passing the `-[MTLBuffer contents]` of a foreign device's buffer violates that constraint.

Ah. Good insight! Thanks.

I agree host-mapped memory allocated on another GPU should be fine to import. I've made changes to support `VK_EXTERNAL_MEMORY_HANDLE_TYPE_HOST_MAPPED_FOREIGN_MEMORY_BIT_EXT `.

>  I don't recall Metal allowing the direct sharing of MTLBuffer objects at all.

macOS Metal supports creating a `MTLBuffer` as a view on a foreign `MTLBuffer`, but only for `Private` storage and BLIT operations, so not useful for this extension:

- [-`[MTBuffer newRemoteBufferViewForDevice:]`](https://developer.apple.com/documentation/metal/mtlbuffer/2967415-makeremotebufferview)
- [-`[MTBuffer remoteStorageBuffer]`](https://developer.apple.com/documentation/metal/mtlbuffer/2967416-remotestoragebuffer)

cdavis5e: > macOS Metal supports creating a `MTLBuffer` as a view on a foreign `MTLBuffer`, but only for `Private` storage and BLIT operations, so not useful for this extension:
> 
> * [-`[MTBuffer newRemoteBufferViewForDevice:]`](https://developer.apple.com/documentation/metal/mtlbuffer/2967415-makeremotebufferview)
> * [-`[MTBuffer remoteStorageBuffer]`](https://developer.apple.com/documentation/metal/mtlbuffer/2967416-remotestoragebuffer)

Yep, that's the one I thought you were thinking of. The docs also state that this device and the other device must belong to the same peer group. We'd only need it to implement device groups for real.

cdavis5e (approved):

DUOLabs333:
Am I doing something wrong? I tried enabling this with `ext_names[ext_count++] = VK_KHR_EXTERNAL_MEMORY_EXTENSION_NAME;`, but I get `VK_ERROR_EXTENSION_NOT_PRESENT`.

DUOLabs333:
Am I missing something? I tried enabling it with 
```
ext_names[ext_count++] = VK_EXT_EXTERNAL_MEMORY_HOST_EXTENSION_NAME;
```
but I got `VK_ERROR_EXTENSION_NOT_PRESENT`.

Issue #1838, opened 2023-02-11
Explicit LOD Workaround for Apple silicon regression


italomandara:
I encountered this issue only on Resident Evil 3 Remake so far (using crossover on M1 Max), I have tried few other games but couldn't see any issue so far. This could happen on more games that aren't in my game library, like all re-engine games https://en.wikipedia.org/wiki/RE_Engine.

Probably this issue is related to this commit: 2c01c8b7e066ec9c95c53b4613066e7561a6ac4a

before the commit:
![Before](https://user-images.githubusercontent.com/12135454/218267481-cae8ed1f-4d60-4e5b-aea6-9d3c10fe0c6a.png)

after:
![after](https://user-images.githubusercontent.com/12135454/218267491-676472be-7d39-40d0-a15e-11a29196b5c4.png)

Machine: Apple M1 Max 14" 32GB RAM
os: 13.1 (22C65) Ventura
software: Crossover, Steam, Resident Evil 3 Remake (DX11 version)

italomandara:
This happens also using VKD3D so definitely not a Crossover + DXVK issue

billhollings:
> Probably this issue is related to this commit: [2c01c8b](https://github.com/KhronosGroup/MoltenVK/commit/2c01c8b7e066ec9c95c53b4613066e7561a6ac4a)
> Machine: Apple M1 Max 14" 32GB RAM os: 13.1 (22C65) Ventura software: Crossover, Steam, Resident Evil 3 Remake (DX11 version)

@cdavis5e Are you able to look into this?

cdavis5e:
> @cdavis5e Are you able to look into this?

I don't have access to that game, especially now that I don't work for CW anymore.

@italomandara What happens when you run it with validation enabled?

italomandara:
> > @cdavis5e Are you able to look into this?
> 
> I don't have access to that game, especially now that I don't work for CW anymore.
> 
> @italomandara What happens when you run it with validation enabled?

I'll test and let you know meanwhile can you quickly remind me how to enable the validation please?

cdavis5e:
> I'll test and let you know meanwhile can you quickly remind me how to enable the validation please?

Set the environment variables `METAL_DEVICE_WRAPPER_TYPE=1`, `MTL_DEBUG_LAYER=1`, `MTL_SHADER_VALIDATION=1`.

italomandara:
[...]

PR #1861, opened 2023-04-01
Allow to disable Explicit LOD Workaround for Apple silicon.

italomandara:
This is related to issue #1838.
Since there's a fix being worked, and it could take some time, and probably not a priority for now, it'd be nice to give at least the possibility to disable the workaround via an ENV variable like it's done with other features in the codebase.
I was thinking something like MVK_CONFIG_ENABLE_EXPLICIT_LOD_WORKAROUND=0 to disable it and have it set to 1 by default so that it doesn't change anything compared to the current state of the codebase.

### DIFF
--- a/MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
@@ -1645,7 +1645,7 @@ void MVKPhysicalDevice::initMetalFeatures() {
 			break;
 		case kAppleVendorId:
 			// TODO: Other GPUs?
-			_metalFeatures.needsSampleDrefLodArrayWorkaround = true;
+            MVK_SET_FROM_ENV_OR_BUILD_BOOL(_metalFeatures.needsSampleDrefLodArrayWorkaround, MVK_ENABLE_EXPLICIT_LOD_WORKAROUND);
 			// fallthrough
 		case kIntelVendorId:
 		case kNVVendorId:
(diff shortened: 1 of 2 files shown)


billhollings (changes requested):
Thanks for submitting this fix!

If we are to keep this fix the way it is, I've requested some changes to the documentation and layout.

But the bigger question is whether we really should add a forever config setting to the API, and when it would be employed? Looking at #1838, this appears to be an edge-case that should be resolvable via a fix to SPIRV-Cross, and we haven't finished investigating that yet.

However, I recognize that this is a regression in production, and we need to deal with it immediately. With that in mind, can we internalize the actual tracking variable? Can you rename `MVK_CONFIG_ENABLE_EXPLICIT_LOD_WORKAROUND` to `MVK_ENABLE_EXPLICIT_LOD_WORKAROUND`, and change this line in `MTLDevice.mm`:

	_metalFeatures.needsSampleDrefLodArrayWorkaround = true;

to

	MVK_SET_FROM_ENV_OR_BUILD_BOOL(_metalFeatures.needsSampleDrefLodArrayWorkaround, MVK_ENABLE_EXPLICIT_LOD_WORKAROUND);

This will internalize the variable, and use the build or env var to set it, without changes to the config API. Once the final SPIRV-Cross fix is in, we can simply ignore that env var from then on.

You can then scrap everything about `MVKConfiguration::enableExplicitLodWorkaround` (including the documentation changes I requested in this review).

And, as requested in #1838, can you submit the shader conversions that are breaking in RE3 there, please?

italomandara:
> And, as requested in #1838, can you submit the shader conversions that are breaking in RE3 there, please?
I gave the wrong shader conversion and so I deleted it will add that back soon.

italomandara:
@billhollings I squashed everything in 1 commit for convenience (to revert when the fix is ready)

billhollings (approved):
Thanks. LGTM now.

Issue #1419, opened 2021-08-11
Enhancement assessment: VK_PRIMITIVE_TOPOLOGY_TRIANGLE_FAN


billhollings:
Metal does not support triangle fans (`VK_PRIMITIVE_TOPOLOGY_TRIANGLE_FAN`). This is reported accurately by MoltenVK as part of `VkPhysicalDevicePortabilitySubsetFeaturesKHR` in the `VK_KHR_portability_subset` extension.

However, some apps, including many CTS tests, continue to need triangle fans, which opens the question of whether MoltenVK should attempt to support triangle fans, regardless of Metal support.

Triangle fans can be treated as triangle strips, with a different vertex order. Specifically, a triangle fan can be converted to a triangle strip by reorganizing the vertex indexing using a standard algorithm:

```
    if( vtxIdx % 2 == 0 )
        vtxIdx = vtxIdx / 2;
    else
        vtxIdx = vtxCnt - 1 - vtxIdx / 2;
```

There are a couple of options about where and how this remapping should be done. I've done some initial research into options within MoltenVK, and present them here for discussion. There are no doubt other options, and I may be over-or-under-estimating the complexity of each option in my initial review.

#### Remap Vertices or Vertex Indices Prior to Draw

In this first option, since the vertex data can be updated in the vertex buffer anytime before a draw call is encoded, including via other GPU activity, the remapping of the vertices would need to be done via copying the vertices or vertex indices into a temporary `MTLBuffer` using a Compute Shader encoded just before the draw command.

We could add a `kMVKGraphicsStageVertexRemap` to `MVKGraphicsStage` enumeration, and include it in the results of `MVKGraphicsPipeline::getStages()` when the pipeline requires triangle fan topology. The various `MVKCmdDraw*::encode()` functions could then incorporate this into their stage management operations, similarly to how they incorporate Compute Shaders as part of tessellation operations.

The scheduling of the `kMVKGraphicsStageVertexRemap` would need to take into consideration both tessellated and non-tessellated pipeline handling, and the passing of the temporary vertex buffer in place of the original buffer. This could be quite a complex modification, given the complexity of the existing `MVKCmdDraw*::encode()` functions.

#### Remap Each Vertex in the `vertex` Shader

In this second option, the MSL of the `vertex` shader could be modified to perform the above remapping on each vertex, by modifying the effective value of `glVertexID` before grabbing the vertex data from the array of vertices.

[...]

PR #1864, opened 2023-04-10
WIP: `VK_PRIMITIVE_TOPOLOGY_TRIANGLE_FAN` Support

IsaacMarovitz:
Following suggestions in #1419

### DIFF
--- a/MoltenVK/MoltenVK/Commands/MVKCommandPipelineStateFactoryShaderSource.h
+++ b/MoltenVK/MoltenVK/Commands/MVKCommandPipelineStateFactoryShaderSource.h
@@ -402,5 +402,20 @@ kernel void accumulateOcclusionQueryResults(device VisibilityBuffer& dest [[buff
     if (dest.count < max(oldDestCount, src.count)) { dest.countHigh++; }                                        \n\
 }                                                                                                               \n\
                                                                                                                 \n\
+kernel void remapVertices(device const uint* vertexIndices [[ buffer(0) ]],                                     \n\
+                          device uint* remappedVertices [[ buffer(1) ]],                                        \n\
+                          uint numVertices [[ thread_position_in_grid ]]) {                                     \n\
+    uint vertexCount = numVertices;                                                                             \n\
+    for (uint vertexIndex = 0; vertexIndex < vertexCount; ++vertexIndex) {                                      \n\
+        uint index = vertexIndices[vertexIndex];                                                                \n\
+        if (index % 2 == 0) {                                                                                   \n\
+            index = index / 2;                                                                                  \n\
+        } else {                                                                                                \n\
+            index = vertexCount - 1 - index / 2;                                                                \n\
+        }                                                                                                       \n\
+                                                                                                                \n\
+        remappedVerticies[vertexIndex] = index;                                                                 \n\
+    }                                                                                                           \n\
+}                                                                                                               \n\
 ";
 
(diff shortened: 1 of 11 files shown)


MoltenVK/MoltenVK/Commands/MVKCommandPipelineStateFactoryShaderSource.h
```diff
@@ -402,5 +402,20 @@ kernel void accumulateOcclusionQueryResults(device VisibilityBuffer& dest [[buff
...
+        uint index = vertexIndices[vertexIndex];                                                                \n\
+        if (index % 2 == 0) {                                                                                   \n\
+            index = index / 2;                                                                                  \n\
+        } else {                                                                                                \n\
+            index = vertexCount - 1 - index / 2;                                                                \n\
+        }                                                                                                       \n\
+                                                                                                                \n\
+        remappedVertices[vertexIndex] = index;                                                                  \n\
+    }                                                                                                           \n\
+}                                                                                                               \n\
```
billhollings, line 419: I understand this is a WIP, but I'd like to understand how this shader works.

`thread_position_in_grid` will be a unique index for each instance, and I assume, one for each vertex that is to be remapped? 

I'm not sure you want to be looping here, as the higher instances will tromp on the lower instances. If `thread_position_in_grid` does represent the original index, then the calculation of `index` could just be done once per shader instance.

MoltenVK/MoltenVK/Commands/MVKCommandPipelineStateFactoryShaderSource.h
```diff
@@ -402,5 +402,20 @@ kernel void accumulateOcclusionQueryResults(device VisibilityBuffer& dest [[buff
     if (dest.count < max(oldDestCount, src.count)) { dest.countHigh++; }                                        \n\
 }                                                                                                               \n\
                                                                                                                 \n\
+kernel void remapVertices(device const uint* vertexIndices [[ buffer(0) ]],                                     \n\
+                          device uint* remappedVertices [[ buffer(1) ]],                                        \n\
+                          uint numVertices [[ thread_position_in_grid ]]) {                                     \n\
```
billhollings, line 407: I understand this is a WIP, but I'd like to understand how this shader works.

`thread_position_in_grid` will be a unique index for each instance, and I assume, one for each vertex that is to be remapped? 

I'm not sure you want to be looping here, as the higher instances will tromp on the lower instances. If `thread_position_in_grid` does represent the original index, then the calculation of `index` could just be done once per shader instance.

IsaacMarovitz: To be entirely honest once I started working on this PR I quickly realised I was out of my depth. 😅 I may just close this PR as I doubt I can finish it without hours of digging and probably countless questions.

rcaridade145: @IsaacMarovitz can the code from rpcs3 - https://github.com/RPCS3/rpcs3/blob/master/rpcs3/Emu/RSX/Common/BufferUtils.cpp#L464 - or xenia - https://github.com/xenia-project/xenia/blob/master/src/xenia/gpu/primitive_processor.cc - help you?

billhollings: > I may just close this PR

Leave it open as WIP. You have some great ideas here, and it's a good start! You can return to it later, or I or someone else could pick it up from here. So thanks very much for your contribution, and for at the very least kicking this off!

With that in mind, if anyone is going to work on this, just post here that you are, so that we can all coordinate, and not duplicate efforts.

IsaacMarovitz: > @IsaacMarovitz can the code from rpcs3 - https://github.com/RPCS3/rpcs3/blob/master/rpcs3/Emu/RSX/Common/BufferUtils.cpp#L464 - or xenia - https://github.com/xenia-project/xenia/blob/master/src/xenia/gpu/primitive_processor.cc - help you?

Thanks for the resources I'll take a look through and see if I can gleam anything useful 😄

Issue #1846, opened 2023-03-06
Update buffer data and use buffer device address in shader


maksym-pasichnyk:
I'm trying to update buffer by mapping HostVisible + HostCoherent memory and updating via memcpy, but data inside shader is wrong.

```c++
auto vertices = std::array{
    Vec2<f32>{-1.0F, -1.0F},
    Vec2<f32>{+1.0F, -1.0F},
    Vec2<f32>{-1.0F, +1.0F},
    Vec2<f32>{+1.0F, +1.0F},
};

auto indices = std::array{
    0u, 1u, 2u,
    2u, 1u, 3u
};

GpuBufferInfo index_buffer_info{};
GpuBufferInfo vertex_buffer_info{};

vulkan->AllocateTemporary(&index_buffer_info, sizeof(indices), alignof(u32));
vulkan->AllocateTemporary(&vertex_buffer_info, sizeof(vertices), alignof(Vec2<f32>));
```

Code below not working
```c++
// This code not working
auto memory = vulkan->context.frame_allocators[vulkan->frame_index].allocation.device_memory;
void* mapped = vulkan->context.logical_device.mapMemory(memory, 0, VK_WHOLE_SIZE, vk::MemoryMapFlags());
std::memcpy(reinterpret_cast<u8*>(mapped) + index_buffer_info.offset, indices.data(), sizeof(indices));
std::memcpy(reinterpret_cast<u8*>(mapped) + vertex_buffer_info.offset, vertices.data(), sizeof(vertices));
vulkan->context.logical_device.unmapMemory(memory);
```

If I add line below before mapMemory or after unmapMemory, everything working fine
```c++
u8 tmp = 0;
cmd.updateBuffer(vulkan->context.frame_allocators[vulkan->frame_index].buffer, 0, sizeof(u32), &tmp);
```

Also when I updating buffers with cmd.updateBuffer, everything is working fine
```c++
auto buffer = vulkan->context.frame_allocators[vulkan->frame_index].buffer;
cmd.updateBuffer(buffer, index_buffer_info.offset, sizeof(indices), indices.data());
cmd.updateBuffer(buffer, vertex_buffer_info.offset, sizeof(vertices), vertices.data());
```

billhollings:
Are you sure the device memory is host coherent? It seems to be behaving as if the device memory isn't host coherent. 

Is this working on other Vulkan platforms?

maksym-pasichnyk:
[...]

Issue #1597, opened 2022-05-23
MoltenVK shouldn't check every resource on a barrier


cdavis5e:
Another major source of slowdowns on MoltenVK is that, on every barrier, it dutifully checks every outstanding resource to see if it needs to be synchronized. The slowdown is bad enough that, in CrossOver's Final Fantasy XIV port, the checks are entirely disabled. We haven't seen any obvious problems with this, but that's only because Metal does some of its own synchronization. It is possible to disable this implicit synchronization--something MoltenVK should probably be doing--using the `MTLResourceHazardTrackingModeUntracked` option. But before we can do that, we'll need to do something about the horrendous performance of barriers. It occurs to me that, instead of checking *every* resource, MoltenVK should only be checking resources that may have been written to, i.e. non-input attachments, storage images, and storage buffers. It shouldn't be too much trouble to keep track of those resources on a command buffer.

@js6i This might be something to look at after #1594 is merged.

PR #1868, opened 2023-04-14
Ensure shaders using PhysicalStorageBufferAddresses encode the associated MTLBuffer.

billhollings:
- `MVKDevice` track `VkBuffers` marked with `VK_BUFFER_USAGE_SHADER_DEVICE_ADDRESS_BIT`.
- Add `SPIRVToMSLConversionResultInfo::usesPhysicalStorageBufferAddressesCapability` to detect and track shaders that use `PhysicalStorageBufferAddresses` capability, and track such shader stages within pipeline.
- `MVKResourcesCommandEncoderState` encode usage of `VkBuffers` marked with `VK_BUFFER_USAGE_SHADER_DEVICE_ADDRESS_BIT` when pipeline uses `PhysicalStorageBufferAddresses` capability.
- Rename `MVKResourcesCommandEncoderState::encodeArgumentBufferResourceUsage()` to `encodeResourceUsage()`.
- `MVKDevice` move some functions to public scope and remove friend classes.
- `MVKDeviceMemory` ensure `_vkMemAllocFlags` is always initialized (unrelated).
- Rename `MVKFoundation` template method `contains()` to `mvkContains()` (unrelated).

Fixes issue #1846.

### DIFF
--- a/MoltenVK/MoltenVK/Commands/MVKCommandEncoderState.h
+++ b/MoltenVK/MoltenVK/Commands/MVKCommandEncoderState.h
@@ -355,11 +355,11 @@ public:
 						   MVKArrayRef<uint32_t> dynamicOffsets,
 						   uint32_t& dynamicOffsetIndex);
 
-	/** Encodes the Metal resource to the Metal command encoder. */
-	virtual void encodeArgumentBufferResourceUsage(MVKShaderStage stage,
-												   id<MTLResource> mtlResource,
-												   MTLResourceUsage mtlUsage,
-												   MTLRenderStages mtlStages) = 0;
+	/** Encodes the indirect use of the Metal resource to the Metal command encoder. */
+	virtual void encodeResourceUsage(MVKShaderStage stage,
+									 id<MTLResource> mtlResource,
+									 MTLResourceUsage mtlUsage,
+									 MTLRenderStages mtlStages) = 0;
 
 	void markDirty() override;
 
@@ -548,10 +548,10 @@ public:
                         std::function<void(MVKCommandEncoder*, MVKMTLTextureBinding&)> bindTexture,
                         std::function<void(MVKCommandEncoder*, MVKMTLSamplerStateBinding&)> bindSampler);
 
-	void encodeArgumentBufferResourceUsage(MVKShaderStage stage,
-										   id<MTLResource> mtlResource,
-										   MTLResourceUsage mtlUsage,
-										   MTLRenderStages mtlStages) override;
+	void encodeResourceUsage(MVKShaderStage stage,
+							 id<MTLResource> mtlResource,
+							 MTLResourceUsage mtlUsage,
+							 MTLRenderStages mtlStages) override;
 
 	/** Offset all buffers for vertex attribute bindings with zero divisors by the given number of strides. */
 	void offsetZeroDivisorVertexBuffers(MVKGraphicsStage stage, MVKGraphicsPipeline* pipeline, uint32_t firstInstance);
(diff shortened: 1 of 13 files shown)


MoltenVK/MoltenVK/Commands/MVKCommandEncoderState.h
```diff
@@ -355,11 +355,11 @@ class MVKResourcesCommandEncoderState : public MVKCommandEncoderState {
 						   MVKArrayRef<uint32_t> dynamicOffsets,
 						   uint32_t& dynamicOffsetIndex);
 
-	/** Encodes the Metal resource to the Metal command encoder. */
-	virtual void encodeArgumentBufferResourceUsage(MVKShaderStage stage,
-												   id<MTLResource> mtlResource,
-												   MTLResourceUsage mtlUsage,
-												   MTLRenderStages mtlStages) = 0;
+	/** Encodes the indirect use of the Metal resource to the Metal command encoder. */
```
cdavis5e, line 358: Note that this must *also* be done for resources with `MTLHazardTrackingModeUntracked`, although we don't use that... [yet](https://github.com/KhronosGroup/MoltenVK/issues/1597) 😉 .

MoltenVK/MoltenVK/Commands/MVKCommandEncoderState.mm
```diff
@@ -693,6 +693,11 @@
 
 	encodeMetalArgumentBuffer(stage);
 
+	MVKPipeline* pipeline = getPipeline();
+	if (pipeline && pipeline->usesPhysicalStorageBufferAddressesCapability(stage)) {
+		getDevice()->encodeGPUAddressableBuffers(this, stage);
```
cdavis5e, line 698: I wonder if, eventually, we'll want to do some sort of GC-like tracing and only encode those buffers that are reachable from buffers that were directly passed--see (again) #1597 for why that's desirable. On the other hand, that might be even more difficult, given that it's difficult to tell if a write contained a buffer address...

In particular, I could see this horribly affecting performance in vkd3d, both due to encoding overhead and also due to memory pressure from keeping so many buffers in VRAM at once, since nearly every buffer it creates is going to have the `SHADER_DEVICE_ADDRESS` bit set.

cdavis5e (approved):

MoltenVK/MoltenVK/Commands/MVKCommandEncoderState.h
billhollings: Yes. #1597 would be good. Thanks for the reminder. I've moved that higher on my list of ToDos. 

It's too bad GitHub doesn't have an issue priority system to keep things high on the list. I guess we could fake it with labels. P1, P2, etc.

MoltenVK/MoltenVK/Commands/MVKCommandEncoderState.mm
billhollings: > GC-like tracing

A good idea, but I don't know if we could do that effectively. 

Since an app can pass an arbitrary `gpuAddress` to a shader, it can access any active `VK_BUFFER_USAGE_SHADER_DEVICE_ADDRESS_BIT` buffer, which is what I assume here. 

Similarly, with descriptor-indexing, a shader could be commanded to arbitrarily access any of hundreds of thousands of resources in a huge descriptor-indexed Metal argument buffer, some based on indexes passed in via other buffers, some based on indexes derived from calculations within the shader itself.

But...this is indeed a real bottleneck issue with Metal, and I'm open to any suggestions! 

I was contemplating that `MTLHeaps` might help with streamlining encoding the resource usage, but it would be very hard to map arbitrary resources in an argument buffer to heaps that could be effectively encoded to make the contained resources resident.

Issue #1871, opened 2023-04-18
`VK_ERROR_FEATURE_NOT_PRESENT` in `MVKDevice::enableFeatures` does not actually print the missing feature


IsaacMarovitz:
I'm currently trying my luck at getting Doom (2016) running and quickly ran into a slight issue, specifically with how this error is declared in `MVKDevice`

```obj-c++
void MVKDevice::enableFeatures(VkBool32* pEnabledBools, const VkBool32* pRequestedBools, const VkBool32* pAvailableBools, uint32_t count) {
	for (uint32_t i = 0; i < count; i++) {
		pEnabledBools[i] = pRequestedBools[i] && pAvailableBools[i];
		if (pRequestedBools[i] && !pAvailableBools[i]) {
			setConfigurationResult(reportError(VK_ERROR_FEATURE_NOT_PRESENT, "vkCreateDevice(): Requested feature is not available on this device."));
		}
	}
}
```

This is the only usage of `VK_ERROR_FEATURE_NOT_PRESENT` in MoltenVK that does not actually log the requested feature that is missing and it means I don't actually know what unsupported feature Doom is trying to use.

billhollings:
PR #1905 fixes this issue, and identifies the struct name and member index of the improperly enabled feature flag. Please retest and close this issue if it fixes the problem.

PR #1873, opened 2023-04-19
Better `VK_ERROR_FEATURE_NOT_PRESENT` logging

IsaacMarovitz:
Naïve fix for #1871, definitely a better way to do this but this mostly gets the job done.

### DIFF
--- a/MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
@@ -4575,7 +4575,8 @@ void MVKDevice::enableFeatures(const VkDeviceCreateInfo* pCreateInfo) {
 	if (pCreateInfo->pEnabledFeatures) {
 		enableFeatures(&_enabledFeatures.robustBufferAccess,
 					   &pCreateInfo->pEnabledFeatures->robustBufferAccess,
-					   &pdFeats2.features.robustBufferAccess, 55);
+					   &pdFeats2.features.robustBufferAccess, 55,
+                       (char*)"Robust Buffer Access");
 	}
 
 	for (const auto* next = (VkBaseInStructure*)pCreateInfo->pNext; next; next = next->pNext) {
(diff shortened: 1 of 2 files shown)


MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
```diff
@@ -4694,14 +4719,14 @@ static uint32_t mvkGetEntryProperty(io_registry_entry_t entry, CFStringRef prope
 	enableFeatures((VkBool32*)(&(pEnabled->pNext) + 1),
 				   (VkBool32*)(&(pRequested->pNext) + 1),
 				   (VkBool32*)(&(pAvailable->pNext) + 1),
-				   count);
+				   count, (char*)"");
```
IsaacMarovitz, line 4722: Not sure what do to for this one

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
```diff
@@ -4575,7 +4575,8 @@ static uint32_t mvkGetEntryProperty(io_registry_entry_t entry, CFStringRef prope
 	if (pCreateInfo->pEnabledFeatures) {
 		enableFeatures(&_enabledFeatures.robustBufferAccess,
 					   &pCreateInfo->pEnabledFeatures->robustBufferAccess,
-					   &pdFeats2.features.robustBufferAccess, 55);
+					   &pdFeats2.features.robustBufferAccess, 55,
+                       (char*)"Robust Buffer Access");
```
cdavis5e, line 4579: I was about to reply to your bug and express skepticism at the possibility of fixing it given how the code is structured, but then I realized how to do it.

The `enableFeatures()` function takes an *array* of booleans to check. Its callers reinterpret the feature structs as arrays, relying on the compiler laying out the `VkBool32` members end to end with no padding, as in an array. (This among other things means that [the feature Doom 2016 needs](https://github.com/KhronosGroup/MoltenVK/issues/1872) is not necessarily `robustBufferAccess`, though it is somewhere in this struct.) If you want to print the name of the feature(s) which caused the check to fail, you'll need to pass an array of `char *` or `string`. Each member of this array should correspond to a member of the struct being checked.

billhollings: Yes. Ultimately, we'd need a way to stringify the members of each of those structs. 

Unfortunately, C++ doesn't offer any reasonable way to do this, and I expect we'd have to manually type out a static array of c-strings containing the names of all of the struct members, for each struct that we want to submit to `enableFeatures()`.

PR #1881, opened 2023-05-01
Deprecate the obsolete and non-standard VK_MVK_moltenvk extension.

billhollings:
The `VK_MVK_moltenvk` extension has never been brought inside Vulkan, and the functions have never been supported by the Vulkan Loader and Layers. Most of the functionality has long been replaced by the official `VK_metal_objects` extension.

- Remove `VK_MVK_moltenvk` as an advertised extension.
- Refactor `vk_mvk_moltenvk.h` header file into separate headers files: 
    - `mvk_config.h` - Valid public config functions 
    - `mvk_private_api.h` - Valid development debugging functions used with care 
    - `mvk_deprecated_api.h` - Formally deprecated functions.
- Retain skeleton `vk_mvk_moltenvk.h` header file for legacy compatibility only.
- Update documentation and header comments to explain changes.


richard-lunarg (commented):
Users still need vkGetMoltenVKConfigurationMVK until we get the replacement in, correct? If so, how do they get the pointer to that function now, when linking statically? Without the extension exposed, I can't seem to get a pointer to it any longer.

billhollings:
> Users still need vkGetMoltenVKConfigurationMVK until we get the replacement in, correct? If so, how do they get the pointer to that function now, when linking statically? Without the extension exposed, I can't seem to get a pointer to it any longer.

Uggh. Thanks for catching this. I've updated this PR. Can you try again, please?

richard-lunarg:
The instance extension VK_MVK_moltenvk is still not listed (not sure if you were wanting to roll that back in or not). I still cannot get a pointer to vkGetMoltenVKConfigurationMVK, with or without an instance pointer using vkGetInstanceProcAddr() when linking statically.

richard-lunarg:
Okay, user error... too many copies of MoltenVK on my system<g>. Latest test shows this is working now with your changes. The extension shows up, and I can get the pointer as needed. I think you're good to go now.

billhollings:
> Okay, user error... too many copies of MoltenVK on my system. Latest test shows this is working now with your changes. The extension shows up, and I can get the pointer as needed. I think you're good to go now.

Thanks for retesting. 

One question...the function pointers should indeed work...but the `VK_MVK_moltenvk` should NOT be showing up in the advertised list of extensions. It's been removed. Are you sure you're still seeing it?

richard-lunarg:
>.but the VK_MVK_moltenvk should NOT be showing up in the advertised list of extensions. 
>It's been removed. Are you sure you're still seeing it?

Ah, you are correct, it is indeed missing, but I can still get the function pointer.

PR #1897, opened 2023-05-11
Update release action artifact glob.

Steveice10:
Simple update to the release workflow for the build CI updates. Since we have ```MoltenVK-all```, ```MoltenVK-macos```, and ```MoltenVK-ios``` artifacts in the build workflow, we can capture all of these and upload their artifacts with the glob pattern ```MoltenVK*/*```.

Resulting release: https://github.com/Steveice10/MoltenVK/releases/tag/v1.2.4

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -125,5 +125,5 @@ jobs:
       - name: Create Release
         uses: ncipollo/release-action@v1
         with:
-          artifacts: "MoltenVK/*"
+          artifacts: "MoltenVK*/*"
           artifactErrorsFailBuild: true


billhollings:
@Steveice10 

It looks like we're hitting a [syntax error](https://github.com/KhronosGroup/MoltenVK/actions/runs/4950913582/jobs/8857136749) on the artifact upload for a [Release](https://github.com/KhronosGroup/MoltenVK/releases/tag/test-release-CI-3) on the MoltenVK repo. 

Any thoughts?

Steveice10:
> @Steveice10 
> 
> 
> 
> It looks like we're hitting a [syntax error](https://github.com/KhronosGroup/MoltenVK/actions/runs/4950913582/jobs/8857136749) on the artifact upload for a [Release](https://github.com/KhronosGroup/MoltenVK/releases/tag/test-release-CI-3) on the MoltenVK repo. 
> 
> 
> 
> Any thoughts?

I think what it's saying is not a syntax error but that the release already exists. If you're cutting a tag by creating the release in GitHub that might be why.

There's an option you can turn on, allowUpdates, that would enable this approach by allowing the workflow to update an existing release. By default it does not allow that.

Steveice10:
This should help loosen the strictness a bit: https://github.com/KhronosGroup/MoltenVK/pull/1900

PR #1898, opened 2023-05-12
Respect the bind point supplied to vkCmdBindDescriptorSets / vkCmdPushDescriptorSets

etang-cw:
Prevents calls to `vkCmdPushDescriptorSetKHR` with layouts that specify `VK_SHADER_STAGE_ALL` from applying to things they shouldn't (e.g. having `VK_PIPELINE_BIND_POINT_COMPUTE` binds trash the graphics bindings)

### DIFF
--- a/MoltenVK/MoltenVK/GPUObjects/MVKDescriptor.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKDescriptor.mm
@@ -20,6 +20,16 @@
 #include "MVKDescriptorSet.h"
 #include "MVKBuffer.h"
 
+#define BIND_GRAPHICS_OR_COMPUTE(cmdEncoder, bind, pipelineBindPoint, stage, ...) \
+	do { \
+		if ((stage) == kMVKShaderStageCompute) { \
+			if ((cmdEncoder) && (pipelineBindPoint) == VK_PIPELINE_BIND_POINT_COMPUTE) \
+				(cmdEncoder)->_computeResourcesState.bind(__VA_ARGS__); \
+		} else { \
+			if ((cmdEncoder) && (pipelineBindPoint) == VK_PIPELINE_BIND_POINT_GRAPHICS) \
+				(cmdEncoder)->_graphicsResourcesState.bind(static_cast<MVKShaderStage>(stage), __VA_ARGS__); \
+		} \
+	} while (0)
 
 #pragma mark MVKShaderStageResourceBinding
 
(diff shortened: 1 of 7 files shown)


MoltenVK/MoltenVK/GPUObjects/MVKDescriptor.mm
```diff
@@ -272,9 +273,11 @@ void mvkPopulateShaderConversionConfig(mvk::SPIRVToMSLConversionConfiguration& s
                     if (_applyToStage[i]) {
                         bb.index = mtlIdxs.stages[i].bufferIndex + rezIdx;
                         if (i == kMVKShaderStageCompute) {
-							if (cmdEncoder) { cmdEncoder->_computeResourcesState.bindBuffer(bb); }
+							if (cmdEncoder && pipelineBindPoint == VK_PIPELINE_BIND_POINT_COMPUTE)
+								cmdEncoder->_computeResourcesState.bindBuffer(bb);
                         } else {
-							if (cmdEncoder) { cmdEncoder->_graphicsResourcesState.bindBuffer(MVKShaderStage(i), bb); }
+							if (cmdEncoder && pipelineBindPoint == VK_PIPELINE_BIND_POINT_GRAPHICS)
+								cmdEncoder->_graphicsResourcesState.bindBuffer(MVKShaderStage(i), bb);
```
billhollings, line 280: Is there a reason why you applied this only to the `UNIFORM_BUFFER` and `STORAGE_BUFFER` descriptor types?  Presumably this also applies to all the other descriptor types?

And should this change also apply to similar functionality in subclasses of `MVKDescriptor`, to cover the bind point included in `vkCmdBindDescriptorSets()`?

If yes to these questions, rather than copying and pasting this change everywhere, I think we should take the repeating code block:

```
if (i == kMVKShaderStageCompute) {
	if (cmdEncoder) { cmdEncoder->_computeResourcesState.bindXXX(bb); }
} else {
	if (cmdEncoder) { cmdEncoder->_graphicsResourcesState.bindXXX(MVKShaderStage(i), bb); }
}
```

that appears everywhere, and refactor it into either a template function, or macro definition (to handle the `bindXXX` part).

If this is the case, we can bring this PR in as-is, to get it functional, and I can take care of the generalized refactoring later.

billhollings (commented):
Thanks for you submission!

Before pulling this in, I'd like to get the questions below answered, regarding expanding this to other use cases later.

MoltenVK/MoltenVK/GPUObjects/MVKDescriptor.mm
etang-cw: 1. Because I forgot the rest existed, I'll get that fixed
2. I think `vkCmdBindDescriptorSets` is already handled by https://github.com/KhronosGroup/MoltenVK/blob/e50cb32127ea8f1e784098ef4fc5b3477d4b7f6d/MoltenVK/MoltenVK/Commands/MVKCommandBuffer.mm#L674-L694 unless you can think of another case that needs this

etang-cw: Ahh never mind, that only binds the descriptor set itself, not the things inside of it

billhollings (approved):
LGTM now. Thanks!

PR #1903, opened 2023-05-12
Fix simulator feature discovery

ChibiDenDen:
Fix depth clamp and texture swizzle feature discovery on simulator builds

Both of these features are not supported by the iPhone simulator and leads to crashes when used


billhollings (approved):
Thanks for catching this, and for the fix!

Issue #1880, opened 2023-04-28
Registry ID not constant across reboots


rcombs:
I was just reading through the code implementing `VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_ID_PROPERTIES`, and noticed that MoltenVK currently populates the `deviceUUID` field with:
- The vendor ID (PCIID-style)
- The device ID (PCIID or `getHighestGPUCapability`)
- The registry ID

This last part is seemingly a problem: in my testing, the registry ID of the internal GPU on the M1 Max on my 2021 16-inch MacBook Pro appears to vary across reboots (or at least across OS versions), which is not allowed in `deviceUUID` by the Vulkan spec.

In my own code, I'm solving this problem by replacing the registry ID with the `location` (builtin vs slot vs external) and `locationNumber` (integrated vs discrete for builtin, slot number for slot, port number for external). I believe this still satisfies the uniqueness requirements for the `deviceUUID` field, while also providing a stronger stability guarantee.

Making this change in MoltenVK would be a one-time violation of the spec (since the ID must be stable across driver versions), but _not_ making it is a continuous violation (since the ID must be stable across reboots).

PR #1908, opened 2023-05-16
Populate deviceUUID from MTLDevice location and peer group info.

billhollings:
`MTLDevice registryID` is not constant across OS reboots, which is not conformant with `deviceUUID` requirements.

Replace with combination of `MTLDevice` `location`, `locationNumber`, `peerGroupID`, and `peerIndex`, which should define uniqueness, and **_should_** be constant across OS reboots.

I've included device peer group because `locationNumber` is [not unique](https://developer.apple.com/documentation/metal/mtldevice/3114006-locationnumber?language=objc) in the case of slots and external ports that support multiple GPUs. However, I am not 100% certain that the peer group info is constant across a reboot, although it will be for cases where a peer group is not defined.

Fixes issue #1880.

### DIFF
--- a/MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
@@ -727,14 +727,20 @@ void MVKPhysicalDevice::getProperties(VkPhysicalDeviceProperties2* properties) {
 	}
 }
 
+// Since these are uint8_t arrays, use Big-Endian byte ordering,
+// so a hex dump of the array is human readable in its parts.
 void MVKPhysicalDevice::populateDeviceIDProperties(VkPhysicalDeviceVulkan11Properties* pVk11Props) {
 	uint8_t* uuid;
 	size_t uuidComponentOffset;
 
-	//  ---- Device ID ----------------------------------------------
+	//  ---- Device UUID ----------------------------------------------
 	uuid = pVk11Props->deviceUUID;
 	uuidComponentOffset = 0;
 	mvkClear(uuid, VK_UUID_SIZE);
+
+	// From Vulkan spec: deviceUUID must be universally unique for the device,
+	// AND must be immutable for a given device across instances, processes,
+	// driver APIs, driver versions, and system reboots.
 
 	// First 4 bytes contains GPU vendor ID
 	uint32_t vendorID = _properties.vendorID;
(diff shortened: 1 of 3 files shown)


rcombs:
I'm unfortunately not sure whether or not peer group IDs are stable across reboots; I don't have access to a system that has peer groups. I'd assume that peer _index_ would be stable, but for all I know, group IDs might be unpredictable large values like registry IDs (them being conspicuously 64-bit has me a little worried).

I'd missed that locationNumber wasn't necessarily unique; what I'm doing in cross-platform code is keeping track of a mapping from GPU ID strings (generated from whatever information is available on a platform) to an integer value, which I increment each time that ID is seen; I then append that to the ID. You could probably allocate a couple bytes to something similar. This means that two GPUs with the same observed location+number+etc would always get unique IDs; they wouldn't necessarily be stable across reboots (it'd depend on device enumeration order), but usually in that sort of situation you have 2 identical devices anyway, so selection between them isn't especially important.

If you wanted to provide a stronger guarantee that a given ID is guaranteed to refer to at least the same device _model_ (or vendor, at the very least), I suppose it'd be possible to hash in the `name` field (is this completely stable? is it localized? the documentation is unclear); alternately, what I do is look up the device in the IORegistry tree using `IORegistryEntryIDMatching` + `IOServiceGetMatchingService`, then get its parent (`IORegistryEntryGetParentEntry`), then try to read `vendor-id`, `device-id`, `subsystem-vendor-id`, and `subsystem-id` properties (`IORegistryEntryCreateCFProperty`). The `vendor-id` is sometimes available on the device itself (not the parent); barring that, `IOObjectGetClass` gives a string that at least identifies the underlying driver and vendor.

cdavis5e:
> I'm unfortunately not sure whether or not peer group IDs are stable across reboots; I don't have access to a system that has peer groups.

I *do* have such a system--in fact, purchased one in part for the *express purpose* of exploring peer groups as a potential implementation of the `VK_KHR_device_group` extension's functionality--and I have verified that peer groups *are* in fact stable across reboots, at least, in practice.

As for `registryID`... I wonder if we couldn't use it to set the `deviceLUID` property. It serves largely the same purpose that an LUID does on Windows: it's a locally unique ID used to identify the device across APIs for the purpose of sharing resources.

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
```diff
@@ -746,10 +750,25 @@
...
+
+	// Next 2 bytes contains GPU location number
+	uint16_t dvcLocNum = mvkGetLocationNumber(_mtlDevice);
+	*(uint16_t*)&uuid[uuidComponentOffset] = NSSwapHostShortToBig(dvcLocNum);
+	uuidComponentOffset += sizeof(deviceID);
+
+	// Next 2 bytes contains GPU peer group ID
+	uint16_t pgID = mvkGetPeerGroupID(_mtlDevice);
+	*(uint16_t*)&uuid[uuidComponentOffset] = NSSwapHostShortToBig(pgID);
+	uuidComponentOffset += sizeof(deviceID);
```
cdavis5e, line 766: From what I've seen, all eight bytes of the peer group ID are significant.

cdavis5e (changes requested):

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
billhollings: Fixed by hashing this as `uint64_t` into a final `uint64_t` ID that includes the other properties too.

billhollings:
> vendor-id, device-id

These are already included in the first part of `deviceUUID`.

> I *do* have such a system--in fact, purchased one in part for the *express purpose* of exploring peer groups as a potential implementation of the `VK_KHR_device_group` extension's functionality--and I have verified that peer groups *are* in fact stable across reboots, at least, in practice.

> As for `registryID`... I wonder if we couldn't use it to set the `deviceLUID` property. It serves largely the same purpose that an LUID does on Windows: it's a locally unique ID used to identify the device across APIs for the purpose of sharing resources.

Ah. Thanks for the research. Nice that you have a larger system to test on.

I've updated this PR to build one `uint64_t` hash from the four `MTLDevice` properties, to allow `peerGroupID` to contribute all of its bytes. I've also added support for `deviceLUID` per your suggestion.

cdavis5e (approved):

PR #1912, opened 2023-05-17
WIP: Manual Frame Capture

IsaacMarovitz:
**What this adds:**
This PR allows you to trigger a manual frame capture of any length. This capture is started and stopped with the shortcut `CMD + Shift + C`, and saves to the path set in the environment variable `MVK_CONFIG_GPU_CAPTURE_OUTPUT_FILE`. If this environment variable is not set, the keyboard event listener will not be started.

This feature requires both macOS 10.15 or newer and for the GPU capture environment variable to be enabled.

**Why?**
This feature has the potential to be incredibly useful for those debugging games running under Wine, which often do not cooperate well with Xcode, and want the flexibility of starting a frame capture at any point during gameplay.

`Debug Tools` commit was working, my attempt to clean it up introduced segfaults not entirely sure they're coming from I think the added environment variable definition might be to blame.

### DIFF
--- a/MoltenVK/MoltenVK/API/mvk_config.h
+++ b/MoltenVK/MoltenVK/API/mvk_config.h
@@ -707,6 +707,27 @@ typedef struct {
 	 */
 	const char* autoGPUCaptureOutputFilepath;
 
+    /**
+     * The path to a file where the manual GPU capture should be saved.
+     * In this case, the Xcode scheme must have Metal GPU capture enabled, but does
+     * not need not be run under Xcode's control at all. This is useful in case the app
+     * cannot be run under Xcode's control. A path starting with '~' can be used to place it in a
+     * user's home directory, as in the shell. This feature requires Metal 3.0 (macOS 10.15).
+     *
+     * If this parameter is NULL or an empty string, the capture will not be started.
+     *
+     * Capture will commence with the shortcut CMD + Shift + C, and will stop
+     * when the shortcut is entered a second time.
+     *
+     * The value of this parameter must be changed before creating a VkDevice,
+     * for the change to take effect.
+     *
+     * The initial value or this parameter is set by the
+     * MVK_CONFIG_GPU_CAPTURE_OUTPUT_FILE
+     * runtime environment variable or MoltenVK compile-time build setting.
+     */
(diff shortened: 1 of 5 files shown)


billhollings (changes requested):
I like this idea! It should prove very helpful at runtimes. Thanks for thinking of it and working on it!

I made some inline requests to refactor this to integrate this a bit better with existing behaviour, and use existing configs. Let me know what your thoughts are about that.

MoltenVK/MoltenVK/API/mvk_config.h
```diff
@@ -707,6 +707,27 @@ typedef struct {
...
+     * when the shortcut is entered a second time.
+     *
+     * The value of this parameter must be changed before creating a VkDevice,
+     * for the change to take effect.
+     *
+     * The initial value or this parameter is set by the
+     * MVK_CONFIG_GPU_CAPTURE_OUTPUT_FILE
+     * runtime environment variable or MoltenVK compile-time build setting.
+     */
+    const char* gpuCaptureOutputFilepath;
```
billhollings, line 729: Instead of a different file path, I'd like to reuse `autoGPUCaptureOutputFilepath`, combined with an additional enum in `MVKConfigAutoGPUCaptureScope`. Something like `MVK_CONFIG_AUTO_GPU_CAPTURE_SCOPE_FRAME_USER_TRIGGERED`. You should then be able to reuse much of the existing scaffolding that exists for the other auto scopes.

Regarding names, we can either think of the "auto" component of the member and enum names as just meaning ("Xcode not required"), even though some of them might involve user interaction.

Or we can rename the members and enums to remove the "auto" parts, while still retaining backwards compatibility. See how `semaphoreUseMTLEvent` was renamed to `semaphoreSupportStyle`, and how Vulkan renames structs and enums.

IsaacMarovitz:
> I like this idea! It should prove very helpful at runtimes. Thanks for thinking of it and working on it!
> 
> I made some inline requests to refactor this to integrate this a bit better with existing behaviour, and use existing configs. Let me know what your thoughts are about that.

Thanks for the review Bill! I'll work on integrating your changes. There's a segfault somewhere I think maybe related to the env var I added so I'll try to sus that out.

Issue #1870, opened 2023-04-17
Invalid Resource errors when storage buffers are bound across stages (w/ Argument Buffers)


riperiperi:
When running on mac with MoltenVK, Ryujinx uses the metal argument buffers config option to take advantage of higher limits for descriptor sets. However, there seems to be an issue where binding usage flags are recorded multiple times and the resource can be left with the incorrect usage flags on use, which causes this error:

```
[mvk-error] VK_ERROR_DEVICE_LOST: MTLCommandBuffer "vkQueueSubmit CommandBuffer on Queue 0-0" execution failed (code 9): Invalid Resource (00000009:kIOGPUCommandBufferCallbackErrorInvalidResource)
```

An example would be:
- Game draws with a program, storage buffer "sb" is bound in the vertex stage.
- ... other draws...
- Switch program, storage buffer "sb" is bound in the fragment stage, attempt to draw
- Error during execution

Switching the stage flags on the storage descriptor set bindings to match all stages does help a lot of these cases, but it's not a perfect solution. This issue also seems to extend to cases with rasterizer discard, as the buffer is only bound for the vertex stage in that case, and this seems to force the buffer usage to only be recorded for vertex. Typically for us, this happens when games use transform feedback, which we emulate by using rasterizer discard and storage buffer writes.  We could avoid the resource issue in this case by detecting this case and splitting the render pass, but there is a lot of overhead in that, and weird to have in our Vulkan backend.

In Ryujinx, buffers that are contiguous in memory are often combined, so it's pretty common for a buffer to be used in more than one situation (with different sub-range), including across stages.

I noticed last year and planned to create some kind of way to reproduce this, but haven't been able to find the time and figured I should just report what I saw when using Ryujinx. That does mean this is more of an observation rather than knowing exactly what's going wrong here.

billhollings:
> Switching the stage flags on the storage descriptor set bindings to match all stages does help a lot of these cases, but it's not a perfect solution. This issue also seems to extend to cases with rasterizer discard, as the buffer is only bound for the vertex stage in that case, and this seems to force the buffer usage to only be recorded for vertex. Typically for us, this happens when games use transform feedback, which we emulate by using rasterizer discard and storage buffer writes.

I'm trying to understand the overall flow in the situation with rasterization disabled. 

[...]

Issue #1811, opened 2023-01-07
VK_ERROR_DEVICE_LOST in MacOS 12.6.1 Apple M1, but identical app works fine in MacOS 10.14.6


elmar-k:
Dear MoltenVK-Team,

if I copy our Intel x86_64 app from a MacBook Pro (Late 2013) with MacOS 10.14.6 Mojave (where it runs fine) to an Apple M1 MacBook Air (2020) with MacOS 12.6.1 Monterey, I get the following error in the terminal:

```
[mvk-error] VK_ERROR_DEVICE_LOST: MTLCommandBuffer "vkQueueSubmit CommandBuffer on Queue 0-0
" execution failed (code 9): Invalid Resource (00000009:kIOGPUCommandBufferCallbackErrorInva
lidResource)
```

Here are additional helpful infos:
*) I'm using libMoltenVK.dylib and libvulkan.1.dylib from the latest LunarG SDK 1.3.236.0
*) Of course I've the validation layer VK_LAYER_KHRONOS_validation enabled, which doesn't report errors.
*) The problem is 100% reproducible.
*) The same code runs fine in Linux and Windows with lots of different Vulkan drivers.
*) We are working hard on a native ARM version of our app. I tried this one too and could reproduce the bug. So the issue is not related to Rosetta 2 code translation or other issues with Intel code on M1.
*) While I cannot come up with a small example to reproduce, I can of course send you the app, or replace my libMoltenVK.dylib and libvulkan.1.dylib with debugging versions of yours and return the output.
*) Shortly after app startup, I see this warning message in the terminal:
```
2023-01-07 06:18:06.747638+0100 my.app[1351:30955] [Metal Compiler Warning] Warning: Com
pilation succeeded with: 
program_source:96:18: warning: writable resources in non-void vertex function
vertex main0_out main0(constant objdata& command [[buffer(0)]], constant ubo& frame [[buffer
(1)]], device drawcomdata& draw [[buffer(5)]], const device srcballdata& src [[buffer(7)]], 
uint gl_VertexIndex [[vertex_id]], uint gl_InstanceIndex [[instance_id]])
                 ^
program_source:96:120: note: writable buffer defined here
vertex main0_out main0(constant objdata& command [[buffer(0)]], constant ubo& frame [[buffer
(1)]], device drawcomdata& draw [[buffer(5)]], const device srcballdata& src [[buffer(7)]], 
[...]

PR #1918, opened 2023-05-23
Accumulate render stages when a resource is used by multiple descriptor bindings.

billhollings:
Should fix issues #1870 & #1811.

### DIFF
--- a/Docs/Whats_New.md
+++ b/Docs/Whats_New.md
@@ -31,6 +31,7 @@ Released 2023/05/23
   large for `MTLCounterSampleBuffer`, and fall back to emulation via CPU timestamps.
 - Ensure shaders that use `PhysicalStorageBufferAddresses` encode the use of the associated `MTLBuffer`.
 - Disable pipeline cache compression prior to macOS 10.15 and iOS/tvOS 13.0.
+- Accumulate render stages when a resource is used by multiple descriptor bindings.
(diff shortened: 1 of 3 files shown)


riperiperi:
I don't know how, but this seems to make the issue worse on Ryujinx. On a build that sets stage flags for all storage buffers to equal all stages used by the shader, and deliberately splits render pass when a storage buffer is used on two shaders with a different combination of stages, this triggers the error in cases where it didn't before.

```
[mvk-error] VK_ERROR_DEVICE_LOST: MTLCommandBuffer "vkQueueSubmit CommandBuffer on Queue 0-0" execution failed (code 9): Invalid Resource (00000009:kIOGPUCommandBufferCallbackErrorInvalidResource)
```

On mvk main branch it doesn't do this. (though the game doesn't render due to a weird new spir-v translation error)

I'm not sure why this is, because if the issue were that it only respects the usage set the first time the resource is recorded, it should be behaving the same.

billhollings:
> I'm not sure why this is, because if the issue were that it only respects the usage set the first time the resource is recorded, it should be behaving the same.

Thanks for the feedback.

Agreed. This should be no worse. I'm going to pull this in, because it does fix the situation where two descriptor bindings contradict each other. We can continued to work on your specific environment to see what the root causes might be.

Issue #1925, opened 2023-05-30
VK_PRESENT_MODE_IMMEDIATE_KHR locked to display refresh rate on macOS


SRSaunders:
Using mode `VK_PRESENT_MODE_IMMEDIATE_KHR` I expect no-delay presentation (i.e. tearing) with an unlocked frame rate.  Instead I observe a frame rate locked to my display refresh rate (60 Hz).  Using the Optick profiler, it appears the presentation engine is still using all swapchain buffers and waiting for vsync, which is counter to the spec.

In contrast, the same code behaves properly on Windows and Linux - i.e. immediate presentation with an unlocked frame rate.

Is there something special about the setup requirements for this presentation mode on macOS with MoltenVK?  Note that I am using triple buffering (3 GPU frames + 3 swapchain buffers) with `synchronousQueueSubmits = VK_FALSE`.  Using the Optick profiler, I can see the application is waiting on GPU graphics queue completion and not on acquire, submit or present, which is a bit puzzling since everything is idle (CPU + GPU) for a long period of time between vsync events.

This may be a duplicate of #581 (which is still open with no apparent resolution).

billhollings:
`VK_PRESENT_MODE_IMMEDIATE_KHR` should be working okay.

PR #1927 enhances the macOS Cube demo to permit enabling `VK_PRESENT_MODE_IMMEDIATE_KHR` in the `macOS/ DemoViewController.m` file. Try that out and see if it behaves as you expect. If so, we'll need to figure out how your app is using it differently than the demo. There may be some subtle secondary interactions that are showing up.

SRSaunders:
Thanks for updating the Cube demo.  I will try it out asap as reference point.

However, in the mean time I have been doing some digging and think I have found a trigger for the issue, but not the underlying cause.  As soon as I define an instance of `vk::PresentTimesInfoGOOGLE` in the pNext chain of `Queue::presentKHR()`, the problem arises and **VK_PRESENT_MODE_IMMEDIATE_KHR** locks to my display refresh rate.  When it is not included in the pNext chain, FPS is unlocked and present behaves as I would expect.

I don't know if this is intrinsic to the way the feature works, but it does correspond to my observed platform differences: on Windows and Linux (no FPS locking) the feature is not available for my AMD card and thus is not included in the pNext chain, however it is with MoltenVK on macOS where FPS locking is evident.

Strange coincidence given our recent interaction on improving the present times feature :)

[...]

PR #1936, opened 2023-06-06
Support VK_PRESENT_MODE_IMMEDIATE_KHR if VkPresentTimeGOOGLE::desiredPresentTime is zero.

billhollings:
- `[MTLDrawable presentAtTime:]` syncs to display vsync. To support `VK_PRESENT_MODE_IMMEDIATE_KHR` while using `VkPresentTimeGOOGLE::presentID`, only call `presentAtTime:` if `VkPresentTimeGOOGLE::desiredPresentTime` has been explicitly set to a non-zero value.
- Clarify initially clearing `MVKImagePresentInfo` to all zeros.

Fixes issue #1925.

### DIFF
--- a/Docs/Whats_New.md
+++ b/Docs/Whats_New.md
@@ -19,6 +19,7 @@ MoltenVK 1.2.5
 Released TBD
 
 - Ensure non-dispatch compute commands don't interfere with compute encoding state used by dispatch commands.
+- Support `VK_PRESENT_MODE_IMMEDIATE_KHR` if `VkPresentTimeGOOGLE::desiredPresentTime` is zero.
 - Add support for `VK_PRESENT_MODE_IMMEDIATE_KHR` to macOS Cube demo.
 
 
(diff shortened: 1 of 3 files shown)


SRSaunders:
A quick follow-up to this fix. I decided to upgrade my dev system from Monterey to Ventura and noticed some interesting differences related to this issue, fortunately on the positive side:

1. Independent of this issue, Apple seems to have made a change to the compositing/present handling in Ventura that now properly respects vsync frame rate for windowed environments with MoltenVK. Previously (Monterey and earlier), when using MoltenVK in a windowed environment with vsync on, the frame rate would appear double the monitor refresh rate. Only in fullscreen mode would the frame rate equal the monitor refresh rate. This issue now seems fixed in Ventura, with windowed environments showing correct vsync behaviour and frame rate.
2. Specifically regarding this fix, it seems that without it and when `presentInfo.hasPresentTime = true` and `desiredPresentTime = 0,` the new Ventura compositing/present handling behaves poorly and causes severe display flickering, almost like it is skipping display frames or inserting blank frames. This did not happen in Monterey and earlier. I was able to observe this behaviour using the stock 1.3.250 SDK (which does not yet include your fix) and also by reverting your fix in the latest dev stream of MoltenVK standalone. So it appears that `desiredPresentTime = 0` does have meaning to `presentAtTime`, but as far as I can tell it is not good behaviour.
 
In summary, this fix was good in Monterey, but even better in Ventura.

billhollings:
Thanks for returning here with an update!

> In summary, this fix was good in Monterey, but even better in Ventura.

Wonderful! Actually, I built the fix in Ventura 13.3. So, that might have been a factor in our back-and-forth discussion on the issue itself?

> UPDATE: I am speculating here, but I wonder if presentAtTime: presentationTime = 0 now under certain situations means drop this frame on the floor and don't present it? If so, this perhaps could give you a mechanism (coupled with vsync timing calculation logic) to support VK_PRESENT_MODE_MAILBOX_KHR. Maybe a bit of a stretch but interesting to think about.

Ah. An interesting idea, if it works.

PR #1944, opened 2023-06-10
Allows maximizing the concurrent executing compilation tasks.

RaafatAkkad:
https://developer.apple.com/videos/play/wwdc2023/10127/?time=540

On an M1 Pro Macbook Pro 16" maximumConcurrentCompilationTaskCount goes from 2 to 8

### DIFF
--- a/MoltenVK/MoltenVK/API/mvk_config.h
+++ b/MoltenVK/MoltenVK/API/mvk_config.h
@@ -913,6 +913,16 @@ typedef struct {
 	 */
 	MVKConfigCompressionAlgorithm shaderSourceCompressionAlgorithm;
 
+	/**
+	 * Maximize the concurrent executing compilation tasks.
+	 *
+	 * The initial value or this parameter is set by the
+	 * MVK_CONFIG_SHOULD_MAXIMIZE_CONCURRENT_COMPILATION
+	 * runtime environment variable or MoltenVK compile-time build setting.
+	 * This setting requires macOS 13.3 & is disabled by default.
+	 */
+	VkBool32 shouldMaximizeConcurrentCompilation;
+
 } MVKConfiguration;
 
 
--- a/MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
@@ -1596,6 +1596,7 @@ MVKPhysicalDevice::MVKPhysicalDevice(MVKInstance* mvkInstance, id<MTLDevice> mtl
 	_supportedExtensions(this, true),
 	_pixelFormats(this) {				// Set after _mtlDevice
 
+	initMTLDevice();
 	initProperties();           		// Call first.
 	initMetalFeatures();        		// Call second.
 	initFeatures();             		// Call third.
@@ -1606,6 +1607,15 @@ MVKPhysicalDevice::MVKPhysicalDevice(MVKInstance* mvkInstance, id<MTLDevice> mtl
 	initCounterSets();
 	initVkSemaphoreStyle();
 	logGPUInfo();
+}
+
+void MVKPhysicalDevice::initMTLDevice() {
+#if MVK_XCODE_14_3 && MVK_MACOS && !MVK_MACCAT
+	if ([_mtlDevice respondsToSelector: @selector(setShouldMaximizeConcurrentCompilation:)]) {
+		[_mtlDevice setShouldMaximizeConcurrentCompilation: mvkConfig().shouldMaximizeConcurrentCompilation];
+		MVKLogInfoIf(mvkConfig().debugMode, "maximumConcurrentCompilationTaskCount %lu", _mtlDevice.maximumConcurrentCompilationTaskCount);
+	}
+#endif
 }
 
 // Initializes the physical device properties (except limits).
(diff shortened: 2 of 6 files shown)


MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
```diff
@@ -3629,8 +3629,15 @@ static uint32_t mvkGetEntryProperty(io_registry_entry_t entry, CFStringRef prope
 		id<MTLDevice> sysMTLDevice = MTLCreateSystemDefaultDevice();
 		if (mvkGetRegistryID(sysMTLDevice) == mvkGetRegistryID(mtlDevice)) {
 			_physicalDevice->replaceMTLDevice(sysMTLDevice);
+			mtlDevice = _physicalDevice->getMTLDevice();
```
RaafatAkkad, line 3632: I don't have the hardware to test the above scenario but this is required if replaceMTLDevice changes the _mtlDevice variable

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
```diff
@@ -3629,8 +3629,15 @@ static uint32_t mvkGetEntryProperty(io_registry_entry_t entry, CFStringRef prope
...
 		}
 	}
+    
+#if MVK_XCODE_14_3 && TARGET_OS_OSX
+    [mtlDevice setShouldMaximizeConcurrentCompilation: mvkConfig().shouldMaximizeConcurrentCompilation];
+    MVKLogInfoIf(mvkConfig().debugMode, "maximumConcurrentCompilationTaskCount %lu", mtlDevice.maximumConcurrentCompilationTaskCount);
+#endif
```
billhollings, line 3639: In addition to the Xcode version test, these calls must be wrapped in a `respondsToSelector:` test as well (as found elsewhere), because the runtime might be an older OS version that doesn't support the calls. 

One test for `shouldMaximizeConcurrentCompilation` should be sufficient since `maximumConcurrentCompilationTaskCount` was introduced at the same time.

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
billhollings: This is performed as part of creating a swap chain, which really isn't the appropriate place to do set the shader compilation options. The notes to the commit for the replacement code seems to indicate the pointer won't actually change anyway.

So, let's move the code setting `shouldMaximizeConcurrentCompilation` to the `MVKPhysicalDevice` constructor. This is the first time we're actually configuring something on the `MTLDevice`, so I suggest we create a member function called something like `initMTLDevice()`, and call it in the `MVKPhysicalDevice` constructor, above the call to `initProperties()`.

------------------- 

@cdavis5e In reviewing your note from your older commit, it indicates the `MTLDevice` pointers are the same. If that's the case, is the call to `replaceMTLDevice:` even needed? Based on your note, if I understand the flow correctly, the `MTLDevice` pointer will always be the same as what's already in `MVKPhysicalDevice`? If that's not the case, then a lot could have happened with the old `MTLDevice` (settings and possible resources) that will now be out of scope?

billhollings (changes requested):
Thanks for submitting this. I have a few questions and refactoring suggestions.

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
cdavis5e: They should be the same. Frankly, I don't see how Metal could work on Mac if they *weren't* the same. From what I understand, Metal creates `MTLDevice` objects for all GPUs when it initializes; the array of those devices is what `MTLCopyAllDevices()` returns. All `MTLCreateSystemDefaultDevice()` does on Mac is:
1. force Macs with a mux to select the high-performance GPU for presentation instead of the integrated one, and
2. return a +1 reference to the "default" device, which on Macs with a mux will be the high-performance GPU, and on Macs with multiple video cards will be the device for the primary display.

MoltenVK/MoltenVK/API/mvk_config.h
```diff
@@ -913,6 +913,16 @@ typedef struct {
...
 
+    /**
+     * Maximize the concurrent executing compilation tasks.
+     *
+     * The initial value or this parameter is set by the
+     * MVK_CONFIG_SHOULD_MAXIMIZE_CONCURRENT_COMPILATION
+     * runtime environment variable or MoltenVK compile-time build setting.
+     * This setting requires macOS 13.3 & is disabled by default.
+     */
+    VkBool32 shouldMaximizeConcurrentCompilation;
```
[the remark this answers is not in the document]
RaafatAkkad: I think it should be configurable as Apple has made it opt-in.

I don't have any sources about the cons (Apple recently updated Blender to always MaximizeConcurrentCompilation (https://projects.blender.org/blender/blender/pulls/106625) but didn't mention the drawbacks. I guess it adds extra Thread overhead which the app might not benefit from.

For me it has slightly reduced shader compile stutter so I think it's worth it.

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
RaafatAkkad: Thanks for your feedback, I've created https://github.com/KhronosGroup/MoltenVK/pull/1945 based on your suggestions but I'm unable to test this. Sorry.

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
RaafatAkkad: Thanks, updated.

billhollings (approved):
LGTM now. Thanks for changes.

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
billhollings: > Thanks for your feedback, I've created #1945 based on your suggestions but I'm unable to test this. Sorry.

Thanks. The call to `MTLCreateSystemDefaultDevice()` in #1945 needs work. Since this PR is otherwise the same as #1945, I'm going to pull this one in.

PR #1945, opened 2023-06-14
Cleanup createSwapchain based on https://github.com/KhronosGroup/MoltenVK/pull/1944#discussion_r1228171186

RaafatAkkad:
I'm unable to test this scenario due to lack of hardware.


MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
```diff
@@ -3628,7 +3638,7 @@ static uint32_t mvkGetEntryProperty(io_registry_entry_t entry, CFStringRef prope
 	if (mvkConfig().switchSystemGPU && !(mtlDevice.isLowPower || mtlDevice.isHeadless) ) {
 		id<MTLDevice> sysMTLDevice = MTLCreateSystemDefaultDevice();
 		if (mvkGetRegistryID(sysMTLDevice) == mvkGetRegistryID(mtlDevice)) {
-			_physicalDevice->replaceMTLDevice(sysMTLDevice);
+			MTLCreateSystemDefaultDevice();
```
billhollings, line 3641: Thanks for the effort, but this is not what I was contemplating in my notes in #1944. `MTLCreateSystemDefaultDevice()` should only be called once, and I'd like to move this entire code block out of `createSwapchain()`. I've accepted #1944, and we'll handle the changes to this code in a separate PR. I'll take care of that.

PR #1955, opened 2023-06-22
Support the `VK_EXT_4444_formats` extension.

cdavis5e:
This turned out to be a little bit more involved than I had hoped. But, with this, we can now use the `VK_FORMAT_A4R4G4B4_UNORM_PACK16` and `VK_FORMAT_A4B4G4R4_UNORM_PACK16` formats from shaders, use them as blit sources, and even clear them. Storage images and render targets of these formats aren't supported, however. To support the latter would require the insertion of a swizzle into the fragment shader before returning. The former cannot be reasonably supported.

### DIFF
--- a/MoltenVK/MoltenVK/Commands/MVKCommandResourceFactory.h
+++ b/MoltenVK/MoltenVK/Commands/MVKCommandResourceFactory.h
@@ -37,20 +37,31 @@ class MVKQueryPool;
  * This structure can be used as a key in a std::map and std::unordered_map.
  */
 typedef struct MVKRPSKeyBlitImg {
-	uint16_t srcMTLPixelFormat = 0;			/**< as MTLPixelFormat */
-	uint16_t dstMTLPixelFormat = 0;			/**< as MTLPixelFormat */
-	uint8_t srcMTLTextureType = 0;			/**< as MTLTextureType */
+	uint16_t srcMTLPixelFormat : 12;		/**< as MTLPixelFormat */
+	uint16_t dstMTLPixelFormat : 12;		/**< as MTLPixelFormat */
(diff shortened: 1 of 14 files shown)


MoltenVK/MoltenVK/Commands/MVKCommandResourceFactory.h
```diff
@@ -37,20 +37,31 @@ class MVKQueryPool;
...
+	uint16_t dstMTLPixelFormat : 12;		/**< as MTLPixelFormat */
+	uint8_t srcMTLTextureType : 4;			/**< as MTLTextureType */
+	uint8_t srcFilter : 4;					/**< as MTLSamplerMinMagFilter */
 	uint8_t srcAspect = 0;					/**< as VkImageAspectFlags */
-	uint8_t srcFilter = 0;					/**< as MTLSamplerMinMagFilter */
 	uint8_t dstSampleCount = 0;
+	uint8_t srcSwizzleR : 4;				/**< as VkComponentSwizzle */
+	uint8_t srcSwizzleG : 4;				/**< as VkComponentSwizzle */
+	uint8_t srcSwizzleB : 4;				/**< as VkComponentSwizzle */
+	uint8_t srcSwizzleA : 4;				/**< as VkComponentSwizzle */
```
billhollings, line 49: First time we've employed bitfields, I think. Nice introduction!

billhollings (approved):
LGTM. Thanks for submitting!

This doesn't require a SPIRV-Cross upgrade?

cdavis5e:
> LGTM. Thanks for submitting!
> 
> This doesn't require a SPIRV-Cross upgrade?

Nope.

PR #1958, opened 2023-06-23
Support the `VK_EXT_calibrated_timestamps` extension.

cdavis5e:
This extension has a direct Metal equivalent in the `-[MTLDevice sampleTimestamps:gpuTimestamp:]` method. However, that method returns CPU timestamps in the Mach absolute time domain, which is *not* that of `CLOCK_MONOTONIC_RAW` but of `CLOCK_UPTIME_RAW`. The function that corresponds to `CLOCK_MONOTONIC_RAW` is `mach_continuous_time()`. Therefore, this implementation uses the `mach_continuous_time()` function for the CPU timestamp. Perhaps we should lobby the WG for `VK_TIME_DOMAIN_CLOCK_UPTIME_RAW_EXT`.


billhollings (approved):
LGTM

PR #1960, opened 2023-06-26
Advertise the `VK_KHR_shader_non_semantic_info` extension.

cdavis5e:
This just provides support for the `SPV_KHR_non_semantic_info` extension, which supports extended instruction sets that do not affect the semantics of a SPIR-V shader (e.g. debug info). SPIRV-Cross already handles these instruction sets, so no additional work is required on our part to support this extension.


billhollings (approved):
LGTM. Thanks!

PR #1962, opened 2023-06-28
Add support for VK_PRIMITIVE_TOPOLOGY_TRIANGLE_FAN.

### LINKED ISSUE #1419: Enhancement assessment: VK_PRIMITIVE_TOPOLOGY_TRIANGLE_FAN

### LINKED ISSUE #1799: VkPrimitiveTopology value 5 is not supported
I'm trying to launch a game but I get a bunch of lines in the log stating that VkPrimitiveTopology value 5 is not supported and Metal does not support triangle fans. Is there any possible way to implement this into moltenVK?

[FullLog.txt](https://github.com/KhronosGroup/MoltenVK/files/10253275/FullLog.txt)

billhollings:
To reduce complexity and repetitive copy-pasted spaghetti code, the design approach here was to implement triangle fan conversion on `MVKCmdDrawIndexedIndirect`, as the most general of the draw commands, and then populate and invoke a synthetic `MVKCmdDrawIndexedIndirect` command from the other draw commands.

- Rename pipeline factory shader `cmdDrawIndexedIndirectMultiviewConvertBuffers()` to `cmdDrawIndexedIndirectConvertBuffers()`, and in addition to original support for modifying indirect content to support multiview, add support for converting triangle fan indirect content and indexes to triangle list.
- Modify `MVKCmdDrawIndexedIndirect` to track need to convert triangle fans to triangle list, and invoke kernel function when needed.
- Modify `MVKCmdDraw`, `MVKCmdDrawIndexed`, and `MVKCmdDrawIndirect` to populate and invoke a synthetic `MVKCmdDrawIndexedIndirect` command to convert triangle fans to triangle lists.
- Add pipeline factory shader `cmdDrawIndirectPopulateIndexes()` to convert non-indexed indirect content to indexed indirect content.
- `MVKCmdDrawIndexedIndirect` add support for zero divisor vertex buffers potentially coming from `MVKCmdDraw` and `MVKCmdDrawIndexed`.

- Rename pipeline factory shader `cmdDrawIndexedIndirectConvertBuffers()` to `cmdDrawIndexedIndirectTessConvertBuffers()` so it will be invoked from `MVKCommandEncodingPool::getCmdDrawIndirectTessConvertBuffersMTLComputePipelineState()` (unrelated).

Fixes #1419.
Fixes #1799.

### DIFF
--- a/Docs/Whats_New.md
+++ b/Docs/Whats_New.md
@@ -25,6 +25,7 @@ Released TBD
 	- `VK_EXT_shader_demote_to_helper_invocation`
 	- `VK_EXT_shader_subgroup_ballot`
 	- `VK_EXT_shader_subgroup_vote`
+- Add support for `VK_PRIMITIVE_TOPOLOGY_TRIANGLE_FAN`.
 - Ensure non-dispatch compute commands don't interfere with compute encoding state used by dispatch commands.
 - Support `VK_PRESENT_MODE_IMMEDIATE_KHR` if `VkPresentTimeGOOGLE::desiredPresentTime` is zero.
 - Support maximizing the concurrent executing compilation tasks via `MVKConfiguration::shouldMaximizeConcurrentCompilation`
--- a/MoltenVK/MoltenVK/Commands/MVKCmdDraw.h
+++ b/MoltenVK/MoltenVK/Commands/MVKCmdDraw.h
@@ -91,6 +91,7 @@ public:
 						uint32_t firstInstance);
 
     void encode(MVKCommandEncoder* cmdEncoder) override;
+	void encodeIndexedIndirect(MVKCommandEncoder* cmdEncoder);
 
 protected:
 	MVKCommandTypePool<MVKCommand>* getTypePool(MVKCommandPool* cmdPool) override;
@@ -120,6 +121,7 @@ public:
 
 protected:
 	MVKCommandTypePool<MVKCommand>* getTypePool(MVKCommandPool* cmdPool) override;
+	void encodeIndexedIndirect(MVKCommandEncoder* cmdEncoder);
 
 	uint32_t _firstIndex;
 	uint32_t _indexCount;
@@ -146,6 +148,7 @@ public:
 
 protected:
 	MVKCommandTypePool<MVKCommand>* getTypePool(MVKCommandPool* cmdPool) override;
+	void encodeIndexedIndirect(MVKCommandEncoder* cmdEncoder);
 
 	id<MTLBuffer> _mtlIndirectBuffer;
 	VkDeviceSize _mtlIndirectBufferOffset;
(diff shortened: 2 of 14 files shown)


billhollings:
@cdavis5e As described above in the PR description , the implementation here delegates from `MVKCmdDraw`, `MVKCmdDrawIndexed`, and `MVKCmdDrawIndirect`, to a synthetic `MVKCmdDrawIndexedIndirect` instance, whenever the topology is triangle fans. I did this to avoid having to duplicate the conversion code across all of the draw commands.

It got me to contemplating whether a similar approach might be useful for the entire tessellation draw flow. Basically, implement any non-native-Metal workflow(s) once in `MVKCmdDrawIndexedIndirect`, and delegate from the other draw commands, leaving those draw commands to just handle behaviour that Metal supports natively.

Thoughts about the practicality of such an approach? You know the tess implementation much better than I do.

IsaacMarovitz:
Screenshot 2023-07-01 at 12 43 45
Can confirm this fixes the UI in No Man's Sky!

PR #1964, opened 2023-06-29
Fix unreachable code in MVKDeferredOperation::join().

billhollings:
- Fix unreachable code in `MVKDeferredOperation::join()`.
- Refactor code so deferred functions call back to `MVKDeferredOperation` instance to update current status, and deferred function execution returns result of individual thread execution.
- `MVKDeferredOperation` use `MVKSmallVector` for `_functionParameters`.
- `MVKDeferredOperation` use a single mutex lock.
- Add additional comments explaining design to developers of future extensions that use deferred operations.

@AntarticCoder Please review as changes to your original code. Most of it is just clean up and documentation, but the way that deferred functions interact with `MVKDeferredOperation::join()` functionally has changed. I also did not have any trouble with `MVKSmallVector` on either Xcode 14 or 15, so I've reverted to using that as originally planned.

### DIFF
--- a/MoltenVK/MoltenVK/GPUObjects/MVKSync.h
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKSync.h
@@ -640,67 +640,85 @@ protected:
 #pragma mark -
 #pragma mark MVKDeferredOperation
 
+/** The maximum number of function parameters for all functions added to MVKDeferredOperationFunctionPointer. */
+static const int kMVKMaxDeferredFunctionParameters = 3;
+
 /** Defines the function pointer for each dependent function. */
-union MVKDeferredOperationFunctionPointer
-{
-    // Empty until deferred functions from other extensions have been defined
-    // Planning to use std::functions
+union MVKDeferredOperationFunctionPointer {
+    // Empty until deferred functions from other extensions have been defined.
+	// Planning to use std::functions
+	// Each function should take a pointer to a MVKDeferredOperation instance, to allow
+	// the function to call setMaxConcurrency() and setOperationResult(), and should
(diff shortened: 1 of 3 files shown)


billhollings:
> however this will only occur in the case of there being no function that has been loaded into the deferred operation, as it is the default return of the switch statement

Yes. My thinking here is that this default switch case will only occur when either an unimplemented deferred type is invoked, or the app invokes it without a deferred operation. Either way, under normal operation, this default switch case will never be encountered.

I chose to return `VK_SUCCESS` in this case, because an invocation that hits here is by definition a no-op, there is no work to complete, and the execution of this "no-work" is "successful". The spec does include the following text, which would indicate that we must return `VK_SUCCESS` from at least one call to `vkDeferredOperationJoinKHR()`, even if it's a no-op:

>Implementations must guarantee forward progress by enforcing the following invariants:
> 1. If only one thread has invoked [vkDeferredOperationJoinKHR](https://registry.khronos.org/vulkan/specs/1.3-extensions/html/vkspec.html#vkDeferredOperationJoinKHR) on a given operation, that thread must execute the operation to completion and return VK_SUCCESS.
> 2. If multiple threads have concurrently invoked [vkDeferredOperationJoinKHR](https://registry.khronos.org/vulkan/specs/1.3-extensions/html/vkspec.html#vkDeferredOperationJoinKHR) on the same operation, then at least one of them must complete the operation and return VK_SUCCESS.

AntarticCoder:
Alright then, this LGTM.

billhollings:
> Alright then, LGTM.

Thanks for the review!

PR #1965, opened 2023-07-03
Fix depth <-> color copies

etang-cw:
Metal doesn't let you make color views of depth textures

### DIFF
--- a/MoltenVK/MoltenVK/Commands/MVKCmdTransfer.mm
+++ b/MoltenVK/MoltenVK/Commands/MVKCmdTransfer.mm
@@ -139,7 +139,7 @@ void MVKCmdCopyImage<N>::encode(MVKCommandEncoder* cmdEncoder, MVKCommandUse com
         bool isDstCompressed = _dstImage->getIsCompressed();
         bool isDstDepth = _dstImage->getIsDepthStencil();
 
-        // If source and destination have different formats and at least one is compressed, use a temporary intermediary buffer
+        // If source and destination have different formats and at least one is compressed, or one is color and the other is depth/stencil, use a temporary intermediary buffer.
         bool useTempBuffer = (srcMTLPixFmt != dstMTLPixFmt) && (isSrcCompressed || isDstCompressed || isSrcDepth != isDstDepth);
 
         if (useTempBuffer) {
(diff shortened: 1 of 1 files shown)


MoltenVK/MoltenVK/Commands/MVKCmdTransfer.mm
```diff
@@ -133,12 +133,14 @@ static inline MTLSize mvkClampMTLSize(MTLSize size, MTLOrigin origin, MTLSize ma
         
         MTLPixelFormat srcMTLPixFmt = _srcImage->getMTLPixelFormat(srcPlaneIndex);
         bool isSrcCompressed = _srcImage->getIsCompressed();
+        bool isSrcDepth = _srcImage->getIsDepthStencil();
 
         MTLPixelFormat dstMTLPixFmt = _dstImage->getMTLPixelFormat(dstPlaneIndex);
         bool isDstCompressed = _dstImage->getIsCompressed();
+        bool isDstDepth = _dstImage->getIsDepthStencil();
 
         // If source and destination have different formats and at least one is compressed, use a temporary intermediary buffer
```
cdavis5e, line 142: You should probably update this comment, too.

cdavis5e (approved):
Pretty straightforward. Just one minor quibble...

MoltenVK/MoltenVK/Commands/MVKCmdTransfer.mm
billhollings, line 142:
```suggestion
        // If source and destination have different formats and at least one is compressed, or one is color and the other is depth/stencil, use a temporary intermediary buffer.
```

billhollings (changes requested):
Thanks for the submission!

Per the observation by @cdavis5e, please update the comment (for which I've made a suggestion, so you can accept it here).

MoltenVK/MoltenVK/Commands/MVKCmdTransfer.mm
billhollings: Meh. I'll just do it here. 😉

billhollings (approved):

etang-cw:
Actually, it looks like depth <-> color copies aren't allowed in Vulkan, and I should be fixing vkd3d

Do we still want this?

cdavis5e:
> Do we still want this?

In that case, it's probably better to close this, since this is undefined behavior anyway. I can't speak for @billhollings, but in cases of undefined behavior, I prefer that the driver do the least amount of work necessary.

billhollings:
> in cases of undefined behavior, I prefer that the driver do the least amount of work necessary

Agreed. If this is forbidden in Vulkan too, then this should be triggering a Validation Layer error, and if not, should be fixed there.

PR #1966, opened 2023-07-05
Added Enum for MSL Version 3.1

AntarticCoder:
In this PR, I've added support for Xcode 15, and added a case for MSL version 3.1. I added this because I noticed xcode was throwing some warnings about an unhandled switch case.

### DIFF
--- a/Common/MVKCommonEnvironment.h
+++ b/Common/MVKCommonEnvironment.h
@@ -88,6 +88,10 @@ extern "C" {
 #endif
 
 /** Building with Xcode versions. iOS version also covers tvOS. */
+#ifndef MVK_XCODE_15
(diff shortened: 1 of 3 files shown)


billhollings (changes requested):
Thanks for submitting this!

For completeness, would you mind also adding the additional changes identified in [this PR](https://github.com/KhronosGroup/MoltenVK/pull/1940/files) for the files:

```
MVKDevice.mm
OSSupport.mm
```
(basically...search for `MTLLanguageVersion3_0` and add a corresponding `MTLLanguageVersion3_1` entry).

But don't include the changes to `MVKPixelFormats.mm`, which is the main purpose of that PR, and it's still a WIP. We'll remove everything else from that PR after this PR is merged.

Much thanks!

AntarticCoder:
Thanks, I didn't notice the extra bits required for MSL 3.1.

billhollings (approved):

Issue #1956, opened 2023-06-23
Implementing Vulkan Acceleration Structures


AntarticCoder:
At the moment MoltenVK does not support raytracing(#427), and to support `VK_KHR_ray_tracing_pipeline` and `VK_KHR_ray_query`, we need to implement acceleration structures. PR #1954 (issue #1953) implemented `VK_KHR_deferred_host_operations` that finishe off the dependincies for `VK_KHR_acceleration_structures`. The only thing left to do is to actually implement it, this issue will provide a place to discuss the design decisions for acceleration structures.

I'm also planning on trying to implement this myself.

natevm:
If it's any help, consolidating some information

In vulkan I have examples of AABB accels [here](https://github.com/gprt-org/GPRT/blob/90f538eab76b44b9f0d871383e7bd4cbe23d1d67/gprt/gprt.cpp#L3293), Triangular geometry accels [here](https://github.com/gprt-org/GPRT/blob/90f538eab76b44b9f0d871383e7bd4cbe23d1d67/gprt/gprt.cpp#L2669), and Instance accels [here](https://github.com/gprt-org/GPRT/blob/90f538eab76b44b9f0d871383e7bd4cbe23d1d67/gprt/gprt.cpp#L3893) 

It appears that these types roughly translate to 
[MTLAccelerationStructureBoundingBoxGeometryDescriptor](https://developer.apple.com/documentation/metal/mtlaccelerationstructureboundingboxgeometrydescriptor),
[MTLAccelerationStructureTriangleGeometryDescriptor](https://developer.apple.com/documentation/metal/mtlaccelerationstructuretrianglegeometrydescriptor), and [MTLInstanceAccelerationStructureDescriptor](https://developer.apple.com/documentation/metal/mtlinstanceaccelerationstructuredescriptor)



https://developer.apple.com/documentation/metal/mtlaccelerationstructure

natevm:
One place that might be a good starting point is populating the structures of data for acceleration structure features and properties. In my code I do that [here](https://github.com/gprt-org/GPRT/blob/90f538eab76b44b9f0d871383e7bd4cbe23d1d67/gprt/gprt.cpp#L2639-L2655)

In `VkPhysicalDeviceAccelerationStructureFeaturesKHR`, we could probably just return true for the "accelerationStructure" field, and false for all the other fields.

[...]

Issue #1953, opened 2023-06-16
VK_KHR_deferred_host_operations needs to be implemented before adding Raytracing Support


AntarticCoder:
In issue #427 there has been discussion about the implementation of `VK_KHR_ray_tracing_pipeline` or `VK_KHR_ray_query`. Both of these extensions need the `VK_KHR_acceleration_structure` extension to work. To implement acceleration structures, we need the `VK_KHR_deferred_host_operation`, which allows for heavy operations that can take longer to do to be deferred until later. So as a small step towards raytracing, deferred_host_operations could be implemented. 

This issue is dependent on issue #427

Edit: I'd like to try and implement this, could somebody point me to which files would need to be modified as I'm not quite sure where to start

natevm:
I’m not that familiar with MoltenVK, but I’m guessing the best first steps here would be to add a `vkCreateDeferredOperationKHR`and a `vkDestroyDeferredOperationKHR` that allocate / free an opaque `VkDeferredOperationKHR` object.

AntarticCoder:
Yeah that seems simple but looking through the source files, I'm not able to find a file to put `VkDeferredOperationKHR`, I know it would have to go in the `/GPUObjects` subdirectory though, but it seems unclear to me what files `VkDeferredOperationKHR` would go into.

AntarticCoder:
Would an opaque handle of `VkDeferredOperationKHR` fit into `GPUObjects/MVKSync.h`?

natevm:
Personally I think that’s a good fit. This is a synchronization mechanism for sure, just a bit unique in that it’s about host-side asynchronous operations. 

Worst case we can move it around later on in a pull request.

AntarticCoder:
Thanks for the insigh, I plan to start a PR on Monday/Tuesday

AntarticCoder:
Opened a pull request, please link the issue with it: #1954

natevm:
IMO, a union seems like a decent way to do it. So long as there is a clear organization that can be extended.

AntarticCoder:
That's how I plan to do it, everytime a new vulkan command is implemented requiring deferred operations, you add the operations functions pointer to the union and also add the function to an enum that will manage with variable in the union should be used.

natevm:
Indirectly related to this issue, does Metal RT support host-side acceleration structure construction? If it doesn’t, then is it possible to still support vk_khr_acceleration_structure?

cdavis5e:
[...]

Issue #427, opened 2018-12-26
Raytracing support


teaglin:
Hi, I was curious if there is any support for ray tracing. I am interested in prototyping raytracing applications on macOS with Metal 2 and then scaling them up for production on Nvidia's RTX GPUs.

cdavis5e:
Theoretically, it should be possible to support the `VK_NV_ray_tracing` extension on Mac, using the Metal Performance Shaders framework which uses a very similar technique to accelerate ray tracing. At this time, though, ray tracing isn't supported.

billhollings:
If anyone wants to take a run at implementing this...it would be welcome.

teaglin:
I might be interested, but I probably won't be free for a few weeks to a month for other priorities.

Lichtso:
Why not go straight for the newest (vendor independent) `VK_KHR_ray_tracing` extension?
https://www.khronos.org/news/press/khronos-group-releases-vulkan-ray-tracing

billhollings:
> Why not go straight for the newest (vendor independent) `VK_KHR_ray_tracing` extension?

This is an older thread. But yes...we should now aim for the KHR version.

Lichtso:
One of the dependencies of `VK_KHR_ray_tracing` is `VK_KHR_buffer_device_address`. Which requires the host to know the physical addresses (pointers) inside the device. However, it seems that Metal does not expose that information to the user.

oscarbg:
Came here searching for VK_KHR_buffer_device_address references, as is a recommended extension for current VKD3D Proton 2.0..
let me point out that now raytracing on Metal (on iOS 14.0 and Mac 11.0) seems more similar to VK_KHR_ray_tracing having added support for raytracing shaders (intersection, etc..) (i.e. integrated into Metal shading language)..
seems this new support requires on Mac:
*RDNA GPU or later (no Vega earlier support)
*Intel Skylake GPU or later
and on IOS: A13 or later..
recalling from memory so might be wrong..

krupitskas:
Im quite interested to add Vulkan raytracing api support with metal performance shaders, I will take a look into this,

dictoon:
Any news on this?

krupitskas:
Not yet, working on this in free time, when I will have any results to share and questions to ask, I will create a draft pull request. However I expect "fun" with shaders.

oscarbg:
@krupitskas so you will work on spirv cross raytracing metal support?
[...]

PR #1967, opened 2023-07-06
WIP: Implementing Acceleration Structures

AntarticCoder:
This PR provides an implementation of the `VK_KHR_acceleration_structure` extension which provides a gateway to ray queries and ray tracing pipelines. This PR is still very WIP due to not being anywhere close to done. The reason for opening this PR so early on is to allow for more concrete discussion of the implementation of acceleration structures, and also keeps people up to date on the implementation.

This PR is related to:
#427 
#1953 - Not directly related but may have some slight discussion on Acceleration Structures
#1956

### DIFF
--- a/MoltenVK/MoltenVK/Commands/MVKCmdAccelerationStructure.mm
+++ b/MoltenVK/MoltenVK/Commands/MVKCmdAccelerationStructure.mm
@@ -21,27 +21,28 @@
 #include "MVKCommandBuffer.h"
 #include "MVKCommandPool.h"
 #include "MVKAccelerationStructure.h"
+#include "MVKFoundation.h"
 
 #include <Metal/Metal.h>
 
 #pragma mark -
 #pragma mark MVKCmdBuildAccelerationStructure
 
-VkResult MVKCmdBuildAccelerationStructure::setContent(MVKCommandBuffer*                                       cmdBuff,
-                                                      uint32_t                                                infoCount,
-                                                      const VkAccelerationStructureBuildGeometryInfoKHR*      pInfos,
-                                                      const VkAccelerationStructureBuildRangeInfoKHR* const*  ppBuildRangeInfos) {
+VkResult MVKCmdBuildAccelerationStructure::setContent(MVKCommandBuffer* cmdBuff,
+                                                      uint32_t infoCount,
+                                                      const VkAccelerationStructureBuildGeometryInfoKHR* pInfos,
+                                                      const VkAccelerationStructureBuildRangeInfoKHR* const* ppBuildRangeInfos) {
+    _buildInfos.clear();
     _buildInfos.reserve(infoCount);
-    for (uint32_t i = 0; i < infoCount; i++)
-    {
+    for (uint32_t i = 0; i < infoCount; i++) {
         MVKAccelerationStructureBuildInfo& info = _buildInfos.emplace_back();
         info.info = pInfos[i];
 
         // TODO: ppGeometries
         info.geometries.reserve(pInfos[i].geometryCount);
         info.ranges.reserve(pInfos[i].geometryCount);
-        memcpy(info.geometries.data(), pInfos[i].pGeometries, pInfos[i].geometryCount);
-        memcpy(info.ranges.data(), ppBuildRangeInfos[i], pInfos[i].geometryCount);
+        memcpy(info.geometries.data(), pInfos[i].pGeometries, pInfos[i].geometryCount * sizeof(VkAccelerationStructureGeometryKHR));
+        memcpy(info.ranges.data(), ppBuildRangeInfos[i], pInfos[i].geometryCount * sizeof(VkAccelerationStructureBuildRangeInfoKHR));
 
         info.info.pGeometries = info.geometries.data();
     }
@@ -50,25 +51,12 @@ VkResult MVKCmdBuildAccelerationStructure::setContent(MVKCommandBuffer*         
 }
 
 void MVKCmdBuildAccelerationStructure::encode(MVKCommandEncoder* cmdEncoder) {
-    id<MTLAccelerationStructureCommandEncoder> accStructEncoder = cmdEncoder->getMTLAccelerationStructureEncoder(kMVKCommandUseBuildAccelerationStructure);
-    
-    for (MVKAccelerationStructureBuildInfo& entry : _buildInfos)
-    {
-        VkAccelerationStructureBuildGeometryInfoKHR& buildInfo = entry.info;
+    for (MVKAccelerationStructureBuildInfo& entry : _buildInfos) {
+        const auto& buildInfo = entry.info;
+        const auto& ranges = entry.ranges;
 
-        MVKAccelerationStructure* mvkSrcAccStruct = (MVKAccelerationStructure*)buildInfo.srcAccelerationStructure;
         MVKAccelerationStructure* mvkDstAccStruct = (MVKAccelerationStructure*)buildInfo.dstAccelerationStructure;
-
-        id<MTLAccelerationStructure> srcAccStruct = mvkSrcAccStruct->getMTLAccelerationStructure();
         id<MTLAccelerationStructure> dstAccStruct = mvkDstAccStruct->getMTLAccelerationStructure();
-        
-        id<MTLHeap> srcAccStructHeap = mvkSrcAccStruct->getMTLHeap();
-        id<MTLHeap> dstAccStructHeap = mvkDstAccStruct->getMTLHeap();
-        
-        // Should we throw an error here?
-        // https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/vkCmdBuildAccelerationStructuresKHR.html#VUID-vkCmdBuildAccelerationStructuresKHR-pInfos-03667
-        if(buildInfo.mode == VK_BUILD_ACCELERATION_STRUCTURE_MODE_UPDATE_KHR && !mvkDstAccStruct->getAllowUpdate())
-            continue;
         
         MVKDevice* mvkDevice = cmdEncoder->getDevice();
         MVKBuffer* mvkBuffer = mvkDevice->getBufferAtAddress(buildInfo.scratchData.deviceAddress);
@@ -78,32 +66,95 @@ void MVKCmdBuildAccelerationStructure::encode(MVKCommandEncoder* cmdEncoder) {
         id<MTLBuffer> scratchBuffer = mvkBuffer->getMTLBuffer();
         NSInteger scratchBufferOffset = mvkBuffer->getMTLBufferOffset();
         
-        if (buildInfo.mode == VK_BUILD_ACCELERATION_STRUCTURE_MODE_BUILD_KHR)
-        {
-            MTLAccelerationStructureDescriptor* descriptor = mvkDstAccStruct->populateMTLDescriptor(
-                mvkDevice,
-                buildInfo,
-                entry.ranges.data(),
-                nullptr
-            );
+        if (buildInfo.mode == VK_BUILD_ACCELERATION_STRUCTURE_MODE_BUILD_KHR) {
+            MTLAccelerationStructureDescriptor* descriptor = mvkDstAccStruct->newMTLAccelerationStructureDescriptor(buildInfo, entry.ranges.data(), nullptr);
+
+            id<MTLFence> fence = nil;
+            if (buildInfo.type == VK_ACCELERATION_STRUCTURE_TYPE_TOP_LEVEL_KHR) {
+                // Only one geometry, validated in populateMTLDescriptor
+
+                // TODO: handle instances.arrayOfPointers == true
+                // TODO: ppGeometries
+                const VkAccelerationStructureGeometryInstancesDataKHR& instancesData = buildInfo.pGeometries[0].geometry.instances;
+                uint64_t instancesBDA = instancesData.data.deviceAddress;
+                MVKBuffer* mvkInstancesBuffer = cmdEncoder->getDevice()->getBufferAtAddress(instancesBDA);
+                NSUInteger bOffset = (instancesBDA - mvkInstancesBuffer->getMTLBufferGPUAddress()) + mvkInstancesBuffer->getMTLBufferOffset();
+
+                // Allocate a transient buffer to store converted instance data
+                NSUInteger tmpBuffSize = sizeof(MTLAccelerationStructureInstanceDescriptor) * ranges[0].primitiveCount;
+                const MVKMTLBufferAllocation* tmpBuff = cmdEncoder->getTempMTLBuffer(tmpBuffSize, true);
+
+                ((MTLInstanceAccelerationStructureDescriptor*)descriptor).instanceDescriptorBuffer = tmpBuff->_mtlBuffer;
+                ((MTLInstanceAccelerationStructureDescriptor*)descriptor).instanceDescriptorBufferOffset = tmpBuff->_offset;
+
+                // Dispatch compute pipeline to convert instance data
+                uint32_t srcStride = sizeof(VkAccelerationStructureInstanceKHR);
+                uint32_t instanceCount = ranges[0].primitiveCount;
+                id<MTLComputeCommandEncoder> mtlConvertEncoder = cmdEncoder->getMTLComputeEncoder(kMVKCommandUseBuildAccelerationStructureConvertBuffers);
+                id<MTLComputePipelineState> mtlConvertState = cmdEncoder->getCommandEncodingPool()->getCmdBuildAccelerationStructureConvertBuffersMTLComputePipelineState();
+                [mtlConvertEncoder setComputePipelineState: mtlConvertState];
+                [mtlConvertEncoder setBuffer: mvkInstancesBuffer->getMTLBuffer()
+                                      offset: bOffset
+                                     atIndex: 0];
+                [mtlConvertEncoder setBuffer: tmpBuff->_mtlBuffer
+                                      offset: tmpBuff->_offset
+                                     atIndex: 1];
+                cmdEncoder->setComputeBytes(mtlConvertEncoder,
+                                            &srcStride,
+                                            sizeof(srcStride),
+                                            2);
+                cmdEncoder->setComputeBytes(mtlConvertEncoder,
+                                            &instanceCount,
+                                            sizeof(instanceCount),
+                                            3);
+
+                if (cmdEncoder->getMetalFeatures().nonUniformThreadgroups) {
+                    [mtlConvertEncoder dispatchThreads: MTLSizeMake(instanceCount, 1, 1)
+                                 threadsPerThreadgroup: MTLSizeMake(mtlConvertState.threadExecutionWidth, 1, 1)];
+                } else {
+                    [mtlConvertEncoder dispatchThreadgroups: MTLSizeMake(mvkCeilingDivide<NSUInteger>(instanceCount, mtlConvertState.threadExecutionWidth), 1, 1)
+                                      threadsPerThreadgroup: MTLSizeMake(mtlConvertState.threadExecutionWidth, 1, 1)];
+                }
+    
+                // We (probably) need to insert an explicit fence here. It is not exactly clear what is going on, but based on
+                // the Metal debugger, Metal does not recognize the read-write dependency between the tmpBuff and the build command.
+                // Thus, we need to explicitly synchronize.
+                fence = [cmdEncoder->getMTLDevice() newFence];
+                [mtlConvertEncoder updateFence:fence];
+                [cmdEncoder->_mtlCmdBuffer addCompletedHandler: ^(id<MTLCommandBuffer>) { [fence release]; }];
+            }
+
+            id<MTLAccelerationStructureCommandEncoder> accStructEncoder = cmdEncoder->getMTLAccelerationStructureEncoder(kMVKCommandUseBuildAccelerationStructure);
+
+            if (fence)
+                [accStructEncoder waitForFence:fence];
 
             [accStructEncoder buildAccelerationStructure:dstAccStruct
-                                                descriptor:descriptor
-                                                scratchBuffer:scratchBuffer
-                                                scratchBufferOffset:scratchBufferOffset];
+                                              descriptor:descriptor
+                                           scratchBuffer:scratchBuffer
+                                     scratchBufferOffset:scratchBufferOffset];
+
+            [descriptor release];
         }
-        else if (buildInfo.mode == VK_BUILD_ACCELERATION_STRUCTURE_MODE_UPDATE_KHR)
-        {
+        else if (buildInfo.mode == VK_BUILD_ACCELERATION_STRUCTURE_MODE_UPDATE_KHR) {
+            // TODO: check if we are allowed to update, not sure if validation layers handle this
+
             MTLAccelerationStructureDescriptor* descriptor = [MTLAccelerationStructureDescriptor new];
+
+            MVKAccelerationStructure* mvkSrcAccStruct = (MVKAccelerationStructure*)buildInfo.srcAccelerationStructure;
+            id<MTLAccelerationStructure> srcAccStruct = mvkSrcAccStruct->getMTLAccelerationStructure();
             
             if (mvkIsAnyFlagEnabled(buildInfo.flags, VK_BUILD_ACCELERATION_STRUCTURE_PREFER_FAST_BUILD_BIT_KHR))
                 descriptor.usage += MTLAccelerationStructureUsagePreferFastBuild;
             
+            id<MTLAccelerationStructureCommandEncoder> accStructEncoder = cmdEncoder->getMTLAccelerationStructureEncoder(kMVKCommandUseBuildAccelerationStructure);
             [accStructEncoder refitAccelerationStructure:srcAccStruct
                                               descriptor:descriptor
                                              destination:dstAccStruct
                                            scratchBuffer:scratchBuffer
                                      scratchBufferOffset:scratchBufferOffset];
+
+            [descriptor release];
         }
     }
 }
@@ -111,9 +162,9 @@ void MVKCmdBuildAccelerationStructure::encode(MVKCommandEncoder* cmdEncoder) {
 #pragma mark -
 #pragma mark MVKCmdCopyAccelerationStructure
 
-VkResult MVKCmdCopyAccelerationStructure::setContent(MVKCommandBuffer*                  cmdBuff,
-                                                     VkAccelerationStructureKHR         srcAccelerationStructure,
-                                                     VkAccelerationStructureKHR         dstAccelerationStructure,
+VkResult MVKCmdCopyAccelerationStructure::setContent(MVKCommandBuffer* cmdBuff,
+                                                     VkAccelerationStructureKHR srcAccelerationStructure,
+                                                     VkAccelerationStructureKHR dstAccelerationStructure,
                                                      VkCopyAccelerationStructureModeKHR copyMode) {
     
     MVKAccelerationStructure* mvkSrcAccStruct = (MVKAccelerationStructure*)srcAccelerationStructure;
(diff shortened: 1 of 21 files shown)


AntarticCoder:
Acceleration Structure and Raytracing in general does not seem to be supported before MacOS 11, so Xcode 11.7 will always fail.

cdavis5e:
> Acceleration Structure and Raytracing in general does not seem to be supported before MacOS 11, so Xcode 11.7 will always fail.

Then the parts of MoltenVK that deal with Acceleration Structures need to be inside `MVK_XCODE_12` blocks.

billhollings:
> > Acceleration Structure and Raytracing in general does not seem to be supported before MacOS 11, so Xcode 11.7 will always fail.
> 
> Then the parts of MoltenVK that deal with Acceleration Structures need to be inside `MVK_XCODE_12` blocks.

Agreed. But before forcing that, we should discuss whether that makes sense at this point. Xcode 11 is now 4 years old, and at some point, we must give up support for it, from a practicality perspective (like this one). Retaining support for Xcode 11 was added a couple of years ago because some devs required it for their internal processes.

A few months ago, [I reached out to the community](https://github.com/KhronosGroup/MoltenVK/discussions/1850) about this exact question, and received no responses. Unless we can determine a good reason for maintaining Xcode 11, maybe now is the time to drop support for it.

AntarticCoder:
MacOS 11 seems to have support from devices as old as 2013 and newer. So it's a matter of dropping support of these pre-2013 devices, as well as that, some people stay on MacOS 10 for support of 32 bit applications and other reasons. This is just something to take into consideration.

billhollings:
> A few months ago, [I reached out to the community](https://github.com/KhronosGroup/MoltenVK/discussions/1850) about this exact question, and received no responses. Unless we can determine a good reason for maintaining Xcode 11, maybe now is the time to drop support for it.

I have added a [ping post](https://github.com/KhronosGroup/MoltenVK/discussions/1850#discussioncomment-6386483) to that feedback request thread.

@AntarticCoder Hold off wrapping your code in any `MVK_XCODE_12` guards while this PR remains a WIP. When this PR is ready to go, based on any feedback we receive to my query ping, we can decide whether we need to actually implement those guard wraps, or abandon Xcode 11.

AntarticCoder:
@billhollings Alright, I'll hold off on the `MVK_XCODE_12` guards. Thanks

billhollings:
> @billhollings MacOS 11 seems to have support from devices as old as 2013 and newer. So it's a matter of dropping support of these pre-2013 devices, as well as that, some people stay on MacOS 10 for support of 32 bit applications and other reasons. This is just something to take into consideration.

The `MVK_XCODE_12` guard is strictly for API compilation during MoltenVK builds (ie- _will it build with the Metal API supported by Xcode 11_). Support for older OS runtimes is handled independently, through things like `respondsToSelector:`.

AntarticCoder:
@billhollings Ah yes my mistake. Also just a thought but only about 120 people actually watch this repository, so I'm not sure how many people will see your message.

K0bin:
> and when you call vkGetBufferDeviceAddress, we could push the address along with buffer

It's probably better to do that at buffer creation time and keep vkGetBufferDeviceAddress fast.

AntarticCoder:
@K0bin But not every created buffer will be used via the device address. So if you pushed it on `vkGetBufferDeviceAddress`, you would effectivly be keeping uneeded buffers out of the map.

K0bin:
Base it off of VK_BUFFER_USAGE_SHADER_DEVICE_ADDRESS_BIT.

AntarticCoder:
Okay then, I'll get started on the implementation. Thanks @K0bin

billhollings:
> Base it off of VK_BUFFER_USAGE_SHADER_DEVICE_ADDRESS_BIT.

Search for `VK_BUFFER_USAGE_SHADER_DEVICE_ADDRESS_BIT` in existing MoltenVK code. There is already an `MVKSmallVector` containing a list of these in `MVKDevice::_gpuAddressableBuffers`. Perhaps this could be modified to use a `std::unordered_map` and used to serve both purposes?

AntarticCoder:
@billhollings That seems like a good idea, I'll go ahead and use that for now, and we can change it in the future if it's not getting the job done.

AntarticCoder:
6/12 commands have been implemented, so I'm halfway there. 🎉

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
```diff
@@ -3640,6 +3640,20 @@ static uint32_t mvkGetEntryProperty(io_registry_entry_t entry, CFStringRef prope
 	*pMaxDeviation = cpuEnd - cpuStart;
 }
 
+MVKBuffer* MVKDevice::getBufferAtAddress(uint64_t address)
```
K0bin, line 3642: This doesn't cut it.

What if the application tries to use a VkBuffer at an offset?
It's perfectly legal to add an offset to the VA as long as its still within a buffer.

AntarticCoder: @K0bin I'm lost on what we would need to allow for offsets when getting the address.

K0bin: Here's what my toy renderer does for example: 
Allocate a single huge VkBuffer for all vertex data in the scene and get the BDA of that.
Then when I build the BLAS, I calculate the VA for where my geometry actually sits which is the BDA + offset.
That then gets passed to the BLAS build. So if you just look for the base BDA of a buffer, you wouldn't find anything in that case.

AntarticCoder: Thanks @K0bin, I almost forgot that I'm storing buffer addresses in a map, so I can't access offset addresses smoothly. My only question is how you get the ending of the address for the buffer?

AntarticCoder: Thanks, I didn't think it was that simple.

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
```diff
@@ -3641,16 +3641,24 @@ static uint32_t mvkGetEntryProperty(io_registry_entry_t entry, CFStringRef prope
 
 MVKBuffer* MVKDevice::getBufferAtAddress(uint64_t address)
 {
-    auto mvkBufIt = _gpuBufferAddressMap.find(address);
+    lock_guard<mutex> lock(_rezLock);
     
-    if(mvkBufIt != _gpuBufferAddressMap.end())
+    std::unordered_map<std::pair<uint64_t, uint64_t>, MVKBuffer*>::iterator it;
```
K0bin, line 3648: Not particularly pretty (a named struct instead of std::pair would improve readability) and pretty slow but that should work now.

rcaridade145: Can't we treat the BDA + offset as a buffer? Instead of just storing the ranges?

I mean if it is a "view" of a buffer can't we add it to the unordered_map?

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
AntarticCoder: > EDIT: I read this as "why would we need to allow for offsets".
> 
> Here's what my toy renderer does for example: 
> Allocate a single huge VkBuffer for all vertex data in the scene and get the BDA of that.
> Then when I build the BLAS, I calculate the VA for where my geometry actually sits which is the BDA + offset.
> That then gets passed to the BLAS build. So if you just look for the base BDA of a buffer, you wouldn't find anything in that case.
> 
> For what to do: store buffer ranges and iterate over them to find one that contains the requested address. To speed it up, you could combine this with a map where you mask off some of the lower bits of the VA as a key.

This buffer is just a single giant buffer, which is then suballocated to then accessed via the offset. I'm not sure how we would go about adding these suballocated buffers as regular buffers.

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
AntarticCoder: > EDIT: I read this as "why would we need to allow for offsets".
> 
> Here's what my toy renderer does for example: 
> Allocate a single huge VkBuffer for all vertex data in the scene and get the BDA of that.
> Then when I build the BLAS, I calculate the VA for where my geometry actually sits which is the BDA + offset.
> That then gets passed to the BLAS build. So if you just look for the base BDA of a buffer, you wouldn't find anything in that case.
> 
> For what to do: store buffer ranges and iterate over them to find one that contains the requested address. To speed it up, you could combine this with a map where you mask off some of the lower bits of the VA as a key.

This buffer is just a single giant buffer, which is then suballocated to then accessed via the offset. I'm not sure how we would go about adding these suballocated buffers as regular buffers.

rcaridade145: I meant that just to increase the hit rate . If i understood correctly we're always using the base buffer and then always calculating to find the inner buffer.  
This reminds me of Java ByteBuffers and the method slice - https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/nio/ByteBuffer.html#slice() .

K0bin: @rcaridade145 I think you misunderstand the problem.

The application passes a VA (virtual address) and Metal needs a buffer object. It's essentially a raw pointer for the GPU that has to be mapped back to an opaque buffer object.

AntarticCoder:
How should we store the device address of an acceleration structure. I was thinking we could give the acceleration structures a seperate "imaginary" address space, which we could then refer to then if we need to get it in the future.

Basically, we would create a map in `MVKDevice` that holds acceleration structure addreses, and refer to it when it is needed. We would also not need to hold the address ranges of the acceleration structure, because as far as I'm aware, we don't need an offset for acceleration structures.

AntarticCoder:
Could anyone give me some feedback on the memory address system for acceleration structure. I'd just like a second pair of eyes on the implementation. 

Thanks

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
```diff
@@ -4165,6 +4251,30 @@ static uint32_t mvkGetEntryProperty(io_registry_entry_t entry, CFStringRef prope
...
+    return accStruct;
+}
+
+void MVKDevice::removeAccelerationStructure(MVKAccelerationStructure* accStruct) {
+    std::unordered_map<uint64_t, MVKAccelerationStructure*>::iterator accStructIt = _gpuAccStructAddressMap.find(accStruct->getDeviceAddress());
+    uint64_t addressOffset = accStructIt->second->getMTLSize();
+    _gpuAccStructAddressMap.erase(accStructIt);
+    
+    // This can lead to fragmentation over time, so I'll just push all keys after this back
+    // This, however is also another performance issue
```
cdavis5e, line 4268: I wonder, then, if it wouldn't be better to use an ordered (RB-tree-based) `std::map<K, V>` instead.

MoltenVK/MoltenVK/GPUObjects/MVKDevice.h
```diff
@@ -907,6 +931,9 @@ class MVKDevice : public MVKDispatchableVulkanAPIObject {
 	MVKSmallVector<MVKSmallVector<MVKQueue*, kMVKQueueCountPerQueueFamily>, kMVKQueueFamilyCount> _queuesByQueueFamilyIndex;
 	MVKSmallVector<MVKResource*, 256> _resources;
 	MVKSmallVector<MVKBuffer*, 8> _gpuAddressableBuffers;
+    std::unordered_map<MVKBufferAddressRange, MVKBuffer*, MVKHash_uint64_t_pair> _gpuBufferAddressMap;
```
cdavis5e, line 934: The perfectionist in me wants to use a data structure specifically designed for mapping ranges to values. There's gotta be such a structure out there somewhere...

cdavis5e (commented):
Well, your implementation so far looks pretty straightforward.

MoltenVK/MoltenVK/GPUObjects/MVKDevice.h
rcaridade145: Something like a IntervalTree https://en.m.wikipedia.org/wiki/Interval_tree ? Java can use something like RangeMap https://www.baeldung.com/guava-rangemap . In cpp i believe boost has something.

AntarticCoder: Boost is probaly to heavy for use in MoltenVK, unless it's standalone. I found an interval tree implementation on github [here](https://github.com/5cript/interval-tree). This could be something to consider for the future, but I'm not sure if we have to implement it immediately.

MoltenVK/MoltenVK/Commands/MVKCmdAccelerationStructure.mm
```diff
@@ -0,0 +1,207 @@
...
+VkResult MVKCmdCopyMemoryToAccelerationStructure::setContent(MVKCommandBuffer* cmdBuff,
+                                                                uint64_t srcAddress,
+                                                                VkAccelerationStructureKHR dstAccelerationStructure,
+                                                                VkCopyAccelerationStructureModeKHR copyMode) {
+    MVKAccelerationStructure* mvkDstAccStruct = (MVKAccelerationStructure*)dstAccelerationStructure;
+    _mvkDevice = mvkDstAccStruct->getDevice();
+    _srcAddress = srcAddress;
+    _copyMode = copyMode;
+    
+    MVKAccelerationStructure* mvkSrcAccStruct = (MVKAccelerationStructure*)_mvkDevice->getAccelerationStructureAtAddress(_srcAddress);
```
cdavis5e, line 185: And *vice versa* for `vkCmdCopyMemoryToAccelerationStructure()`. Note that here, the `mode` must be `VK_COPY_ACCELERATION_STRUCTURE_MODE_DESERIALIZE_KHR`.

cdavis5e (changes requested):

MoltenVK/MoltenVK/Commands/MVKCmdAccelerationStructure.mm
```diff
@@ -0,0 +1,171 @@
...
+                                                             VkAccelerationStructureKHR         srcAccelerationStructure,
+                                                             uint64_t                           dstAddress,
+                                                             VkCopyAccelerationStructureModeKHR copyMode) {
+    
+    MVKAccelerationStructure* mvkSrcAccStruct = (MVKAccelerationStructure*)srcAccelerationStructure;
+    _mvkDevice = mvkSrcAccStruct->getDevice();
+    _dstAddress = dstAddress;
+    _copyMode = copyMode;
+    
+    MVKAccelerationStructure* mvkDstAccStruct = (MVKAccelerationStructure*)_mvkDevice->getAccelerationStructureAtAddress(_dstAddress);
```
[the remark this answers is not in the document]
AntarticCoder: Oh, that was a bit of dumb mistake, I probaly should read the documentation closer. Anyway I'll get that fixed.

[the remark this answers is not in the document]
AntarticCoder: @cdavis5e Should I copy the acceleration structure into the buffer using [`newBufferWithBytes:length:options:`](https://developer.apple.com/documentation/metal/mtldevice/1433429-newbufferwithbytes?language=objc), then just read the buffer when needed?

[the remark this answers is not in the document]
AntarticCoder: @cdavis5e How should I implement `VkCmdCopyAccelerationStructureToMemory`. It's easy enough to read the data from the buffer, but how would I upload data to the buffer. Could I create a new MTLbuffer and then copy that "staging buffer" to the target buffer?

MoltenVK/MoltenVK/Commands/MVKCmdAccelerationStructure.mm
AntarticCoder: @cdavis5e I've resolved this

MoltenVK/MoltenVK/Commands/MVKCmdAccelerationStructure.mm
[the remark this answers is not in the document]
cdavis5e: > How should I implement `VkCmdCopyAccelerationStructureToMemory`. It's easy enough to read the data from the buffer, but how would I upload data to the buffer. Could I create a new MTLbuffer and then copy that "staging buffer" to the target buffer?

No. You would get the buffer underlying the destination memory address from the `MVKDevice`, similar to how you were trying to get the acceleration structure, and you would copy to that.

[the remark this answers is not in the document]
AntarticCoder: @cdavis5e Could you check if the way I'm doing it now is correct.

MoltenVK/MoltenVK/Commands/MVKCmdAccelerationStructure.mm
```diff
@@ -0,0 +1,200 @@
...
+                                        
+void MVKCmdCopyAccelerationStructureToMemory::encode(MVKCommandEncoder* cmdEncoder) {
+    id<MTLAccelerationStructureCommandEncoder> accStructEncoder = cmdEncoder->getMTLAccelerationStructureEncoder(kMVKCommandUseNone);
+    _mvkDevice = cmdEncoder->getDevice();
+    
+    if(_copyMode != VK_COPY_ACCELERATION_STRUCTURE_MODE_SERIALIZE_KHR || !_dstBuffer->getDeviceMemory()->isMemoryHostAccessible()){
+        return;
+    }
+    
+    memcpy(_dstBuffer->getDeviceMemory()->getHostMemoryAddress(), (void*)_srcAccelerationStructure, sizeof(_srcAccelerationStructure));
```
cdavis5e, line 161: The copy is supposed to be performed on the device, especially since it is possible that one or both of the resources are in device-local (i.e. `MTLStorageModePrivate`) memory. The `-[MTLBlitCommandEncoder copyFromBuffer:sourceOffset:toBuffer:destinationOffset:size:]` method is what you want here.

Also, you'll need a method on the `MVKAccelerationStructure` to get its underlying `MVKBuffer`. And, you'll need to make sure that the `MTLAccelerationStructure` and `MTLBuffer` underlying both objects have their underlying GPU storage aliased to each other for this to work right. The only way I can think of to ensure that is to use a placement `MTLHeap`. Unfortunately, this means this functionality will now be limited to macOS Ventura or iOS 16, in which the `-[MTLHeap newAccelerationStructureWithDescriptor:offset:]` method you need for that was introduced.

Note that, although MoltenVK has support for using placement `MTLHeap`s, it's currently disabled due to some rendering artifacts with render targets in heaps. (I really need to check if that's been fixed.) But that doesn't matter, because you need to use an `MTLHeap` even in the dedicated allocation case.

MoltenVK/MoltenVK/Commands/MVKCmdAccelerationStructure.mm
```diff
@@ -0,0 +1,200 @@
...
+    
+    void* serializedAccStruct = _srcBuffer->getHostMemoryAddress();
+    if(!serializedAccStruct){
+        return; // Should I remove this? For this to work, the memory can't be device only, but the spec does not seem to restrict this
+    }
+
+    MVKAccelerationStructure* mvkSrcAccStruct = (MVKAccelerationStructure*)serializedAccStruct;
+    id<MTLAccelerationStructure> srcAccelerationStructure = mvkSrcAccStruct->getMTLAccelerationStructure();
+    
+    [accStructEncoder
```
cdavis5e, line 197: Er, uh, you're still assuming the destination resource has a `VkAccelerationStructure` created at that address. You'll need to make an analogous change here to the one you need to make for `vkCmdCopyAccelerationStructureToMemory()` above.

MoltenVK/MoltenVK/Commands/MVKCmdAccelerationStructure.mm
```diff
@@ -0,0 +1,200 @@
...
+    
+    _infoCount = infoCount;
+    _geometryInfos = &geoInfo;
+    _buildRangeInfos = *ppBuildRangeInfos;
+    
+    return VK_SUCCESS;
+}
+
+void MVKCmdBuildAccelerationStructure::encode(MVKCommandEncoder* cmdEncoder) {
+    id<MTLAccelerationStructureCommandEncoder> accStructEncoder = cmdEncoder->getMTLAccelerationStructureEncoder(kMVKCommandUseNone);
```
cdavis5e, line 43: You should add an enumerator to `MVKCommandUse` for each of these new commands. This is used to set the debug label for the command encoder, which aids debugging.

cdavis5e (changes requested):

MoltenVK/MoltenVK/Commands/MVKCmdAccelerationStructure.mm
K0bin: If I understand this correctly, you're proposing to implement Acceleration Structure copies using memory aliasing with a buffer and copying the buffer contents.

Why not just AccelerationStructureCommandEncoder::copy?

cdavis5e: This particular copy is for copying (i.e. serializing) acceleration structures to buffers.

billhollings:
@AntarticCoder @cdavis5e

This PR seems to have stalled. I've received a request from a game studio using MoltenVK who would like to see `VK_KHR_acceleration_structure` completed, and is willing to fund that work.

If either of you have spare time, and are interested in receiving compensation to work on finishing `VK_KHR_acceleration_structure`, let me know and I'll put you in touch with them. Their schedule is not rushed, so this is something that could start anytime in the next month or two, and be something that could be fit in part-time.

This sponsor is actually interested in seeing general ray tracing added, so if you (or anyone else out there), is interested in working on a funded project to see the following completed (in order or priority, and similar schedule behavior as above), please let me know:

- `VK_KHR_acceleration_structure`
- `VK_KHR_ray_tracing_pipeline`
- `VK_KHR_ray_query`
- `VK_KHR_pipeline_library`

AntarticCoder:
@billhollings 

I’m sorry, I’ve been busy with since I have began school and never got around to finishing this PR. I am interested in receiving compensation for finishing up `VK_KHR_acceleration_structure`. I also would be interested in working on general ray-tracing as well. Could you somehow get me into contact with the game studio?  

Thanks

cdavis5e:
I'm interested in this. I've talked with Holochip, and they're also interested.

billhollings:
> @billhollings
> 
> I’m sorry, I’ve been busy with since I have began school and never got around to finishing this PR. I am interested in receiving compensation for finishing up `VK_KHR_acceleration_structure`. I also would be interested in working on general ray-tracing as well. Could you somehow get me into contact with the game studio?
> 
> Thanks

@AntarticCoder 

I think it definitely makes sense to have you working on completing this PR. Can you shoot me an email at support@brenwill.com, and we'll sort things out. On your email, can you quote an hourly rate you'd like, how much time you have available, and where you are located (for how to best get you actually paid), please?

AntarticCoder:
@billhollings 

Just sent you an email.

zmarlon:
Is there any news regarding this PR? I now own an M3 Max Macbook and could test if this is of interest.

kanerogers:
Our game studio is interested in cross-platform ray tracing with Vulkan, wondering whether there's been any progress here.

K0bin:
How do you intend to work around the fact that Metal needs a list of all bottom level acceleration structures to build the TLAS while Vulkan only needs a GPU buffer address that contains that data?

MoltenVK/MoltenVK/Vulkan/vulkan.mm
```diff
@@ -2865,6 +2866,114 @@ MVK_PUBLIC_VULKAN_SYMBOL VkResult vkSetPrivateData(
...
+    VkDevice                                              device,
+    VkAccelerationStructureBuildTypeKHR                   buildType,
+    const VkAccelerationStructureBuildGeometryInfoKHR*    pBuildInfo,
+    const uint32_t*                                       pMaxPrimitiveCounts,
+    VkAccelerationStructureBuildSizesInfoKHR*             pSizeInfo) {
+    
+    MVKTraceVulkanCallStart();
+    MVKDevice* mvkDev = (MVKDevice*)device;
+    VkAccelerationStructureBuildSizesInfoKHR buildSizes = MVKAccelerationStructure::getBuildSizes(mvkDev, buildType, pBuildInfo);
+    pSizeInfo = &buildSizes;
```
pablode, line 2923: pSizeInfo should be dereferenced

KingVulpes:
Would this work in something like RTX Remix? which uses a modified dxvk to translate DirectX to Vulkan RT. It would be interesting to see something like Portal RTX converted over for Mac

watbulb:
> Would this work in something like RTX Remix? which uses a modified dxvk to translate DirectX to Vulkan RT. It would be interesting to see something like Portal RTX converted over for Mac

Given that MoltenVK targets Apple family of devices, and RTX Remix is not.
No. Also hello.

cdavis5e:
> Given that MoltenVK targets Apple family of devices, and RTX Remix is not.
> No.

Never say "never"...

watbulb:
> > Given that MoltenVK targets Apple family of devices, and RTX Remix is not.
> > No.
> 
> Never say "never"...

heh, I work on the RTX remix codebase a lot, this isn't on the roadmap. But sure, maybe one day ... :)

KingVulpes:
> > > Given that MoltenVK targets Apple family of devices, and RTX Remix is not.
> > > No.
> > 
> > Never say "never"...
> 
> heh, I work on the RTX remix codebase a lot, this isn't on the roadmap. But sure, maybe one day ... :)

Well I wasn't implying the RTX Remix devs would integrate it, given RTX Remix is open sourced, I was interested if it could be ported over for personal purposes

watbulb:
> > > > Given that MoltenVK targets Apple family of devices, and RTX Remix is not.
> > > > No.
> > > 
> > > 
> > > Never say "never"...
> > 
> > 
> > heh, I work on the RTX remix codebase a lot, this isn't on the roadmap. But sure, maybe one day ... :)
> 
> Well I wasn't implying the RTX Remix devs would integrate it, given RTX Remix is open sourced, I was interested if it could be ported over for personal purposes

And I am here to tell you that it is unlikely, considering the it would require RTX remix to be ported to apple.

fknfilewalker:
> How do you intend to work around the fact that Metal needs a list of all bottom level acceleration structures to build the TLAS while Vulkan only needs a GPU buffer address that contains that data?

With Metal 4 there is a good chance that it is now possible to directly specify a  buffer containing the AS descriptor instead of the previous list of handles + index.

See [var instanceDescriptorBuffer: MTL4BufferRange](https://developer.apple.com/documentation/metal/mtl4instanceaccelerationstructuredescriptor/instancedescriptorbuffer)

JarrettSJohnson:
Is there a chance we could get a rebase on this PR? I have some changes (~1000 lines or so) that I'd like to apply to this PR/branch but it require some things from `main`.

(Make some adjustments for it to run anyway--sent the PR to your repo.)

billhollings:
> Is there a chance we could get a rebase on this PR? I have some changes (~1000 lines or so) that I'd like to apply to this PR/branch but it require some things from `main`.
> 
> (Make some adjustments for it to run anyway--sent the PR to your repo.)

@AntarticCoder: If you're not actively working on this PR anymore on your own repo, perhaps someone like @JarrettSJohnson could take it over via a separate combined PR to this repo.

TheApplePieGod:
I was working with @AntarticCoder a bit ago on this PR, and there were some changes that never got merged. I spoke with @JarrettSJohnson on their PR, and decided to rebase the current branch and include the other changes we currently had.

AntarticCoder:
@billhollings How would we do that? Would we close out this PR and then open up a new one from @JarrettSJohnson or something else? Thanks

JarrettSJohnson:
As long as I can update @AntarticCoder 's branch, I don't mind just adding to there.

AntarticCoder:
@JarrettSJohnson Great. Would you like me to add you as a collaborator to my fork?

PR #1968, opened 2023-07-07
Support the `VK_EXT_pipeline_creation_feedback` extension.

cdavis5e:
This provides feedback that indicates:
* how long it took to compile each shader stage and the pipeline as a whole;
* whether or not the pipeline or any shader stage were found in any supplied pipeline cache; and
* whether or not any supplied base pipeline were used to accelerate pipeline creation.

This is similar to the performance statistics that MoltenVK already collects.

Since we don't use any supplied base pipeline at all, this implementation never sets
`VK_PIPELINE_CREATION_FEEDBACK_BASE_PIPELINE_ACCELERATION_BIT`. However, I've identified several places where we could probably use the base pipeline to accelerate pipeline creation. One day, I should probably implement that.

Likewise, because we don't yet support using `MTLBinaryArchive`s, `VK_PIPELINE_CREATION_FEEDBACK_APPLICATION_PIPELINE_CACHE_HIT_BIT` is never set on the whole pipeline, though it *is* set for individual stages, on the assumption that any shader found in a cache is likely to be found in Metal's own implicit cache.

In this implementation, shader stage compilation time includes any time needed to build the `MTLComputePipelineState`s needed for vertex and tessellation control shaders in tessellated pipelines.

This patch also changes compilation of the vertex stage `MTLComputePipelineState`s in tessellated pipelines to be eager instead of lazy. We really ought to have been doing this anyway, in order to report pipeline failures at creation time instead of draw time. I'm not happy, though, that we now pay the cost of all three pipeline states all the time, instead of just the ones that are used.

This also gets rid of some fields of `MVKGraphicsPipeline` that were only used during pipeline construction, which should save some memory, particularly for apps that create lots of pipelines.

### DIFF
--- a/MoltenVK/MoltenVK/GPUObjects/MVKPipeline.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKPipeline.mm
@@ -337,42 +337,54 @@ bool MVKGraphicsPipeline::supportsDynamicState(VkDynamicState state) {
 
 static const char vtxCompilerType[] = "Vertex stage pipeline for tessellation";
 
-id<MTLComputePipelineState> MVKGraphicsPipeline::getTessVertexStageState() {
-    MTLComputePipelineDescriptor* plDesc = [_mtlTessVertexStageDesc copy];  // temp retain a copy to be thread-safe.
-    plDesc.computeFunction = _mtlTessVertexFunctions[0];
-    id<MTLComputePipelineState> plState = getOrCompilePipeline(plDesc, _mtlTessVertexStageState, vtxCompilerType);
-    [plDesc release];                                                       // temp release
-    return plState;
-}
+bool MVKGraphicsPipeline::compileTessVertexStageState(MTLComputePipelineDescriptor* vtxPLDesc,
+													  id<MTLFunction>* vtxFunctions,
+													  VkPipelineCreationFeedback* pVertexFB) {
+	uint64_t startTime = 0;
+    if (pVertexFB) {
+		startTime = mvkGetTimestamp();
+	}
+    vtxPLDesc.computeFunction = vtxFunctions[0];
+    bool res = !!getOrCompilePipeline(vtxPLDesc, _mtlTessVertexStageState, vtxCompilerType);
 
-id<MTLComputePipelineState> MVKGraphicsPipeline::getTessVertexStageIndex16State() {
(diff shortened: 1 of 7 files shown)


cdavis5e:
@AntarticCoder Heads up--when you implement `VK_KHR_ray_tracing_pipeline`, you'll now need to support this as well.

AntarticCoder:
Thanks I'll keep that in mind

MoltenVK/MoltenVK/GPUObjects/MVKPipeline.mm
```diff
@@ -337,42 +337,54 @@
...
+    vtxPLDesc.computeFunction = vtxFunctions[2];
+    vtxPLDesc.stageInputDescriptor.indexType = MTLIndexTypeUInt32;
+    res |= !!getOrCompilePipeline(vtxPLDesc, _mtlTessVertexStageIndex32State, vtxCompilerType);
+
+	if (pVertexFB) {
+		if (!res) {
+			// Compilation of the shader will have enabled the flag, so I need to turn it off.
+			mvkDisableFlags(pVertexFB->flags, VK_PIPELINE_CREATION_FEEDBACK_VALID_BIT);
+		}
+		pVertexFB->duration += (mvkGetTimestamp() - startTime) * mvkGetTimestampPeriod();
```
billhollings, line 369: `(mvkGetTimestamp() - startTime) * mvkGetTimestampPeriod()` appears a lot as copy-paste. 

Can we move this to `MVKOSExtensions.h/mm` as a function similar to `mvkGetElapsedMilliseconds()`? 

For example:

```
uint64_t mvkGetElapsedTime(uint64_t startTimestamp = 0, uint64_t endTimestamp = 0);

uint64_t mvkGetElapsedTime(uint64_t startTimestamp, uint64_t endTimestamp) {
	if (endTimestamp == 0) { endTimestamp = mvkGetTimestamp(); }
	return (endTimestamp - startTimestamp) * _mvkTimestampPeriod;
}
```

billhollings (changes requested):
Thanks for doing this. LGTM. Just one quibble about some repetitive code pasting.

billhollings (approved):
Thanks. LGTM.

`mvkGetElapsedNanoseconds()` - Nice. Much more appropriate than the name than I had suggested.

Issue #1874, opened 2023-04-22
Commands using mtlComputeEnc in MVKCmdTransfer.mm polluting the tracked encoder state?


mbechard:
I have an error that seems incorrect, where I have an argument buffer of size 12 at index 1, but metal is saying only 4 bytes are bound to it. Tracing through the code, I can see the code is avoiding re-binding at that index because it thinks it's not dirty.
However just before this command, I'm using `vkCmdFillBuffer`, which I can see is calling setBytes with a size of 4 for index 1.
I do have a command before the vkCmdFillBuffer that is binding the buffer to index 1, which is why the state tracking thinks it's already bound with the correct size of 12.

Isn't `vkCmdFillBuffer` (and likely other similar ones in the file), changing the encoder state, but those changes aren't being tracked in `MVKResourcesCommandEncoderState`?

billhollings:
PR #1922 should fix this. Please retest with latest MoltenVK and close this issue if it corrects the problem.

mbechard:
This commit seems to introduce a regression, as I now get
`-[MTLDebugComputeCommandEncoder dispatchThreadgroups:threadsPerThreadgroup:]:1047: failed assertion `computePipelineState must not be nil.'`
On startup in my app, which worked fine before. Seems like there is a new case where the compute pipeline encoder isn't executing.

PR #1980, opened 2023-07-19
Fix regression caused by #1922

mbechard:
Needed to complete fix for #1874

We can't wait until getMTLComputeEncoder() is called to dirty the state, because this call will be avoided by dirty checks themselves. Those checks are comparing against leftover and now incorrect state since the previous encoder has already ended.
It needs to be dirtied on encoder end.

### DIFF
--- a/MoltenVK/MoltenVK/Commands/MVKCommandBuffer.mm
+++ b/MoltenVK/MoltenVK/Commands/MVKCommandBuffer.mm
@@ -839,6 +839,10 @@ void MVKCommandEncoder::endMetalRenderEncoding() {
 void MVKCommandEncoder::endCurrentMetalEncoding() {
 	endMetalRenderEncoding();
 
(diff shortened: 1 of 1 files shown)


billhollings (changes requested):
Good catch. Thanks for submitting this.

Just one change request. Now that `endCurrentMetalEncoding()` always marks the compute states as dirty, can you change the line:

https://github.com/KhronosGroup/MoltenVK/blob/42fa5117d94dcdb3bda9ccff5ea1a3cf88445500/MoltenVK/MoltenVK/Commands/MVKCommandBuffer.mm#L859

to the opposite logic:

```
	markCurrentComputeStateDirty = false;	// Already marked dirty above in endCurrentMetalEncoding()
```

Marking dirty, particularly on resource state, can be expensive, so no point in doing it twice.

mbechard:
ok done, thanks

PR #1981, opened 2023-07-19
Fix bad access regression to prematurely-deallocated MTLFunction.

billhollings:
When compiling tessellation vertex shaders, `MVKGraphicsPipeline` pass array of `MVKMTLFunction` instead of `MTLFunctions` to retain `MTLFunctions` for duration of processing.

Fixes a regression introduced in PR #1968.


cdavis5e (approved):
Huh... I have to admit, I was worried about the lifetime of the `MTLFunction` objects when I wrote the original change, but no problem ever showed up when I ran the `dEQP-VK.tessellation.*` tests. I guess I should've listened to my instincts there.

Anyway, thanks for catching and fixing this.

billhollings:
> no problem ever showed up when I ran the `dEQP-VK.tessellation.*` tests

Huh. Strange, because it was quite consistent on the test that caught it, which FWIW was...

`dEQP-VK.pipeline.monolithic.push_constant.graphics_pipeline.overlap_4_shaders_vert_tess_frag`

PR #1982, opened 2023-07-28
Don't fetch a submodule if the commit is already known

stefand:
This allows rebuilding without an internet connection and avoids random build failures on flaky connections.


billhollings (approved):
Good idea! Thanks for the submission.

Issue #1983, opened 2023-07-31
Move CPU/GPU timestamp correlation to vkGetPhysicalDeviceProperties()


billhollings:
It [turns out](https://github.com/KhronosGroup/MoltenVK/issues/1979#issuecomment-1658806275) that `MTLDevice sampleTimestamps:gpuTimestamp:` executes synchronously with other queue activities, and can block GPU execution if it is called frequently.

Move timestamp correlation to the execution of `vkGetPhysicalDeviceProperties()` so it is called only on demand by the app.

Furthermore, `MTLDevice sampleTimestamps:gpuTimestamp:` should only be invoked if all of these conditions are met (otherwise the `VkPhysicalDeviceLimits::timestampPeriod` should be set to `1.0`):

- CPU is not Apple Silicon (`!supportsMTLGPUFamily(Apple1)`).
- `MVKPhysicalDevice::_timestampMTLCounterSet` exists.
-  `MVKPhysicalDeviceMetalFeatures::counterSamplingPoints` is empty.

billhollings:
PR #1992 fixes this issue. Please retest to verify, and close this issue if it fixes the problem you encountered.

PR #1992, opened 2023-08-12
Fix sync delay between calls to vkQueueSubmit() on non-Apple-Silicon devices.

billhollings:
The `[MTLDevice sampleTimestamps:gpuTimestamp:]` function turns out to be synchronized with other queue activities, and can block GPU execution if it is called between `MTLCommandBuffer` submissions. On non-Apple-Silicon devices, it was called before and after every `vkQueueSubmit()` submission, to track the correlation between GPU and CPU timestamps, and was delaying the start of GPU work on the next submission (on Apple Silicon, both CPU & GPU timestamps are specified in nanoseconds, and the call was bypassed).

Move timestamp correlation from `vkQueueSubmit()` to `vkGetPhysicalDeviceProperties()`, where it is used to update `VkPhysicalDeviceLimits::timestampPeriod` on non-Apple-Silicon devices.

Delegate `MVKPhysicalDevice::getProperties(VkPhysicalDeviceProperties2*)` to `MVKPhysicalDevice::getProperties(VkPhysicalDeviceProperties*)`, plus minimize wasted effort if `pNext` is empty (unrelated).

Move the declaration of several `MVKPhysicalDevice` member structs to potentially reduce member spacing (unrelated).

Fixes issue #1983.

### DIFF
--- a/Docs/Whats_New.md
+++ b/Docs/Whats_New.md
@@ -363,6 +363,7 @@ Released 2022/02/07
 - Fix issue where *MSL 2.3* only available on *Apple Silicon*, even on *macOS*.
 - Fix memory leak of dummy `MTLTexture` in render subpasses that use no attachments.
 - Fix Metal object retain-release errors in assignment operators.
+- Fix sync delay between calls to `vkQueueSubmit()` on non-Apple-Silicon devices.
 - Fix use of GPU counter sets on older versions of iOS running on the simulator.
 - `mvk::getShaderOutputs()` in `SPRIVReflection.h` support flattening nested structures.
 - Replaced ASL logging levels with `MVKConfigLogLevel`.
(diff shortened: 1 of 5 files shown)


cdavis5e (approved):

SRSaunders:
Quick follow-up.  It may be beneficial to document this change with a brief release note that recommends refreshing `physicalDeviceProperties` from time to time to ensure `physicalDeviceProperties.limits.timestampPeriod` is updated with up-to-date runtime values for non-Apple GPUs.  I believe non-Apple GPUs have different power states that can change over time and result in different timestamp calibrations.  I can see this happening in RBDoom3BFG with my AMD 6600XT GPU.  I now need to call `vkGetPhysicalDeviceProperties()` periodically to refresh `timestampPeriod`.  No big deal, and this does not seems to interfere with encoding start and the reason for the fix, but something to be aware of nonetheless.  Of course this is a non-issue for Apple GPUs.

billhollings:
> It may be beneficial to document this change

Good suggestion. I've added a note in `MoltenVK_Runtime_Users_Guide.md` to that effect.

PR #2008, opened 2023-08-31
Fix crash in Dota 2 on macOS < 10.15 - closes #2006

### LINKED ISSUE #2006: Dota 2 crashing after updating MoltenVK on macOS 10.13/10.14
See https://github.com/ValveSoftware/Dota-2/issues/2438, the source of the problem appears to be:

```
2023-08-31 14:20:09.344 dota2[34750:13062847] *** Terminating app due to uncaught exception 'NSInvalidArgumentException', reason: '-[BronzeMtlDevice sampleTimestamps:gpuTimestamp:]: unrecognized selector sent to instance 0x7fd098010000'
```

It looks like this is a regression calling sampleTimestamps on macOS < 10.15: https://developer.apple.com/documentation/metal/mtldevice/3194378-sampletimestamps

I think I see the problem, I'll issue a PR in a bit.

danginsburg:


danginsburg:
Users have confirmed that this fixed the crash, can you please merge?

billhollings:
Thanks. Good catch. Apologies for the recent regression.

Issue #1990, opened 2023-08-10
Flickering horizontal line output on macOS 12


Steveice10:
Over in Citra we've been getting reports since starting testing for our new Vulkan backend that specifically macOS 12 users are frequently getting strange flickering horizontal line video output covering the entire render area.

I don't know if anyone has tested on macOS 11 (our minimum), but no one on macOS 13 or above has encountered this issue. We haven't seen it on any other platform either.

We have tried enabling the Metal validation layer and didn't get any errors out, and we don't think any of the Vulkan validation errors we've seen are related. Primarily because they've come and gone over the course of development on this backend, but this issue remains constant.

I'm not sure if this is a MoltenVK issue, a Metal bug, or something about our usage of Vulkan, but I figured it would be a good idea to reach out and see if you guys have any insights into why this might be happening.

Issue describing the problem: https://github.com/citra-emu/citra/issues/6562

Source with Vulkan support: https://github.com/GPUCode/citra/tree/vulkan-but-better

Builds with Vulkan support: https://github.com/citra-emu/citra-canary/releases

Example screenshot from a user:
![image](https://github.com/KhronosGroup/MoltenVK/assets/1269164/54849e70-a16b-4480-bd35-559c9a27428c)

Steveice10:
There's also a video someone sent us a while back demonstrating the issue **(warning: rapid flashing)**:

https://github.com/KhronosGroup/MoltenVK/assets/1269164/d69a2f4a-2fd2-4328-ae97-d6dc0c1b7ff3

billhollings:
Is this happening across multiple users, or just the user that filed the Citra issue you posted above?

What value are you setting in `VkSwapchainCreateInfoKHR::presentMode`? Using `VK_PRESENT_MODE_IMMEDIATE_KHR` may sometimes cause tearing.

Steveice10:
> Is this happening across multiple users, or just the user that filed the Citra issue you posted above?
> 
> What value are you setting in `VkSwapchainCreateInfoKHR::presentMode`? Using `VK_PRESENT_MODE_IMMEDIATE_KHR` may sometimes cause tearing.

It’s happening to several users, always ones running macOS 12.

If the user has Vsync enabled we use FIFO. Otherwise we check for Mailbox, and if that isn’t present we use Immediate. According to the user log they have Vsync enabled, so they should be using FIFO.

Steveice10:
[...]

PR #2012, opened 2023-09-06
Guard against CAMetalDrawable with invalid pixel format.

billhollings:
- Calling `nextDrawable` may result in a nil drawable, or a drawable with no pixel format. Attempt several times to retrieve a drawable with a valid pixel format, and if unsuccessful, return an error from `vkQueuePresentKHR()` and `vkAcquireNextImageKHR()`, to force swapchain to be re-created.
- Reorganize `MVKQueuePresentSurfaceSubmission::execute()` to detect drawable with invalid format, attach `MTLCommandBuffer` completion handler just before commit, and delay enqueuing `MTLCommandBuffer` until commit.
- Refactor `mvkOSVersionIsAtLeast()` for clarity (unrelated).

Possible fix for #1990.

### DIFF
--- a/Common/MVKOSExtensions.h
+++ b/Common/MVKOSExtensions.h
@@ -39,27 +39,30 @@ static const MVKOSVersion kMVKOSVersionUnsupported = std::numeric_limits<MVKOSVe
 MVKOSVersion mvkOSVersion();
 
 /** Returns a MVKOSVersion built from the version components. */
-inline MVKOSVersion mvkMakeOSVersion(uint32_t major, uint32_t minor, uint32_t patch) {
+static inline MVKOSVersion mvkMakeOSVersion(uint32_t major, uint32_t minor, uint32_t patch) {
 	return (float)major + ((float)minor / 100.0f) + ((float)patch / 10000.0f);
 }
 
 /** Returns whether the operating system version is at least minVer. */
-inline bool mvkOSVersionIsAtLeast(MVKOSVersion minVer) { return mvkOSVersion() >= minVer; }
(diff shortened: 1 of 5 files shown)


Common/MVKOSExtensions.h
```diff
@@ -39,27 +39,30 @@ static const MVKOSVersion kMVKOSVersionUnsupported = std::numeric_limits<MVKOSVe
...
 /** Returns whether the operating system version is at least minVer. */
-inline bool mvkOSVersionIsAtLeast(MVKOSVersion minVer) { return mvkOSVersion() >= minVer; }
+static inline bool mvkOSVersionIsAtLeast(MVKOSVersion minVer) { return mvkOSVersion() >= minVer; }
 
 /**
  * Returns whether the operating system version is at least the appropriate min version.
- * The constant kMVKOSVersionUnsupported can be used for either value to cause the test
- * to always fail on that OS, which is useful for indidicating functionalty guarded by
+ * The constant kMVKOSVersionUnsupported can be used for any of the values to cause the test
+ * to always fail on that OS, which is useful for indidicating that functionalty guarded by
```
cdavis5e, line 52: Since you're making changes to this comment...
```suggestion
 * to always fail on that OS, which is useful for indicating that functionalty guarded by
```

cdavis5e (approved):

Common/MVKOSExtensions.h
billhollings: Thanks. Hmmm...it would be cool if Xcode spell-checked comments.

Issue #2011, opened 2023-09-03
Wrong fragCmdBlitImage shader in 1.2.5 version


rokuz:
I started to get the following error:
```
Validation layer:
VK_ERROR_INITIALIZATION_FAILED: Could not compile support shader from MSL source (Error code 3):
#include <metal_stdlib>
using namespace metal;

typedef struct {
    float4 v_position [[position]];
    float3 v_texCoord;
} VaryingsPosTex;

typedef struct {
    float depth [[depth(any)]];
} FragmentOutputs;

typedef struct {
    uint slice;
    float lod;
} TexSubrez;

constexpr sampler ce_sampler(mip_filter::nearest, filter::nearest);
fragment FragmentOutputs fragCmdBlitImage(VaryingsPosTex varyings [[stage_in]],
                         depth2d<float> tex [[texture(0)]],
                         constant TexSubrez& subRez [[buffer(0)]]) {
    FragmentOutputs out;
    out.depth = tex.sample(ce_sampler, varyings.v_texCoord.xy, level(subRez.lod)).x;
    return out;
}

program_source:23:82: error: member reference base type 'float' is not a structure or union
    out.depth = tex.sample(ce_sampler, varyings.v_texCoord.xy, level(subRez.lod)).x;
                ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~
```

# After some investigation:

This commit https://github.com/KhronosGroup/MoltenVK/commit/5d173d0e093eed2b42994332ab2c735ff0c8eb8a introduced swizzling for depth textures:
```
		if (mvkIsAnyFlagEnabled(blitKey.srcAspect, (VK_IMAGE_ASPECT_DEPTH_BIT))) {
			[msl appendFormat: @"    out.depth = tex.sample(ce_sampler, varyings.v_texCoord%@%@, level(subRez.lod)).%c;", coordArg, sliceArg, swizzleArg[0]];
			[msl appendLineMVK];
		}
```
The problem is sampling from `depth2d<float>` gives `float`. See [MSL spec](https://developer.apple.com/metal/Metal-Shading-Language-Specification.pdf) page 214.

cc @cdavis5e

billhollings:
PR #2018 fixes this.

PR #2015, opened 2023-09-09
Ensure objects retained for life of MTLCommandBuffer.

### LINKED ISSUE #2014: Switch vkQueuePresent to use a retainedRefs command buffer?
In the last two Dota releases, we've encountered crashes with the Steam Overlay.  The most recent one was reproducible on Apple M1 and occurred without any Metal Validation errors (https://github.com/ValveSoftware/Dota-2/issues/2449).  The first time this happened on 03/24/23, I worked around it locally by having `MVKQueuePresentSurfaceSubmittion::execute` add a parameter true to getMTLCommandBuffer on this line:

https://github.com/KhronosGroup/MoltenVK/blob/473ea0c2f18e4d13389cfe1412b4042c5f3b83ba/MoltenVK/MoltenVK/GPUObjects/MVKQueue.mm#L625

...so that the command buffer was created with retainedReferences.  I shipped that fix, but then after some more investigation we found a bug in the Steam Overlay (caught by Metal validation) where the Steam Overlay command buffer was relying on the command buffer to retain references to its MTLRenderPassDescriptor.  We fixed that by doing a retain on it and releasing it in a MtlCommandBuffer completion handler.

This time (our latest update shipping with latest MoltenVK) with no changes we are aware of to the Steam Overlay a similar crash happened on Apple M1 (the bug linked above). There were no validation errors so I don't know what the problem could be with the Overlay.  It was intermittent.  However, I again applied the local fix to use retainedReferences on the vkQueuePresent command buffer and that worked around it.

I'm going to need to carry that workaround locally now.  I wonder if it should just be the default behavior?  It seems there is little downside and it will prevent crashes/hangs with any misbehaving injected layers?

billhollings:
- `vkCmdBlitImage()` ensure swizzle texture view is retained for life of `MTLCommandBuffer`.
- `vkQueuePresentKHR()` use `MTLCommandBuffer` that retains references.
- Update MoltenVK version to 1.2.6.

Fixes #2014.

### DIFF
--- a/Docs/Whats_New.md
+++ b/Docs/Whats_New.md
@@ -20,6 +20,7 @@ Released TBD
 
 - Fix rare case where vertex attribute buffers are not bound to Metal 
   when no other bindings change between pipelines.
+- Ensure objects retained for life of `MTLCommandBuffer` during `vkCmdBlitImage()` & `vkQueuePresentKHR()`.
 - Fix case where a `CAMetalDrawable` with invalid pixel format causes onscreen flickering.
 - Improve behavior of swapchain image presentation stalls caused by Metal regression.
 - Add several additional performance trackers, available via logging, or the `mvk_private_api.h` API.
--- a/MoltenVK/MoltenVK/API/mvk_config.h
+++ b/MoltenVK/MoltenVK/API/mvk_config.h
@@ -45,7 +45,7 @@ extern "C" {
  */
 #define MVK_VERSION_MAJOR   1
 #define MVK_VERSION_MINOR   2
-#define MVK_VERSION_PATCH   5
+#define MVK_VERSION_PATCH   6
 
 #define MVK_MAKE_VERSION(major, minor, patch)    (((major) * 10000) + ((minor) * 100) + (patch))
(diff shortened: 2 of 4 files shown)


cdavis5e (approved):

cdavis5e:
```/me smacks forehead yet again```

How did I miss *that*?? Anyway, thanks for cleaning up yet another one of the broken parts of my changes. There's still one other part that I know is broken (#2011), but I'm hoping to get to fixing it soon.

danginsburg:
LGTM, thanks for the fix!

billhollings:
> `/me smacks forehead yet again`
> 
> How did I miss _that_?? Anyway, thanks for cleaning up yet another one of the broken parts of my changes. 

No worries. The CTS tests were actually passing, despite the validation errors. I just cleaned it up.

> There's still one other part that I know is broken (#2011), but I'm hoping to get to fixing it soon.

Actually, I'm going through a cleansing phase to clean up a few small bugs right now. I'm planning to have a look at #2011 in the process of that in the next day or two.

Issue #2013, opened 2023-09-07
Remove or disable timestampPeriod lowpass filter on non-Apple GPUs


SRSaunders:
Is there any way to disable the `timestampPeriod` low pass filter that was implemented right at the end of the 1.2.5 release cycle (commit 530bde199b0e8ca3b20e0f7c0b6fb08d27d8284b)?

See the following new code in `MVKPhysicalDevice::updateTimestampsAndPeriod()`:

```
// Basic lowpass filter Y = (1 - a)Y + a*X.
// The lower a is, the slower Y will change over time.
static const float a = 0.05;
_properties.limits.timestampPeriod = ((1.0 - a) * _properties.limits.timestampPeriod) + (a * tsPeriod);
```

This negatively affects my embedded Optick profiler integration which depends on an accurate point-in-time measurement of `timestampPeriod` for GPU/CPU timeline calibration just before beginning a profiling trace. Averaging this value ruins the timeline calibration and GPU traces are now misaligned with CPU traces.

Could this filter be preferably removed or even disabled via a config setting (e.g. on/off or setting the `a` value to 1.0)?

billhollings:
Huh. Interesting. 

Since `vkGetPhysicalDeviceProperties()` is mostly about fixed platform characteristics, and extracts a ton of info, I didn't expect it to be something that an app would be calling frequently, so I figured volatile point-in-time values would be unhelpful, hence the averaging.

Perhaps we can add a config option for the `a` value, to allow an app to choose how volatile it wants the `timestampPeriod` to be?

SRSaunders:
> Perhaps we can add a config option for the `a` value, to allow an app to choose how volatile it wants the `timestampPeriod` to be?

That would work for me.  Thx.

billhollings:
PR #2017 add a configurable alpha for the `timestampPeriod`. 

To disable the filter, set env var `MVK_CONFIG_TIMESTAMP_PERIOD_LOWPASS_ALPHA=1.0`.

If this works for your needs, please close this issue.

SRSaunders:
Thanks for implementing this.  It looks like it will solve my issue but unfortunately I cannot test for 2-3 weeks since I am currently away on vacation until the end of Sept.  I will test and hopefully close when I return.

SRSaunders:
[...]

PR #2017, opened 2023-09-12
Add configurable lowpass filter for VkPhysicalDeviceLimits::timestampPeriod.

billhollings:
- Add `MVKConfiguration::timestampPeriodLowPassAlpha`, along with matching `MVK_CONFIG_TIMESTAMP_PERIOD_LOWPASS_ALPHA` env var.
- Add `MVKConfigMembers.def` file to describe `MVKConfiguration` members, to support consistent batch handling of members.
- Add env var & build settings `MVK_CONFIG_DEBUG`, plus legacy `MVK_CONFIG_ALLOW_METAL_EVENTS` & `MVK_CONFIG_ALLOW_METAL_FENCES`.
- Simplify environment variable retrieval functions and macros.
- Rename `MVKDevice::updateTimestampsAndPeriod()` to `updateTimestampPeriod()`.

Fixes issue #2013.

### DIFF
--- a/MoltenVK/MoltenVK/API/mvk_config.h
+++ b/MoltenVK/MoltenVK/API/mvk_config.h
@@ -191,7 +191,7 @@ typedef struct {
 	 * and the changed value will immediately effect subsequent MoltenVK behaviour.
 	 *
 	 * The initial value or this parameter is set by the
-	 * MVK_DEBUG
+	 * MVK_CONFIG_DEBUG
 	 * runtime environment variable or MoltenVK compile-time build setting.
 	 * If neither is set, the value of this parameter is false if MoltenVK was
 	 * built in Release mode, and true if MoltenVK was built in Debug mode.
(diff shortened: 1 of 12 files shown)


MoltenVK/MoltenVK/API/mvk_config.h
```diff
@@ -191,7 +191,7 @@ typedef struct {
 	 * and the changed value will immediately effect subsequent MoltenVK behaviour.
 	 *
 	 * The initial value or this parameter is set by the
-	 * MVK_DEBUG
+	 * MVK_CONFIG_DEBUG
```
cdavis5e, line 194: This is a change in MoltenVK's external interface, and it should probably be documented somewhere.

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
```diff
@@ -1582,9 +1582,9 @@
 		if (elapsedCPUNanos && elapsedGPUTicks) {		// Ensure not zero
 			float tsPeriod = elapsedCPUNanos / elapsedGPUTicks;
 			
-			// Basic lowpass filter Y = (1 - a)Y + a*X.
-			// The lower a is, the slower Y will change over time.
-			static const float a = 0.05;
+			// Basic lowpass filter TPout = (1 - A)TPout + (A * TPin).
+			// The lower A is, the slower Y will change over time.
```
cdavis5e, line 1586:
```suggestion
			// The lower A is, the slower TPout will change over time.
```

MoltenVK/MoltenVK/Utility/MVKEnvironment.cpp
```diff
@@ -27,67 +27,35 @@ static void mvkInitConfigFromEnvVars() {
...
-	MVK_SET_FROM_ENV_OR_BUILD_INT32 (evCfg.apiVersionToAdvertise,                  MVK_CONFIG_API_VERSION_TO_ADVERTISE);
-	MVK_SET_FROM_ENV_OR_BUILD_INT32 (evCfg.advertiseExtensions,                    MVK_CONFIG_ADVERTISE_EXTENSIONS);
-	MVK_SET_FROM_ENV_OR_BUILD_BOOL  (evCfg.resumeLostDevice,                       MVK_CONFIG_RESUME_LOST_DEVICE);
-	MVK_SET_FROM_ENV_OR_BUILD_INT32 (evCfg.useMetalArgumentBuffers,                MVK_CONFIG_USE_METAL_ARGUMENT_BUFFERS);
-	MVK_SET_FROM_ENV_OR_BUILD_INT32 (evCfg.shaderSourceCompressionAlgorithm,       MVK_CONFIG_SHADER_COMPRESSION_ALGORITHM);
-	MVK_SET_FROM_ENV_OR_BUILD_BOOL  (evCfg.shouldMaximizeConcurrentCompilation,    MVK_CONFIG_SHOULD_MAXIMIZE_CONCURRENT_COMPILATION);
+#define STR(name) #name
+
+#define MVK_CONFIG_MEMBER(member, mbrType, name) \
+	evCfg.member = (mbrType)mvkGetEnvVarNumber(STR(MVK_CONFIG_##name), MVK_CONFIG_##name);
```
cdavis5e, line 33: ...Will the preprocessor expand this macro *before*, or *after* stringification?

The normal behavior is to expand macros *after* stringification, so that stringified macro names don't get expanded at all. I know the usual pattern is to do this:

```c++
#define __STR(a) #a
#define STR(a) __STR(a)
```
and that forces the preprocessor to expand the macro *before* stringification. Here, we don't want that, because we want the string `"MVK_CONFIG_FOO"` and not `"bar"` (given `#define MVK_CONFIG_FOO bar`). I'm just not sure that the preprocessor won't expand the macro anyway.

cdavis5e (changes requested):

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
billhollings: Good catch. Fixed.

MoltenVK/MoltenVK/API/mvk_config.h
billhollings: Good catch! Thanks for noticing this. I didn't fully deal with the legacy part of that change.

I've added additional documentation about the change, and some env var code to continue to accept `MVK_DEBUG` as a legacy env var.

MoltenVK/MoltenVK/Utility/MVKEnvironment.cpp
billhollings: Hmmm...interesting points.

This flow is working, but I can't say that I explicitly took into consideration the issues you raise, so it may be working by accident more than by design.

The design, and operation, of the flow is...

```
1. ##: MVK_CONFIG_##name --> MVK_CONFIG_DEBUG
2. STR(): MVK_CONFIG_DEBUG --> "MVK_CONFIG_DEBUG"
3. (VkBool32)mvkGetEnvVarNumber("MVK_CONFIG_DEBUG", MVK_CONFIG_DEBUG);
4. exp: (VkBool32)mvkGetEnvVarNumber("MVK_CONFIG_DEBUG", 0);
```

where `MVK_CONFIG_DEBUG` is not expanded until the last step, and only where it was not stringified.

There's an explanation [here](https://stackoverflow.com/a/8754665) that corroborates this as "by design", vis-a-vis: 

> Macro arguments are completely macro-expanded before they are substituted into a macro body, unless they are stringified or pasted with other tokens.

so `STR(a)` does not expand its argument because it stringifies it.

cdavis5e: All right, I'm satisfied.

cdavis5e (approved):

Issue #1514, opened 2022-01-25
Add support for VK_KHR_synchronization2


MennoVink:
Would it be possible to add support for this extension to mvk itself?

On windows i can choose to use the sdk layer if the driver doesn't support the extension. With mvk however this wont work for my app. There's two ways i can link mvk. Either statically as framework, or dynamically through the loader. In order to get access to layers on macos i need to link dynamically. I dont think i can do that because as i understand it [the runtime user guide](https://github.com/KhronosGroup/MoltenVK/blob/master/Docs/MoltenVK_Runtime_UserGuide.md) mentions this will make it so that i can only use vkGetMoltenVKConfigurationMVK and vkSetMoltenVKConfigurationMVK.
My application requires access to the other functions as well. My application has some low level metal code in order to make vulkan<->opengl interop work.
The functions i need:
- vkGetPhysicalDeviceMetalFeaturesMVK
- vkGetMTLDeviceMVK
- vkSetMTLTextureMVK
- vkGetMTLTextureMVK
In the future i might also need vkGetIOSurfaceMVK if/when opengl is dropped and [Syphon](https://github.com/Syphon/Syphon-Framework) needs to be implemented with the low level io surfaces or metal directly.

Are there any alternatives i could use to get both sync2 support as well as access to the metal id's that are used under the hood?

billhollings:
Thanks for this enhancement request.

We can certainly look to implement `VK_KHR_synchronization2`. It should be fairly straightforward, and implementable within the short term, although, as with existing synchronization, we may not be able to provide full Vulkan functionality.

--------------

> On windows i can choose to use the sdk layer if the driver doesn't support the extension

I'm confused about what is happening when you do this. How does the Vulkan layer environment implement `VK_KHR_synchronization2` if the underlying driver does not?

--------------

> `vkGetMoltenVKConfigurationMVK()` and `vkSetMoltenVKConfigurationMVK()`.

These should be currently useable even with the SDK, as long as `libMoltenVK.a/dylib` is linked, because the `VkInstance` parameter in these calls is actually ignored. In addition, you can use the equivalent environment variables during app launch and completely bypass using `vkGetMoltenVKConfigurationMVK()` and `vkSetMoltenVKConfigurationMVK()`

--------------

>`vkGetMTLDeviceMVK()`
>`vkSetMTLTextureMVK()`
>`vkGetMTLTextureMVK()`

[...]

PR #2021, opened 2023-09-19
Add support for extension VK_KHR_synchronization2

billhollings:
- `MVKPhysicalDevice` add support for `VkPhysicalDeviceSynchronization2Features`.
- Pass sync2 structs to `MVKPipelineBarrier`, `MVKCmdPipelineBarrier`, `MVKCmdSetEvent`, `MVKCmdResetEvent`, `MVKCmdWaitEvents`, `MVKRenderPass`, `MVKQueue` & `MVKQueueSubmission`.
- Replace use of `VkPipelineStageFlags` & `VkAccessFlags` with `VkPipelineStageFlags2` & `VkAccessFlags2`.
- Add stage masks to `MVKPipelineBarrier`, and redefine `apply*MemoryBarrier()` functions to remove separately passing stage masks.
- Add `MVKSemaphoreSubmitInfo` to track semaphores in `MVKQueueSubmission`.
- Add `MVKCommandBufferSubmitInfo` to track command buffers in `MVKQueueCommandBufferSubmission`.
- Add `MVKSubpassDependency` to combine `VkSubpassDependency` & `VkMemoryBarrier2` in `MVKRenderPass`.
- Remove abstract `MVKCmdSetResetEvent` superclass.
- Streamline code in `MVKMTLFunction::operator=` (unrelated).

Implements issue #1514.

### DIFF
--- a/MoltenVK/MoltenVK/API/mvk_datatypes.h
+++ b/MoltenVK/MoltenVK/API/mvk_datatypes.h
@@ -414,13 +414,13 @@ MTLWinding mvkMTLWindingFromSpvExecutionMode(uint32_t spvMode);
 MTLTessellationPartitionMode mvkMTLTessellationPartitionModeFromSpvExecutionMode(uint32_t spvMode);
 
 /**
- * Returns the combination of Metal MTLRenderStage bits corresponding to the specified Vulkan VkPiplineStageFlags,
+ * Returns the combination of Metal MTLRenderStage bits corresponding to the specified Vulkan VkPipelineStageFlags2,
  * taking into consideration whether the barrier is to be placed before or after the specified pipeline stages.
  */
-MTLRenderStages mvkMTLRenderStagesFromVkPipelineStageFlags(VkPipelineStageFlags vkStages, bool placeBarrierBefore);
+MTLRenderStages mvkMTLRenderStagesFromVkPipelineStageFlags(VkPipelineStageFlags2 vkStages, bool placeBarrierBefore);
 
-/** Returns the combination of Metal MTLBarrierScope bits corresponding to the specified Vulkan VkAccessFlags. */
-MTLBarrierScope mvkMTLBarrierScopeFromVkAccessFlags(VkAccessFlags vkAccess);
+/** Returns the combination of Metal MTLBarrierScope bits corresponding to the specified Vulkan VkAccessFlags2. */
+MTLBarrierScope mvkMTLBarrierScopeFromVkAccessFlags(VkAccessFlags2 vkAccess);
 
 #pragma mark -
 #pragma mark Geometry conversions
--- a/MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
@@ -320,6 +320,11 @@ void MVKPhysicalDevice::getFeatures(VkPhysicalDeviceFeatures2* features) {
 				subgroupSizeFeatures->computeFullSubgroups = _metalFeatures.simdPermute || _metalFeatures.quadPermute;
 				break;
 			}
+			case VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_SYNCHRONIZATION_2_FEATURES: {
+				auto* synch2Features = (VkPhysicalDeviceSynchronization2Features*)next;
+				synch2Features->synchronization2 = true;
+				break;
+			}
 			case VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_TEXTURE_COMPRESSION_ASTC_HDR_FEATURES: {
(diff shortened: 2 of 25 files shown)


MoltenVK/MoltenVK/API/mvk_datatypes.h
```diff
@@ -414,13 +414,13 @@ MTLWinding mvkMTLWindingFromSpvExecutionMode(uint32_t spvMode);
...
+ * Returns the combination of Metal MTLRenderStage bits corresponding to the specified Vulkan VkPipelineStageFlags2,
  * taking into consideration whether the barrier is to be placed before or after the specified pipeline stages.
  */
-MTLRenderStages mvkMTLRenderStagesFromVkPipelineStageFlags(VkPipelineStageFlags vkStages, bool placeBarrierBefore);
+MTLRenderStages mvkMTLRenderStagesFromVkPipelineStageFlags(VkPipelineStageFlags2 vkStages, bool placeBarrierBefore);
 
-/** Returns the combination of Metal MTLBarrierScope bits corresponding to the specified Vulkan VkAccessFlags. */
-MTLBarrierScope mvkMTLBarrierScopeFromVkAccessFlags(VkAccessFlags vkAccess);
+/** Returns the combination of Metal MTLBarrierScope bits corresponding to the specified Vulkan VkAccessFlags2. */
+MTLBarrierScope mvkMTLBarrierScopeFromVkAccessFlags(VkAccessFlags2 vkAccess);
```
cdavis5e, line 423: This is a breaking change.

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
```diff
@@ -4172,16 +4177,15 @@ static uint32_t mvkGetEntryProperty(io_registry_entry_t entry, CFStringRef prope
...
+void MVKDevice::applyMemoryBarrier(MVKPipelineBarrier& barrier,
 								   MVKCommandEncoder* cmdEncoder,
 								   MVKCommandUse cmdUse) {
-	if (!mvkIsAnyFlagEnabled(dstStageMask, VK_PIPELINE_STAGE_HOST_BIT) ||
-		!mvkIsAnyFlagEnabled(barrier.dstAccessMask, VK_ACCESS_HOST_READ_BIT) ) { return; }
 	lock_guard<mutex> lock(_rezLock);
 	for (auto& rez : _resources) {
-		rez->applyMemoryBarrier(srcStageMask, dstStageMask, barrier, cmdEncoder, cmdUse);
+		if (mvkIsAnyFlagEnabled(barrier.dstStageMask, VK_PIPELINE_STAGE_HOST_BIT) &&
+			mvkIsAnyFlagEnabled(barrier.dstAccessMask, VK_ACCESS_HOST_READ_BIT) ) {
```
cdavis5e, line 4186: This loop has long been a performance bottleneck (cf. #592, #1597). Is it really necessary to check that condition on every iteration, instead of checking once before the entire loop like we were before?

cdavis5e (changes requested):

MoltenVK/MoltenVK/API/mvk_datatypes.h
billhollings: > This is a breaking change.

It shouldn't be. The spec includes:

> The VkPipelineStageFlags2 bitmask goes beyond the 31 individual bit flags allowable within a C99 enum, which is how [VkPipelineStageFlagBits](https://registry.khronos.org/vulkan/specs/1.3-extensions/html/vkspec.html#VkPipelineStageFlagBits) is defined. The first 31 values are common to both, and are interchangeable.

> The VkAccessFlags2 bitmask goes beyond the 31 individual bit flags allowable within a C99 enum, which is how [VkAccessFlagBits](https://registry.khronos.org/vulkan/specs/1.3-extensions/html/vkspec.html#VkAccessFlagBits) is defined. The first 31 values are common to both, and are interchangeable.

So `VkPipelineStageFlagBits ` and `VkAccessFlagBits` should pass transparently, and be consistent with previous behaviour.

Am I missing something?

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
billhollings: Thanks! Good catch (again)!

I have no idea why I made that change...I can only guess that I must have brain-farted and thought I was iterating barriers instead of resources.

I've reverted it now.

MoltenVK/MoltenVK/API/mvk_datatypes.h
cdavis5e: On *x86*, loading a 32-bit register clears the upper 32 bits of the corresponding 64-bit register. Does *ARM* work that way, too?

/me looks

...apparently, it does. Never mind, then.

cdavis5e (approved):

PR #2025, opened 2023-09-27
Misc Cleanup

etang-cw:
Minor things I changed while working on other features
You can check the individual commits for details, the biggest one is a modification of MVKArrayRef so you use `MVKArrayRef<const X>` instead of `const MVKArrayRef<X>`, as you can't accidentally un-const the former by copying it into a local variable, (which you can do with ArrayRefs, since they don't own their storage)

### DIFF
--- a/MoltenVK/MoltenVK/Utility/MVKFoundation.h
+++ b/MoltenVK/MoltenVK/Utility/MVKFoundation.h
@@ -598,7 +598,7 @@ bool mvkAreEqual(const T* pV1, const T* pV2, size_t count = 1) {
  * which works on individual chars or char arrays, not strings.
  * Returns false if either string is null.
  */
-static constexpr bool mvkStringsAreEqual(const char* pV1, const char* pV2, size_t count = 1) {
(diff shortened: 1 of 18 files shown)


MoltenVK/MoltenVK/Utility/MVKFoundation.h
```diff
@@ -478,20 +478,21 @@ std::size_t mvkHash(const N* pVals, std::size_t count = 1, std::size_t seed = 53
...
+private:
+	Type* _data;
+	size_t _size;
+
+public:
+	constexpr Type* begin() const { return _data; }
+	constexpr Type* end() const { return &_data[_size]; }
+	constexpr Type* data() const { return _data; }
+	constexpr size_t size() const { return _size; }
+	constexpr size_t byte_size() const { return _size * sizeof(Type); }
```
billhollings, line 490: For consistency with code base, use camelCase: `byteSize()`.

billhollings (changes requested):
Thanks for the submission! Nice fixes.

LGTM. Just a couple of syntax nits.

billhollings (approved):
LGTM. Thanks!

PR #2029, opened 2023-10-03
In-shader vertex loading support

etang-cw:
MoltenVK counterpart to https://github.com/KhronosGroup/SPIRV-Cross/pull/2204 (Note: Won't pass CI until SPIRV-cross changes are merged)

Now fully supports all parts of Vulkan vertex loading
- No more minVertexInputBindingStrideAlignment
- No more restrictions on alignment of vertex buffer offsets (which there doesn't seem to be a portability flag for, but definitely didn't work before)
- Support (almost) all the vertex formats
- Support larger-than-stride elements
- Fully working divisor 0 support


cdavis5e:
> Any chance to push this forward?

It's held up on KhronosGroup/SPIRV-Cross#2204, and from what I've read lately, H-K probably thinks that's too invasive (i.e. it doesn't belong in SPIRV-Cross). You'll have to take it up with CW, though.

FernandoS27:
For us, it's a must have.

Issue #1930, opened 2023-06-01
Vulkan 1.3 requirements


billhollings:
[Appendix D of the Vulkan spec](https://registry.khronos.org/vulkan/specs/1.3-extensions/html/vkspec.html#versions-1.3) and the [VK_VERSION_1_3 man page](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VK_VERSION_1_3.html) summarize the extension list below, and adds additional Vulkan 1.3 enhancement requirements.

Extension | MoltenVK<br/>Support | Status Notes |
| :--| :-: |:-- 
|`VK_KHR_copy_commands2`| :white_check_mark: | |
|`VK_KHR_dynamic_rendering`| :white_check_mark: | |
|`VK_KHR_format_feature_flags2`| :white_check_mark: | Added (#2108) |
|`VK_KHR_maintenance4`| :white_check_mark: | Added (#2484) |
|`VK_KHR_shader_integer_dot_product`| :white_check_mark: | Added (#2119) |
|`VK_KHR_shader_non_semantic_info`| :white_check_mark: | |
|`VK_KHR_shader_terminate_invocation`| :white_check_mark: | Added ( #2424)  |
|`VK_KHR_synchronization2`| :white_check_mark: | Added (#2021) |
|`VK_KHR_zero_initialize_workgroup_memory`| :white_check_mark: | Added (#2450) |
|`VK_EXT_4444_formats`| :white_check_mark: | |
|`VK_EXT_extended_dynamic_state`| :white_check_mark: | Added (#2036) |
|`VK_EXT_extended_dynamic_state2`| :white_check_mark: | Added (#2036) |
|`VK_EXT_image_robustness`| :white_check_mark: | |
|`VK_EXT_inline_uniform_block`| :white_check_mark: | |
|`VK_EXT_pipeline_creation_cache_control`| :white_check_mark: | |
|`VK_EXT_pipeline_creation_feedback`| :white_check_mark: | |
|`VK_EXT_private_data`| :white_check_mark: | |
|`VK_EXT_shader_demote_to_helper_invocation`| :white_check_mark: | |
|`VK_EXT_subgroup_size_control`| :white_check_mark: | |
|`VK_EXT_texel_buffer_alignment`| :white_check_mark: | |
|`VK_EXT_texture_compression_astc_hdr`| :white_check_mark: | |
|`VK_EXT_tooling_info`| :white_check_mark: | Added (#2435) |
[...]

Issue #1739, opened 2022-10-10
Support VK_EXT_extended_dynamic_state, VK_EXT_extended_dynamic_state2 & VK_EXT_extended_dynamic_state3


billhollings:
Not all of the functionality of `VK_EXT_extended_dynamic_state`, `VK_EXT_extended_dynamic_state2` & `VK_EXT_extended_dynamic_state3` can be supported in Metal but the following dynamic state functionality can be supported:

`VK_EXT_extended_dynamic_state`:

- VK_DYNAMIC_STATE_CULL_MODE
- VK_DYNAMIC_STATE_FRONT_FACE
- VK_DYNAMIC_STATE_PRIMITIVE_TOPOLOGY
- VK_DYNAMIC_STATE_VIEWPORT_WITH_COUNT
- VK_DYNAMIC_STATE_SCISSOR_WITH_COUNT
- ~VK_DYNAMIC_STATE_VERTEX_INPUT_BINDING_STRIDE~ _(unsupported)_
- VK_DYNAMIC_STATE_DEPTH_TEST_ENABLE
- VK_DYNAMIC_STATE_DEPTH_WRITE_ENABLE
- VK_DYNAMIC_STATE_DEPTH_COMPARE_OP
- ~VK_DYNAMIC_STATE_DEPTH_BOUNDS_TEST_ENABLE~  _(unsupported but `VkPhysicalDeviceFeatures::depthBounds` disabled already)_
- VK_DYNAMIC_STATE_STENCIL_TEST_ENABLE
- VK_DYNAMIC_STATE_STENCIL_OP

`VK_EXT_extended_dynamic_state2`:

- VK_DYNAMIC_STATE_DEPTH_BIAS_ENABLE
- ~VK_DYNAMIC_STATE_PRIMITIVE_RESTART_ENABLE~ _(unsupported)_
- ~VK_DYNAMIC_STATE_RASTERIZER_DISCARD_ENABLE~ _(unsupported)_
- ~VK_DYNAMIC_STATE_LOGIC_OP_EXT~  _(unsupported but can be disabled and `VkPhysicalDeviceFeatures::logicOp` disabled already)_
- VK_DYNAMIC_STATE_PATCH_CONTROL_POINTS_EXT

`VK_EXT_extended_dynamic_state3` **_(each unsupported dynamic state can be individually disabled)_**:

- ~VK_DYNAMIC_STATE_ALPHA_TO_COVERAGE_ENABLE_EXT~
[...]

PR #2036, opened 2023-10-10
Add support for VK_EXT_extended_dynamic_state & VK_EXT_extended_dynamic_state2 extensions.

billhollings:
- Add `MVKRenderingCommandEncoderState` to consolidate handling
  of static and dynamic rasterizing states in a consistent manner.
- Rework `MVKDepthStencilCommandEncoderState` to consolidate handling
  of static and dynamic depth states in a consistent manner.
- `MVKMTLDepthStencilDescriptorData` clean up content setting, and struct layout.
- Add `MVKRenderStateType` to enumerate render state types.
- Add `MVKRenderStateFlags` to track binary info about states (enabled, dirty, etc).
- Add `MVKMTLBufferBinding::stride`.
- Add `MVKPhysicalDeviceMetalFeatures::dynamicVertexStride.
- Set `MVKPhysicalDeviceMetalFeatures::vertexStrideAlignment`
  to 1 for Apple5+ GPUs (unrelated).
- Set VkPhysicalDeviceLimits::maxVertexInputBindingStride
  to unlimited for Apple2+ GPUs (unrelated).
- Add `mvkVkRect2DFromMTLScissorRect()` and simplify
  `mvkMTLViewportFromVkViewport()` and `mvkMTLScissorRectFromVkRect2D()`.
- `MVKFoundation`:
  - Add `mvkEnableAllFlags(` and `mvkDisableAllFlags()`.
  - Improve performance of `mvkClear()`, `mvkCopy()` & `mvkAreEqual()`
    when content is a single simple primitive type (unrelated).
  - Declare more functions as `static constexpr` (unrelated).

- Add `MVKPipelineCommandEncoderState` subclasses
  `MVKGraphicsPipelineCommandEncoderState` & `MVKComputePipelineCommandEncoderState`,
  track patch control points in `MVKGraphicsPipelineCommandEncoderState`,
  and add `getGraphicsPipeline()` & `getComputePipeline()` to simplify casting.
- Rename `MVKCmdRenderPass.h/mm to MVKCmdRendering.h/mm.
- Move `MVKCmdExecuteCommands` from `MVKCmdRenderPass.h/mm` to `MVKCmdPipeline.h/mm`.
- While working on `vkCmdSetLogicOpEXT()`, add support for
  `vkCmdSetLogicOpEnableEXT()` from `VK_EXT_extended_dynamic_state3`.

Implements primary chunks of #1739 and impacts #1930.


MoltenVK/MoltenVK/API/mvk_datatypes.h
```diff
@@ -378,10 +378,13 @@ MTLMultisampleStencilResolveFilter mvkMTLMultisampleStencilResolveFilterFromVkRe
 #endif
 
 /** Returns the Metal MTLViewport corresponding to the specified Vulkan VkViewport. */
-MTLViewport mvkMTLViewportFromVkViewport(VkViewport vkViewport);
+MTLViewport mvkMTLViewportFromVkViewport(const VkViewport vkViewport);
```
cdavis5e, line 381: Why the `const` on a struct passed by value?

MoltenVK/MoltenVK/Commands/MVKCommandEncoderState.mm
```diff
@@ -337,130 +266,255 @@
...
+		if (_cmdEncoder->_pDeviceFeatures->multiViewport) {
+#if MVK_MACOS_OR_IOS
+			[rendEnc setViewports: mtlViewports.viewports count: mtlViewports.viewportCount];
+#endif
+		} else {
+			[rendEnc setViewport: mtlViewports.viewports[0]];
+		}
+	}
+
+	// If rasterizing discard has been dynamically enabled, or culling has been dynamically
```
cdavis5e, line 487: Clever hack.

MoltenVK/MoltenVK/Commands/MVKCommandEncoderState.mm
```diff
@@ -812,23 +871,32 @@
...
-                                                              b.mtlBytes,
-                                                              b.size,
-                                                              b.index);
+							   NSUInteger mtlStride = isDynamicVertexStride ? b.stride : MTLAttributeStrideStatic;
+							   if (b.isInline) {
+								   [cmdEncoder->_mtlRenderEncoder setVertexBytes: b.mtlBytes
+																		  length: b.size
+#if MVK_XCODE_15
+																 attributeStride: mtlStride
+#endif
```
cdavis5e, line 885: What happens if MoltenVK is built with Xcode 15, but deployed on a system older than macOS 14/iOS 17?

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
```diff
@@ -387,6 +387,18 @@
 				formatFeatures->formatA4B4G4R4 = canSupport4444;
 				break;
 			}
+			case VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_EXTENDED_DYNAMIC_STATE_FEATURES_EXT: {
+				auto* extDynState = (VkPhysicalDeviceExtendedDynamicStateFeaturesEXT*)next;
+				extDynState->extendedDynamicState = true;
```
cdavis5e, line 392: Can we advertise support for this prior to Metal 3.1? It looks like dynamic vertex strides are required by this feature.

MoltenVK/MoltenVK/GPUObjects/MVKPipeline.h
```diff
@@ -218,6 +218,49 @@ struct MVKStagedDescriptorBindingUse {
 	MVKBitArray stages[4] = {};
 };
 
+/** Enumeration identifying different state content types. */
+enum MVKRenderStateType {
```
cdavis5e, line 222: Should this be an `enum class`? I guess it doesn't matter so much since this isn't a public API...

MoltenVK/MoltenVK/GPUObjects/MVKPipeline.mm
```diff
@@ -509,40 +494,37 @@
...
-
 	// Blending - must ignore allowed bad pColorBlendState pointer if rasterization disabled or no color attachments
 	if (_isRasterizingColor && pCreateInfo->pColorBlendState) {
-		memcpy(&_blendConstants, &pCreateInfo->pColorBlendState->blendConstants, sizeof(_blendConstants));
+		mvkCopy(_blendConstants, pCreateInfo->pColorBlendState->blendConstants, 4);
+
+		// Metal does not support blending with logic operations.
+		if (pCreateInfo->pColorBlendState->logicOpEnable && pCreateInfo->pColorBlendState->logicOp != VK_LOGIC_OP_COPY) {
+			setConfigurationResult(reportError(VK_ERROR_FEATURE_NOT_PRESENT, "Metal does not support blending using logic operations."));
+		}
```
cdavis5e, line 504: There's no need to check this. We don't even report the `logicOp` feature to begin with. The validation layer should catch this.

MoltenVK/MoltenVK/GPUObjects/MVKPipeline.mm
```diff
@@ -509,40 +494,37 @@
...
-		if (_rasterInfo.depthClampEnable) {
-			if (_device->_enabledFeatures.depthClamp) {
-				_mtlDepthClipMode = MTLDepthClipModeClamp;
-			} else {
-				setConfigurationResult(reportError(VK_ERROR_FEATURE_NOT_PRESENT, "This device does not support depth clamping."));
-			}
+	_hasRasterInfo = mvkSetOrClear(&_rasterInfo, pCreateInfo->pRasterizationState);
+	if (_hasRasterInfo) {
+		if (_rasterInfo.depthClampEnable && !_device->_enabledFeatures.depthClamp) {
+			setConfigurationResult(reportError(VK_ERROR_FEATURE_NOT_PRESENT, "This device does not support depth clamping."));
```
cdavis5e, line 527: Again, this should be caught by the validation layer. I know this check was already present, but since you're changing it regardless...

MoltenVK/MoltenVK/GPUObjects/MVKPipeline.mm
```diff
@@ -557,26 +539,81 @@
...
+
+	for (uint32_t i = 0; i < pDS->dynamicStateCount; i++) {
+		auto dynStateType = getRenderStateType(pDS->pDynamicStates[i]);
+		bool isDynamic = true;
+
+		// Some dynamic states have other restrictions
+		switch (dynStateType) {
+			case VertexStride:
+				isDynamic = _device->_pMetalFeatures->dynamicVertexStride;
+				if ( !isDynamic ) { setConfigurationResult(reportError(VK_ERROR_FEATURE_NOT_PRESENT, "This device and platform does not support VK_DYNAMIC_STATE_VERTEX_INPUT_BINDING_STRIDE (macOS 14.0 or iOS/tvOS 17.0, plus either Apple4 or Mac2 GPU).")); }
```
cdavis5e, line 607: This check on the other hand is probably necessary because there's no way for the validation layer to catch it. Maybe we need another feature in the `VkPhysicalDevicePortabilityFeaturesKHR` struct.

MoltenVK/MoltenVK/Utility/MVKFoundation.h
```diff
@@ -384,24 +387,24 @@ static constexpr bool mvkVkComponentMappingsMatch(VkComponentMapping cm1, VkComp
...
 /** Rounds the value to nearest integer using half-to-even rounding. */
 static inline double mvkRoundHalfToEven(const double val) {
-	return val - std::remainder(val, 1.0);	// remainder() uses half-to-even rounding, and unfortunately isn't constexpr until C++23.
+	return val - std::remainder(val, 1.0);	// remainder() uses half-to-even rounding, but unfortunately isn't constexpr until C++23.
 }
 
 /** Returns whether the value will fit inside the numeric type. */
 template<typename T, typename Tval>
-const bool mvkFits(const Tval& val) {
+static constexpr const bool mvkFits(const Tval& val) {
```
cdavis5e, line 395: The `const` on the return type is useless.

MoltenVK/MoltenVK/Utility/MVKFoundation.h
```diff
@@ -278,21 +281,21 @@ void mvkFlipVertically(void* rowMajorData, uint32_t rowCount, size_t bytesPerRow
...
+static constexpr  int64_t kMVKUndefinedLargePositiveInt64 =  mvkEnsurePowerOfTwo(std::numeric_limits<int64_t>::max() / 2);
+static constexpr  int64_t kMVKUndefinedLargeNegativeInt64 = -kMVKUndefinedLargePositiveInt64;
+static constexpr uint64_t kMVKUndefinedLargeUInt64        =  kMVKUndefinedLargePositiveInt64;
 
 
 #pragma mark Vulkan structure support functions
 
 /** Returns a VkExtent2D created from the width and height of a VkExtent3D. */
-static inline VkExtent2D mvkVkExtent2DFromVkExtent3D(VkExtent3D e) { return {e.width, e.height }; }
+static constexpr VkExtent2D mvkVkExtent2DFromVkExtent3D(VkExtent3D e) { return {e.width, e.height }; }
```
cdavis5e, line 295: Does `constexpr` imply `inline`?

/me looks

...Apparently, it does. Carry on.

cdavis5e (changes requested):
Whew! What a large change!

MoltenVK/MoltenVK/API/mvk_datatypes.h
billhollings: Good point. Fixed.

MoltenVK/MoltenVK/Commands/MVKCommandEncoderState.mm
billhollings: Thanks. I was not at all sure it would work. But it got me through CTS successfully.

I am hopeful that Apple Silicon can optimize to apply the scissor test prior shader processing (without needing to force it with `early_fragment_tests`), which would help avoid any side-effects of shader execution (including performance degradation). It seems likely that it would optimize ownership and scissor tests that way.

MoltenVK/MoltenVK/GPUObjects/MVKPipeline.h
billhollings: Yeah. Our pattern for internal `enums` so far is to just use `enum`.

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
billhollings: Unlike DS3's`VkPhysicalDeviceExtendedDynamicState3FeaturesEXT `, `VkPhysicalDeviceExtendedDynamicStateFeaturesEXT` only has a single enablement flag, so the entire extension has a binary enablement.

Users had been asking for `VK_EXT_extended_dynamic_state`, knowing that Metal did not support dynamic vertex stride, and I figured it would be one of the least-used dynamic states anyway, so I figured it wouldn't be right to limit the entire extension based on vertex stride limitations.

When it comes to it, we'll have to introduce a portability flag. In the meantime, the limitation is documented.

MoltenVK/MoltenVK/GPUObjects/MVKPipeline.mm
billhollings: Yes, I guess this is redundant. Removed.

MoltenVK/MoltenVK/GPUObjects/MVKPipeline.mm
billhollings: Removed.

MoltenVK/MoltenVK/GPUObjects/MVKPipeline.mm
billhollings: > Maybe we need another feature in the VkPhysicalDevicePortabilityFeaturesKHR struct.

Yes. My thinking exactly. For the Vulkan 1.3 version of that struct.

MoltenVK/MoltenVK/Utility/MVKFoundation.h
billhollings: Thanks. Copy-pasto during mods.

spnda:
Will there be an exhaustive list of all the dynamic statea that are "invalidly" unsupported in the MVK documentation? I see some are noted in the original issue on these extension but it doesnt seem up to date anymore.

MoltenVK/MoltenVK/Commands/MVKCommandEncoderState.mm
billhollings: 🤦🏻‍♂️ Thanks! Good catch! 

I clearly tried to outsmart myself there, with minimizing the changes.

I've consolidated the changes into `MVKCommandEncoder::encodeVertexAttributeBuffer()`, to avoid rampant repeated spaghetti code.

billhollings:
> Whew! What a large change!

Thanks for all the feedback. I appreciate it.

> Will there be an exhaustive list of all the dynamic statea that are "invalidly" unsupported in the MVK documentation?

The list is actually not so exhaustive. There are only two real deviations from conformance (others are actually permitted in the spec via feature flags). I've updated `MoltenVK_Runtime_UserGuide.md` to document them. In the process, I've structured the extensions section of that document to make it easier to be consistent in specifying limitations for each extension in a readable way, and to encourage us devs to keep that up.

cdavis5e:
You didn't push anything new for me to review.

billhollings:
> You didn't push anything new for me to review.

Oops! Pushed now.

cdavis5e (approved):

Issue #2041, opened 2023-10-15
Dynamic vertex input binding stride breaks rendering 


tycho:
My application uses ANGLE + MoltenVK for rendering on macOS.

After updating to the current HEAD (61b871217838effcf74346509dba9f67109a8abd), my app was no longer rendering anything correctly. The framebuffer clear worked as intended, but that's about all. No other geometry renders.

Attaching the Xcode debugger with Metal API validation enabled results in this during the first draw or so:

```
validateCommonDrawErrors:instanceCount:baseInstance:maxVertexID:]:5708: failed assertion `Draw Errors Validation
providing `attributeStride` value of 64 when stride at buffer-layout index 25 is static.
providing `attributeStride` value of 64 when stride at buffer-layout index 26 is static.
providing `attributeStride` value of 64 when stride at buffer-layout index 27 is static.
providing `attributeStride` value of 64 when stride at buffer-layout index 28 is static.
providing `attributeStride` value of 16 when stride at buffer-layout index 29 is static.
providing `attributeStride` value of 16 when stride at buffer-layout index 30 is static.
```

I did discover a workaround for this problem. If I disable the ANGLE feature named `useVertexInputBindingStrideDynamicState`, all the remaining dynamic state functionality appears to function as intended and the game renders perfectly.

I am not certain how to debug this, or what further data I can provide. I can provide a demo app if needed to reproduce the problem.

tycho:
Here's a redistributable demo build of the game in question:

https://www.uplinklabs.net/files/Darwinia-macOS-demo-2.3.2-53-g7625f44ff.zip

You'll have to remove the `com.apple.quarantine` xattr in order for it to read the overrides.txt and enable the repro case. (By default the game currently uses a workaround to disable the feature that breaks rendering.)

cdavis5e:
Ah, dammit, I should've caught this in my review. When dynamic vertex strides are used, this has to be explicitly enabled in the vertex descriptor, by setting the stride there to `MTLBufferLayoutStrideDynamic`. (Conversely, when using the new array method to set the vertex buffer and the stride is *not* dynamic, the stride in the method must be `MTLAttributeStrideStatic`. But we don't use the array method yet.)

billhollings:
Thanks for submitting this, and the _**very helpful**_ demo app!

PR #2042 fixes this. Please retest with latest MoltenVK and close this issue if it fixes the issue you encountered.

billhollings:
[...]

PR #2042, opened 2023-10-16
Fixes for VK_EXT_extended_dynamic_state.

billhollings:
- `MVKPipeline` only work around zero stride if stride is static.
- Ensure dynamic vertex stride is not enabled on builds before Xcode 15.
- Add `MVKRenderStateType::LineWidth` for track all default options (unrelated).

Fixes issue #2041.

### DIFF
--- a/MoltenVK/MoltenVK/GPUObjects/MVKPipeline.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKPipeline.mm
@@ -560,6 +560,7 @@ static MVKRenderStateType getRenderStateType(VkDynamicState vkDynamicState) {
 		case VK_DYNAMIC_STATE_DEPTH_TEST_ENABLE:           return DepthTestEnable;
 		case VK_DYNAMIC_STATE_DEPTH_WRITE_ENABLE:          return DepthWriteEnable;
 		case VK_DYNAMIC_STATE_FRONT_FACE:                  return FrontFace;
+		case VK_DYNAMIC_STATE_LINE_WIDTH:                  return LineWidth;
 		case VK_DYNAMIC_STATE_LOGIC_OP_EXT:                return LogicOp;
 		case VK_DYNAMIC_STATE_LOGIC_OP_ENABLE_EXT:         return LogicOpEnable;
 		case VK_DYNAMIC_STATE_PATCH_CONTROL_POINTS_EXT:    return PatchControlPoints;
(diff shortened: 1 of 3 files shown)


MoltenVK/MoltenVK/GPUObjects/MVKPipeline.mm
```diff
@@ -1377,7 +1378,7 @@ static MTLVertexFormat mvkAdjustFormatVectorToSize(MTLVertexFormat format, uint3
 			maxBinding = max(pVKVB->binding, maxBinding);
 			uint32_t vbIdx = getMetalBufferIndexForVertexAttributeBinding(pVKVB->binding);
 			auto vbDesc = inputDesc.layouts[vbIdx];
-			if (pVKVB->stride == 0) {
+			if (isVtxStrideStatic && pVKVB->stride == 0) {
```
cdavis5e, line 1381: On line 1373 there should probably be another check for this. (I can't actually annotate 1373 with this comment.)

cdavis5e (changes requested):
Just one more check that you missed.

MoltenVK/MoltenVK/GPUObjects/MVKPipeline.mm
billhollings: Ah, yes. Thanks. Good catch. Fixed now.

billhollings: Ah yes. Thanks. Good catch. Fixed now.

cdavis5e (approved):

MoltenVK/MoltenVK/GPUObjects/MVKPipeline.mm
```diff
@@ -1366,18 +1367,16 @@ static MTLVertexFormat mvkAdjustFormatVectorToSize(MTLVertexFormat format, uint3
...
 
-			// Vulkan allows any stride, but Metal only allows multiples of 4.
-            // TODO: We could try to expand the buffer to the required alignment in that case.
-			VkDeviceSize mtlVtxStrideAlignment = _device->_pMetalFeatures->vertexStrideAlignment;
-            if ((pVKVB->stride % mtlVtxStrideAlignment) != 0) {
-				setConfigurationResult(reportError(VK_ERROR_INITIALIZATION_FAILED, "Under Metal, vertex attribute binding strides must be aligned to %llu bytes.", mtlVtxStrideAlignment));
+			// Vulkan allows any stride, but Metal requires multiples of 4 on older GPUs.
+            if (isVtxStrideStatic && (pVKVB->stride % _device->_pMetalFeatures->vertexStrideAlignment) != 0) {
+				setConfigurationResult(reportError(VK_ERROR_INITIALIZATION_FAILED, "Under Metal, vertex attribute binding strides must be aligned to %llu bytes.", _device->_pMetalFeatures->vertexStrideAlignment));
                 return false;
```
tycho, line 1373: FYI, there's some significant indentation inconsistency here.

billhollings: It's the typical tab-spacing issue. Everything is aligned when tab-spacing is set to 4. Github uses 8.

We never got around to mechanically enforce formatting. Formatting consistency is maintained by the devs, and some use spaces instead of tabs.

Issue #2045, opened 2023-10-17
New [mvk-warn] message: Metal does not support disabling primitive restart


SRSaunders:
Following commit 61b871217838effcf74346509dba9f67109a8abd MoltenVK now generates the following warning at runtime:

`[mvk-warn] VK_ERROR_FEATURE_NOT_PRESENT: vkCreateGraphicsPipeline(): Metal does not support disabling primitive restart.`

This is a bit odd, because if you try to correct this and set `primitiveRestartEnable` to true in `VkPipelineInputAssemblyStateCreateInfo`, then the Vulkan Validation layer (by using SDK 1.3.261.1) complains with:

`[Vulkan] ERROR location=0xbfc2d693 code=0, layerPrefix='Validation'] Validation Error: [ VUID-VkPipelineInputAssemblyStateCreateInfo-topology-00428 ] | MessageID = 0xbfc2d693 | vkCreateGraphicsPipelines() pCreateInfo[0]: topology is VK_PRIMITIVE_TOPOLOGY_TRIANGLE_LIST and primitiveRestartEnable is VK_TRUE. It is invalid.`

Something is conflicting between MoltenVK's warning system and the Validation layer.  I realize that I am testing with two different versions of MoltenVK (1.2.6 standalone for the [mvk-warn] case, and SDK 1.3.261.1 for the Validation Layer case), but I am concerned this issue will persist even after the new Vulkan loader is built.  I can't test it, but I wanted to provide a heads up just in case.

I found a link here that seems to say Metal's expectation re `primitiveRestartEnable` may be dependent on `topology` which also closely matches Vulkan's expectation (as enforced by validation layer): https://github.com/KhronosGroup/glTF/issues/1142#issuecomment-433717774.  Should your MoltenVK warning message also be dependent on `topology`?

billhollings:
You are correct. Thanks for catching this.

PR #2046 fixes this and only emits the warning for strip topology.

richard-lunarg:
@billhollings If you could put this in Rc2, it's causing failures in our testing too.

SRSaunders:
This seems to correct the extraneous warning messages.  Thanks.

billhollings:
@richard-lunarg 

> If you could put this in Rc2, it's causing failures in our testing too.

[v1.2.6-rc2](https://github.com/KhronosGroup/MoltenVK/releases/tag/v1.2.6-rc2) is available.

SRSaunders:
[...]

PR #2046, opened 2023-10-18
Emit primitiveRestartEnable disabled warning only for strip topology.

billhollings:
- Move check and warning to `MVKRenderingCommandEncoderState`.
- Pass `primitiveRestartEnable` to `MVKRenderingCommandEncoderState`.
- Warn only if `primitiveRestartEnable` disabled and strip topology is used.

Fixes issue #2045.

### DIFF
--- a/MoltenVK/MoltenVK/Commands/MVKCmdRendering.h
+++ b/MoltenVK/MoltenVK/Commands/MVKCmdRendering.h
@@ -584,6 +584,8 @@ public:
 
 protected:
 	MVKCommandTypePool<MVKCommand>* getTypePool(MVKCommandPool* cmdPool) override;
+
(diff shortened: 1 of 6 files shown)


cdavis5e:
What about triangle fans?

billhollings:
> What about triangle fans?

Good catch!

I had tried working that in, but because there is no Metal triangle fan topology value, I ended up having to double-track both Metal and Vulkan topology values in `MVKRenderingCommandEncoderState`, which felt like I was adding a bunch of confusing special-case-outlier code and tracking, all for an unenforceable warning message, so I dropped it.

Triangle fan restart will work (it's handled in the index conversion), but like strips in Metal, cannot be disabled. The app log just won't see a warning for TriStrips.

PR #2047, opened 2023-10-24
Reduce disk space consumed after running fetchDependencies.

billhollings:
- Add `--keep-cache` option to control whether or not to retain the `External/build/Intermediates` directory (default not retained).
- Export `KEEP_CACHE` & `SKIP_PACKAGING` to be available within scripts used by `ExternalDependencies` Xcode builds.
- Move `BLD_SPECIFIED` to `build()` instead of `build_impl()` to avoid updating it from background thread (which will fail).
- Update MoltenVK version to 1.2.7 (unrelated).
- Add `CompilerMSL::Options::replace_recursive_inputs` to pipeline cache (unrelated).
- Update GitHub CI to Xcode 15.0.
- Update `Whats_new.md` document.

### DIFF
--- a/fetchDependencies
+++ b/fetchDependencies
@@ -67,6 +67,12 @@
 #      --no-parallel-build
 #              Build the external libraries serially instead of in parallel. This is the default.
 #
+#      --keep-cache
+#              Do not remove the External/build/Intermediates cache directory after building.
+#              Removing the Intermediates directory returns significant disk space after the
+#              build, and is the default behaviour. Use this option if you intend to run this
+#              script repeatedly to incrementally build one platform at a time.
+#
 #      --glslang-root path
 #              "path" specifies a directory path to a KhronosGroup/glslang repository.
 #              This repository does need to be built and the build directory must be in the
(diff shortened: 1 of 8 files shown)


fetchDependencies
```diff
@@ -512,9 +525,10 @@ if [ "$XC_USE_BCKGND" != "" ]; then
 fi
 
 if [ "$BLD_SPECIFIED" != "" ]; then
-	# Build XCFrameworks, update latest symlink, and clean MoltenVK for rebuild
+	# Build XCFrameworks, update latest symlink, remove intermediates, and clean MoltenVK for rebuild
 	PROJECT_DIR="."
 	CONFIGURATION=${XC_CONFIG}
+	export SKIP_PACKAGING=""
```
cdavis5e, line 531: Since you already `export` `SKIP_PACKAGING` above at top level, it shouldn't be necessary to do so here as well.

fetchDependencies
```diff
@@ -191,6 +198,10 @@ while (( "$#" )); do
          XC_USE_BCKGND=""
          shift 1
          ;;
+       --keep-cache)
+         export KEEP_CACHE="Y"
```
cdavis5e, line 202: Ditto.

cdavis5e (approved):
Looks good. Just a few minor quibbles.

fetchDependencies
billhollings: Done. Thanks.

fetchDependencies
billhollings: Ditto.

PR #2060, opened 2023-11-06
Fix regression error in argument buffer runtime arrays.

billhollings:
- Update to latest SPIRV-Cross contining the fix.
- Modify CTS options in runcts script to avoid outputting full CTS log, and use less file caching, all to reduce memory and filespace consumption, and possibly improve performance (unrelated).
- Update `MoltenVKShaderConverter` tool to include Metal 3.1 support and improved argument buffer settings (unrelated).
- Small unrelated non-functional edits.


cdavis5e (approved):
I remember reading that [the `distutils` Python package is deprecated](https://docs.python.org/3.11/library/distutils.html). It looks like they made good on their promise to remove it in 3.12.

juan-lunarg:
> I remember reading that [the `distutils` Python package is deprecated](https://docs.python.org/3.11/library/distutils.html). It looks like they made good on their promise to remove it in 3.12.

I'd recommend using `actions/setup-python` as a temporary workaround.

EX:

```yml
      - uses: actions/setup-python@v4
        with:
          python-version: '3.11'
```

billhollings:
> > I remember reading that [the `distutils` Python package is deprecated](https://docs.python.org/3.11/library/distutils.html). It looks like they made good on their promise to remove it in 3.12.
> 
> I'd recommend using `actions/setup-python` as a temporary workaround.
> 
> EX:
> 
> ```yaml
>       - uses: actions/setup-python@v4
>         with:
>           python-version: '3.11'
> ```

Thanks for the detective work and suggestion!

I've [reported this](https://github.com/KhronosGroup/glslang/issues/3393) to `glslang`, and implemented the suggested patch here. 

Looks like `glslang` has a PR ready already, so we might be flipping back soon!

PR #2061, opened 2023-11-07
Don't make texture views where not necessary

etang-cw:
This is mainly to work around a bug in macOS Sonoma where Radeon Pro 580Xs read black if you attempt to sample a D24S8 texture through a texture view that was previously copied from (so texture goes from copy source → shader read with texture view) [FB13342692]

But less texture views also makes things easier to track in the Xcode shader debugger, so might as well apply it always

### DIFF
--- a/MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
@@ -91,7 +91,10 @@ id<MTLTexture> MVKImagePlane::getMTLTexture(MTLPixelFormat mtlPixFmt) {
         lock_guard<mutex> lock(_image->_lock);
         mtlTex = _mtlTextureViews[mtlPixFmt];
         if ( !mtlTex ) {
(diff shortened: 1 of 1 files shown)


billhollings (changes requested):
Thanks very much for catching this, and suggesting an improvement.

This is a bit odd though. Under what conditions does `[baseTexture pixelFormat] == mtlPixFmt` work, but the check, a couple of lines above, here doesn't?

https://github.com/KhronosGroup/MoltenVK/blob/ab84de541d1f0ec6a1f54b15eae7a72383ce6855/MoltenVK/MoltenVK/GPUObjects/MVKImage.mm#L86



Also a couple of nits about with formatting.

etang-cw:
Looks like the issue was actually "fixed" by an update to the game that happened while I was testing and slightly changed how the game renders, hiding the issue.  The actual reason for the texture view was that it needed a R001 swizzle, so we'll have to wait for Apple to fix the bug.

PR #2065, opened 2023-11-14
Fix MVKBitArray::getIndexOfFirstSetBit() skipping over entries.

js6i:
Reset startIndex so that we only skip over bits in the section that contains it.


billhollings (approved):
Hmmm...good catch!

Issue #2057, opened 2023-11-02
`vkCmdClearAttachments` flips viewport in SDK `1.3.268.1`


mokafolio:
Moltenvk shipping with SDK `1.3.268.1` seems to break the viewport orientation after a call to `vkCmdClearAttachments` (this is using opengl style viewports with `VK_KHR_Maintenance1`). Please see the attached screenshots. The correct one is with SDK `1.3.261.1` while the flipped one is with SDK `1.3.268.1`. This issue only happens when using `vkCmdClearAttachments` and I assume it might be due to some internal state tracking of moltenvk that goes wrong?

For what it's worth, the pinned moltenvk revision of 1.3.268.1 is  9e4ee9e679a6e1eb612bf82824ea0f68f43e558c. The revision that still works from the 1.3.261.1 sdk is 02a8c011a8ca077dfaa09f3b7ebf68be869dc26f.

<img width="912" alt="Screenshot 2023-11-02 at 11 40 59 AM" src="https://github.com/KhronosGroup/MoltenVK/assets/10217168/1c9ed565-a83d-4164-9e49-cf3a5e91a743">
<img width="912" alt="Screenshot 2023-11-02 at 11 38 38 AM" src="https://github.com/KhronosGroup/MoltenVK/assets/10217168/b9977f98-12b4-413f-bce6-07b72784e820">

billhollings:
Thanks for catching and reporting this.

This was caused during the addition of `VK_EXT_extended_dynamic_state` & `VK_EXT_extended_dynamic_state2`.

I'll include a fix for this when support for `VK_EXT_extended_dynamic_state3` is added, within the next week or two.

billhollings:
PR #2066 includes a fix for this. Please retest using latest version, and close this issue if the problem is fixed.

mokafolio:
can confirm that this is fixed when using the current tip of `main`. thanks!

Issue #2056, opened 2023-11-02
polygonMode line broken on SDK `1.3.268.1`


mokafolio:
Moltenvk shipping with SDK `1.3.268.1` seems to have broken `VkPipelineRasterizationStateCreateInfo` `polygonMode`. `VK_POLYGON_MODE_LINE` used to work previously but now just renders an opaque blob (see screenshots attached). For what it's worth, the pinned moltenvk revision of `1.3.268.1` is 9e4ee9e679a6e1eb612bf82824ea0f68f43e558c. The revision that still works from the `1.3.261.1` sdk is 02a8c011a8ca077dfaa09f3b7ebf68be869dc26f

<img width="570" alt="Screenshot 2023-11-02 at 10 32 58 AM" src="https://github.com/KhronosGroup/MoltenVK/assets/10217168/951f0852-cc46-4570-802e-be89251a64b2">
<img width="570" alt="Screenshot 2023-11-02 at 10 35 33 AM" src="https://github.com/KhronosGroup/MoltenVK/assets/10217168/f0a45820-b517-4a77-aebc-4a0713083d8d">

billhollings:
Thanks for catching and reporting this.

This was caused during the addition of `VK_EXT_extended_dynamic_state` & `VK_EXT_extended_dynamic_state2`. Polygon mode is actually part of `VK_EXT_extended_dynamic_state3`, so it wasn't testing properly.

I'll include a fix for this when support for `VK_EXT_extended_dynamic_state3` is added, within the next week or two.

billhollings:
PR #2066 includes a fix for this. Please retest using latest version, and close this issue if the problem is fixed.

mokafolio:
can confirm that this is fixed in current tip of `main`.

xxxmonsterxxx:
> PR #2066 includes a fix for this. Please retest using latest version, and close this issue if the problem is fixed.

Funny situation. I didn't use this option exactly after you broke it =) And today I got confused with it using 1.3.268.1... Yeah, I spent half a day searching for a bug in my code, but it turned out to be simple. I usually try to find a solution myself first, and only then resort to Google. Thank for report and for solution.

PR #2066, opened 2023-11-15
Add support for VK_EXT_extended_dynamic_state3 extension.

billhollings:
- Move patch point tracking from pipeline state to render state, and remove `MVKPipelineCommandEncoderState` subclasses no longer needed.
- Move sample location tracking from renderpass input to pipeline static or dynamic state tracking.
- Restart Metal render pass when sample locations change, and enable `VkPhysicalDeviceSampleLocationsPropertiesEXT::variableSampleLocations`.
- Fix regression that broke `VK_POLYGON_MODE_LINE` (unrelated).
- Fix regression in marking `MVKRenderingCommandEncoderState` dirty after `vkCmdClearAttachments()` (unrelated).

This PR includes fixes for issues #2056 and #2057.


MoltenVK/MoltenVK/Commands/MVKCommandEncoderState.mm
```diff
@@ -433,17 +402,118 @@
...
+	bool wasSLEnblDirty = _dirtyStates.isEnabled(SampleLocationsEnable);
+	
+	_dirtyStates.enableAll();
+
+	_dirtyStates.set(SampleLocations, wasSLDirty);
+	_dirtyStates.set(SampleLocationsEnable, wasSLEnblDirty);
+}
+
+// Don't call parent beginMetalRenderPass() because it may cause 
+// will call local markDirty() which is too aggressive.
```
cdavis5e, line 504: "may cause will call"??
```suggestion
// Don't call parent beginMetalRenderPass() because it
// will call local markDirty() which is too aggressive.
```

MoltenVK/MoltenVK/GPUObjects/MVKDevice.h
```diff
@@ -73,16 +73,21 @@ class MVKPrivateDataSlot;
...
-const static uint32_t kMVKMaxViewportScissorCount = 16;
-const static uint32_t kMVKMaxDescriptorSetCount = SPIRV_CROSS_NAMESPACE::kMaxArgumentBuffers;
+static constexpr uint32_t   kMVKQueueFamilyCount = 4;
+static constexpr uint32_t   kMVKQueueCountPerQueueFamily = 1;		// Must be 1. See comments in MVKPhysicalDevice::getQueueFamilies()
+static constexpr uint32_t   kMVKMinSwapchainImageCount = 2;
+static constexpr uint32_t   kMVKMaxSwapchainImageCount = 3;
+static constexpr uint32_t   kMVKMaxColorAttachmentCount = 8;
+static constexpr uint32_t   kMVKMaxViewportScissorCount = 16;
+static constexpr uint32_t   kMVKMaxDescriptorSetCount = SPIRV_CROSS_NAMESPACE::kMaxArgumentBuffers;
+static constexpr uint32_t   kMVKSampleLocationCount = 16;
```
cdavis5e, line 86: I'm not sold on this name. It implies that you can have up to 16 distinct sample locations, when this constant is about the granularity of the sample grid. It becomes especially confusing in the definition of `kMVKMaxSampleLocation` just below.

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
```diff
@@ -747,11 +779,11 @@
 			case VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_SAMPLE_LOCATIONS_PROPERTIES_EXT: {
 				auto* sampLocnProps = (VkPhysicalDeviceSampleLocationsPropertiesEXT*)next;
 				sampLocnProps->sampleLocationSampleCounts = _metalFeatures.supportedSampleCounts;
-				sampLocnProps->maxSampleLocationGridSize = kMetalSamplePositionGridSize;
-				sampLocnProps->sampleLocationCoordinateRange[0] = 0.0;
-				sampLocnProps->sampleLocationCoordinateRange[1] = (15.0 / 16.0);
+				sampLocnProps->maxSampleLocationGridSize = kMetalSampleLocationGridSize;
+				sampLocnProps->sampleLocationCoordinateRange[0] = kMVKMinSampleLocation;
+				sampLocnProps->sampleLocationCoordinateRange[1] = kMVKMaxSampleLocation;
 				sampLocnProps->sampleLocationSubPixelBits = 4;
```
cdavis5e, line 785: It should be possible to derive this value from the constants above as well--4 is the binary logarithm of 16, because 2<sup>4</sup> = 16.

MoltenVK/MoltenVK/GPUObjects/MVKPipeline.mm
```diff
@@ -305,7 +312,7 @@
 				cmdEncoder->_renderingState.setFrontFace(_rasterInfo.frontFace, false);
 				cmdEncoder->_renderingState.setPolygonMode(_rasterInfo.polygonMode, false);
 				cmdEncoder->_renderingState.setDepthBias(_rasterInfo);
-				cmdEncoder->_renderingState.setDepthClipEnable( !_rasterInfo.depthClampEnable, false );
+				cmdEncoder->_renderingState.setDepthClipEnable( !_rasterInfo.depthClampEnable, false);
```
cdavis5e, line 315:
```suggestion
				cmdEncoder->_renderingState.setDepthClipEnable(!_rasterInfo.depthClampEnable, false);
```

MoltenVK/MoltenVK/GPUObjects/MVKPipeline.mm
```diff
@@ -1635,7 +1641,7 @@ static MTLVertexFormat mvkAdjustFormatVectorToSize(MTLVertexFormat format, uint3
 
     // Multisampling - must ignore allowed bad pMultisampleState pointer if rasterization disabled
     if (_isRasterizing && pCreateInfo->pMultisampleState) {
-        plDesc.sampleCount = mvkSampleCountFromVkSampleCountFlagBits(pCreateInfo->pMultisampleState->rasterizationSamples);
+        plDesc.rasterSampleCount = mvkSampleCountFromVkSampleCountFlagBits(pCreateInfo->pMultisampleState->rasterizationSamples);
```
cdavis5e, line 1644: This is not always available. It probably needs a magic property like the other ones in `MoltenVK/MoltenVK/OS` in case  this property isn't supported.

MoltenVK/MoltenVK/Utility/MVKFoundation.h
```diff
@@ -102,8 +102,8 @@ typedef enum : uint8_t {
 
 /** Represents a given stage of a graphics pipeline. */
 enum MVKGraphicsStage {
-	kMVKGraphicsStageVertex = 0,	/**< The vertex shader stage. */
-	kMVKGraphicsStageTessControl,	/**< The tessellation control shader stage. */
+	kMVKGraphicsStageVertex = 0,	/**< The tessellation vertex compute shader stage. */
```
cdavis5e, line 105: Hmm, I'm going to have to update this line in the Transform Feedback change...

cdavis5e (changes requested):
Some minor feedback, and one potential problem.

MoltenVK/MoltenVK/Commands/MVKCommandEncoderState.mm
billhollings: Fixed.

MoltenVK/MoltenVK/GPUObjects/MVKDevice.h
billhollings: Good point. Changed.

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
billhollings: Ah. Nice idea. Done.

MoltenVK/MoltenVK/GPUObjects/MVKPipeline.mm
billhollings: One of my style quirks is that I deliberately try to include a space between `( !`  to call out the `!` during reading. Nothing worse than missing a "not" in any communication! 😉

 I've fixed this the other way, and added a matching space at the end.

MoltenVK/MoltenVK/GPUObjects/MVKPipeline.mm
billhollings: I hear you. I looked into that when I made the change. Oddly, `rasterSampleCount` is documented both in header and docs as being supported from the beginning (iOS 8/macOS 10.11). I assume it was publicly available back then, but not documented for some unknown reason.

Since it's explicitly documented that way, I'm not expecting it to cause problems with either runtime on older OS's, or app store filters...but Apple is certainly not perfect in its consistency...so we'll see.

billhollings:
> Some minor feedback, and one potential problem.

Thanks for the feedback. Should be cleaned up now.

cdavis5e (approved):

PR #2072, opened 2023-11-21
Enable some mandatory `VK_EXT_descriptor_indexing` features.

cdavis5e:
These features must be supported if the extension or `descriptorIndexing` feature is.

Fixes `dEQP-VK.info.device_mandatory_features`.

### DIFF
--- a/MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
@@ -128,9 +128,9 @@ void MVKPhysicalDevice::getFeatures(VkPhysicalDeviceFeatures2* features) {
 		.shaderInputAttachmentArrayDynamicIndexing = _metalFeatures.arrayOfTextures,
 		.shaderUniformTexelBufferArrayDynamicIndexing = _metalFeatures.arrayOfTextures,
 		.shaderStorageTexelBufferArrayDynamicIndexing = _metalFeatures.arrayOfTextures,
-		.shaderUniformBufferArrayNonUniformIndexing = false,
+		.shaderUniformBufferArrayNonUniformIndexing = true,
 		.shaderSampledImageArrayNonUniformIndexing = _metalFeatures.arrayOfTextures && _metalFeatures.arrayOfSamplers,
-		.shaderStorageBufferArrayNonUniformIndexing = false,
+		.shaderStorageBufferArrayNonUniformIndexing = true,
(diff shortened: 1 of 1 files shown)


MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
```diff
@@ -126,14 +126,14 @@
 		.shaderInt8 = true,
 		.descriptorIndexing = _vulkan12FeaturesNoExt.descriptorIndexing,
 		.shaderInputAttachmentArrayDynamicIndexing = _metalFeatures.arrayOfTextures,
-		.shaderUniformTexelBufferArrayDynamicIndexing = _metalFeatures.arrayOfTextures,
```
billhollings, line 129: When does `_metalFeatures.arrayOfTextures` fail? Are you running CTS on an ancient devices or OS?

Perhaps we should disable the `VK_EXT_descriptor_indexing` extension completely in these cases. Perhaps test `_metalFeatures.arrayOfTextures` in `MVKPhysicalDevice::initExtensions()`.

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
```diff
@@ -126,14 +126,14 @@
...
-		.shaderStorageTexelBufferArrayDynamicIndexing = _metalFeatures.arrayOfTextures,
-		.shaderUniformBufferArrayNonUniformIndexing = false,
```
billhollings, line 131: What are the ramifications of indicating such complete support?

Will it cause other CTS tests, that check these flags, to fail if we actually don't support them?

billhollings (commented):
A couple of inline questions.

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
cdavis5e: I had thought that we were enabling `shaderSampledImageDynamicIndexing` unconditionally, but, on closer inspection, we weren't. I've changed it to disable the extension if the requisite Metal features aren't supported, as you suggested.

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
billhollings: > I don't see why non-uniform indexing can't work.

Ah. Good point.

billhollings (approved):

PR #2073, opened 2023-11-21
MVKDevice: Clamp max per-set descriptor limit to minimum 1024.

cdavis5e:
As required by the Vulkan spec.

Fixes the CTS tests
`dEQP-VK.api.info.vulkan1p2_limits_validation.khr_maintenance_3` and
`dEQP-VK.api.maintenance3_check.maintenance3_properties`.

### DIFF
--- a/MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
@@ -525,9 +525,7 @@ void MVKPhysicalDevice::getProperties(VkPhysicalDeviceProperties2* properties) {
 	supportedProps11.maxMultiviewViewCount = 32;
 	supportedProps11.maxMultiviewInstanceIndex = canUseInstancingForMultiview() ? uintMax / 32 : uintMax;
 	supportedProps11.protectedNoFault = false;
-	supportedProps11.maxPerSetDescriptors = 4 * (_metalFeatures.maxPerStageBufferCount +
-												 _metalFeatures.maxPerStageTextureCount +
-												 _metalFeatures.maxPerStageSamplerCount);
+	supportedProps11.maxPerSetDescriptors = getMaxPerSetDescriptorCount();
 	supportedProps11.maxMemoryAllocationSize = _metalFeatures.maxMTLBufferSize;
 
 	// Create a SSOT for these Vulkan 1.2 properties, which can be queried via two mechanisms here.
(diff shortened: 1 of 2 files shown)


billhollings:
This feels arbitrary. Unless arg buffers are in play, this minimum total is never going to be met. 

Walk me through this. By inflating this value, what are we telling an app that might be dangerous?

billhollings (commented):
This feels arbitrary. Unless arg buffers are in play, this minimum total is never going to be met.

Walk me through this. By inflating this value, what are we telling an app that might be dangerous?

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
```diff
@@ -525,9 +525,9 @@
 	supportedProps11.maxMultiviewViewCount = 32;
 	supportedProps11.maxMultiviewInstanceIndex = canUseInstancingForMultiview() ? uintMax / 32 : uintMax;
 	supportedProps11.protectedNoFault = false;
-	supportedProps11.maxPerSetDescriptors = 4 * (_metalFeatures.maxPerStageBufferCount +
+	supportedProps11.maxPerSetDescriptors = max(4 * (_metalFeatures.maxPerStageBufferCount +
```
juan-lunarg, line 528: If you intent is to clamp it would be a bit more clear to use std::clamp

https://en.cppreference.com/w/cpp/algorithm/clamp

juan-lunarg: Also minor preference ideally the clamp was a in a helper function to make the 1024 less of a magic number. Since you could comment the helper function.

cdavis5e:
> This feels arbitrary. Unless arg buffers are in play, this minimum total is never going to be met.
> 
> Walk me through this. By inflating this value, what are we telling an app that might be dangerous?

This is the maximum number of total descriptors of all types across all stages that may be contained in a descriptor set at once. It's meant to be a constraint on the size of a descriptor set itself. Even if we were to raise it, an app would still be constrained by the maximum per-type descriptor counts.

This is the *minimum* maximum that is [mandated by the Vulkan spec][1]. The CTS actually tests that the limit is at least 1024.

[1]: https://registry.khronos.org/vulkan/specs/1.3-extensions/html/chap48.html#limits-required

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
cdavis5e: Isn't that just `max(min(value, hi), lo)`? What would I use for `hi`? I feel like that would be more verbose, because I'd have to say `std::numeric_limits<uint32_t>::max()`, just to satisfy the function's prototype.

You're probably right that I ought to factor this into a helper function--I don't know why you deleted that comment...

juan-lunarg: Apologies you are correct std::clamp isn't suitable in this scenario.

cdavis5e: > I don't know why you deleted that comment...

Figures. GitHub claims to support live updates of PRs, but for some reason, after a while it just stops for me...

Oh well. Anyway, I've factored this into a helper function.

billhollings (approved):

PR #2076, opened 2023-11-21
MVKDevice: Correct some required limits.

cdavis5e:
If the `wideLines` feature isn't supported, then `lineWidthGranularity`
must be zero.

If the `fullDrawIndexUint32` feature is supported, then
`maxDrawIndexedIndexValue` must be `UINT32_MAX`. I had originally done
this when I turned the feature on, but for a while now, we've been
setting it to one less, because primitive restart can't be disabled and
the value is defined to exclude primitive restart.

The wording in the spec is ambiguous. The description of
`maxDrawIndexedIndexValue` says:

> * `maxDrawIndexedIndexValue` is the maximum index value that **can**
>   be used for indexed draw calls when using 32-bit indices. *This
>   excludes the primitive restart index value of 0xFFFFFFFF.* [emphasis
>   added]

But the description of `fullDrawIndexUint32` says:

> * `fullDrawIndexUint32` specifies the full 32-bit range of indices is
>   supported for indexed draw calls when using a VkIndexType of
>   `VK_INDEX_TYPE_UINT32`. `maxDrawIndexedIndexValue` is the maximum
>   index value that **may** [sic] be used *(aside from the primitive
>   restart index, which is always 2<sup>32</sup>-1 when the VkIndexType
>   is `VK_INDEX_TYPE_UINT32`)*. If this feature is supported,
>   `maxDrawIndexedIndexValue` **must** be 2<sup>32</sup>-1; otherwise
>   it **must** be no smaller than 2<sup>24</sup>-1. [emphasis added]

It's unclear whether it means that the primitive restart index is to be
ignored, or the maximum draw index must account for it.

The alternative is to disable `fullDrawIndexUint32` because we cannot
set `maxDrawIndexedIndexValue` to `UINT32_MAX`; but that might mislead
programs into thinking that we only support 24-bit vertex indices.

Fixes the `dEQP-VK.info.device_properties` test.

### DIFF
--- a/MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
@@ -2748,7 +2748,7 @@ void MVKPhysicalDevice::initLimits() {
     _properties.limits.maxComputeWorkGroupCount[1] = kMVKUndefinedLargeUInt32;
     _properties.limits.maxComputeWorkGroupCount[2] = kMVKUndefinedLargeUInt32;
 
-    _properties.limits.maxDrawIndexedIndexValue = numeric_limits<uint32_t>::max() - 1;	// Support both fullDrawIndexUint32 and automatic primitive restart.
+    _properties.limits.maxDrawIndexedIndexValue = numeric_limits<uint32_t>::max();
     _properties.limits.maxDrawIndirectCount = kMVKUndefinedLargeUInt32;
 
 
--- a/MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
@@ -2692,7 +2692,7 @@ void MVKPhysicalDevice::initLimits() {
     _properties.limits.pointSizeGranularity = 1;
     _properties.limits.lineWidthRange[0] = 1;
     _properties.limits.lineWidthRange[1] = 1;
-    _properties.limits.lineWidthGranularity = 1;
+    _properties.limits.lineWidthGranularity = 0;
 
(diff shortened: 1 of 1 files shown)


billhollings:
I'm confused.

As you mention, for `fullDrawIndexUint32`, the spec says:

> If this feature is supported,   `maxDrawIndexedIndexValue` **must** be 2<sup>32</sup>-1

This should mean we are doing it correctly already. Where does the conflicting requirement that `maxDrawIndexedIndexValue` must be `UINT32_MAX`, come from? If it's coming from the CTS test, could the test be in error?

cdavis5e:
> I'm confused.
> 
> As you mention, for `fullDrawIndexUint32`, the spec says:
> 
> > If this feature is supported,   `maxDrawIndexedIndexValue` **must** be 2<sup>32</sup>-1
> 
> This should mean we are doing it correctly already. Where does the conflicting requirement that `maxDrawIndexedIndexValue` must be `UINT32_MAX`, come from? If it's coming from the CTS test, could the test be in error?

`UINT32_MAX` *is* 2<sup>32</sup>-1. 2<sup>32</sup> is 0x100000000 (one followed by eight zeroes in hex), which cannot fit in a 32-bit integer. We were setting it to 2<sup>32</sup>-**2**, i.e. `UINT32_MAX`-1.

billhollings (approved):

billhollings:
> `UINT32_MAX` *is* 2<sup>32</sup>-1. 2<sup>32</sup> is 0x100000000 (one followed by eight zeroes in hex), which cannot fit in a 32-bit integer. We were setting it to 2<sup>32</sup>-**2**, i.e. `UINT32_MAX`-1.

Well! I guess that was my official math-brain-fart for this week! Of course you are correct. 🤦🏻‍♂️😂

Issue #1670, opened 2022-07-30
Expose experimental Vulkan features using internal Metal API


spnda:
As noted by @cdavis5e in #1657, there's a few things in Metal which are not documented or exposed. These features could possibly be used to experiment supporting some Vulkan features that would not be supportable otherwise. The initial idea is to have a compile-time option which would enable these experimental features, as using these private APIs would not allow shipping an app with MoltenVK into the App Store and should therefore not be the default. This issue shall act as some sort of a tracker of potential features and their implementations by looking into the symbols and Objective-C metadata.

5 patches have already been written to support some features (although not entirely stable):
https://github.com/KhronosGroup/MoltenVK/issues/1657#issuecomment-1197438815

I have also just gone ahead and extracted the symbols of the `Metal` binary found in the `Metal.framework` shipped with the latest macOS 13 beta, meaning with support for Metal 3 and Metal IO. I couldn't get `class-dump` to not crash with this binary, so this is just the basic output of `nm`:
[metal_symbols.txt](https://github.com/KhronosGroup/MoltenVK/files/9224185/metal_symbols.txt)
[metalkit_symbols.txt](https://github.com/KhronosGroup/MoltenVK/files/9224704/metalkit_symbols.txt)

Some few things I've found in metal_symbols.txt:
- `-[MTLHeap _gpuAddress]`
- `MTLDeviceFeatureQueries` contains A LOT of interesting queries about GPU features, of which most are not exposed normally.
- Metal acceleration structures seem to be built using `MTLGPUBVHBuilder`, which seems to be more like something from MPS.
- `_MTLDevice` has a lot of other interesting queries about GPU capabilities.
- `_MTLPipelineCache` is a thing.
- `setLineWidth:` (as used by one of the patches linked above) is not listed anymore.
- `-[MTLRenderPipelineColorAttachmentDescriptor setLogicOpEnabled:]` is not listed anymore. However, `-[MTLRenderPipelineDescriptor setLogicOperation:]` is available.
- `-[MTLToolsRenderCommandEncoder useResourceGroup:usage:stages:]`... I wonder what resource groups are. (found in MetalTools)

billhollings:
> `setLineWidth:` (as used by one of the patches linked above) is not listed anymore.

Nice sleuthing.

Interestingly, `-[_MTLDevice maxLineWidth]` _**is**_ still there.

[...]

Issue #1657, opened 2022-07-23
Add line width support as optional compilation macro


M1Minecraft:
Hello, I'm trying to play No Man's Sky through Crossover and it says, "Does not support Wide Line". Does Metal 3 allow the use of Wide Lines or is it still not possible to implement?

cdavis5e:
There is no *public* API for wide line support in Metal.

<br><br><br>
<sub>[This patch should make it work.](https://github.com/KhronosGroup/MoltenVK/files/9174751/0001-Support-wide-lines.patch.txt) I can't put this into `master`, though, because that would prevent apps bundling MoltenVK from submitting to the App Store. You're on your own here; we can provide no support. Maybe you can convince CW to put it in their fork; I no longer have much pull with them.</sub>

M1Minecraft:
> There is no *public* API for wide line support in Metal.
> 
> 
> 
> <br><br><br>
> 
> <sub>[This patch should make it work.](https://github.com/KhronosGroup/MoltenVK/files/9174751/0001-Support-wide-lines.patch.txt) I can't put this into `master`, though, because that would prevent apps bundling MoltenVK from submitting to the App Store. You're on your own here; we can provide no support. Maybe you can convince CW to put it in their fork; I no longer have much pull with them.</sub>

I will try my best 😁, thank you for the support.

M1Minecraft:
> There is no _public_ API for wide line support in Metal.
> 
> [This patch should make it work.](https://github.com/KhronosGroup/MoltenVK/files/9174751/0001-Support-wide-lines.patch.txt) I can't put this into `master`, though, because that would prevent apps bundling MoltenVK from submitting to the App Store. You're on your own here; we can provide no support. Maybe you can convince CW to put it in their fork; I no longer have much pull with them.

I'm kinda new to this and can't seem to figure out how to apply this patch? What steps would be required?

cdavis5e:
> I'm kinda new to this and can't seem to figure out how to apply this patch? What steps would be required?

If you're using Git, `git apply 0001-Support-wide-lines.patch.txt` will work.

If not, use `patch -p1 <0001-Support-wide-lines.patch.txt`.

In both cases, you need to be in MoltenVK's root source directory.

billhollings:
[...]

PR #2091, opened 2023-12-10
Feature widelines

jarrodnorwell:


jarrodnorwell:
Ignore, sent PR to wrong repo

billhollings:
@jarrodnorwell 

Interesting PR.

```
void MVKLineWidthCommandEncoderState::encodeImpl(uint32_t stage) {
    if (stage != kMVKGraphicsStageRasterization) { return; }
    if (!_cmdEncoder->_pDeviceFeatures->wideLines) { return; }
    if (![_cmdEncoder->_mtlRenderEncoder respondsToSelector: @selector(setLineWidth:)]) { return; }
    [(id<MVKMTLRenderCommandEncoderLineWidth>)_cmdEncoder->_mtlRenderEncoder setLineWidth: _lineWidth];
}
```

I assume that `[MTLRenderCommandEncoder setLineWidth:]` is a private Metal API method?

> Ignore, sent PR to wrong repo

If you have a fork repo where you are accumulating use of Metal's private API, it might be good to bring them into the main MoltenVK repo, with an appropriate compile guard macro, such as `MVK_USE_METAL_PRIVATE`, that could be enabled when building MoltenVK.

Enabling that build setting would almost certainly disqualify an app from distribution via the App Store, but for macOS games that are distributed directly, these features would be functional. And it would give us a basis for requesting Apple to open up some of those private API's.

Are you interested in discussing this?

jarrodnorwell:
Hey @billhollings,

This is actually an accumulation of patches from @cdavis5e and eventually @js6i with their Geometry Shader fork, I'm hoping to bring better support to my Yuzu port for iOS (as well as my Citra port to the same platform)

My plan wasn't to merge it into the main repository as it does use private API's as you mentioned and my apps won't be made available on the AppStore

billhollings:
> This is actually an accumulation of patches from @cdavis5e and eventually @js6i with their Geometry Shader fork

@jarrodnorwell @cdavis5e @js6i 

As I mention above, it might be helpful to be able to pull these into MoltenVK main, wrapped in a guard macro.

Can we get access to these patches? And/or a list of private functionality that has been added?

billhollings:
Hmmm...I had forgotten that we'd travelled down this road before in #1657 and #1670. 🤦🏻‍♂️

I'll do further follow up there.

billhollings:
PR #2137 has added this functionality.

PR #2098, opened 2023-12-16
Fix heapUsage query for non-unified memory devices

### LINKED ISSUE #2082: Incorrect heapUsage values from VkPhysicalDeviceMemoryBudgetPropertiesEXT query on non-unified memory machines
When calling `vkGetPhysicalDeviceMemoryProperties2()` with `VkPhysicalDeviceMemoryBudgetPropertiesEXT` in the pNext chain, I observe the following for the returned `heapUsage`:

1. `heapUsage[0]` seems to be the total allocated gpu memory, and for my application yields the same value as found in `mvkPerfStats.device.gpuMemoryAllocated.latest`
2. `heapUsage[1]` appears to be something else related to MoltenVK's CPU/system memory footprint.

Looking at VulkanCapsViewer on macOS, Memory heap 0 shows `DEVICE_LOCAL_BIT, HOST_VISIBLE_BIT, and HOST_CACHED_BIT`, while Memory heap 1 shows `HOST_VISIBLE_BIT, HOST_COHERENT_BIT, and HOST_CACHED_BIT`.

As on Vulkan for other platforms, the different heaps correspond to various types of GPU physical device memory, each supporting different access capabilities (e.g. device-only, host-visible, host-coherent, etc).  What do the different heaps represent for MoltenVK?  The reason I am confused is that for my application I would have expected `heapUsage[0] + heapUsage[1]` to represent the total GPU memory in use.  However, the results seem to indicate the total is stored in `heapUsage[0]` and I am left wondering what is stored in `heapUsage[1]` and how to use both pieces of data.

If I am interested in getting the total allocated GPU memory of all types (e.g. similar to what would be displayed in MSI Afterburner on Windows, or MangoHUD on Linux), how should I be using the above `heapUsage` values?

**UPDATE**:  Here is some data collected directly from memory allocators used in my application vs. the query:

Memory Allocator Information (buffers and textures):
Heap 0: 1076 MB
Heap 1:   636 MB
Total:      1712 MB

`vkGetPhysicalDeviceMemoryProperties2()` query (with `VkPhysicalDeviceMemoryBudgetPropertiesEXT`):
Heap 0: 1804 MB from `heapUsage[0]`
Heap 1:  1296 MB from `heapUsage[1]`

As you can see, the `heapUsage[0]` query (1804 MB) is fairly close to the sum of the heap allocations (1712 MB).  However, the `heapUsage[1]` number does not make sense.  Why doesn't the `heapUsage` data from the query better align with the allocator information?  And the sum of the `heapUsage` data should add up to the total allocations, and not be placed in the `heapUsage[0]` slot.  Can you explain what is happening here?

SRSaunders:
Fixes #2082 for non-unified memory machines.  No change for unified memory machines.

1. Replaces `task_vm_info.phys_footprint` with `task_vm_info.ledger_tag_graphics_footprint` in `mvkGetUsedMemorySize()`.  This provides a more accurate estimate of mapped memory for `heapUsage[1]` on non-unified memory machines.  As an aside, it also provides a more accurate fall-back within `MVKPhysicalDevice::getCurrentAllocatedSize()` if the selector is not available for any machine.
2. Fixes the calculation for `heapUsge[0]` on non-unified memory machines, where it represents the device-only memory.  The calculation remains the same for unified memory machines (where `heapUsage[1] = 0`).

The only edge case I see is for non-unified memory machines where the `currentAllocatedSize` selector is not available.  In this case `heapUsage[0]` will return 0 for device memory and `heapUsage[1]` will return mapped memory.  I think this is a reasonable fallback if the total memory information is not available from Metal.

### DIFF
--- a/Common/MVKOSExtensions.mm
+++ b/Common/MVKOSExtensions.mm
@@ -134,7 +134,11 @@ uint64_t mvkGetUsedMemorySize() {
 	task_vm_info_data_t task_vm_info;
 	mach_msg_type_number_t task_size = TASK_VM_INFO_COUNT;
 	if (task_info(mach_task_self(), TASK_VM_INFO, (task_info_t)&task_vm_info, &task_size) == KERN_SUCCESS) {
+#ifdef TASK_VM_INFO_REV3_COUNT	// check for rev3 version of task_vm_info
+		return task_vm_info.ledger_tag_graphics_footprint;
+#else
 		return task_vm_info.phys_footprint;
+#endif
 	}
 	return 0;
 }
--- a/Common/MVKOSExtensions.mm
+++ b/Common/MVKOSExtensions.mm
@@ -135,10 +135,12 @@ uint64_t mvkGetUsedMemorySize() {
 	mach_msg_type_number_t task_size = TASK_VM_INFO_COUNT;
 	if (task_info(mach_task_self(), TASK_VM_INFO, (task_info_t)&task_vm_info, &task_size) == KERN_SUCCESS) {
 #ifdef TASK_VM_INFO_REV3_COUNT	// check for rev3 version of task_vm_info
-		return task_vm_info.ledger_tag_graphics_footprint;
-#else
-		return task_vm_info.phys_footprint;
+		if (task_size >= TASK_VM_INFO_REV3_COUNT) {
+			return task_vm_info.ledger_tag_graphics_footprint;
+		}
+		else
 #endif
+			return task_vm_info.phys_footprint;
 	}
 	return 0;
 }
--- a/MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
@@ -1650,12 +1650,12 @@ VkResult MVKPhysicalDevice::getMemoryProperties(VkPhysicalDeviceMemoryProperties
 				auto* budgetProps = (VkPhysicalDeviceMemoryBudgetPropertiesEXT*)next;
 				mvkClear(budgetProps->heapBudget, VK_MAX_MEMORY_HEAPS);
 				mvkClear(budgetProps->heapUsage, VK_MAX_MEMORY_HEAPS);
-				budgetProps->heapBudget[0] = (VkDeviceSize)getRecommendedMaxWorkingSetSize();
-				budgetProps->heapUsage[0] = (VkDeviceSize)getCurrentAllocatedSize();
 				if (!getHasUnifiedMemory()) {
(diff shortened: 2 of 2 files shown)


Common/MVKOSExtensions.mm
```diff
@@ -134,7 +134,11 @@ uint64_t mvkGetUsedMemorySize() {
 	task_vm_info_data_t task_vm_info;
 	mach_msg_type_number_t task_size = TASK_VM_INFO_COUNT;
 	if (task_info(mach_task_self(), TASK_VM_INFO, (task_info_t)&task_vm_info, &task_size) == KERN_SUCCESS) {
+#ifdef TASK_VM_INFO_REV3_COUNT	// check for rev3 version of task_vm_info
+		return task_vm_info.ledger_tag_graphics_footprint;
```
cdavis5e, line 138: What happens if MoltenVK is built with an SDK that has `REV3` struct, but run on a system without it?

SRSaunders: Good catch.  I wonder if this can be tested at runtime?  Let me look into it.  I also just pushed a small defensive guard change.  I may need one more to address your question.

SRSaunders: I submitted a change which hopefully solves your runtime question.  Let me know...

cdavis5e (approved):

SRSaunders:
One little change while you merge:

`uint64_t currentAllocatedSize` should be declared either `auto` or `VkDeviceSize`.

Sorry for missing this.

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
```diff
@@ -1650,12 +1650,17 @@
...
 				mvkClear(budgetProps->heapBudget, VK_MAX_MEMORY_HEAPS);
 				mvkClear(budgetProps->heapUsage, VK_MAX_MEMORY_HEAPS);
-				budgetProps->heapBudget[0] = (VkDeviceSize)getRecommendedMaxWorkingSetSize();
-				budgetProps->heapUsage[0] = (VkDeviceSize)getCurrentAllocatedSize();
 				if (!getHasUnifiedMemory()) {
 					budgetProps->heapBudget[1] = (VkDeviceSize)mvkGetAvailableMemorySize();
 					budgetProps->heapUsage[1] = (VkDeviceSize)mvkGetUsedMemorySize();
 				}
+				budgetProps->heapBudget[0] = (VkDeviceSize)getRecommendedMaxWorkingSetSize();
+				uint64_t currentAllocatedSize = (VkDeviceSize)getCurrentAllocatedSize();
```
cdavis5e, line 1658:
```suggestion
				auto currentAllocatedSize = (VkDeviceSize)getCurrentAllocatedSize();
```

PR #2099, opened 2023-12-16
Fix system memory size of tvOS

karelrooted:
use sysctlbyname("hw.memsize") to get sytem memory size instead of host_info, host_info is not available on tvOS, so we change to sysctlbyname which support all apple platforms

the sytem memory size of tvOS is hardcoded as 0 currently, found this [bug](https://github.com/haasn/libplacebo/issues/229#issuecomment-1857622686) while using libplacebo's vulkan render to play video on tvOS 

apple [document](https://developer.apple.com/documentation/kernel/1387446-sysctlbyname
) didn't stat this API is available on tvOS though

### DIFF
--- a/Common/MVKOSExtensions.mm
+++ b/Common/MVKOSExtensions.mm
@@ -23,6 +23,7 @@
 #include <mach/mach_time.h>
 #include <mach/task.h>
 #include <os/proc.h>
+#include <sys/sysctl.h>
 
 #import <Foundation/Foundation.h>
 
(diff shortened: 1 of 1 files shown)


billhollings (changes requested):
Thanks for the submission!

So strange that tvOS is the only platform that doesn't support `host_info()`.

However, in testing your submission here, `sysctlbyname()` is available on all platforms, and gives the same numeric results as `host_info()` on both macOS and iOS.

So, can you modify this PR to remove the platform compile macros, and just use your new code for all platforms, please?

karelrooted:
PR is updated to use sysctlbyname to get system memory size for all apple platforms

billhollings (approved):
Thanks!

Issue #2096, opened 2023-12-15
Regression with 13998affe, point rendering no longer working


mbechard:
I noticed that point rendering no longer works as of 13998affe. I'm not using polygons with POLYGON_MODE_POINT, I'm using point primitive topology. Tracing through the MoltenVK code, everything seems fine at first glance, cmdEncoder->_mtlRenderEncoder drawIndexedPrimitives is being given the correct primitive type.
Any ideas where else to look as to how this commit broke point rendering?

Further info, if I stop using dynamic primitive topology, then point rendering works again. Not surprising, although unclear how this commit breaks it still

billhollings:
PR #2100 should fix this. Please retest with latest MoltenVK and close this issue if the problem is fixed.

mbechard:
Confirmed it fixes my issue, thanks!

PR #2100, opened 2023-12-17
Fix regression error and properties in point-topology/polygon-mode behavior.

billhollings:
- `MVKGraphicsPipeline::isRenderingPoints()` fix regression error to support dynamically setting point topology.
- Move emulation of `VK_POLYGON_MODE_POINT` in Metal, if the polygon model is static, to `MVKRenderingCommandEncoderState`, to handle dynamic setting of topology.
- `MVKDevice` populate `VkPhysicalDeviceExtendedDynamicState3PropertiesEXT`.
- Move error reporting in `mvkMTLTriangleFillModeFromVkPolygonModeInObj()` to encoding.
- `MVKRenderingCommandEncoderState::setCullMode()` use `getContent()` for readability and consistency (unrelated).
- Pass pipeline object to datatype functions for reporting accuracy (unrelated).

Fixes issue #2096.

### DIFF
--- a/MoltenVK/MoltenVK/Commands/MVKCommandEncoderState.mm
+++ b/MoltenVK/MoltenVK/Commands/MVKCommandEncoderState.mm
@@ -302,7 +302,7 @@ void MVKDepthStencilCommandEncoderState::encodeImpl(uint32_t stage) {
 void MVKRenderingCommandEncoderState::setCullMode(VkCullModeFlags cullMode, bool isDynamic) {
 	auto mtlCullMode = mvkMTLCullModeFromVkCullModeFlags(cullMode);
 	setMTLContent(CullMode);
-	_cullBothFaces[isDynamic ? StateScope::Dynamic : StateScope::Static] = (cullMode == VK_CULL_MODE_FRONT_AND_BACK);
+	getContent(_cullBothFaces, isDynamic) = (cullMode == VK_CULL_MODE_FRONT_AND_BACK);
 }
 
 void MVKRenderingCommandEncoderState::setFrontFace(VkFrontFace frontFace, bool isDynamic) {
@@ -313,6 +313,7 @@ void MVKRenderingCommandEncoderState::setFrontFace(VkFrontFace frontFace, bool i
 void MVKRenderingCommandEncoderState::setPolygonMode(VkPolygonMode polygonMode, bool isDynamic) {
 	auto mtlPolygonMode = mvkMTLTriangleFillModeFromVkPolygonMode(polygonMode);
 	setMTLContent(PolygonMode);
+	getContent(_isPolygonModePoint, isDynamic) = (polygonMode == VK_POLYGON_MODE_POINT);
 }
 
 void MVKRenderingCommandEncoderState::setBlendConstants(float blendConstants[4], bool isDynamic) {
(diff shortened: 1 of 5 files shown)


MoltenVK/MoltenVK/Commands/MVKCommandEncoderState.mm
```diff
@@ -420,8 +421,20 @@
...
-	return getMTLContent(PrimitiveTopology);
+	if (isDynamicState(PolygonMode) &&
+		_isPolygonModePoint[StateScope::Dynamic] &&
+		!_isPolygonModePoint[StateScope::Static]) {
+		
+		reportWarning(VK_ERROR_FEATURE_NOT_PRESENT, "vkCmdSetPolygonMode(): Metal does not support setting VK_POLYGON_MODE_POINT dynamically.");
+		return getMTLContent(PrimitiveTopology);
+	}
+
+	return getContent(_isPolygonModePoint, PolygonMode) ? MTLPrimitiveTypePoint : getMTLContent(PrimitiveTopology);
```
cdavis5e, line 437: If we can't actually set it to `VK_POLYGON_MODE_POINT` dynamically... can we even claim to support `extendedDynamicState3PolygonMode`?

MoltenVK/MoltenVK/GPUObjects/MVKPipeline.mm
```diff
@@ -1581,7 +1582,7 @@ static MTLVertexFormat mvkAdjustFormatVectorToSize(MTLVertexFormat format, uint3
 	if (pCreateInfo->pInputAssemblyState) {
 		plDesc.inputPrimitiveTopologyMVK = isRenderingPoints(pCreateInfo)
 												? MTLPrimitiveTopologyClassPoint
-												: mvkMTLPrimitiveTopologyClassFromVkPrimitiveTopology(pCreateInfo->pInputAssemblyState->topology);
+												: mvkMTLPrimitiveTopologyClassFromVkPrimitiveTopologyInObj(pCreateInfo->pInputAssemblyState->topology, this);
```
cdavis5e, line 1585: I thought we had a macro that did that, like we do for the other ones...

MoltenVK/MoltenVK/GPUObjects/MVKPipeline.mm
```diff
@@ -1949,7 +1955,7 @@ static MTLVertexFormat mvkAdjustFormatVectorToSize(MTLVertexFormat format, uint3
 			((pCreateInfo->pRasterizationState->rasterizerDiscardEnable && !isDynamicState(RasterizerDiscardEnable)) ||
 			 ((pCreateInfo->pRasterizationState->cullMode == VK_CULL_MODE_FRONT_AND_BACK) && !isDynamicState(CullMode) &&
 			  pCreateInfo->pInputAssemblyState &&
-			  (mvkMTLPrimitiveTopologyClassFromVkPrimitiveTopology(pCreateInfo->pInputAssemblyState->topology) == MTLPrimitiveTopologyClassTriangle))));
+			  (mvkMTLPrimitiveTopologyClassFromVkPrimitiveTopologyInObj(pCreateInfo->pInputAssemblyState->topology, this) == MTLPrimitiveTopologyClassTriangle))));
```
cdavis5e, line 1958: This one, too.

cdavis5e (changes requested):

MoltenVK/MoltenVK/Commands/MVKCommandEncoderState.mm
billhollings: Wireframe is still a valid (IMHO more so than points) use of polygon mode, so supporting it is still useful.

Making a distinction was the purpose of the `VkPhysicalDevicePortabilitySubsetFeaturesKHR::pointPolygons` (in final to be named `rasterizationPolygonModePoints`). With the static emulation in place, we can probably now limit this to `rasterizationPolygonModePointsDynamic`

MoltenVK/MoltenVK/GPUObjects/MVKPipeline.mm
billhollings: Thanks. You are of course correct. I got confused, forgot about the macro design, and thought we were calling the public API versions. 🤦🏻‍♂️

Fixed now.

MoltenVK/MoltenVK/GPUObjects/MVKPipeline.mm
billhollings: Ditto. Fixed now.

cdavis5e (approved):

MoltenVK/MoltenVK/Commands/MVKCommandEncoderState.mm
cdavis5e: Very well, then.

PR #2121, opened 2024-01-09
Update dependency libraries to match Vulkan SDK 1.3.275.

billhollings:
- Hide static _Vulkan_ API symbols by default (build setting `MVK_HIDE_VULKAN_SYMBOLS=0` by default).
- Update documentation.

The SDK release has been moved from the original 1.3.273 temp branches, to 1.3.275.

The rationale for hiding **MoltenVK** static symbols by default was driven by:
1. _Cube_ demo now uses _Volk_, which requires _Vulkan_ symbols to be hidden, which means the _Cube_ demo couldn't be run unless the user modified the **MoltenVK** build settings.
2. **MoltenVK** as distributed via the SDK, is built with symbols hidden.
3. Most users use the _Vulkan_ Loader.
4. Many users use  _Volk_.


cdavis5e (approved):
> Add `Templates/spirv-tools/build.zip`

What's that for?

cdavis5e:
@js6i Heads up--I think you're OK because Wine gets Vulkan function pointers with `vkGetInstanceProcAddr()`/`vkGetDeviceProcAddr()`, but I also recall that CrossOver Wine at least linked to MoltenVK directly.

Docs/Whats_New.md
```diff
@@ -25,6 +25,7 @@ Released 2024/01/08
 	- `VK_EXT_extended_dynamic_state3` *(Metal does not support `VK_POLYGON_MODE_POINT`)*
 	- `VK_EXT_headless_surface`
 	- `VK_EXT_layer_settings`
+- **MoltenVK** build now hides static _Vulkan_ API symbols by default (build setting `MVK_HIDE_VULKAN_SYMBOLS=0` by default).
```
spnda, line 28: You meant =1 here.

spnda (changes requested):

spnda (approved):

billhollings:
> > Add `Templates/spirv-tools/build.zip`
> 
> What's that for?

It's a cache of auto-gen files that result from a SPIRV-Tools build. 

A SPIRV-Tools build can take a while, and we only use the few auto-gen files, so instead of requiring everyone do that build as part of `fetchDependencies`, we cache the files in that small ZIP file, and extract it in `fetchDependencies`. Plus the SPIRV-Tools build uses CMake instead of Xcode, so it reduces a user dev tool requirement for someone just interested in building MoltenVK to use it.

It's been there forever. Creating the `build.zip` is handled by running `packagePregenSpirvToolsHeaders` as a manual part of the release process after the `glslang` version is updated.

PR #2126, opened 2024-01-12
Fix translation of fully remapped multi vertex attribute bindings

aitor-lunarg:
Fix crash on CTS test:
dEQP-VK.robustness.robustness1_vertex_access.out_of_bounds_stride_30_middle_of_buffer

Bindings with multiple vertex attributes would not correctly detect if the whole binding was remapped to new location/s. This meant original binding's stride was never set to 0 which would report an error and in some cases crash.

### DIFF
--- a/MoltenVK/MoltenVK/GPUObjects/MVKPipeline.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKPipeline.mm
@@ -1491,7 +1491,7 @@ bool MVKGraphicsPipeline::addVertexInputToPipeline(T* inputDesc,
 					vbXltdDesc.stride = vbDesc.stride;
 					vbXltdDesc.stepFunction = vbDesc.stepFunction;
 					vbXltdDesc.stepRate = vbDesc.stepRate;
-					xldtVACnt++;
+					xldtVACnt += xltdBind.mappedAttributeCount;
 				}
 			}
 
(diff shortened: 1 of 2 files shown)


MoltenVK/MoltenVK/GPUObjects/MVKPipeline.mm
```diff
@@ -1533,13 +1533,14 @@ static MTLVertexFormat mvkAdjustFormatVectorToSize(MTLVertexFormat format, uint3
 	// See if a translated binding already exists (for example if more than one VA needs the same translation).
 	for (auto& xltdBind : _translatedVertexBindings) {
 		if (xltdBind.binding == binding && xltdBind.translationOffset == translationOffset) {
+			xltdBind.mappedAttributeCount += 1u;
```
billhollings, line 1536: The style pattern in MoltenVK would use `++` here.

Is there an argument that the `+= 1` idiom might be better?

billhollings (changes requested):
Thanks for submitting this.

LGTM. Just a minor style nit.

MoltenVK/MoltenVK/GPUObjects/MVKPipeline.mm
aitor-lunarg: Changed!

There's no reason other than personal preference, and being unsure of the code-style when I submitted the PR.

billhollings (approved):
Thanks!

Issue #1975, opened 2023-07-10
MoltenVK Enhancement Priorities


billhollings:
This list provides a place to consolidate the priorities and status of some of the larger MoltenVK enhancements whose development is either underway, or will be soon. This list will also help to prevent multiple MoltenVK devs unknowingly working independently on significant enhancements. 

If you are, or soon plan to be, working on a significant MoltenVK enhancement, please comment below, and we'll add it to the list (not every enhancement needs to go on this list, if you are banging out a small enhancement). If you want to take over one of the un-started enhancements, let us know below and it can be assigned to you.

If you are waiting on a significant enhancement, and want to request that it be added to the list, please request that in the issue for that enhancement.

The list is in order of attack. This list does not include bug fixing work.

| | In Dev | Merged | Developer | Enhancement| Deps | Issue# | PR# |
| :-: | :-: | :-: |:-- |:--  |:-: |:-: |:-: 
| 1 | :white_check_mark: |  | @billhollings  | Metal 3 Arg Buffs (enhancing `VK_EXT_descriptor_indexing`) | | | |
| 2 | :white_check_mark: |  | @AntarticCoder  | Vulkan Acceleration Structures | | #1956 | #1967 |
| 3 | :white_check_mark: |  | @gpx1000 | Transform Feedback | | | #1943 |
| 4 |  |  | @billhollings  | (macOS 14+ only) Integrate Apple's Shader Converter tech (including Geom & Tess shaders) | 1 |  | |
| 5 |  |  | @billhollings  | `VK_KHR_synchronization2` | | #1514 | |
| 6 |  |  | @billhollings  | `VK_EXT_descriptor_buffer` | 1, 5 | #1776 | |
| 7 | :white_check_mark: |  | @billhollings  | Pipeline caching using `MTLBinaryArchives` | | #1765 | |

AntarticCoder:
If anyone's interested in implementing this, the spec is [here](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VK_KHR_maintenance4.html). `VK_KHR_maintenance4` is also considered core in Vulkan 1.3, so implementing it is almost a no brainer.

billhollings:
> VK_KHR_maintenance4 is also considered core in Vulkan 1.3

Yes. Vulkan 1.3 requirements are being tracked separately in issue #1930. I've clarified that above, along with how this list interacts with that list.

oscarbg:
Hi,
[...]

PR #2137, opened 2024-01-25
Add optional support for VkPhysicalDeviceFeatures::wideLines.

### LINKED ISSUE #1657: Add line width support as optional compilation macro

### LINKED ISSUE #1670: Expose experimental Vulkan features using internal Metal API

billhollings:
- Add `MVK_USE_METAL_PRIVATE_API` build setting to allow MoltenVK to be built with access to Metal private API calls.
- Add support for VkPhysicalDeviceFeatures::wideLines feature when `MVK_USE_METAL_PRIVATE_API` is enabled in a MoltenVK build.
- Set `lineWidthRange` and `lineWidthGranularity` to reasonable arbitrary values.

Fixes #1657.
Fixes #1670.
Replaces #2091.

Future PR's can leverage `MVK_USE_METAL_PRIVATE_API` to add additional private functionality, as it is uncovered.


MoltenVK/MoltenVK/Commands/MVKCommandEncoderState.h
```diff
@@ -322,7 +324,8 @@ class MVKRenderingCommandEncoderState : public MVKCommandEncoderState {
 	MTLPrimitiveType _mtlPrimitiveTopology[StateScope::Count] = { MTLPrimitiveTypePoint, MTLPrimitiveTypePoint };
 	MTLDepthClipMode _mtlDepthClipEnable[StateScope::Count] = { MTLDepthClipModeClip, MTLDepthClipModeClip };
 	MTLTriangleFillMode _mtlPolygonMode[StateScope::Count] = { MTLTriangleFillModeFill, MTLTriangleFillModeFill };
-	uint32_t _mtlPatchControlPoints[StateScope::Count] = {};
+	float _mtlLineWidth[StateScope::Count] = { 1, 1 };
+	uint32_t _mtlPatchControlPoints[StateScope::Count] = { 0, 0};
```
spnda, line 328: Nitpick, but slight formatting error.

README.md
```diff
@@ -294,27 +298,37 @@ or
...
 statically bound, to provide the application with the ability to retrieve the remaining
 _Vulkan_ API calls as function pointers.
 
 
+<a name="metal_private_api"></a>
+### Accessing _Metal_ Private API calls
+
+You can optionally build **MoltenVK** with access to _Metal_ private API calls. 
+Doing so will allow **MoltenVK** to extend functionality by using certain private _Metal_ API calls, 
+however, doing so will disqualify the app from being distributed via _Apple_ App Stores.
```
spnda, line 322: I somehow really dislike how `_Metal_ private API calls` sounds and the repetition of the `doing so`. I feel like this reads much better:
> You can optionally build **MoltenVK** with access to private _Metal_ API calls.
Doing so will allow **MoltenVK** to extend its functionality by using certain private _Metal_ API calls, but it will also disqualify the app from being distributed via _Apple_ App Stores.

MoltenVK/MoltenVK/Commands/MVKCommandEncoderState.mm
```diff
@@ -322,6 +322,11 @@
 	getContent(_isPolygonModePoint, isDynamic) = (polygonMode == VK_POLYGON_MODE_POINT);
 }
 
+void MVKRenderingCommandEncoderState::setLineWidth(float lineWidth, bool isDynamic) {
+	auto mtlLineWidth = lineWidth;
+	setMTLContent(LineWidth);
```
spnda, line 327: unused local variable `mtlLineWidth`?

spnda (changes requested):

README.md
billhollings: Thanks. I struggled a bit with wording there. Fixed now.

MoltenVK/MoltenVK/Commands/MVKCommandEncoderState.h
billhollings: Fixed.

MoltenVK/MoltenVK/Commands/MVKCommandEncoderState.mm
billhollings: It's part of the setup for the `setMTLContent(LineWidth)` macro pattern. See other nearby uses.

spnda: Just checked again and saw it was a macro specific to the file. I mixed it up with the `setContent` functions from the command buffer classes.

spnda (approved):

Docs/Whats_New.md
```diff
@@ -18,6 +18,8 @@ MoltenVK 1.2.8
 
 Released TBD
 
+- Add `MVK_USE_METAL_PRIVATE_API` build setting to allow **MoltenVK** to be built with access to _Metal_ private API calls.
```
cdavis5e, line 21: This may be bikeshedding, but Apple typically uses the IBM jargon "Service Provider Interface" (SPI) for a private interface like this.

MoltenVK/MoltenVK/Commands/MVKCmdRendering.h
```diff
@@ -607,6 +607,25 @@ class MVKCmdSetPolygonMode : public MVKCommand {
...
+
+	void encode(MVKCommandEncoder* cmdEncoder) override;
+
+protected:
+	MVKCommandTypePool<MVKCommand>* getTypePool(MVKCommandPool* cmdPool) override;
+
+	float _lineWidth;
+};
+
+
```
cdavis5e, line 628: I wonder if all the line width stuff should be hidden behind the compile-time switch.

MoltenVK/MoltenVK/Utility/MVKEnvironment.h
```diff
@@ -73,6 +73,19 @@
...
+ * Enabling this build setting during a MoltenVK build will allow MoltenVK to
+ * extend its functionality by using certain private Metal API calls, but it 
+ * will also disqualify the app from being distributed via Apple App Stores.
+ *
+ * Disabled by default.
+ */
+#ifndef MVK_USE_METAL_PRIVATE_API
+#	define MVK_USE_METAL_PRIVATE_API		0
+#endif
+
```
cdavis5e, line 88: I had actually envisioned a second setting that can be set at run time and controls whether Metal SPIs would be used at run time.

cdavis5e (changes requested):
Dang, you beat me to the punch. :)

I actually have similar changes in my own tree, complete with settings to enable/disable them at build time and at run time. After this one goes in, I'm going to try to get my other changes upstreamed.

Docs/Whats_New.md
billhollings: Fair enough. Although AFAICT, colloquially, everyone seems to refer to them as "private API's", so I think we should continue to include use of that terminology. Although, when you do your add, feel free to add something like an "(aka SPI calls)" into these comments if you feel SPI is well known.

MoltenVK/MoltenVK/Commands/MVKCmdRendering.h
billhollings: I thought of that, but figured it would add clutter with more conditional compiling macros. Plus, this way makes it trivial to adjust if Apple makes this call public in some future release. And if the app ignores `wideLines`, it will just result in a no-op as far as functionality is concerned.

Is your concern just the cost of carrying compiled code that cannot be used?

MoltenVK/MoltenVK/Utility/MVKEnvironment.h
billhollings: Interesting. What's the use case for that? And how would it work (given that the `setLineWidth:` must be completely stripped from any binaries hitting the App Store)?

billhollings:
> Dang, you beat me to the punch. :)

I hate when that happens! 😖

To help with this risk, I try to either assign myself to the related issue (although I didn't do that here because it was a quick fix), or add it to the [upcoming task list](https://github.com/KhronosGroup/MoltenVK/issues/1975) (which I did do on this one).

But I'm open to other mechanisms for us to communicate what's being worked on. It would be good to have some mechanism that we can all collaborate on.

> I actually have similar changes in my own tree, complete with settings to enable/disable them at build time and at run time. After this one goes in, I'm going to try to get my other changes upstreamed.

Okay. Looking forward to seeing it.

MoltenVK/MoltenVK/Commands/MVKCmdRendering.h
spnda: I wonder how the private API usage is determined when uploading to the App Store. I feel like if they see that at some point we invoke private API, no matter if its used at runtime or not, it gets rejected? Something like objdump but for linked symbols would show setLineWidth for the MoltenVK dylib, which could cause issues I think.

Please correct me if this is not the case.

cdavis5e: > Is your concern just the bloat cost of carrying compiled binary code that will never be used?

Yes.

MoltenVK/MoltenVK/Utility/MVKEnvironment.h
cdavis5e: (Huh, I can't see @spnda's comment...)

IIRC, many Objective-C symbols aren't imported the way regular symbols are. Class descriptors and non-fragile i-var descriptors are, but protocol descriptors, property descriptors, and method selectors aren't. Ultimately, due to the dynamic nature of Objective-C, every Objective-C method and property access resolves to a call to `objc_msgSend()`. It would thus not be possible to ascertain which Objective-C interfaces we are using by inspecting the `dyld` info. That can only be determined at run time, when the methods and properties are actually used.

The other reason I wanted a separate switch for run time is that I was afraid that using Metal SPIs might cause other problems, and having that switch would make it easier to test that without having to recompile MoltenVK.

billhollings: > (Huh, I can't see @spnda's comment...)

Neither can I. I assume they must have deleted it.

> It would thus not be possible to ascertain which Objective-C interfaces we are using by inspecting the `dyld` info. That can only be determined at run time, when the methods and properties are actually used.

Are you referring to the App Store figuring out what calls we're making?

If so, we ran into a private-API issue early on with `IOSurface` on iOS. Are you thinking the `IOSurface` class was undefined, and the App Store could tell we were using it, but the `setLineWidth:` method would not be traceable? 

I can't say yes or no to that until we try, but even so, I'm wondering if it might be traceable if it's a unique symbol.

Regardless, I definitely want to avoid a cat-and-mouse game with Apple about what they can or cannot detect in their App Store filtering. I'd rather be clean about it, and ensure we only send legal builds to the App Store.

How about we get this build-time level of functionality in, and then explore and test options for adding runtime sneakiness in afterwards?

cdavis5e: > Are you referring to the App Store figuring out what calls we're making?

Yes.

> If so, we ran into a private-API issue early on with `IOSurface` on iOS. Are you thinking the `IOSurface` class was undefined, and the App Store could tell we were using it, but the `setLineWidth:` method would not be traceable?

Yes.



> Regardless, I definitely want to avoid a cat-and-mouse game with Apple about what they can or cannot detect in their App Store filtering. I'd rather be clean about it, and ensure we only send legal builds to the App Store.
> 
> How about we get this build-time level of functionality in, and then explore and test options for adding runtime sneakiness in afterwards?

Very well. I actually wanted both methods anyway. I'm sorry I wasn't clear about that earlier.

cdavis5e (approved):

PR #2139, opened 2024-01-26
Handle depth/stencil swizzle sample correctly

aitor-lunarg:
Fixes CTS tests:
`dEQP-VK.pipeline.monolithic.sampler.border_swizzle.d*.opaque_white.*`

### DIFF
--- a/MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
@@ -1698,35 +1698,68 @@ VkResult MVKImageViewPlane::initSwizzledMTLPixelFormat(const VkImageViewCreateIn
 			break;
 	}
 
+	// Metal requires special handling when sampling depth/stencil textures in the following cases:
+	// 1. Sampling stencil from a depth/stencil format
+	// 2. Metal's undefined values for depth/stencil sample to vec4 conversion
+	if (mvkIsAnyFlagEnabled(aspectMask, VK_IMAGE_ASPECT_DEPTH_BIT | VK_IMAGE_ASPECT_STENCIL_BIT) &&
+		mvkIsAnyFlagEnabled(_imageView->_usage, (VK_IMAGE_USAGE_SAMPLED_BIT | VK_IMAGE_USAGE_INPUT_ATTACHMENT_BIT)))
+	{
+		// 1. Sampling stencil from a depth/stencil format
+		// To sample the stencil value from a depth/stencil format, Metal requires to drop the depth for the view format.
+		// Meaning that MTLPixelFormatDepth32Float_Stencil8 needs to be MTLPixelFormatX32_Stencil8 for the view according to spec:
+		// You can't directly read the stencil value of a texture with the MTLPixelFormatDepth32Float_Stencil8 format.
+		// To read stencil values from a texture with the MTLPixelFormatDepth32Float_Stencil8 format, create a texture view
+		// of that texture using the MTLPixelFormatX32_Stencil8 format, and sample the texture view instead.
+		if (aspectMask == VK_IMAGE_ASPECT_STENCIL_BIT) {
+			if (_mtlPixFmt == MTLPixelFormatDepth32Float_Stencil8) {
+				_mtlPixFmt = MTLPixelFormatX32_Stencil8;
+			}
+#if MVK_MACOS
+			else if (_mtlPixFmt == MTLPixelFormatDepth24Unorm_Stencil8) {
+				_mtlPixFmt = MTLPixelFormatX24_Stencil8;
+			}
+#endif
+		}
+		
+		// 2. Metal's undefined values for depth/stencil sample to vec4 conversion
+		// Due to differences in Metal and Vulkan specification for sampling depth/stencil textures into vec4, we need to
+		// provide the correct mapping from Vulkan to Metal
+		// Metal states:
+		// "For a texture with depth or stencil pixel format (such as MTLPixelFormatDepth24Unorm_Stencil8 or MTLPixelFormatStencil8),
+		// the default value for an unspecified component is undefined." which means all values but R will be undefined.
+		// Vulkan states: https://registry.khronos.org/vulkan/specs/1.3-extensions/html/vkspec.html#textures-conversion-to-rgba
+		// So we need to ensure that G and B are set to 0 and A to 1 for the swizzle in Metal
+		if (enableSwizzling()) {
+			if (_componentSwizzle.r == VK_COMPONENT_SWIZZLE_A) {
+				_componentSwizzle.r = VK_COMPONENT_SWIZZLE_ONE;
+			} else if (_componentSwizzle.r == VK_COMPONENT_SWIZZLE_G || _componentSwizzle.r == VK_COMPONENT_SWIZZLE_B) {
+				_componentSwizzle.r = VK_COMPONENT_SWIZZLE_ZERO;
+			}
+			if (_componentSwizzle.g == VK_COMPONENT_SWIZZLE_A) {
+				_componentSwizzle.g = VK_COMPONENT_SWIZZLE_ONE;
+			} else if (_componentSwizzle.g == VK_COMPONENT_SWIZZLE_G || _componentSwizzle.g == VK_COMPONENT_SWIZZLE_B || _componentSwizzle.g == VK_COMPONENT_SWIZZLE_IDENTITY) {
+				_componentSwizzle.g = VK_COMPONENT_SWIZZLE_ZERO;
+			}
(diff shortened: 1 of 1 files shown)


MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
```diff
@@ -1697,36 +1697,62 @@ static void signalAndUntrack(const MVKSwapchainSignaler& signaler) {
...
 #define VK_COMPONENT_SWIZZLE_ANY       VK_COMPONENT_SWIZZLE_MAX_ENUM
 
 	// If we have an identity swizzle, we're all good.
 	if (SWIZZLE_MATCHES(R, G, B, A)) {
-
-		// Identity swizzles of depth/stencil formats can require special handling.
-		if (mvkIsAnyFlagEnabled(_imageView->_usage, (VK_IMAGE_USAGE_SAMPLED_BIT | VK_IMAGE_USAGE_INPUT_ATTACHMENT_BIT))) {
-
-			// If only stencil aspect is requested, possibly change to stencil-only format.
-			if (aspectMask == VK_IMAGE_ASPECT_STENCIL_BIT) {
```
aitor-lunarg, line 1711: @billhollings is there any reason this was inside the `SWIZZLE_MATCHES` condition? I tracked down the commits where this was introduced and there's nothing that comes to mind that would make us have this inside that condition. Let me know if I missed anything

MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
```diff
@@ -1838,30 +1864,6 @@ static void signalAndUntrack(const MVKSwapchainSignaler& signaler) {
 			}
 			break;
 
-		case MTLPixelFormatDepth32Float_Stencil8:
```
aitor-lunarg, line 1841: @billhollings I was hesitant to remove this for the same reason as the other comment above. Let me know if I missed something!

MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
```diff
@@ -1697,36 +1697,62 @@ static void signalAndUntrack(const MVKSwapchainSignaler& signaler) {
...
+		if (_componentSwizzle.b == VK_COMPONENT_SWIZZLE_A) {
+			_componentSwizzle.b = VK_COMPONENT_SWIZZLE_ONE;
+		}
+		else if (_componentSwizzle.b == VK_COMPONENT_SWIZZLE_G || _componentSwizzle.b == VK_COMPONENT_SWIZZLE_B || _componentSwizzle.b == VK_COMPONENT_SWIZZLE_IDENTITY) {
+			_componentSwizzle.b = VK_COMPONENT_SWIZZLE_ZERO;
+		}
+		if (_componentSwizzle.a == VK_COMPONENT_SWIZZLE_A || _componentSwizzle.a == VK_COMPONENT_SWIZZLE_IDENTITY) {
+			_componentSwizzle.a = VK_COMPONENT_SWIZZLE_ONE;
+		}
+		else if (_componentSwizzle.a == VK_COMPONENT_SWIZZLE_G || _componentSwizzle.a == VK_COMPONENT_SWIZZLE_B ) {
```
billhollings, line 1744: Minor style issue for consistency. On each of these. please pull the `else` up to the bracket line:  `} else {`. 

Line 1707 is fine because of the conditional compile.

MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
billhollings: Hmmm...nothing springs to mind. We might have long ago made the mistaken assumption that D/S can't be swizzled.

billhollings (changes requested):
Thanks. LGTM. Just a minor style nit.

MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
billhollings: Hmmm...now that you mention it, the code in this section might be about handling swizzles back in the old days before swizzling was available, where we just relied on a swap to a compatible format.

If possible, we should try to retain that for older systems.

billhollings (changes requested):
Actually...a couple of comments about other changes.

MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
billhollings: Sorry...my earlier comments got mixed up because I did two reviews, and you were also editing. Starting over:

> is there any reason this was inside the SWIZZLE_MATCHES condition?

Hmmm...nothing springs to mind. We might have long ago made the mistaken assumption that D/S can't be swizzled.

But  you seem to have removed the part that does:

```
if (SWIZZLE_MATCHES(R, G, B, A)) {
   ...
   return VK_SUCCESS;
}
for non-DS formats.
```

cdavis5e: > Hmmm...nothing springs to mind. We might have long ago made the mistaken assumption that D/S can't be swizzled.

I can't think of any good reason, either. The fact that the code below, which you also removed, also changed the format in the swizzled case suggests this is the right thing to do.

> 
> But you seem to have removed the part that does:
> 
> ```
> if (SWIZZLE_MATCHES(R, G, B, A)) {
>    ...
>    return VK_SUCCESS;
> }
> for non-DS formats.
> ```

No, it's still there.

MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
cdavis5e: The check for the swizzle should probably be retained, especially for the case where `enableSwizzle()` is false. The format switch is handled by the newly refactored code above.

MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
```diff
@@ -1697,36 +1697,62 @@ static void signalAndUntrack(const MVKSwapchainSignaler& signaler) {
...
+	if (aspectMask == VK_IMAGE_ASPECT_STENCIL_BIT) {
+		if (_mtlPixFmt == MTLPixelFormatDepth32Float_Stencil8) {
+			_mtlPixFmt = MTLPixelFormatX32_Stencil8;
+		}
+#if MVK_MACOS
+		else if (_mtlPixFmt == MTLPixelFormatDepth24Unorm_Stencil8) {
+			_mtlPixFmt = MTLPixelFormatX24_Stencil8;
+		}
+#endif
+	}
```
cdavis5e, line 1711: The original code had a check that the view usage included `SAMPLED` or `INPUT_ATTACHMENT` before switching the format. You should probably check that here, too.

cdavis5e (changes requested):

MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
aitor-lunarg: Will do! When I started the review I was not sure why the check was there, but after some research I see it's required. I will add a comment in code for the future.

MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
aitor-lunarg: Fixed

MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
```diff
@@ -1698,35 +1698,68 @@ static void signalAndUntrack(const MVKSwapchainSignaler& signaler) {
...
+		}
+		
+		// 2. Metal's undefined values for depth/stencil sample to vec4 conversion
+		// Due to differences in Metal and Vulkan specification for sampling depth/stencil textures into vec4, we need to
+		// provide the correct mapping from Vulkan to Metal
+		// Metal states:
+		// "For a texture with depth or stencil pixel format (such as MTLPixelFormatDepth24Unorm_Stencil8 or MTLPixelFormatStencil8),
+		// the default value for an unspecified component is undefined." which means all values but R will be undefined.
+		// Vulkan states: https://registry.khronos.org/vulkan/specs/1.3-extensions/html/vkspec.html#textures-conversion-to-rgba
+		// So we need to ensure that G and B are set to 0 and A to 1 for the swizzle in Metal
```
billhollings, line 1731: For readability, Maybe reword:

```suggestion
		// Vulkan requires that all components be defined, with `G` and `B` set to `0` and `A` set to `1`.
		// See https://registry.khronos.org/vulkan/specs/1.3-extensions/html/vkspec.html#textures-conversion-to-rgba.
```

billhollings (approved):
LGTM now. Just one minor comment rewording suggestion.

cdavis5e (approved):
LGTM

PR #2141, opened 2024-01-26
Don't update currentExtent of headless surface when swapchain attached.

### LINKED ISSUE #2129: Once a swapchain is bound to a headless surface, its currentExtent is no longer -1
I'm not sure that this is the intended behavior of headless surfaces, but: when a headless surface is created, it has a currentExtent of (-1, -1) (of course, since VkDeviceSize is unsigned, this is just 0xFFFF...). However, once a swapchain is created on the surface, the `currentExtent` returned by `vkGetPhysicalDeviceSurfaceCapabilitiesKHR` is now fixed to the `imageExtent` of the swapchain. This means that if you wanted to attach a swapchain with a larger extent to the surface, you can't, and would have to create a new surface.

Since the surface is headless, there shouldn't be a limit to the extent of any images presented to it.

billhollings:
Fixes #2129.

### DIFF
--- a/Docs/Whats_New.md
+++ b/Docs/Whats_New.md
@@ -24,6 +24,8 @@ Released TBD
 - Ensure buffers available for buffer addresses in push constants.
 - Support `libMoltenVK.dylib` for _iOS Simulator_ architecture.
 - Restore support for _iOS Simulator_ destination in recent update to _Cube_ demo that uses dynamic-linking.
+- Don't update `currentExtent` of headless surface when swapchain attached.
+
 
 
 MoltenVK 1.2.7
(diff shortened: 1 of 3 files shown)


Docs/Whats_New.md
```diff
@@ -23,7 +23,10 @@ Released TBD
 - Fix potential crash when using multi-planar images.
 - Ensure buffers available for buffer addresses in push constants.
 - Support `libMoltenVK.dylib` for _iOS Simulator_ architecture.
-- Restore support for _iOS Simulator_ destination in recent update to _Cube_ demo that uses dynamic-linking.
+- Restore support for _iOS Simulat
+Speor_ destination in recent update to _Cube_ demo that uses dynamic-linking.
```
cdavis5e, line 27:
```suggestion
- Restore support for _iOS Simulator_ destination in recent update to _Cube_ demo that uses dynamic-linking.
```

I take it you didn't mean to insert those spurious characters there.

cdavis5e (approved):
Just one small issue I caught.

Docs/Whats_New.md
billhollings: WTF? I must have sneezed with my hands over the keyboard!

Thanks. You're proofreading is, as always, impeccable!

PR #2142, opened 2024-01-27
Add more Metal SPI support

cdavis5e:
This change adds several more Metal private APIs (aka "Service Provider Interfaces" or SPIs), including:

* Framebuffer logic operations
* MIP LoD bias on samplers
* Native sample masking, using a property on `MTLRenderPipelineDescriptor` instead of emulating in the fragment shader, which doesn't always work quite right (as @aitor-lunarg could probably attest)!
* Depth bounds test (requires an AMD GPU)

It also adds a run-time configuration parameter corresponding to the build-time `MVK_USE_METAL_PRIVATE_API`, which will allow disabling the use of these SPIs without recompiling MoltenVK as a troubleshooting measure.


cdavis5e:
It has occurred to me that we should probably add a CI configuration that builds with `MVK_USE_METAL_PRIVATE_API=1`, so it doesn't succumb to build rot.

alyssarosenzweig:
Notes for AGX:

* Sampler LOD bias is implemented in shader. Does this code work on Apple Silicon? the pipeline needs to be keyed for it. MoltenVK should probably emulate itself, at least on AGX.
* No hw support for depth bounds test AFAIK.
* Logic ops is in shader. MoltenVK should probably emulate itself, no private API needed and the private API will be no faster on Apple.
* Sample masking is in shader but in theory the driver has enough control to do it correctly, which the metal app doesn't, so this one makes sense to me.

MoltenVK/MoltenVK/OS/MTLRenderPipelineColorAttachmentDescriptor+MoltenVK.m
```diff
@@ -0,0 +1,99 @@
...
+			MTLColorWriteMask writeMask : 4;
+			BOOL logicOpEnabled : 1;
+			MTLLogicOperation logicOp : 4;
+			MTLPixelFormat pixelFormat : 28;
+		};
+		struct {
+			uint64_t bits;
+		};
+	};
+} MTLRenderPipelineAttachmentDescriptorPrivate;
```
billhollings, line 47: What's going on here? Is this a private Metal struct, that has bitwise density, and that we've reversed engineered? 

If so, how was that even done? And the obvious concern is that this adds a whole different level of fragility to the builds and runtimes.

MoltenVK/MoltenVK/OS/MTLRenderPipelineColorAttachmentDescriptor+MoltenVK.m
```diff
@@ -0,0 +1,99 @@
...
+- (BOOL)isLogicOpEnabledMVK {
+	if ([self respondsToSelector:@selector(isLogicOpEnabled)]) {
+		return [self isLogicOpEnabled];
+	}
+	if (![self respondsToSelector:@selector(_descriptorPrivate)]) {
+		return NO;
+	}
+	return [self _descriptorPrivate]->logicOpEnabled;
+}
+
```
billhollings, line 70: Is this saying there are cases where `logicOpEnabled` is actually a supported (private) Metal property, but other cases we have to reach into `_descriptorPrivate`? 

Is there rhyme or reason to it (eg. OS version, GPU type, etc)? Is it worth supporting the former, but not the latter (again, to reduce fragility)?

billhollings (changes requested):
This is cool and interesting. But IMHO some of it is introducing a level of unofficial SPI plumbing and potential fragility that I'm concerned about.

Specific suggestions and concerned listed below.

billhollings:
> It has occurred to me that we should probably add a CI configuration that builds with `MVK_USE_METAL_PRIVATE_API=1`, so it doesn't succumb to build rot.

I agree. Particularly since parts of these changes are adding fragility. 

Even so, doing so will only test current build and runtime environments. But I guess we can leave legacy OS version support testing to users who need to support it.

MoltenVK/MoltenVK/API/mvk_private_api.h
```diff
@@ -240,6 +240,8 @@ typedef struct {
 	MVKConfigCompressionAlgorithm shaderSourceCompressionAlgorithm;            /**< MVK_CONFIG_SHADER_COMPRESSION_ALGORITHM */
 	VkBool32 shouldMaximizeConcurrentCompilation;                              /**< MVK_CONFIG_SHOULD_MAXIMIZE_CONCURRENT_COMPILATION */
 	float timestampPeriodLowPassAlpha;                                         /**< MVK_CONFIG_TIMESTAMP_PERIOD_LOWPASS_ALPHA */
+	VkBool32 useMetalPrivateAPI;                                               /**< MVK_CONFIG_USE_METAL_PRIVATE_API */
+	uint32_t _reserved;
```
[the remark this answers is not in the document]
cdavis5e: The compiler insisted on making the struct size a multiple of 8, so the `static_assert()` in `MoltenVK/MoltenVK/Utility/MVKEnvironment.cpp` fails unless this is present in `MVKConfigMembers.def`, and therefore it must be present here as well.

MoltenVK/MoltenVK/OS/MTLRenderPipelineColorAttachmentDescriptor+MoltenVK.m
cdavis5e: It was present in the Objective-C metadata for `Metal.framework`. At the time I wrote that, I couldn't see a method for setting this. Frankly, I don't like it either, because it *is* an internal implementation detail of Metal. Now there is such a method, so if you like, I can remove this.

MoltenVK/MoltenVK/OS/MTLRenderPipelineColorAttachmentDescriptor+MoltenVK.m
cdavis5e: > Is this saying there are cases where `logicOpEnabled` is actually a supported (private) Metal property, but other cases we have to reach into `_descriptorPrivate`?

Yes, pretty much.

> Is there rhyme or reason to it (eg. OS version, GPU type, etc)?

I believe it is based on OS version.

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
```diff
@@ -366,7 +366,7 @@
 				portabilityFeatures->multisampleArrayImage = _metalFeatures.multisampleArrayTextures;
 				portabilityFeatures->mutableComparisonSamplers = _metalFeatures.depthSampleCompare;
 				portabilityFeatures->pointPolygons = false;
-				portabilityFeatures->samplerMipLodBias = false;
+				portabilityFeatures->samplerMipLodBias = MVK_USE_METAL_PRIVATE_API && getMVKConfig().useMetalPrivateAPI;
```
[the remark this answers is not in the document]
cdavis5e: Done.

MoltenVK/MoltenVK/API/mvk_private_api.h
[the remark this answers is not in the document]
billhollings: Ah. So I just got lucky with struct boundaries when I added that assertion test!

That assertion is just a dev safety to ensure we keep `MVKConfigMembers.def` in sync with the definition of `MVKConfiguration`.

Darn. If `MVKConfiguration` wasn't part of the public API, we could just let its structure be defined by `MVKConfigMembers.def`. Maybe we should anyway, since this is really a private API now that `VK_EXT_layer_settings` is active.

But that's a larger discussion for another time.

For now, I'm imagining another dev coming by months down the road, and thinking that member is reserved by someone, and adding another config member on after it. So...to clarify intended usage, maybe change it from `uint32_t _reserved` to something like `uint32_t _unused_struct_padding`, so its purpose is clearer? And maybe add a comment indicating it's purpose.

MoltenVK/MoltenVK/Utility/MVKConfigMembers.def
```diff
@@ -81,6 +81,8 @@ MVK_CONFIG_MEMBER(useMetalArgumentBuffers,                MVKUseMetalArgumentBuf
 MVK_CONFIG_MEMBER(shaderSourceCompressionAlgorithm,       MVKConfigCompressionAlgorithm,            SHADER_COMPRESSION_ALGORITHM)
 MVK_CONFIG_MEMBER(shouldMaximizeConcurrentCompilation,    VkBool32,                                 SHOULD_MAXIMIZE_CONCURRENT_COMPILATION)
 MVK_CONFIG_MEMBER(timestampPeriodLowPassAlpha,            float,                                    TIMESTAMP_PERIOD_LOWPASS_ALPHA)
+MVK_CONFIG_MEMBER(useMetalPrivateAPI,                     VkBool32,                                 USE_METAL_PRIVATE_API)
+MVK_CONFIG_MEMBER(_reserved,                              uint32_t,                                 USE_METAL_PRIVATE_API)
```
billhollings, line 85: This padding member will need a unique name. Per my note above, perhaps something like:

    MVK_CONFIG_MEMBER(_unused_struct_padding,                              uint32_t,                                 UNUSED_STRUCT_PADDING)

MoltenVK/MoltenVK/OS/MTLRenderPipelineColorAttachmentDescriptor+MoltenVK.m
billhollings: Yeah. I respect your deep research, but at least for now, I think I'd feel much better relying on just the private `logicOpEnabled` property (which you use below), and save this level of intrusion for if in future that we get complaints that the private `logicOpEnabled` property doesn't cover enough platforms or OS versions.

MoltenVK/MoltenVK/OS/MTLRenderPipelineColorAttachmentDescriptor+MoltenVK.m
billhollings: > I believe it is based on OS version.

Yeah...probably. See my follow up note above. For now, until proven necessary by customer complaints, I'd feel better leaving `_descriptorPrivate` out, and just relying only on the `logicOpEnabled` property.

billhollings (changes requested):
Some further thoughts on robustness & clarity future-proofing.

cdavis5e:
I only converted `MVKCmdSetDepthBoundsTestEnable` to derive from `MVKSimpleValueCommand<T>`, because `MVKCmdSetDepthBounds` actually has two values.

PTAL

billhollings (approved):
LGTM.

oscarbg:
Hi,
curious now that Metal 4 on Apple10 family devices (A19 and M5?) support depth bounds test via "public API".. any work to support in MoltenVK?
note also Apple10 Metal family support also includes:
0)8x MSAA
1)32kx32k framebuffers/textures
2)depth bounds testing
3)sampler lod bias
but 0) and 1) should be supported automatically in MoltenVK right now, right?
seen https://developer.apple.com/metal/Metal-Feature-Set-Tables.pdf

thanks..

PR #2144, opened 2024-01-27
Simplify repetitive copy-paste declarations of single-value Vulkan commands.

billhollings:
There are many `MVKCommand` subclasses that simply encode a single value, resulting in loads of repetitive copy-pasted code.
- Add abstract template superclass `MVKSingleValueCommand` for these concrete command classes.
- Modify `MVKCmdSetBlendConstants` to hold an `MVKColor32` so it can subclass from `MVKSingleValueCommand`, and modify `MVKGraphicsPipeline` and `MVKRenderingCommandEncoderState` to use `MVKColor32`.
- Remove trivial comments for simple concrete Vulkan command classes.

@cdavis5e This has interactions with your PR #2142, particularly your new `MVKCmdSetDepthBounds` and `MVKCmdSetDepthBoundsTestEnable` classes, which should both become subclasses of `MVKSingleValueCommand ` (for `MVKCmdSetDepthBounds`, by passing & tracking an `MVKDepthBounds`, similar to how `MVKCmdSetBlendConstants` passes and tracks a `MVKColor32`).

So we'll need to merge these two PRs in order. We can either merge this PR first, and you modify PR #2142, or merge #2142 first, once we've resolved the discussion there, and then I'll add `MVKCmdSetDepthBounds` and `MVKCmdSetDepthBoundsTestEnable` to this PR.

### DIFF
--- a/MoltenVK/MoltenVK/Commands/MVKCmdDebug.h
+++ b/MoltenVK/MoltenVK/Commands/MVKCmdDebug.h
@@ -42,7 +42,6 @@ protected:
 #pragma mark -
 #pragma mark MVKCmdDebugMarkerBegin
 
-/** Vulkan command to begin a marker region into the command buffer. */
 class MVKCmdDebugMarkerBegin : public MVKCmdDebugMarker {
 
 public:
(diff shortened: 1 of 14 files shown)


cdavis5e (approved):
I think this should go in first, and then I'll update my PR.

PR #2147, opened 2024-01-30
runcts script also output a file containing a list of the failed CTS tests.

billhollings:
- Add `get_failing_cts_tests.py` script to extract failures from `runcts` output, and call automatically from `runcts`.


aitor-lunarg (approved):
LGTM!

PR #2155, opened 2024-02-08
Consolidation and fixes for MVK_USE_METAL_PRIVATE_API functionality.

billhollings:
- Expose `MTLRenderPipelineDescriptor.sampleMaskMVK` and `MTLSamplerDescriptor.lodBiasMVK` properties when `MVK_USE_METAL_PRIVATE_API` build setting is disabled.
- `MVKCmdSetDepthBias` & `MVKCmdSetDepthBounds` subclass from `MVKSingleValueCommand`.
- `MVKRenderingCommandEncoderState` simplify calls to set depth bias and depth bounds, and make consistent with other settings.
- `MVKPipeline` remove test for depth bounds support since it's checked before GPU encoding.
- `MVKDepthBias` member order same as in Vulkan calls.
- `Whats_New.md` consolidate notes about `MVK_USE_METAL_PRIVATE_API`.


MoltenVK/MoltenVK/Commands/MVKCommandEncoderState.mm
```diff
@@ -333,33 +333,16 @@
...
-													 float depthBiasSlopeFactor,
-													 float depthBiasClamp) {
-	bool isDynamic = true;
-	MVKDepthBias mtlDepthBias = {
-		.depthBiasConstantFactor = depthBiasConstantFactor,
-		.depthBiasSlopeFactor = depthBiasSlopeFactor,
-		.depthBiasClamp = depthBiasClamp
-	};
+void MVKRenderingCommandEncoderState::setDepthBias(MVKDepthBias depthBias, bool isDynamic) {
+	MVKDepthBias mtlDepthBias = depthBias;
```
cdavis5e, line 341: I wonder if we are now at the point where we could refactor `setMTLContent()` not to need this copy.

cdavis5e (approved):

cdavis5e (approved):

PR #2164, opened 2024-02-23
Enable use of native texture atomics.

js6i:
This commit conditionally skips the emulated image atomics paths if native texture atomics are available and a configuration option is set.

Apart from unlocking some potential performance benefits from not having to force some textures to be linear, it also makes texture atomics work with argument buffers.

### DIFF
--- a/MoltenVK/MoltenVK/API/mvk_private_api.h
+++ b/MoltenVK/MoltenVK/API/mvk_private_api.h
@@ -359,6 +359,7 @@ typedef struct {
 	VkDeviceSize hostMemoryPageSize;				/**< The size of a page of host memory on this platform. */
 	VkBool32 dynamicVertexStride;					/**< If true, VK_DYNAMIC_STATE_VERTEX_INPUT_BINDING_STRIDE is supported. */
 	VkBool32 needsCubeGradWorkaround;				/**< If true, sampling from cube textures with explicit gradients is broken and needs a workaround. */
+	VkBool32 nativeTextureAtomics;                  /**< If true, atomic operations on textures are supported natively. */
 } MVKPhysicalDeviceMetalFeatures;
 
 
--- a/MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
@@ -1810,6 +1810,8 @@ void MVKPhysicalDevice::initMetalFeatures() {
 #if MVK_XCODE_15
 	// Dynamic vertex stride needs to have everything aligned - compiled with support for vertex stride calls, and supported by both runtime OS and GPU.
 	_metalFeatures.dynamicVertexStride = mvkOSVersionIsAtLeast(14.0, 17.0, 1.0) && (supportsMTLGPUFamily(Apple4) || supportsMTLGPUFamily(Mac2));
+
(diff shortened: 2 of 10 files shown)


MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
```diff
@@ -2328,6 +2328,11 @@
 	}
 #endif
 
+#if MVK_XCODE_15
+	if (mslVersionIsAtLeast(MTLLanguageVersion3_1)) {
+		_metalFeatures.nativeTextureAtomics = true;
+	}
+#endif
```
billhollings, line 2335: Combine this section with existing `MVK_XCODE_15`:

https://github.com/KhronosGroup/MoltenVK/blob/37361ccd5595ff08c050981856903713f96221d2/MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm#L1810C1-L1814C1

MoltenVK/MoltenVK/API/mvk_private_api.h
```diff
@@ -241,7 +241,8 @@ typedef struct {
 	VkBool32 shouldMaximizeConcurrentCompilation;                              /**< MVK_CONFIG_SHOULD_MAXIMIZE_CONCURRENT_COMPILATION */
 	float timestampPeriodLowPassAlpha;                                         /**< MVK_CONFIG_TIMESTAMP_PERIOD_LOWPASS_ALPHA */
 	VkBool32 useMetalPrivateAPI;                                               /**< MVK_CONFIG_USE_METAL_PRIVATE_API */
-	uint32_t _unused_struct_padding;
+	VkBool32 useMetalNativeTextureAtomics;                                     /**< MVK_CONFIG_USE_METAL_NATIVE_TEXTURE_ATOMICS */
```
billhollings, line 244: What is the need for a config parameter? Are there use cases where the app would want to override `MVKPhysicalDeviceMetalFeatures::nativeTextureAtomics` for some reason?

billhollings (changes requested):
Thanks for submitting.

Question about need for config parameter, and minor nit on code org.

MoltenVK/MoltenVK/API/mvk_private_api.h
js6i: I guess I was being overly tentative about possible cases which would be supported by the emulation, but not natively. I mostly had cubes in mind, but they're not supported either way.. I got rid of the parameter.

billhollings (approved):
Thanks. LGTM now.

js6i:
Apart from #2168, I found a second issue when testing another patch - the validation layer complains about `MTLTextureUsageShaderAtomic` when the pixel format isn't an integer, which currently triggers for buffer views, and images in the `_mutableFormat` case. Maybe I was not _overly_ tentative after all.

Perhaps, for image views, we could force the primary pixel format of the texture to be an integer, and then make views on that..

@billhollings

billhollings:
> Apart from #2168, I found a second issue when testing another patch - the validation layer complains about `MTLTextureUsageShaderAtomic` when the pixel format isn't an integer, which currently triggers for buffer views, and images in the `_mutableFormat` case. Maybe I was not _overly_ tentative after all.
> 
> Perhaps, for image views, we could force the primary pixel format of the texture to be an integer, and then make views on that..
> 
> @billhollings

Can you open an issue tracker for this second valuation issue?

PR #2166, opened 2024-02-25
OLD:  Improve support for iOS App Store rules by using dynamic XCFramework.

billhollings:
Apple's iOS App Store does not permit an app to link to naked `dylibs`. Instead, these must be placed in frameworks, which are embedded in a dynamic version of `MoltenVK.xcframework`.

- Move static XCFramework, containing `libMoltenVK.a` static libraries, to `Package/Latest/MoltenVK/static/MoltenVK.xcframework`.
- Generate dynamic XCFramework, containing `MoltenVK.framework` dynamic libraries, in `Package/Latest/MoltenVK/dynamic/MoltenVK.xcframework`.
- Always run MoltenVK build scripts, to ensure all components are added to the XCFrameworks, and `MoltenVK/Package` is always refreshed, even if code compilation is not required.
- Consolidate scripts `copy_to_staging.sh` and `copy_lib_to_staging.sh`.
- Fix `make install` to correctly install `/usr/local/lib/libMoltenVK.dylib` on macOS.
- Cube demo link to dynamic `MoltenVK.framework` through `MoltenVK.xcframework`, instead of to naked `libMoltenVK.dylib`.
- Update `README.md` and `MoltenVK_Runtime_UserGuide.md` documents.

@richard-lunarg

PR #2170, opened 2024-03-01
Improve support for iOS App Store rules by using Xcode-built dynamic XCFramework.

billhollings:
Apple's iOS App Store does not permit an app to link to naked `dylibs`.
Instead, these must be placed in frameworks, which are embedded in a 
dynamic version of `MoltenVK.xcframework`.

- Use Xcode to directly generate a MoltenVK.framework for each platform,
  and remove `create_dylib*.sh` scripts.
- Move static `XCFramework`, containing `libMoltenVK.a` static libraries, 
  to `Package/Latest/MoltenVK/static/MoltenVK.xcframework`.
- Generate dynamic `XCFramework`, containing `MoltenVK.framework` dynamic 
  libraries, in `Package/Latest/MoltenVK/dynamic/MoltenVK.xcframework`.
- Add macro `MVK_VERSION_STRING` to create version string at compile time, 
  use it to validate the `CURRENT_PROJECT_VERSION` build setting at compile time,
  and use it at runtime instead of `mvkGetMoltenVKVersionString()` function.
- Add `-w` to `OTHER_LDFLAGS` to dynamic framework builds to suppress 
  spurious linker warnings of the type 
  `ld: warning: no platform load command found in '...', assuming: iOS` 
  issued from the new linker introduced in Xcode 15.
- Add `MoltenVK-MacCat` Xcode target and _MoltenVK Package (MacCat only)_ 
  Xcode scheme to avoid building dynamic `MoltenVK.framework` for the 
  Mac Catalyst platform, because Xcode does not support doing so.
- Always run MoltenVK build scripts, to ensure all components are 
  added to the `XCFrameworks`, and `MoltenVK/Package` is always refreshed, 
  even if code compilation is not required.
- Cube demo link to dynamic `MoltenVK.framework` through 
  `dynamic/MoltenVK.xcframework`, instead of to naked `libMoltenVK.dylib`.
- Update the version of Volk used by the Cube demo, to support 
  loading MoltenVK from dynamic frameworks inside Volk.
- Update `README.md` and `MoltenVK_Runtime_UserGuide.md` documents.
- Update `MVK_PRIVATE_API_VERSION` to `40`.
- Fix make install to install `/usr/local/lib/libMoltenVK.dylib` on macOS (unrelated).
- Remove unused `MTLAttributeStrideStatic` declaration prior to Xcode 15 (unrelated). 

This PR replaces #2166.


billhollings:
@richard-lunarg 

This PR replaces #2166. This PR builds dynamic frameworks directly from Xcode, rather than from a script. Among efficiencies and stability, doing so allows tvOS Simulator to be supported with a dynamic framework.

richard-lunarg (approved):
I've tested this. Built with XCode 15.3 and it passes App store validation. As they say, "Ship it!".

PR #2173, opened 2024-03-06
Don't set MTLTextureUsageShaderAtomic if the pixel format does not support it.

### LINKED ISSUE #2172: Issues with native atomics
The only Metal pixel formats that support `MTLTextureUsageShaderAtomic` are R32Int, R32UInt and RG32UInt. Currently other textures can be assigned that usage, causing an assertion. That's fallout from #2164.

First case is buffer views. We should only add atomic usage to the supported formats. VK_EXT_shader_atomic_float needs to be handled in some other way - maybe we could just use the old emulated path. However, we don't know the format when setting up the layout, which maybe is a problem in some cases..

Second is images which support integer views, but the base pixel format does not support atomics. My initial idea was to make the base Metal texture be integer instead, and return views on that, which seems to work, but I'm a bit worried about it breaking something, and we could instead try just not setting `MTLTextureUsageShaderAtomic` - that's wrong by the docs, but in my testing, Metal does not seem to sound any alarms.

This issue was initially reported [here](https://github.com/KhronosGroup/MoltenVK/issues/2168#issuecomment-1974894333).

@billhollings @cdavis5e

js6i:
Partially fixes #2172. Outstanding problems are float atomics and that we're not handling image views correctly, as described in the issue.

### DIFF
--- a/MoltenVK/MoltenVK/GPUObjects/MVKPixelFormats.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKPixelFormats.mm
@@ -722,7 +722,7 @@ MTLTextureUsage MVKPixelFormats::getMTLTextureUsage(VkImageUsageFlags vkImageUsa
 	}
 
 #if MVK_XCODE_15
-	if (supportAtomics) {
+	if (supportAtomics && (mtlFormat == MTLPixelFormatR32Uint || mtlFormat == MTLPixelFormatR32Sint || mtlFormat == MTLPixelFormatRG32Uint)) {
 		mvkEnableFlags(mtlUsage, MTLTextureUsageShaderAtomic);
 	}
 #endif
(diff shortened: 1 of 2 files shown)


MoltenVK/MoltenVK/GPUObjects/MVKPixelFormats.mm
```diff
@@ -722,7 +722,7 @@
 	}
 
 #if MVK_XCODE_15
-	if (supportAtomics) {
+	if (supportAtomics && (mtlFormat == MTLPixelFormatR32Uint || mtlFormat == MTLPixelFormatR32Sint || mtlFormat == MTLPixelFormatRG32Uint)) {
```
billhollings, line 725: Does `supportAtomics` not include these additional checks already, via the setting of `MVKImage::_shouldSupportAtomics`?

Maybe add `VK_FORMAT_R32G32_UINT` to the `_shouldSupportAtomics` setting test, and also review and confirm the applicability of the additional tests for view-class there?

I'd like to avoid having same-but-not-the same logic in different places, unless there is some necessary reason why it's needed. And then document the difference if there is one.

billhollings (changes requested):
Question about the pixel format addition.

MoltenVK/MoltenVK/GPUObjects/MVKPixelFormats.mm
js6i: Not quite - this test fails the `_hasMutableFormat` case where pixel format is different, but the image support integer views. We could instead disable `_shouldSupportAtomics`. My rationale for putting it here was that it still *should* support atomics, we just can't set the usage bit. Like explained in the issue, that seems to work fine, but may not on another device or after an update, and if so we'll need to do something more intrusive.

billhollings: Okay. We'll see what happens with CTS testing.

billhollings (approved):

PR #2178, opened 2024-03-08
Document GPU capture.

billhollings:


skalarproduktraum:
Thanks for adding this, @billhollings! I had unfortunately completely forgotten about my PR :(

Issue #2151, opened 2024-02-04
VK_EXT_metal_objects crashes (reference counting?)


MennoVink:
I am compiling my app with ARC enabled. I'm by no means an obj-c expert but what i understand from the id<> type this automatically releases the referenced object. This means that any use of this extension crashes my app because MVK doesn't retain references before returning the ids.

If i modify MVK to contain retain calls on everything being returned from `MVKDevice::getMetalObjects` i do not get these crashes anymore.

I dont even have to use the returned object for this to crash. Calling this simple utility function in my app (from within @autoreleasepool) is enough to crash:

```
id< MTLDevice > VulkanMetalUtils::GetMetalDevice( const VulkanLogicalDevice& vkDevice )
{
	vk::StructureChain< vk::ExportMetalObjectsInfoEXT, vk::ExportMetalDeviceInfoEXT > exportInfoChain;
	vk::ExportMetalObjectsInfoEXT& exportInfo = exportInfoChain.get< vk::ExportMetalObjectsInfoEXT >();
	vk::ExportMetalDeviceInfoEXT& deviceInfo = exportInfoChain.get< vk::ExportMetalDeviceInfoEXT >();

	vkDevice.GetHandle().exportMetalObjectsEXT( &exportInfo );
	
	return deviceInfo.mtlDevice;
}
```

XCode wont allow be to add retain calls on my side as I'm compiling with ARC so retain/release calls are not allowed.

billhollings:
I'm not able to replicate this in the Cube demo, where I enabled ARC, and added similar code to what you described:

```
@autoreleasepool {
	id<MTLDevice> mtlDev = self.getMetalDevice;
	NSLog(@"MTLDevice: %@", mtlDev.name);
}

-(id<MTLDevice>) getMetalDevice {
	id<MTLDevice> mtlDev;
	VkExportMetalDeviceInfoEXT mtlDevInfo = {VK_STRUCTURE_TYPE_EXPORT_METAL_DEVICE_INFO_EXT, nil, nil};
	VkExportMetalObjectsInfoEXT mtlObjInfo = {VK_STRUCTURE_TYPE_EXPORT_METAL_OBJECTS_INFO_EXT, &mtlDevInfo };
	vkExportMetalObjectsEXT(demo.device, &mtlObjInfo);
	return mtlDevInfo.mtlDevice;
}
``` 

All seems to work as expected, with no crashing.

[...]

PR #2179, opened 2024-03-11
Fixed crashes when using VK_EXT_metal_objects from ARC enabled apps

MennoVink:
Fixing https://github.com/KhronosGroup/MoltenVK/issues/2151

While this fixes ARC enabled apps, it will require non-ARC apps to manually release. I think that is to be expected, how else can the pointers be usable for an app-determined amount of time. Should i add some documentation for this somewhere?

Could you review if there need to be checks for making sure there is any output? For example:
```
auto* pBuffInfo = (VkExportMetalBufferInfoEXT*)next;
auto* mvkDevMem = (MVKDeviceMemory*)pBuffInfo->memory;
pBuffInfo->mtlBuffer = mvkDevMem->getMTLBuffer();
[pBuffInfo->mtlBuffer retain];
```
It's not checking if `mvkDevMem` is even valid before calling `getMTLBuffer` on it. Consequently i didn't check if `getMTLBuffer`'s result is valid before calling retain on it. I'm assuming this is part of Vulkan's VU specs.

### DIFF
--- a/MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
@@ -4699,18 +4699,21 @@ void MVKDevice::getMetalObjects(VkExportMetalObjectsInfoEXT* pMetalObjectsInfo) 
 			case VK_STRUCTURE_TYPE_EXPORT_METAL_DEVICE_INFO_EXT: {
 				auto* pDvcInfo = (VkExportMetalDeviceInfoEXT*)next;
 				pDvcInfo->mtlDevice = getMTLDevice();
+				[pDvcInfo->mtlDevice retain];
 				break;
 			}
 			case VK_STRUCTURE_TYPE_EXPORT_METAL_COMMAND_QUEUE_INFO_EXT: {
 				auto* pQInfo = (VkExportMetalCommandQueueInfoEXT*)next;
 				MVKQueue* mvkQ = MVKQueue::getMVKQueue(pQInfo->queue);
 				pQInfo->mtlCommandQueue = mvkQ->getMTLCommandQueue();
+				[pQInfo->mtlCommandQueue retain];
 				break;
 			}
 			case VK_STRUCTURE_TYPE_EXPORT_METAL_BUFFER_INFO_EXT: {
 				auto* pBuffInfo = (VkExportMetalBufferInfoEXT*)next;
 				auto* mvkDevMem = (MVKDeviceMemory*)pBuffInfo->memory;
 				pBuffInfo->mtlBuffer = mvkDevMem->getMTLBuffer();
+				[pBuffInfo->mtlBuffer retain];
 				break;
 			}
 			case VK_STRUCTURE_TYPE_EXPORT_METAL_TEXTURE_INFO_EXT: {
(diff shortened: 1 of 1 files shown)


cdavis5e:
> Fixing #2151
> 
> While this fixes ARC enabled apps, it will require non-ARC apps to manually release. I think that is to be expected, how else can the pointers be usable for an app-determined amount of time. Should i add some documentation for this somewhere?

Probably. Really, it ought to be documented as part of the `VK_EXT_metal_objects` extension itself. But since AFAICT you're not a member of Khronos, we'll have to get someone else to update the extension...

> 
> Could you review if there need to be checks for making sure there is any output? For example:
> 
> ```
> auto* pBuffInfo = (VkExportMetalBufferInfoEXT*)next;
> auto* mvkDevMem = (MVKDeviceMemory*)pBuffInfo->memory;
> pBuffInfo->mtlBuffer = mvkDevMem->getMTLBuffer();
> [pBuffInfo->mtlBuffer retain];
> ```
> 
> It's not checking if `mvkDevMem` is even valid before calling `getMTLBuffer` on it. Consequently i didn't check if `getMTLBuffer`'s result is valid before calling retain on it. I'm assuming this is part of Vulkan's VU specs.

Generally speaking, if you were to pass a null handle to a Vulkan function where the VU says that the handle **must** [original emphasis] be valid, the behavior from that point is undefined, and _**literally anything**_ is allowed to happen--from the program crashing to the program working fine to the command `rm -rf /` being executed as root to (as compiler engineers like to say) demons spontaneously flying out of your nose. So if there's a VU saying that `VkExportMetalBufferInfoEXT::memory` **must** be valid--which, of course, [there is][1]--you can avoid checking for null. Not requiring all these null checks allows Vulkan drivers to run faster--the constant validation that OpenGL implementations were required to perform was a notorious performance bottleneck.

[1]: https://registry.khronos.org/vulkan/specs/1.3-extensions/html/chap11.html#VUID-VkExportMetalBufferInfoEXT-memory-parameter

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
```diff
@@ -4726,12 +4729,14 @@ static uint32_t mvkGetEntryProperty(io_registry_entry_t entry, CFStringRef prope
...
 					pImgInfo->mtlTexture = mvkBuffView->getMTLTexture();
 				}
+				[pImgInfo->mtlTexture retain];
 				break;
 			}
 			case VK_STRUCTURE_TYPE_EXPORT_METAL_IO_SURFACE_INFO_EXT: {
 				auto* pIOSurfInfo = (VkExportMetalIOSurfaceInfoEXT*)next;
 				auto* mvkImg = (MVKImage*)pIOSurfInfo->image;
 				pIOSurfInfo->ioSurface = mvkImg->getIOSurface();
+				CFRetain(pIOSurfInfo->ioSurface);
```
cdavis5e, line 4739: This is a curious case here. `IOSurface` was originally a Core Foundation-style object, which means that you had to use manual retain/release even from ARC-enabled Objective-C. But now there's an actual `IOSurface` class. Further, Swift is capable of automatically managing Core Foundation objects, and I believe it will actually wind up importing any `IOSurfaceRef` as an instance of the Objective-C `IOSurface` class. So, you really do need the `CFRetain()` here for Swift. (From Objective-C that means that the caller would either have to `CFRelease()` it manually, or with ARC use a `__bridge_transfer` cast or a call to `CFBridgingRelease()` to tell ARC about it.)

cdavis5e (approved):

billhollings:
Sorry for the delay in addressing this PR. I wanted to wait until I had a some time to dig in and better understand how ARC is handling these objects under the covers. I didn't want to replace the ARC break with a non-ARC break.

Rather than add the additional retains, if we add `__unsafe_unretained` to the declarations of the struct members, it works for both ARC and non-ARC callers. I've requested some follow-up on issue #2151, to see if this will fix the problem you were encountering.

billhollings (commented):
Sorry for the delay in addressing this PR. I wanted to wait until I had a some time to dig into, and better understand, how ARC is handling these objects under the covers. I didn't want to replace the ARC break with the non-ARC break, without a better understanding of why this is happening in the first place.

Rather than add the additional `retains`, if we add `__unsafe_unretained` to the declarations of the struct members, it works for both ARC and non-ARC callers. I've requested some follow-up on issue #2151, to see if this will fix the problem you were encountering.

billhollings:
Replaced by #2200.

PR #2185, opened 2024-03-14
Fix compilation issue due to new project version

jeroenbakker-atmind:
MoltenVK has updated to a new version, but the build scripts still
mentioned the previous version, which failed during build.

```
MVKInstance.mm:774:2: error: 
      static assertion failed due to requirement 'std::string_view("1.2.8") ==
      ("1.2.9")': Xcode build setting CURRENT_PROJECT_VERSION must be identical
      to the MoltenVK version (MVK_VERSION_STRING).
  ...static_assert(string_view(MVK_STRINGIFY(MVK_FRAMEWORK_VERSION)) == MVK_VERSION_STRING...
     ^             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
1 error generated.
```

This is fixed by updating the `CURRENT_PROJECT_VERSION`. Developers updating
should first clear their build folder (`make clean`) before the new changes become
into effect.


billhollings:
Ah. Thanks!

I hadn't compiled when I published the 1.2.9 update. 🤦🏻‍♂️

At least this proves the value of that static assertion! 👍🏻

PR #2190, opened 2024-03-19
Development support for future VK_KHR_portability_subset_metal extension.

billhollings:
- `fetchDependencies` script pulls `VK_KHR_portability_subset_metal` content from private Khronos repo.
- Add `VK_KHR_portability_subset_metal` extension.
- Include `VkPhysicalDevicePortabilitySubsetMetalFeaturesKHR` in `pNext` of `VkPhysicalDeviceFeatures2`.

### DIFF
--- a/MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
@@ -375,6 +375,19 @@ void MVKPhysicalDevice::getFeatures(VkPhysicalDeviceFeatures2* features) {
 				portabilityFeatures->vertexAttributeAccessBeyondStride = true;	// Costs additional buffers. Should make configuration switch.
 				break;
 			}
+			case VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_PORTABILITY_SUBSET_METAL_FEATURES_KHR: {
+				auto* portabilityFeaturesMetal = (VkPhysicalDevicePortabilitySubsetMetalFeaturesKHR*)next;
+				portabilityFeaturesMetal->imageView2DOn3DImage = false;
+				portabilityFeaturesMetal->rasterizationDynamicPolygonModePoints = false;
+				portabilityFeaturesMetal->samplerMipLodBias = getMVKConfig().useMetalPrivateAPI;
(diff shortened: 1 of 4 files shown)


billhollings:
GitHub CI is failing because it doesn't have access to the private Khronos repo. This is expected.

aitor-lunarg (approved):
LGTM! Thank you!

alyssarosenzweig:
> shaderImageGatherExtendedConstOffsets

This is straightforward to lower, as most native drivers do?

cdavis5e:
> > shaderImageGatherExtendedConstOffsets
> 
> This is straightforward to lower, as most native drivers do? https://gitlab.freedesktop.org/mesa/mesa/-/blob/main/src/compiler/nir/nir_lower_tex.c?ref_type=heads#L1253

Sadly, not as straightforward as one would like. It's simple enough to expand it to four `texture::read()` calls, but the problem is handling sampler coordinate wrapping. We can't introspect the `sampler` object for its wrap modes, so we'd have to pass this information alongside the sampler. This becomes particularly problematic when not using argument buffers, which we don't at all by default.

alyssarosenzweig:
> > > shaderImageGatherExtendedConstOffsets
> > 
> > 
> > This is straightforward to lower, as most native drivers do? https://gitlab.freedesktop.org/mesa/mesa/-/blob/main/src/compiler/nir/nir_lower_tex.c?ref_type=heads#L1253
> 
> Sadly, not as straightforward as one would like. It's simple enough to expand it to four `texture::read()` calls, but the problem is handling sampler coordinate wrapping. We can't introspect the `sampler` object for its wrap modes, so we'd have to pass this information alongside the sampler. This becomes particularly problematic when not using argument buffers, which we don't at all by default.

See the linked lowering. You turn it into 4 gathers

billhollings:
> > See the linked lowering. You turn it into 4 gathers
> 
> Thanks for the suggestion, and example. I've added it to the full [issue discussion](https://github.com/KhronosGroup/Vulkan-Portability/issues/31#issuecomment-2009450000).

SPIRV-Cross PR [#2325](https://github.com/KhronosGroup/SPIRV-Cross/pull/2325) adds support for image gather with ConstOffsets.

Issue #2186, opened 2024-03-14
Invalid driver dylib generated since #2170


spnda:
Checking out 9082ca839f4eb83d183837fe656059ac12e4ee03, which is the last commit before the merge of #2170, and building the dylib with `make macos` my application runs just as it used to.

Now if I checkout eff757144614a20d7b452c8388019783d9511a25 (from #2170), build the driver with `make macos` I get this output:
```
install_name_tool: warning: changes being made to the file will invalidate the code signature in: /Users/sean/Documents/git/spnda/MoltenVK/Package/Release/MoltenVK/dynamic/dylib/macOS/libMoltenVK.dylib (for architecture x86_64)
install_name_tool: warning: changes being made to the file will invalidate the code signature in: /Users/sean/Documents/git/spnda/MoltenVK/Package/Release/MoltenVK/dynamic/dylib/macOS/libMoltenVK.dylib (for architecture arm64)
...
note: Run script build phase 'Package MoltenVK' will be run during every build because the option to run the script phase "Based on dependency analysis" is unchecked. (in target 'MoltenVK-macOS' from project 'MoltenVKPackaging')
```
After changing the symlink to the new dylib location the application crashes with an EXC_BAD_ACCESS while loading the dylib file and this reason:
```
Termination Reason:    Namespace CODESIGNING, Code 2 Invalid Page
```

If I update past #2175, build it and run my application, instead of an application crash I now get this output (running with `VK_LOADER_DEBUG=error,warn,info`:
```
ERROR:             loader_scanned_icd_add: Attempt to retrieve either 'vkGetInstanceProcAddr' or 'vk_icdGetInstanceProcAddr' from ICD /usr/local/lib/libMoltenVK.dylib failed.
ERROR:             loader_scanned_icd_add: Attempt to retrieve either 'vkGetInstanceProcAddr' or 'vk_icdGetInstanceProcAddr' from ICD /usr/local/lib/libMoltenVK.dylib failed.
ERROR:             loader_scanned_icd_add: Attempt to retrieve either 'vkGetInstanceProcAddr' or 'vk_icdGetInstanceProcAddr' from ICD /usr/local/lib/libMoltenVK.dylib failed.
ERROR:             loader_scanned_icd_add: Attempt to retrieve either 'vkGetInstanceProcAddr' or 'vk_icdGetInstanceProcAddr' from ICD /usr/local/lib/libMoltenVK.dylib failed.
[...]

PR #2196, opened 2024-03-30
Fix issue where incremental dynamic builds do not link to static dependencies.

billhollings:
After the recent change to building dynamic frameworks and dylibs as targets built by linking to a static library dependency, incremental builds do not correctly link the dynamic frameworks and dylibs to the rebuilt static library.

- Delete dynamic target outputs during static target builds.
- Run platform schemes sequentially in manual order.
- Build the dynamic framework before the dylib.
- Clean up Xcode Scheme organization.

It is not entirely obvious what the dependency problem is, and why these particular changes work. A lot of trial-and-error was required. It's possible that Xcode dependency analysis does not correctly analyze the flow used in MoltenVK builds.

This should fix issue #2186.


spnda (approved):
This seems to fix the issue locally for me. Though I have no idea about the Xcode project changes themselves.

SRSaunders:
I wonder if the problem is not having any source files to compile for the dylib target.  As a quick and dirty test, I removed the libMoltenVK.a static library from the Link Binary with Libraries area under Build Phases, and added all the MoltenVK source and header files directly to the MoltenVK-macOS-dylib target.  This may be duplicating things (I'm not sure actually), but incremental builds do begin working properly after that.

billhollings:
@SRSaunders  Thanks for looking into options and testing them.

> UPDATE - Unfortunately it does look like XCode is duplicating the work. I was hoping that it would be a bit smarter about compilation of like units for different targets. Darn.

Yeah. It would be nice, especially for the individual object-file compiles. But it appears that in...

`~/Library/Developer/Xcode/DerivedData/MoltenVKPackaging-xxxx/Build/Intermediates.noindex/MoltenVK.build/Release`

..._Xcode_ tracks each target build separately.

PR #2199, opened 2024-04-03
WIP: Calculate argument buffer allocation size taking variable counts into account.

js6i:
When allocating descriptors from the pool, we should only request as much as needed for the variable number of descriptors, if applicable.

It looks slightly dicey to use the fixed size argument encoder when encoding variable sized bindings. Since here we're creating the encoder to calculate the size, maybe we could use it for encoding too..

### DIFF
--- a/MoltenVK/MoltenVK/GPUObjects/MVKDescriptor.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKDescriptor.mm
@@ -190,14 +190,17 @@ void mvkPopulateShaderConversionConfig(mvk::SPIRVToMSLConversionConfiguration& s
 
 MVKVulkanAPIObject* MVKDescriptorSetLayoutBinding::getVulkanAPIObject() { return _layout; };
 
-uint32_t MVKDescriptorSetLayoutBinding::getDescriptorCount(MVKDescriptorSet* descSet) const {
+uint32_t MVKDescriptorSetLayoutBinding::getDescriptorCount(MVKDescriptorSet* descSet, uint32_t variableDescriptorCount) const {
 
 	if (_info.descriptorType == VK_DESCRIPTOR_TYPE_INLINE_UNIFORM_BLOCK_EXT) {
 		return 1;
 	}
 
-	if (descSet && hasVariableDescriptorCount()) {
-		return descSet->_variableDescriptorCount;
+	if (hasVariableDescriptorCount()) {
+		if (descSet)
+			return descSet->_variableDescriptorCount;
+		if (variableDescriptorCount != std::numeric_limits<uint32_t>::max())
+			return variableDescriptorCount;
 	}
 
 	return _info.descriptorCount;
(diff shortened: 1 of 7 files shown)


MoltenVK/MoltenVK/GPUObjects/MVKDescriptorSet.mm
```diff
@@ -266,12 +269,26 @@
...
 			NSMutableArray<MTLArgumentDescriptor*>* args = [NSMutableArray arrayWithCapacity: _bindings.size()];
-			for (auto& dslBind : _bindings) { dslBind.addMTLArgumentDescriptors(args); }
+			for (auto& dslBind : _bindings) { dslBind.addMTLArgumentDescriptors(args, dslBind.getDescriptorCount()); }
 			_mtlArgumentEncoder.init(args.count ? [getMTLDevice() newArgumentEncoderWithArguments: args] : nil);
 		}
 	}
 }
 
+size_t MVKDescriptorSetLayout::getEncodedArgumentBufferLength(uint32_t variableDescriptorCount) {
+	if (_hasVariableDescriptorCount) {
```
billhollings, line 279: Since only the last descriptor binding in a descriptor set layout can be variable, instead of tracking `_hasVariableDescriptorCount`, querying the last binding (if at least one exists) by calling `hasVariableDescriptorCount()` is simple. 

With that idea, this could also test that `variableDescriptorCount` is less than `getDescriptorCount()` of that same binding, to determine if calculating with a temporary argument encoder is necessary.

billhollings (changes requested):
Thanks for submitting this.

I've requested a couple of changes.

> It looks slightly dicey to use the fixed size argument encoder when encoding variable sized bindings. Since here we're creating the encoder to calculate the size, maybe we could use it for encoding too..

Good point. And good catch.

Having each descriptor set carry its own encoder sounds like a lot. Perhaps we could have the descriptor set defer to the layout's encoder, unless the variable descriptor count is less than the descriptor count of the variable-length binding.

However, we are going to move away from argument encoders very soon. So, I don't think we should bother making things that complex unless we actually determine it is causing a problem.

MoltenVK/MoltenVK/GPUObjects/MVKDescriptor.mm
```diff
@@ -203,6 +203,19 @@ void mvkPopulateShaderConversionConfig(mvk::SPIRVToMSLConversionConfiguration& s
...
+	if (_info.descriptorType == VK_DESCRIPTOR_TYPE_INLINE_UNIFORM_BLOCK_EXT) {
+		return 1;
+	}
+
+	if (hasVariableDescriptorCount()) {
+		return variableDescriptorCount;
+	}
+
+	return _info.descriptorCount;
+}
```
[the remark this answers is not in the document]
js6i: Yes, I think that works.

MoltenVK/MoltenVK/GPUObjects/MVKDescriptorSet.mm
js6i: Good points, done.

js6i:
> Having each descriptor set carry its own encoder sounds like a lot. Perhaps we could have the descriptor set defer to the layout's encoder, unless the variable descriptor count is less than the descriptor count of the variable-length binding.

I think we could do that. Are they expensive though? I would have thought it's fine to create them on demand, for single use.

> However, we are going to move away from argument encoders very soon. So, I don't think we should bother making things that complex unless we actually determine it is causing a problem.

It's causing a validation error. That said, there's also another one, when the shader has the maximum size array declared, but the buffer that gets bound is smaller. Let's hold off merging for now, I'll see what can be done about the other one.

billhollings:
This PR is almost certainly subsumed or made obsolete by the recent PRs #2260 & #2268.

Closing this as a result. Please reopen if needed.

billhollings:
This PR is almost certainly subsumed or made obsolete by the recent PRs #2260 & #2268.

Closing this as a result. Please reopen if needed.

billhollings:
Replaced by PR #2273.

PR #2208, opened 2024-04-15
Add support for VK_EXT_host_image_copy extension.

### LINKED ISSUE #2197: Add support for VK_EXT_host_image_copy extension
This appears to be a fairly straightforward mapping from Vulkan to Metal functionality.

billhollings:
- Add support for `VK_EXT_host_image_copy` extension.
- `MVKResource::getHostMemoryAddress()` return `nullptr` if `MVKDeviceMemory::getHostMemoryAddress()` returns `nullptr`, regardless of local offset.
- Remove unnecessary  enum value `kMVKVkFormatFeatureFlagsTexTransfer` to reduce redundancy between read and transfer feature flag options
- `MVKResource` remove unnecessary inline qualifiers (unrelated).
- `MVKDevice` remove some obsolete commentary (unrelated).

Fixes #2197.


ovvldc:
Why include unrelated changes in a PR, why not separate them in a housekeeping PR or such? Doesn't that make the commit history difficult to trace, or I am missing something?

MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
```diff
@@ -561,6 +561,142 @@
...
+			}
+		}
+	}
+	return VK_SUCCESS;
+}
+
+// Create concrete implementations of the variations of the copyContent() template function.
+// This is required since the template is called from outside this file (compilation unit).
+template VkResult MVKImage::copyContent(const VkCopyMemoryToImageInfoEXT* pCopyInfo, bool toImage);
+template VkResult MVKImage::copyContent(const VkCopyImageToMemoryInfoEXT* pCopyInfo, bool toImage);
```
cdavis5e, line 633: The design purist in me doesn't like that the `toImage` parameter is necessary here  to avoid duplicating all the setup code, since it will always be `true` in the first case and `false` in the second. Maybe if the parts that performed the actual copy (inside the `if (toImage)` blocks above) were factored out into separate methods, and then the right method would get called in `copyContent()` based on the type of the copy info...

cdavis5e (changes requested):
You forgot to add this to the list of all supported extensions in the User's Guide.

Also, FYI, it looks like GitHub doesn't recognize the keyword "completes" as indicating that an issue should be closed after a PR is merged. You may have to close the issue manually.

I'm also worried about the implications of the `VK_HOST_IMAGE_COPY_MEMCPY_EXT` flag. I don't think we can support that for nonlinear images. `Managed` textures on Intel are where this would crop up.

cdavis5e:
> To help keep the code tidy over the long run, we try to have a "clean as you go" approach, when it comes to non-functional maintenance, since those kind of maintenance projects never seem to get prioritized.

Fair point. Coming from Wine, however, which is extremely strict about this, I actually prefer to at least split the housekeeping changes into their own commits. That way, they show up separately in the history, which makes bisection easier. The PR could still include the main change and the unrelated cleanups together.

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
```diff
@@ -845,6 +854,87 @@
...
+		VK_IMAGE_LAYOUT_VIDEO_DECODE_SRC_KHR,
+		VK_IMAGE_LAYOUT_VIDEO_DECODE_DPB_KHR,
+		VK_IMAGE_LAYOUT_SHARED_PRESENT_KHR,
+		VK_IMAGE_LAYOUT_FRAGMENT_DENSITY_MAP_OPTIMAL_EXT,
+		VK_IMAGE_LAYOUT_FRAGMENT_SHADING_RATE_ATTACHMENT_OPTIMAL_KHR,
+		VK_IMAGE_LAYOUT_RENDERING_LOCAL_READ_KHR,
+		VK_IMAGE_LAYOUT_VIDEO_ENCODE_DST_KHR,
+		VK_IMAGE_LAYOUT_VIDEO_ENCODE_SRC_KHR,
+		VK_IMAGE_LAYOUT_VIDEO_ENCODE_DPB_KHR,
+		VK_IMAGE_LAYOUT_ATTACHMENT_FEEDBACK_LOOP_OPTIMAL_EXT,
```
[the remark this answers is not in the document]
billhollings: A fair point. It could be confusing to users. I've removed them.

I've added a TODO comment to try to help encourage maintenance of this list, but, I do remain concerned that this will not be maintained as extensions with new layouts are added.

MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
```diff
@@ -561,6 +561,142 @@
...
+					codec->decompress(decompBuffer.get(), pImgBytes, linearLayout, imgLayout, imgRgn.imageExtent);
+					pImgBytes = decompBuffer.get();
+					imgLayout = linearLayout;
+				}
+#endif
+				[mtlTex replaceRegion: mtlRgn
+						  mipmapLevel: imgSubRez.mipLevel
+								slice: imgSubRez.baseArrayLayer + imgLyrIdx
+							withBytes: pImgBytes
+						  bytesPerRow: (isPVRTC || is1D) ? 0 : imgLayout.rowPitch
```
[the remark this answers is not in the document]
billhollings: I've tested it both ways, but unfortunately, CTS does not seem to bother testing 1D images directly for host-copy.

However, it should be fine. The `is1D` flag is back-calculated from `MVKImage::_mtlTextureType`, the value of which is set taking into consideration `MVKConfiguration::texture1DAs2D`.

MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
billhollings: Great idea. I'm also a fan of design elegance, and I've implemented what you suggested.

billhollings:
> That way, they show up separately in the history, which makes bisection easier. The PR could still include the main change and the unrelated cleanups together.

This makes sense. I'll try to remember that approach on future PR's.

billhollings:
> You forgot to add this to the list of all supported extensions in the User's Guide.

D'oh! Good catch. Fixed.

> Also, FYI, it looks like GitHub doesn't recognize the keyword "completes" as indicating that an issue should be closed after a PR is merged. You may have to close the issue manually.

Yeah. I noticed that after the fact, too. I've modified it above now.

> I'm also worried about the implications of the `VK_HOST_IMAGE_COPY_MEMCPY_EXT` flag. I don't think we can support that for nonlinear images. `Managed` textures on Intel are where this would crop up.

I don't know why it didn't make it into the spec proper, but the [extension proposal](https://github.com/KhronosGroup/Vulkan-Docs/blob/main/proposals/VK_EXT_host_image_copy.adoc#2-proposal) includes:

> ...`flags` may be `VK_HOST_IMAGE_COPY_MEMCPY_EXT`, in which case the data in host memory should have the same swizzling layout as the image. This is mainly useful for embedded systems where this swizzling is known and well defined outside of Vulkan.

This leads me to interpret `VK_HOST_IMAGE_COPY_MEMCPY_EXT` as a nice-to-have optimization, but that falling back to a linear extraction for both linear and optimal layouts (basically ignoring `VK_HOST_IMAGE_COPY_MEMCPY_EXT`) is acceptable.

The CTS tests all seem to test both with and without `VK_HOST_IMAGE_COPY_MEMCPY_EXT`, and so far so good. However, I have not tested on Intel GPU's. Why do you feel any issues would be limited to that platform?

cdavis5e:
> The CTS tests all seem to test both with and without `VK_HOST_IMAGE_COPY_MEMCPY_EXT`, and so far so good. However, I have not tested on Intel GPU's. Why do you feel any issues would be limited to that platform?

Because on Intel Macs, the GPUs support `Managed` textures which are host-accessible but do not have to be linear. I suppose you could get around that by putting the texture in a placement heap and aliasing it with a buffer, but I'm not sure Metal will let you create a `Managed` `MTLHeap`... I suppose we could also forbid setting `VK_IMAGE_USAGE_HOST_TRANSFER_BIT_EXT` with `Managed` textures, which we expose as Vulkan `HOST_CACHED` memory, but then we'd lose out on being able to use `VK_EXT_host_image_copy` with these images.

cdavis5e (changes requested):
The changes so far LGTM, but I won't approve this until you've answered my objections about `VK_HOST_IMAGE_COPY_MEMCPY_EXT` with `Managed` textures on Intel Macs.

Maybe I should run the CTS against this change. I have a couple of Intel Macs.

billhollings:
> Maybe I should run the CTS against this change. I have a couple of Intel Macs.

If you could, that would be helpful. Without an understanding of what may/will go wrong, it's hard to develop and test a solution for how to handle it.

I've been testing using...

```
cd <mvk-parent-dir>/MoltenVK/Scripts
./runcts <cts-parent-dir>/VK-GL-CTS/external/vulkancts/mustpass/main/vk-default/image/host-image-copy.txt
```

billhollings:
> > Maybe I should run the CTS against this change. I have a couple of Intel Macs.
> 
> If you could, that would be helpful. 

NM. I've dug out my old Intel/Radeon and have managed to get things running on it.

Integrated Intel GPU is fine, but it's definitely misbehaving on the discrete Radeon GPU.

Looking into it.

MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
```diff
@@ -561,6 +561,161 @@
...
+		 mipmapLevel: mipLevel
+			   slice: slice];
+	return VK_SUCCESS;
+}
+
+// Host-copy from a MTLTexture to memory.
+VkResult MVKImage::copyContent(id<MTLTexture> mtlTex,
+							   VkMemoryToImageCopyEXT imgRgn, uint32_t mipLevel, uint32_t slice,
+							   void* pImgBytes, size_t rowPitch, size_t depthPitch) {
+	VkSubresourceLayout imgLayout = { 0, 0, rowPitch, 0, depthPitch};
```
spnda, line 586: If I am not mistaken the two comments here are swapped.

billhollings: Thanks. Fixed now.

billhollings:
> > >The changes so far LGTM, but I won't approve this until you've answered my objections about VK_HOST_IMAGE_COPY_MEMCPY_EXT with Managed textures on Intel Macs.
> > > Maybe I should run the CTS against this change. I have a couple of Intel Macs.
> > 
> > 
> > If you could, that would be helpful.
> 
> NM. I've dug out my old Intel/Radeon and have managed to get things running on it.
> 
> The integrated Intel GPU is fine, but it's definitely misbehaving on the discrete Radeon GPU, as you predicted.
> 
> Looking into it.

This is now working properly on both Radeon & Intel.

It turns out the failures were not a linear/optimal issue, but a managed-memory texture sync issue. I've added a `synchronizeTexture:` where appropriate for managed-memory textures on discrete GPUs.

MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
```diff
@@ -716,6 +711,40 @@ static MTLRegion getMTLRegion(const ImgRgn& imgRgn) {
...
+											 level: imgSubRez.mipLevel];
+				}
+			}
+
+			[mtlBlitEnc endEncoding];
+			[mtlCmdBuff commit];
+			[mtlCmdBuff waitUntilCompleted];
+		}
+	}
+#endif
```
cdavis5e, line 739: I wonder if there's a point where we should be calling `-synchronizeTexture:slice:level:` where we aren't in response to a barrier command.

cdavis5e (approved):
Well, I suppose if it works without any special munging beyond the extra synchronization...

MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
billhollings: > I wonder if there's a point where we should be calling `-synchronizeTexture:slice:level:` where we aren't in response to a barrier command.

I'm not sure what you are wondering here?

cdavis5e: I mean, the host has to synchronize its use of the image with the device before calling this function. According to [the Vulkan spec][1]:

> `vkCopyImageToMemoryEXT` does not check whether the device memory associated with `srcImage` is currently in use before performing the copy. The application **must** guarantee that any previously submitted command that writes to the copy regions has completed before the host performs the copy.

It's unclear, though, if that means that the host is responsible for emitting a barrier, or if this function includes an implicit barrier. I wonder, in particular, if the `vkTransitionImageLayoutEXT()` function has something to do with this.

[1]: https://registry.khronos.org/vulkan/specs/1.3-extensions/html/chap20.html#vkCopyImageToMemoryEXT

billhollings: > It's unclear, though, if that means that the host is responsible for emitting a barrier

Yes. I was unsure about that spec wording too. 

The spec wording for `vkCopyImageToMemoryEXT()` also says

> This command (] is functionally similar to `vkCmdCopyImageToBuffer2, except it is executed on the host and writes to host memory instead of a buffer.

which, if you squint hard enough, might imply the same barrier operations apply. But the spec is silent on which barriers to use.

Some of the CTS tests failing on Radeon did indeed emit a pipeline image barrier after rendering, but they only included `VK_ACCESS_TRANSFER_READ_BIT`, and not the `VK_ACCESS_HOST_READ_BIT` that I would have expected when preparing for a host-copy.

I actually did initially experiment with modifying `MVKImageMemoryBinding::needsHostReadSync()`, to force a sync on `VK_ACCESS_TRANSFER_READ_BIT` if the image included `VK_IMAGE_USAGE_HOST_TRANSFER_BIT_EXT`. It did solve the issue, but I was concerned that it would unduly penalize any non-host transfers that the image might be involved with. I wasn't convinced that setting `VK_IMAGE_USAGE_HOST_TRANSFER_BIT_EXT` implies that the app will focus on host-copy transfers only.

My other consideration was that since `vkCopyImageToMemoryEXT()` is not a GPU command, there's an implicit understanding that the CPU is expecting to wait for the GPU to be finished anyway, and so taking the hit when we know it's expected, was more appropriate.

`vkTransitionImageLayoutEXT()` is about layouts, and not memory access, and I couldn't find anything equivalent in Metal. Plus the problematic CTS tests were not issuing a call to that function anyway.

So, we could follow the path of forcing a sync on every `VK_ACCESS_TRANSFER_READ_BIT` barrier, if that is more appropriate, but the spec is definitely not clear on what's expected.

PR #2217, opened 2024-04-25
On macOS Apple Silicon, avoid managed-memory textures, and resource syncs.

billhollings:
Like their iOS/tvOS counterparts, macOS Apple Silicon GPUs support using Shared memory for textures, and do not require resource synchronization, even with Managed memory. This change treats macOS Apple Silicon the same as iOS & tvOS.

- `MVKPhysicalDevice` add `_hasUnifiedMemory` & `_isAppleGPU` flags.
- `MVKDeviceTrackingMixin` add `isUnifiedMemoryGPU()` & `isAppleGPU()`.
- Do not advertise host-visible-but-not-host-coherent Vulkan memory type on macOS Apple Silicon.
- Replace `mvkMTLStorageModeFromVkMemoryPropertyFlags()` with `MVKPhysicalDevice::getMTLStorageModeFromVkMemoryPropertyFlags()`, and return Shared instead of Managed for Apple Silicon, even if coherency is not requested.
- On unified memory devices, avoid needless calls to `didModifyRange:`, `synchronizeResource:`, and `synchronizeTexture:slice:level:`.

### DIFF
--- a/MoltenVK/MoltenVK/GPUObjects/MVKBuffer.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKBuffer.mm
@@ -81,7 +81,11 @@ VkResult MVKBuffer::bindDeviceMemory(MVKDeviceMemory* mvkMem, VkDeviceSize memOf
 
 #if MVK_MACOS
 	if (_deviceMemory) {
-		_isHostCoherentTexelBuffer = !_device->_pMetalFeatures->sharedLinearTextures && _deviceMemory->isMemoryHostCoherent() && mvkIsAnyFlagEnabled(_usage, VK_BUFFER_USAGE_UNIFORM_TEXEL_BUFFER_BIT | VK_BUFFER_USAGE_STORAGE_TEXEL_BUFFER_BIT);
+		_isHostCoherentTexelBuffer = (!isUnifiedMemoryGPU() &&
+									  !_device->_pMetalFeatures->sharedLinearTextures &&
+									  _deviceMemory->isMemoryHostCoherent() &&
+									  mvkIsAnyFlagEnabled(_usage, (VK_BUFFER_USAGE_UNIFORM_TEXEL_BUFFER_BIT |
+																   VK_BUFFER_USAGE_STORAGE_TEXEL_BUFFER_BIT)));
 	}
 #endif
 
@@ -118,7 +122,8 @@ void MVKBuffer::applyBufferMemoryBarrier(MVKPipelineBarrier& barrier,
 // buffer and host memory for the purpose of the host reading texture memory.
 bool MVKBuffer::needsHostReadSync(MVKPipelineBarrier& barrier) {
 #if MVK_MACOS
-	return (mvkIsAnyFlagEnabled(barrier.dstStageMask, (VK_PIPELINE_STAGE_HOST_BIT)) &&
+	return (!isUnifiedMemoryGPU() &&
+			mvkIsAnyFlagEnabled(barrier.dstStageMask, (VK_PIPELINE_STAGE_HOST_BIT)) &&
 			mvkIsAnyFlagEnabled(barrier.dstAccessMask, (VK_ACCESS_HOST_READ_BIT)) &&
 			isMemoryHostAccessible() && (!isMemoryHostCoherent() || _isHostCoherentTexelBuffer));
 #else
@@ -138,9 +143,7 @@ bool MVKBuffer::overlaps(VkDeviceSize offset, VkDeviceSize size, VkDeviceSize &o
     return false;
 }
 
-#if MVK_MACOS
-bool MVKBuffer::shouldFlushHostMemory() { return _isHostCoherentTexelBuffer; }
-#endif
+bool MVKBuffer::shouldFlushHostMemory() { return !isUnifiedMemoryGPU() && _isHostCoherentTexelBuffer; }
 
 // Flushes the device memory at the specified memory range into the MTLBuffer.
 VkResult MVKBuffer::flushToDevice(VkDeviceSize offset, VkDeviceSize size) {
--- a/MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
@@ -1765,9 +1765,7 @@ VkResult MVKPhysicalDevice::getQueueFamilyProperties(uint32_t* pCount,
 // wild temporary changes, particularly during initial queries before much GPU activity has occurred.
 // On Apple GPUs, CPU & GPU timestamps are the same, and timestamp period never changes.
 void MVKPhysicalDevice::updateTimestampPeriod() {
-	if (_properties.vendorID != kAppleVendorId &&
-		[_mtlDevice respondsToSelector: @selector(sampleTimestamps:gpuTimestamp:)]) {
-
+	if ( !_isAppleGPU && [_mtlDevice respondsToSelector: @selector(sampleTimestamps:gpuTimestamp:)]) {
 		MTLTimestamp earlierCPUTs = _prevCPUTimestamp;
 		MTLTimestamp earlierGPUTs = _prevGPUTimestamp;
 		[_mtlDevice sampleTimestamps: &_prevCPUTimestamp gpuTimestamp: &_prevGPUTimestamp];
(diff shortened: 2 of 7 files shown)


spnda (commented):
The reason that we can avoid Managed textures is that the Apple Silicon Mac GPUs use a shared memory architecture. However, I believe that some MacBooks with Intel CPUs also use this architecture and have no dedicated GPU. Therefore, I wonder if this change should not only be for Apple Silicon, but for any GPU with a SMA, at least for buffers?
> MTLStorageModeShared
> This is the default storage mode for [MTLBuffer](https://developer.apple.com/documentation/metal/mtlbuffer) instances on integrated GPUs and both MTLBuffer and [MTLTexture](https://developer.apple.com/documentation/metal/mtltexture) instances on Apple silicon GPUs. On non-Apple family GPUs, the shared storage mode isn’t available for [MTLTexture](https://developer.apple.com/documentation/metal/mtltexture) instances.

Also, does this bring any noticeable performance benefits?

billhollings:
> The reason that we can avoid Managed memory is that the Apple Silicon Mac GPUs use a shared memory architecture. However, I believe that some MacBooks with Intel CPUs also use this architecture and have no dedicated GPU. Therefore, I wonder if this change should not only be for Apple Silicon, but for any GPU with a SMA, at least for buffers?
> 
> > MTLStorageModeShared
> > This is the default storage mode for [MTLBuffer](https://developer.apple.com/documentation/metal/mtlbuffer) instances on integrated GPUs and both MTLBuffer and [MTLTexture](https://developer.apple.com/documentation/metal/mtltexture) instances on Apple silicon GPUs. On non-Apple family GPUs, the shared storage mode isn’t available for [MTLTexture](https://developer.apple.com/documentation/metal/mtltexture) instances.
> 
> Also, does this bring any noticeable performance benefits?

I did investigate this a bit. As mentioned in the `MTLStorageModeShared` spec quote, Intel GPUs do indeed reject attempts to create a `MTLTexture` from `Shared` memory, but also behave as if the sync functions are not needed. That's why I used `getHasUnifiedMemory()` in `MVKImage::copyImageToMemory()` as part of `VK_EXT_host_image_copy`.

So, to cover both Apple & older Intel GPUs, we'd have to have a mix of `isAppleGPU()` and `getHasUnifiedMemory()` queries, and I didn't want to go to that much back and forth effort testing on older integrated GPUs, particularly since a discrete GPU would be the preference for gaming. It felt like a small use-case. I also only have one Intel machine, and I wasn't confident it would apply to all Intel GPUs. 

I've got nothing against eventually getting there, but my goal here is primarily to clean things up for Apple Silicon behaviour, since that's the path of the future.

MoltenVK/MoltenVK/GPUObjects/MVKBuffer.mm
```diff
@@ -79,11 +79,13 @@
 
 	MVKResource::bindDeviceMemory(mvkMem, memOffset);
 
-#if MVK_MACOS
```
cdavis5e, line 82: Should we leave the preprocessor platform checks in? This is all dead code on iOS and tvOS. Then again, I suppose running the simulator on an Intel Mac could result in needing this special handling...

MoltenVK/MoltenVK/GPUObjects/MVKBuffer.mm
```diff
@@ -117,13 +119,10 @@
 // Returns whether the specified buffer memory barrier requires a sync between this
 // buffer and host memory for the purpose of the host reading texture memory.
 bool MVKBuffer::needsHostReadSync(MVKPipelineBarrier& barrier) {
-#if MVK_MACOS
```
cdavis5e, line 120: Ditto.

MoltenVK/MoltenVK/GPUObjects/MVKBuffer.mm
```diff
@@ -138,9 +137,7 @@
     return false;
 }
 
-#if MVK_MACOS
```
cdavis5e, line 141: Ditto.

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
```diff
@@ -3194,13 +3194,13 @@ static uint32_t mvkGetEntryProperty(io_registry_entry_t entry, CFStringRef prope
 	setMemoryType(typeIdx, sharedHeapIdx, sharedTypePropFlags);
 	typeIdx++;
 
-	// Managed storage
+	// Managed storage. On all Apple Silicon, use Shared instead.
 	uint32_t managedBit = 0;
-#if MVK_MACOS
```
cdavis5e, line 3199: Another potential hunk of dead code on iOS/tvOS...

MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
```diff
@@ -468,19 +468,15 @@
 // Returns whether the specified image memory barrier requires a sync between this
 // texture and host memory for the purpose of the host reading texture memory.
 bool MVKImageMemoryBinding::needsHostReadSync(MVKPipelineBarrier& barrier) {
-#if MVK_MACOS
```
cdavis5e, line 471: Another one...

MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
```diff
@@ -858,12 +854,10 @@ static MTLRegion getMTLRegion(const ImgRgn& imgRgn) {
     pMemoryRequirements->memoryTypeBits = (_isDepthStencilAttachment)
                                           ? mvkPD->getPrivateMemoryTypes()
                                           : mvkPD->getAllMemoryTypes();
-#if MVK_MACOS
```
cdavis5e, line 861: Another one...

cdavis5e (changes requested):
Just some concerns about dead code on iOS and tvOS.

MoltenVK/MoltenVK/GPUObjects/MVKBuffer.mm
billhollings: Good point. Fixed.

MoltenVK/MoltenVK/GPUObjects/MVKBuffer.mm
billhollings: I chose to leave this one to avoid an unimplemented function declaration in `.h`. Seemed better than also having to add guard code there.

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
billhollings: Fixed.

MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
billhollings: Fixed.

MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
billhollings: Fixed.

cdavis5e (approved):

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
```diff
@@ -3235,17 +3242,33 @@ static uint32_t mvkGetEntryProperty(io_registry_entry_t entry, CFStringRef prope
...
+	// If visible to the host and coherent: Shared
+	if (mvkAreAllFlagsEnabled(vkFlags, VK_MEMORY_PROPERTY_HOST_COHERENT_BIT)) {
+		return MTLStorageModeShared;
+	}
+
+	// If visible to the host, but not coherent: Shared on Apple Silicon, Managed on other GPUs.
 #if MVK_MACOS
-	return ([_mtlDevice respondsToSelector: @selector(hasUnifiedMemory)]
-			? _mtlDevice.hasUnifiedMemory : _mtlDevice.isLowPower);
+	return _isAppleGPU ? MTLStorageModeShared : MTLStorageModeManaged;
```
spnda, line 3264: Since we now check `isUnifiedMemoryGPU` instead of `isAppleGPU` for the sync stuff, shouldn't this check also use the former here? Otherwise we'd still be using MTLStorageModeManaged, where those sync operations still need to happen iirc.

spnda (changes requested):

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
spnda: Oh, no, never mind. I forgot that the Intel GPUs don't like creating textures from Shared memory. So I guess we have to still use Managed memory on those, but can still avoid the syncs. All good.

spnda (approved):

PR #2221, opened 2024-05-01
Refactor public MVKDevice content into MVKDeviceTrackingMixin functions.

billhollings:
This is a non-functional code-maintenance change.

Previously, `MVKDevice` contained a significant amount of publicly exposed internal content. This patch adds functions to `MVKDeviceTrackingMixin` to better encapsulate, consolidate & streamline access to this content.

- Make `MVKDeviceTrackingMixin` a friend of `MVKDevice` & `MVKPhysicalDevice`.
- Hide public `MVKDevice` content behind `MVKDeviceTrackingMixin` functions.
- Remove similar `MVKDevice` content pointers from `MVKCommandEncoder`.
- `MVKDevice` remove `getPhysicalDevice()`, `getPixelFormats()` & `getMTLDevice()`, to focus access through `MVKDeviceTrackingMixin`.
- Move performance tracking functions to `MVKDeviceTrackingMixin` to remove need to reference `MVKDevice` multiple times when marking performance values.
- Subclass `MVKQueueSubmission`, `MVKMetalCompiler`, `MVKShaderLibrary`, and `MVKShaderLibraryCache` from `MVKBaseDeviceObject` to make use of these changes.

Note to reviewers:
- This is a non-functional change and passes the same CTS tests as previously.
- The primary design changes of interest are in the declarations of `MVKDeviceTrackingMixin` and `MVKDevice` in `MVKDevice.h`, and to `MVKCommandEncoder` in `MVKCommandBuffer.h`.
- The remaining changes are simple find-and-replace syntax changes to delegate to `MVKDeviceTrackingMixin` instead of reaching into `MVKDevice`.

### DIFF
--- a/MoltenVK/MoltenVK/Commands/MVKCmdTransfer.mm
+++ b/MoltenVK/MoltenVK/Commands/MVKCmdTransfer.mm
@@ -319,7 +319,7 @@ VkResult MVKCmdBlitImage<N>::setContent(MVKCommandBuffer* cmdBuff,
 
 	_filter = filter;
 
-	bool isDestUnwritableLinear = MVK_MACOS && !cmdBuff->getDevice()->_pMetalFeatures->renderLinearTextures && _dstImage->getIsLinear();
+	bool isDestUnwritableLinear = MVK_MACOS && !cmdBuff->getMetalFeatures().renderLinearTextures && _dstImage->getIsLinear();
 
 	_vkImageBlits.clear();		// Clear for reuse
 	for (uint32_t rIdx = 0; rIdx < regionCount; rIdx++) {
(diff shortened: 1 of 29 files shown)


MoltenVK/MoltenVK/Commands/MVKCmdTransfer.mm
```diff
@@ -507,7 +507,7 @@ static inline MTLSize mvkClampMTLSize(MTLSize size, MTLOrigin origin, MTLSize ma
         id<MTLTexture> srcMTLTex = _srcImage->getMTLTexture(srcPlaneIndex);
         id<MTLTexture> dstMTLTex = _dstImage->getMTLTexture(dstPlaneIndex);
         if (blitCnt && srcMTLTex && dstMTLTex) {
-			if (cmdEncoder->getDevice()->_pMetalFeatures->nativeTextureSwizzle &&
```
cdavis5e, line 510: Use of `cmdEncoder->getMetalFeatures()` is frequent enough that I wonder if it wouldn't be worth pulling it into a local variable.

cdavis5e (approved):
Sorry about that. Took me a while to get through this massive refactoring.

MoltenVK/MoltenVK/Commands/MVKCmdTransfer.mm
billhollings: Thanks. Good catch. I thought I had found all the multiple-uses already. Fixed now.

billhollings:
> Sorry about that. Took me a while to get through this massive refactoring.

No apology needed. And thanks. 

I wrote the _Note to reviewers_ above so you could skip all the search & replace noise. But I appreciate your epic diligence and accuracy once again! 👍🏻

PR #2222, opened 2024-05-01
Allow vertex buffer binding when they are used disregarding implicit buffer index

### LINKED ISSUE #1609: Vertex attributes missing from attribute array (regression)
PR #1595 (which fixed issue #1592) introduced a smaller secondary regression that causes the following CTS tests to _Fail_ instead of _Pass_:

```
dEQP-VK.pipeline.vertex_input.max_attributes.query_max_attributes.binding_one_to_one.interleaved
dEQP-VK.pipeline.vertex_input.max_attributes.query_max_attributes.binding_one_to_many.sequential
```

Any fix to this should be validated by the posters of the original issue #1592.

aitor-lunarg:
Fixes #1609 

Due to how MoltenVK decides when a buffer can be bound based on its requirements for the implicit buffer, if the application uses all bindings, implicit buffer index will be uint max. This lead to used buffers not being bound.

I did not test with the application mentioned in #1592 so if someone with the set up ready to test could, it would be great.

Although this change fixes the issue, it's a partial fix. I will create a new issue to discuss some concerns about the whole vertex input attachment architecture, since I do believe we would benefit from a rework of the system (I may be wrong tho). Haven't created a PR with a proof of concept since it'll require a considerable time investment I cannot afford currently.

### DIFF
--- a/MoltenVK/MoltenVK/GPUObjects/MVKPipeline.h
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKPipeline.h
@@ -431,7 +431,7 @@ protected:
 	uint32_t _tessCtlLevelBufferIndex = 0;
 
 	static constexpr uint32_t kMVKMaxVertexInputBindingBufferCount = 31u; // Taken from Metal Feature Set Table. Highest value out of all present GPUs
-	bool _isVertexInputBindingUsed[_maxVertexInputBindingBufferCount] = { false };
+	bool _isVertexInputBindingUsed[kMVKMaxVertexInputBindingBufferCount] = { false };
 	bool _primitiveRestartEnable = true;
 	bool _hasRasterInfo = false;
 	bool _needsVertexSwizzleBuffer = false;
--- a/MoltenVK/MoltenVK/GPUObjects/MVKPipeline.h
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKPipeline.h
@@ -430,6 +430,8 @@ protected:
 	uint32_t _tessCtlPatchOutputBufferIndex = 0;
 	uint32_t _tessCtlLevelBufferIndex = 0;
 
+	static constexpr uint32_t _maxVertexInputBindingBufferCount = 31u; // Taken from Metal Feature Set Table. Highest value out of all present GPUs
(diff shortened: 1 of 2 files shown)


MoltenVK/MoltenVK/GPUObjects/MVKPipeline.h
```diff
@@ -430,6 +430,8 @@ class MVKGraphicsPipeline : public MVKPipeline {
 	uint32_t _tessCtlPatchOutputBufferIndex = 0;
 	uint32_t _tessCtlLevelBufferIndex = 0;
 
+	static constexpr uint32_t _maxVertexInputBindingBufferCount = 31u; // Taken from Metal Feature Set Table. Highest value out of all present GPUs
```
billhollings, line 433: The naming pattern for all `static constexpr` is (or should be) `kMVK...`. See top of `MVKDevice.h` for other examples. Suggest making this `kMVKMaxVertexInputBindingBufferCount`.

billhollings (changes requested):
Thanks for taking care of this.

LGTM. Just one minor naming change request.

MoltenVK/MoltenVK/GPUObjects/MVKPipeline.h
billhollings:
```suggestion
	static constexpr uint32_t kMVKMaxVertexInputBindingBufferCount = 31u; // Taken from Metal Feature Set Table. Highest value out of all present GPUs
```

MoltenVK/MoltenVK/GPUObjects/MVKPipeline.h
```diff
@@ -430,6 +430,8 @@ class MVKGraphicsPipeline : public MVKPipeline {
...
+	bool _isVertexInputBindingUsed[_maxVertexInputBindingBufferCount] = { false };
```
billhollings, line 434:
```suggestion
	bool _isVertexInputBindingUsed[kMVKMaxVertexInputBindingBufferCount] = { false };
```

billhollings (approved):
Thanks for submitting this.

I've gone ahead and made the naming change I suggested, so we can get this into the upcoming SDK.

PR #2224, opened 2024-05-03
Return MoltenVK log level string in pMessageIdName field of debug utils callback data

### LINKED ISSUE #2219: pMessageIdName is always set to nullptr in MVKInstance::debugReportMessage()
When working on macOS/iOS fixes and updates in Sascha Willems' Vulkan examples project, I noticed that MoltenVK's `VkDebugUtilsMessengerCallbackDataEXT::pMessageIdName` is always set to **nullptr** in `MVKInstance::debugReportMessage(MVKVulkanAPIObject* mvkAPIObj, MVKConfigLogLevel logLevel, const char* pMessage)`.

Instead it should probably be set to `_debugReportCallbackLayerPrefix` or  `getReportingLevelString(logLevel)`.  This would allow the debug utils callback message to indicate the message is coming from MoltenVK via labels like: `MoltenVK` (the simplest option), or even `mvk-error`, `mvk-warn`, etc.

Here is a before and after example...
Before:
`WARNING: [0] : vkCreateMacOSSurfaceMVK() is deprecated. Use vkCreateMetalSurfaceEXT() from the VK_EXT_metal_surface extension.
`
After:
`WARNING: [0][MoltenVK] : vkCreateMacOSSurfaceMVK() is deprecated. Use vkCreateMetalSurfaceEXT() from the VK_EXT_metal_surface extension.
`
or
`WARNING: [0][mvk-warn] : vkCreateMacOSSurfaceMVK() is deprecated. Use vkCreateMetalSurfaceEXT() from the VK_EXT_metal_surface extension.
`

This analagous functionality to the Vulkan Loader, which populates the `pMessageIdName` field with things like: `Loader Message`.  For example:

`ERROR: [0][Loader Message] : loader_validate_device_extensions: Device extension VK_EXT_descriptor_buffer not supported by selected physical device or enabled layers.
`

I could provide a small PR to add this if you are interested.  Please advise.

SRSaunders:
Fixes #2219.

Previously the `pMessageIdName` field of debug utils callback data was set to **nullptr**, i.e. no messageId name.  To provide more context and information to the user, we now fill this field with the MoltenVK log level string.  This, combined with the `pMessage` string, shows the severity of the message in textual form, and indicates that it comes from MoltenVK.

Note I have changed the `mvkGetReportingLevelString()` function declaration to **inline** - I hope this is appropriate.

Also, after making this change I wonder whether it should also be applied to the debug report callback code, i.e. replace `_debugReportCallbackLayerPrefix` with `mvkGetReportingLevelString()` for consistency?  I don't want to break anything re automated testing scripts, etc. so I didn't make this change yet.  However, I think it's a question worth asking.

### DIFF
--- a/MoltenVK/MoltenVK/Utility/MVKBaseObject.h
+++ b/MoltenVK/MoltenVK/Utility/MVKBaseObject.h
@@ -225,10 +225,8 @@ static inline const MVKConfiguration& mvkGetMVKConfig(MVKBaseObject* mvkObj) {
 	return mvkObj ? mvkObj->getMVKConfig() : getGlobalMVKConfig();
 }
 
-/**
- * Returns the reporting level string associated with the specified MoltenVK log level.
- */
-static inline const char* mvkGetReportingLevelString(MVKConfigLogLevel logLevel) {
+/** Returns the reporting level string associated with the specified MoltenVK log level. */
+static constexpr char* mvkGetReportingLevelString(MVKConfigLogLevel logLevel) {
 	switch (logLevel) {
 		case MVK_CONFIG_LOG_LEVEL_ERROR:    return "mvk-error";
 		case MVK_CONFIG_LOG_LEVEL_WARNING:  return "mvk-warn";
--- a/MoltenVK/MoltenVK/Utility/MVKBaseObject.h
+++ b/MoltenVK/MoltenVK/Utility/MVKBaseObject.h
@@ -224,3 +224,16 @@ protected:
 static inline const MVKConfiguration& mvkGetMVKConfig(MVKBaseObject* mvkObj) {
 	return mvkObj ? mvkObj->getMVKConfig() : getGlobalMVKConfig();
 }
+
+/**
+ * Returns the reporting level string associated with the specified MoltenVK log level.
+ */
+static inline const char* mvkGetReportingLevelString(MVKConfigLogLevel logLevel) {
+	switch (logLevel) {
+		case MVK_CONFIG_LOG_LEVEL_ERROR:    return "mvk-error";
+		case MVK_CONFIG_LOG_LEVEL_WARNING:  return "mvk-warn";
+		case MVK_CONFIG_LOG_LEVEL_INFO:     return "mvk-info";
+		case MVK_CONFIG_LOG_LEVEL_DEBUG:    return "mvk-debug";
+		default:                            return "mvk-unknown";
+	}
+}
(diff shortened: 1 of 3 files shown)


MoltenVK/MoltenVK/Utility/MVKBaseObject.h
```diff
@@ -224,3 +224,16 @@ class MVKConfigurableMixin {
 static inline const MVKConfiguration& mvkGetMVKConfig(MVKBaseObject* mvkObj) {
 	return mvkObj ? mvkObj->getMVKConfig() : getGlobalMVKConfig();
 }
+
+/**
+ * Returns the reporting level string associated with the specified MoltenVK log level.
+ */
```
billhollings, line 230:
```suggestion

#pragma mark -
#pragma mark Support functions

/** Returns the reporting level string associated with the specified MoltenVK log level. */
```

MoltenVK/MoltenVK/Utility/MVKBaseObject.h
```diff
@@ -224,3 +224,16 @@ class MVKConfigurableMixin {
...
+static inline const char* mvkGetReportingLevelString(MVKConfigLogLevel logLevel) {
```
billhollings, line 231:
```suggestion
/** Returns the reporting level string associated with the specified MoltenVK log level. */
static constexpr char* mvkGetReportingLevelString(MVKConfigLogLevel logLevel) {
```

billhollings:
> Also, after making this change I wonder whether it should also be applied to the debug report callback code, i.e. replace `_debugReportCallbackLayerPrefix` with `mvkGetReportingLevelString()` for consistency? I don't want to break anything re automated testing scripts, etc. so I didn't make this change yet. However, I think it's a question worth asking.

Hmmm...I'm not sure. It's kind of arbitrary, but `PFN_vkDebugReportCallbackEXT:: pLayerPrefix` seems to be about layer identification, so `"MoltenVK"` seems appropriate to me.

billhollings (approved):
Thanks for submitting this. LGTM.

SRSaunders:
Thanks for reviewing and merging.

MoltenVK/MoltenVK/Utility/MVKBaseObject.h
SRSaunders: Causes warnings for static library builds (on Ventura, Xcode 15.2):

`ISO C++11 does not allow conversion from string literal to 'char *' `

Issue #2216, opened 2024-04-25
Vulkan examples regression: descriptor indexing and oit


SRSaunders:
I have been working on updating _macOS_ and _iOS_ support and fixing a few defects that I found in Sascha Willems' Vulkan examples repository.  See https://github.com/SaschaWillems/Vulkan/pull/1117 and https://github.com/SaschaWillems/Vulkan/pull/1119.  During this work I observed two examples that expose what appear to regressions across MoltenVK releases:

• The _descriptorindexing_ example runs well on Vulkan SDK 1.3.250.1/MoltenVK 1.2.4, but fails on any later release with:
```
[mvk-error] SPIR-V to MSL conversion error: Argument buffer resource base type could not be determined. When padding argument buffer elements, all descriptor set resources must be supplied with a base type by the app.
[mvk-error] VK_ERROR_INVALID_SHADER_NV: Fragment shader function could not be compiled into pipeline. See previous logged error.
Fatal : VkResult is "ERROR_INVALID_SHADER_NV" in /Users/steve/XcodeProjects/Vulkan/examples/descriptorindexing/descriptorindexing.cpp at line 346
Assertion failed: (res == VK_SUCCESS), function preparePipelines, file descriptorindexing.cpp, line 346.

```
• The _oit_ example runs well up to SDK 1.3.275.0/MoltenVK 1.2.7, but fails on the current release (1.3.280.0/1):
```
-[MTLTextureDescriptorInternal validateWithDevice:]:1344: failed assertion `Texture Descriptor Validation
MTLTextureUsage has unknown bits 0x20.

```

The first of these may be a duplicate of #2180, but I just wanted to report them here to make sure the MoltenVK team is aware.

SRSaunders:
After a bit of debugging, I believe the first one is a **SPIRVCross** defect.  I patched in the old SPIRVCross (from MoltenVK 1.2.4) at commit 12542fc6fc05000e04742daf93892a0b10edbe80 to the most recent head of MoltenVK 1.2.9.  I had to comment out a few newer msl options, but this did not seem to affect things for the `descriptorindexing` example.  I now have a working example with this mashup.

I have not done any research on the second issue above (`oit` example).
![Screenshot 2024-05-05 at 11 22 08 AM](https://github.com/KhronosGroup/MoltenVK/assets/82544213/82d9ce74-8f42-4473-9d4b-0bc0611171b7)

SRSaunders:
[...]

Issue #2180, opened 2024-03-12
SPIR-V to MSL conversion error in MoltenVK but not when using spirv-cross command line


RefuX:
Disclaimer: Sorry for any incorrect assumptions, this is my first time trying to debug SPIR-V issues.

I'm having a go at getting [multi_draw_indirect](https://github.com/KhronosGroup/Vulkan-Samples/tree/main/samples/performance/multi_draw_indirect) in [Vulkan-Samples](https://github.com/KhronosGroup/Vulkan-Samples/tree/main) working, and I'm running into a issue where I'm getting this error:
`SPIR-V to MSL conversion error: Argument buffer resource base type could not be determined. When padding argument buffer elements, all descriptor set resources must be supplied with a base type by the app.`

I turned on the logging in MoltenVK (`MVK_CONFIG_DEBUG=1`, `MVK_CONFIG_LOG_LEVEL=4`) and see I'm getting the error on the [multi_draw_indirect.frag](https://github.com/KhronosGroup/Vulkan-Samples/blob/main/shaders/multi_draw_indirect/multi_draw_indirect.frag) file:

```
[mvk-info] Compiling Metal shader with FastMath enabled.
[mvk-error] SPIR-V to MSL conversion error: Argument buffer resource base type could not be determined. When padding argument buffer elements, all descriptor set resources must be supplied with a base type by the app.
[mvk-info] Converting SPIR-V:
; SPIR-V
; Version: 1.0
; Generator: Khronos Glslang Reference Front End; 11
; Bound: 51
; Schema: 0
               OpCapability Shader
          %1 = OpExtInstImport "GLSL.std.450"
               OpMemoryModel Logical GLSL450
               OpEntryPoint Fragment %4 "main" %9 %18 %28
               OpExecutionMode %4 OriginUpperLeft
               OpSource GLSL 460
               OpName %4 "main"
               OpName %9 "o_color"
               OpName %16 "textures"
               OpName %18 "in_texture_index"
               OpName %28 "in_uv"
               OpDecorate %9 Location 0
               OpDecorate %16 DescriptorSet 0
               OpDecorate %16 Binding 1
               OpDecorate %18 Flat
               OpDecorate %18 Location 2
               OpDecorate %28 Location 1
          %2 = OpTypeVoid
          %3 = OpTypeFunction %2
          %6 = OpTypeFloat 32
          %7 = OpTypeVector %6 4
          %8 = OpTypePointer Output %7
[...]

PR #2231, opened 2024-05-06
Fix shader argument buffer padding and atomics.

billhollings:
- Pull in SPIRV-Cross fix for argument buffer padding.
- Only use `MTLTextureUsageShaderAtomic` if native atomics are supported on the platform.

Fixes issues #2216 and #2180.

### DIFF
--- a/Docs/Whats_New.md
+++ b/Docs/Whats_New.md
@@ -28,6 +28,8 @@ Released 2024/05/07
 - Fix crash when using `VK_EXT_metal_objects` under _ARC_.
 - Fix deadlock when creating a swapchain on a thread other than the main thread.
 - Fix potential memory leak in `vkQueueWaitIdle()`.
+- Fix argument buffer padding on shader conversion.
+- Avoid atomic image usage if native atomics are not supported on the platform.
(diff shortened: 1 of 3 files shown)


cdavis5e (approved):
Dunno why you asked for my review here, and not on the SPIRV-Cross change, since all this does is pull in an updated version of SPIRV-Cross... but, whatever.

cdavis5e:
For the record, the SPIRV-Cross change looks good, too.

billhollings:
> Dunno why you asked for my review here, and not on the SPIRV-Cross change, since all this does is pull in an updated version of SPIRV-Cross... but, whatever.

Normally I would.

With the SDK release coming up, it was a bit of a rush to get everything sorted out. So to expedite, I thought I'd get the SPRIV-Cross MoltenVK branch set up, and populated here, so they could be reviewed together. If it didn't work out, I could then just delete the SPIRV-Cross branch. 

All part of my master plan! 😂

billhollings:
I've updated this to include the fix for the atomic issue in #2216, and to use the new commit hash from the `main` branch in SPIRV-Cross.

cdavis5e (approved):

Issue #2235, opened 2024-05-09
MoltenVK timestamp doesn't report timing precisely


PENGUINLIONG:
In the current implementation MoltenVK doesn't enforce a barrier when sampling a timestamp, in which case the timestamp may be sampled while some previous shaders are still running. As a result, the timings are totally unusable because they don't correctly represent the time spent on my dispatches. Is it possible to give it a `YES` to the `withBarrier` params?

https://github.com/KhronosGroup/MoltenVK/blob/bf097edc74ec3b6dfafdcd5a38d3ce14b11952d6/MoltenVK/MoltenVK/Commands/MVKCommandBuffer.mm#L1004-L1016

PR #2236, opened 2024-05-09
Enforce barrier when sampling timestamps

PENGUINLIONG:
Some context: https://github.com/KhronosGroup/MoltenVK/issues/2235.


SRSaunders:
Will there be a performance hit by doing this?  If so, then perhaps it needs to be an optional parm.

PENGUINLIONG:
@SRSaunders Is there anything I could refer to?

billhollings (approved):
Interesting catch! Good work. And thanks for submitting this fix!

billhollings:
> @SRSaunders Is there anything I could refer to?

This is a good point, and probably the reason we originally set `barrier` to `NO`.

Apple's doc does [indicate](https://developer.apple.com/documentation/metal/mtlrendercommandencoder/3194379-samplecountersinbuffer?language=objc) that it may affect GPU performance, as one would expect from any barrier.

However, the Vulkan spec explicitly [states](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/vkCmdWriteTimestamp.html#_description):

> When `vkCmdWriteTimestamp` is submitted to a queue, it defines an execution dependency on commands that were submitted before it, and writes a timestamp to a query pool.

which would indicate that a barrier is required and expected here, and that timestamp scope accuracy is more important than absolute performance.

PR #2241, opened 2024-05-28
Added mechanism to trigger GPU capture without attaching a debugger.

js6i:
Occasionally I want to capture a single GPU frame but have trouble attaching Xcode. I hacked in something of this sort several times already, so here it is cleaned up.

### DIFF
--- a/MoltenVK/MoltenVK/API/mvk_private_api.h
+++ b/MoltenVK/MoltenVK/API/mvk_private_api.h
@@ -124,9 +124,10 @@ typedef enum MVKConfigTraceVulkanCalls {
 
 /** Identifies the scope for Metal to run an automatic GPU capture for diagnostic debugging purposes. */
 typedef enum MVKConfigAutoGPUCaptureScope {
-	MVK_CONFIG_AUTO_GPU_CAPTURE_SCOPE_NONE     = 0,	/**< No automatic GPU capture. */
-	MVK_CONFIG_AUTO_GPU_CAPTURE_SCOPE_DEVICE   = 1,	/**< Automatically capture all GPU activity during the lifetime of a VkDevice. */
-	MVK_CONFIG_AUTO_GPU_CAPTURE_SCOPE_FRAME    = 2,	/**< Automatically capture all GPU activity during the rendering and presentation of the first frame. */
+	MVK_CONFIG_AUTO_GPU_CAPTURE_SCOPE_NONE      = 0,	/**< No automatic GPU capture. */
+	MVK_CONFIG_AUTO_GPU_CAPTURE_SCOPE_DEVICE    = 1,	/**< Automatically capture all GPU activity during the lifetime of a VkDevice. */
+	MVK_CONFIG_AUTO_GPU_CAPTURE_SCOPE_FRAME     = 2,	/**< Automatically capture all GPU activity during the rendering and presentation of the first frame. */
+	MVK_CONFIG_AUTO_GPU_CAPTURE_SCOPE_ON_DEMAND = 3,	/**< Capture all GPU activity when signaled on a temporary named pipe. */
 	MVK_CONFIG_AUTO_GPU_CAPTURE_SCOPE_MAX_ENUM = 0x7FFFFFFF
 } MVKConfigAutoGPUCaptureScope;
 
--- a/MoltenVK/MoltenVK/GPUObjects/MVKDevice.h
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKDevice.h
@@ -776,7 +776,8 @@ public:
 	 *
 	 * The mtlCaptureObject must be one of:
 	 *   - MTLDevice for scope MVK_CONFIG_AUTO_GPU_CAPTURE_SCOPE_DEVICE
-	 *   - MTLCommandQueue for scope MVK_CONFIG_AUTO_GPU_CAPTURE_SCOPE_FRAME.
+	 *   - MTLCommandQueue for scopes MVK_CONFIG_AUTO_GPU_CAPTURE_SCOPE_FRAME
+	 *       and MVK_CONFIG_AUTO_GPU_CAPTURE_SCOPE_ON_DEMAND.
 	 */
 	void startAutoGPUCapture(MVKConfigAutoGPUCaptureScope autoGPUCaptureScope, id mtlCaptureObject);
 
@@ -839,6 +840,7 @@ protected:
 	void getDescriptorVariableDescriptorCountLayoutSupport(const VkDescriptorSetLayoutCreateInfo* pCreateInfo,
 														   VkDescriptorSetLayoutSupport* pSupport,
 														   VkDescriptorSetVariableDescriptorCountLayoutSupport* pVarDescSetCountSupport);
+	bool readGPUCapturePipe() { char dummy; return _capturePipeFileDesc >= 0 && read(_capturePipeFileDesc, &dummy, 1) > 0; };
 
 	MVKPhysicalDevice* _physicalDevice = nullptr;
 	MVKExtensionList _enabledExtensions;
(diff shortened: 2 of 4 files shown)


MoltenVK/MoltenVK/GPUObjects/MVKDevice.h
```diff
@@ -869,6 +870,8 @@ class MVKDevice : public MVKDispatchableVulkanAPIObject {
 	id<MTLSamplerState> _defaultMTLSamplerState = nil;
 	id<MTLBuffer> _dummyBlitMTLBuffer = nil;
     uint32_t _globalVisibilityQueryCount = 0;
+	int _capturePipeFildes = -1;
+	std::string _capturePipeFileName;
```
billhollings, line 874: Where possible, we do try to organize to minimize dead space in the underlying class structures. Recommend moving `_capturePipeFileName` up a few lines to below `std::mutex`.

And it took me a while to mentally parse `Fildes`! 😉 

I'll suggest renaming `_capturePipeFildes` to `_capturePipeFileDesc`, to match pattern in `_capturePipeFileName`.

MoltenVK/MoltenVK/API/mvk_private_api.h
```diff
@@ -124,9 +124,10 @@ typedef enum MVKConfigTraceVulkanCalls {
 
 /** Identifies the scope for Metal to run an automatic GPU capture for diagnostic debugging purposes. */
 typedef enum MVKConfigAutoGPUCaptureScope {
-	MVK_CONFIG_AUTO_GPU_CAPTURE_SCOPE_NONE     = 0,	/**< No automatic GPU capture. */
-	MVK_CONFIG_AUTO_GPU_CAPTURE_SCOPE_DEVICE   = 1,	/**< Automatically capture all GPU activity during the lifetime of a VkDevice. */
-	MVK_CONFIG_AUTO_GPU_CAPTURE_SCOPE_FRAME    = 2,	/**< Automatically capture all GPU activity during the rendering and presentation of the first frame. */
+	MVK_CONFIG_AUTO_GPU_CAPTURE_SCOPE_NONE      = 0,	/**< No automatic GPU capture. */
+	MVK_CONFIG_AUTO_GPU_CAPTURE_SCOPE_DEVICE    = 1,	/**< Automatically capture all GPU activity during the lifetime of a VkDevice. */
+	MVK_CONFIG_AUTO_GPU_CAPTURE_SCOPE_FRAME     = 2,	/**< Automatically capture all GPU activity during the rendering and presentation of the first frame. */
+	MVK_CONFIG_AUTO_GPU_CAPTURE_SCOPE_ON_DEMAND = 3,	/**< Capture all GPU activity when signaled on a temporary named pipe. */
```
billhollings, line 130: Please add `MVK_CONFIG_AUTO_GPU_CAPTURE_SCOPE_ON_DEMAND` to the comment on the `startAutoGPUCapture()` declaration.

MoltenVK/MoltenVK/GPUObjects/MVKQueue.mm
```diff
@@ -705,6 +705,12 @@
 		setConfigurationResult(_presentInfo[i].presentableImage->presentCAMetalDrawable(mtlCmdBuff, _presentInfo[i]));
 	}
 
+	if (_queue->_queueFamily->getIndex() == getMVKConfig().defaultGPUCaptureScopeQueueFamilyIndex &&
+		_queue->_index == getMVKConfig().defaultGPUCaptureScopeQueueIndex) {
+		getDevice()->stopAutoGPUCapture(MVK_CONFIG_AUTO_GPU_CAPTURE_SCOPE_ON_DEMAND);
+		getDevice()->startAutoGPUCapture(MVK_CONFIG_AUTO_GPU_CAPTURE_SCOPE_ON_DEMAND, _queue->getMTLCommandQueue());
+	}
```
billhollings, line 712: What happens when you restart the capture on each frame? Metal typically refuses to overwrite the capture file.

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
```diff
@@ -4880,6 +4884,20 @@ static uint32_t mvkGetEntryProperty(io_registry_entry_t entry, CFStringRef prope
 
 	startAutoGPUCapture(MVK_CONFIG_AUTO_GPU_CAPTURE_SCOPE_DEVICE, _physicalDevice->_mtlDevice);
 
+	if (getMVKConfig().autoGPUCaptureScope == MVK_CONFIG_AUTO_GPU_CAPTURE_SCOPE_ON_DEMAND) {
+		char pipeName[] = "/tmp/MoltenVKCapturePipe-XXXXXXXXXX";
+		mktemp(pipeName);
```
billhollings, line 4889: `mktemp()` triggers a deprecation warning on build, and recommends using `mkstemp()`. Doing so might also simplify your code here. There is also `mkostemp()` to specify the file access modes, if needed.

If you absolutely need to use `mktemp()`, wrap this line, or this section, in

```
#pragma clang diagnostic push
#pragma clang diagnostic ignored "-Wdeprecated-declarations"
			    mktemp(pipeName);    // Deprecated but needed for interprocess transfers
#pragma clang diagnostic pop
```

See use elsewhere in MoltenVK.

billhollings (changes requested):
Interesting approach. Thanks for submitting.

A few nitpick changes requested. And need to deal with the deprecation warning.

MoltenVK/MoltenVK/GPUObjects/MVKDevice.h
cdavis5e: > And it took me a while to mentally parse `Fildes`! 😉

Yeah. That's actually a somewhat common abbreviation you see in Unix manpages. The name of a file descriptor parameter is often `fildes`.

Personally, I prefer `_capturePipeFD`. It's shorter, and the initialism "FD" is well known.

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
cdavis5e: The problem is that he does in fact need to use `mktemp(3)` here, because he is not creating a regular file or a directory but a FIFO.

Which reminds me...

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
```diff
@@ -4880,6 +4884,20 @@ static uint32_t mvkGetEntryProperty(io_registry_entry_t entry, CFStringRef prope
...
+		if (mkfifo(pipeName, 0600) < 0) {
```
cdavis5e, line 4890: If `errno` contains `EEXIST`, that means there was a collision and you should probably try this again with a new filename.

MoltenVK/MoltenVK/GPUObjects/MVKQueue.mm
js6i: It doesn't restart on each frame. If there's more bytes in the pipe, `stopAutoGPUCapture` consumes a byte and just returns, and then `startAutoGPUCapture` also just bails out because it's still capturing. That way you can write N bytes to capture N consecutive frames. Once the capture is completed, you just move the capture file somewhere else and you're good to make another one.

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
js6i: The other `mk*temp` functions don't work, as far as I could tell there's no way to create a pipe using them. One alternative option is to create a temporary directory and then create a pipe with a static name in it. But to me that seems worse. I'll add the pragma thing.

cdavis5e (changes requested):

MoltenVK/MoltenVK/GPUObjects/MVKDevice.h
js6i: I went with FileDesc, feel free to change.

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
js6i: Seems a little defensive to me, but fine.

MoltenVK/MoltenVK/GPUObjects/MVKQueue.mm
billhollings: Ah. I see now. Very clever.

billhollings (approved):
Thanks for making changes. LGTM now.

There's some documentation still to be done in `MoltenVK_Configuration_Parameters.md`, but I'll take care of that, because I think I need to rethink how that is laid out, and the reference to _Xcode_.

cdavis5e (approved):

PR #2255, opened 2024-06-14
[ios18 support] add new case for Metal Version 3.2 that is on ios 18 -- reroutes MoltenMK to use version 3.1

Daij-Djan:
This adds a new preprocessor macro for xcode16 / ios18 sdk and reroutes the version to 3.1 (additions to MVKDevice are iline with as what we did for other xcode versions in the past)

### DIFF
--- a/Common/MVKCommonEnvironment.h
+++ b/Common/MVKCommonEnvironment.h
@@ -107,6 +107,10 @@ extern "C" {
 #endif
 
 /** Building with Xcode versions. iOS version also covers tvOS. */
+#ifndef MVK_XCODE_16
+#define MVK_XCODE_16			((__MAC_OS_X_VERSION_MAX_ALLOWED >= 150000) || \
+                                    (__IPHONE_OS_VERSION_MAX_ALLOWED >= 180000))
+#endif
 #ifndef MVK_XCODE_15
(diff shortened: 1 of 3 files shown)


billhollings (changes requested):
Thanks for taking care of this, and for submitting this PR!

> reroutes the version to 3.1
> ...
> // We don't explicitly support Metal 3.2 yet

What do you mean by this? Is there a reason why you are not using `MTLLanguageVersion3_2`?  The `MVK_XCODE_16` and `mvkOSVersionIsAtLeast()` guard code should protect against inappropriate build or runtime usage. Have you tried building or running, and are experiencing problems?

In addition, can you also add the equivalent changes to `OSSupport.mm`?

Thanks.

Daij-Djan:
I didnt wanna make any logic changes so I just made it 3.1 -- I didnt test runtime behaviour of moltenVK when I did the change.

billhollings:
Change it to `MTLLanguageVersion3_2` and test whether you can build using Xcode 15, to make sure it doesn't cause a regression break. If that is successful, we should be good.

If you have Xcode 16 up and running, test a build under that as well. If not, it should be good enough to go ahead with `MTLLanguageVersion3_2` anyway.

billhollings:
Merging this into a temp branch to fix the outstanding issues.

Issue #2246, opened 2024-06-01
Question: Does MVK support VK_DESCRIPTOR_BINDING_VARIABLE_DESCRIPTOR_COUNT_BIT_EXT?


SRSaunders:
The _descriptor_indexing_ sample in the official **Khronos Vulkan-Samples** repo fails (https://github.com/KhronosGroup/Vulkan-Samples/blob/main/samples/extensions/descriptor_indexing/descriptor_indexing.cpp), even with argument buffers enabled.

I have tracked down the issue, and it relates to exhaustion of the descriptor pool.  The sample defines bindings with `set_layout_binding.descriptorCount = maxDescriptorSetUpdateAfterBindSampledImages`, i.e. the max.  This seems excessive, but may not be an issue for other platforms that respect the `VK_DESCRIPTOR_BINDING_VARIABLE_DESCRIPTOR_COUNT_BIT_EXT` flag which is set by the sample.

The problem arises when actually allocating descriptor sets, i.e. when calling `vkAllocateDescriptorSets()` extended by the `VkDescriptorSetVariableDescriptorCountAllocateInfoEXT` structure.  The structure properly defines the number of descriptors to be allocated (i.e. `NumDescriptorsStreaming` and `NumDescriptorsNonUniform` in two separate allocations), but MoltenVK's allocate action seems to check if there are `set_layout_binding.descriptorCount` resources available in the pool, which in this case is 500,000 (the max) and of course this fails since it exceeds the pool size.

I have been able to verify this and get the sample working by using `NumDescriptorsStreaming` in the binding definition, and by increasing the pool size to 2 x `NumDescriptorsStreaming` (for 2 allocation calls).  However, I suspect this may violate the spirit of the sample where variable descriptor count allocations are used.

I am wondering whether MoltenVK supports the variable descriptor count flag and structure, and if so, should it instead be checking for `variable_info.pDescriptorCounts` resources in the pool when the variable bit is enabled and the variable descriptor count structure is present in the pNext chain of `allocate_info`.

I am putting this forward as a question, since I may not have this right.  However, I wanted to report it here before going back to the official sample repo and offering any solutions for MoltenVK in that forum.

billhollings:
The use of Metal Argument Buffers in MoltenVK is currently being overhauled. It should be published in the next week or so. At that point, you can retest this, and if it's still an issue, we can fix it within the new argument buffer environment.

SRSaunders:
[...]

PR #2260, opened 2024-07-01
Improvements to bindless resources and descriptor indexing

billhollings:
This is a large PR that adds significant improvements to the handling of bindless resources and descriptor indexing functionality.

- Add support for Metal3 argument buffers.
- Support argument buffers on all platforms, when Metal 3 is available.
- Support argument buffers on macOS when Metal3 is not available.
- Use Metal argument buffers by default when they are available.
- Revert `MVKConfiguration::useMetalArgumentBuffers` and env var `MVK_CONFIG_USE_METAL_ARGUMENT_BUFFERS` to a boolean value, and enable it by default.
- Update max number of bindless buffers and textures per stage to 1M, per Apple Docs.

Further detailed outlines of the additions and changes to functionality are contained in the notes for each commit.

The new use of argument buffers by default, plus the use of either Metal3 argument buffers or argument buffer encoding, all pass the same CTS tests that are passed using discrete resource indexes, and that were being passed prior to these changes.


MoltenVK/MoltenVK/GPUObjects/MVKDescriptorSet.mm
```diff
@@ -483,42 +625,39 @@
 	return nullptr;
 }
 
-// Retieves the first available descriptor set from the pool, and configures it.
```
cdavis5e, line 486: You've gotten better at spotting these typos.

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
```diff
@@ -576,19 +576,19 @@
 	supportedProps12.robustBufferAccessUpdateAfterBind						= _features.robustBufferAccess;
 	supportedProps12.quadDivergentImplicitLod								= false;
 	supportedProps12.maxPerStageDescriptorUpdateAfterBindSamplers			= isTier2 ? maxSamplerCnt : _properties.limits.maxPerStageDescriptorSamplers;
-	supportedProps12.maxPerStageDescriptorUpdateAfterBindUniformBuffers		= isTier2 ? 500000 : _properties.limits.maxPerStageDescriptorUniformBuffers;
-	supportedProps12.maxPerStageDescriptorUpdateAfterBindStorageBuffers		= isTier2 ? 500000 : _properties.limits.maxPerStageDescriptorStorageBuffers;
-	supportedProps12.maxPerStageDescriptorUpdateAfterBindSampledImages		= isTier2 ? 500000 : _properties.limits.maxPerStageDescriptorSampledImages;
-	supportedProps12.maxPerStageDescriptorUpdateAfterBindStorageImages		= isTier2 ? 500000 : _properties.limits.maxPerStageDescriptorStorageImages;
+	supportedProps12.maxPerStageDescriptorUpdateAfterBindUniformBuffers		= isTier2 ? 1e6 : _properties.limits.maxPerStageDescriptorUniformBuffers;
```
cdavis5e, line 579: Integral constants with exponential notation? I thought this was C++, not Go. 😄  (In Go, all numeric constants not explicitly typed have indeterminate context-dependent type.)

Cute trick, though.

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
```diff
@@ -2388,24 +2388,27 @@
...
 		_metalFeatures.argumentBuffersTier = _mtlDevice.argumentBuffersSupport;
 	} else {
 		_metalFeatures.argumentBuffersTier = MTLArgumentBuffersTier1;
 	}
 
+	// Metal argument buffer support for descriptor sets is supported on macOS 11.0 or later,
+	// or on older versions of macOS using an Intel GPU, or on iOS & tvOS 16.0 or later (Metal 3).
+	_metalFeatures.descriptorSetArgumentBuffers = (_metalFeatures.argumentBuffers &&
+												   (mvkOSVersionIsAtLeast(11.0, 16.0, 1.0) ||
+													_properties.vendorID == kIntelVendorId));
```
cdavis5e, line 2402: Something I need to do is, test argument buffers on my Mac Pro with AMD hardware.

MoltenVK/MoltenVK/Utility/MVKLogging.h
```diff
@@ -128,7 +128,7 @@ extern "C" {
 #	define MVK_LOG_LEVEL_INFO		MVK_LOGGING_ENABLED
 #endif
 #ifndef MVK_LOG_LEVEL_DEBUG
-#	define MVK_LOG_LEVEL_DEBUG		(MVK_LOGGING_ENABLED && MVK_CONFIG_DEBUG)
+#	define MVK_LOG_LEVEL_DEBUG		MVK_LOGGING_ENABLED
```
cdavis5e, line 131: Did you mean to add this change? This will result in debug spam anytime logging is enabled.

cdavis5e (changes requested):
Sorry about the wait. This was, as you noted, a large change.

MoltenVK/MoltenVK/API/mvk_private_api.h
```diff
@@ -44,7 +44,7 @@ typedef unsigned long MTLArgumentBuffersTier;
  */
 
 
-#define MVK_PRIVATE_API_VERSION   41
+#define MVK_PRIVATE_API_VERSION   43
```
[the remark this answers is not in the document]
billhollings: Thanks, good catch. Looks like two of my commits here separately incremented it, and I didn't notice.

MoltenVK/MoltenVK/GPUObjects/MVKDescriptor.mm
```diff
@@ -699,6 +740,24 @@ void mvkPopulateShaderConversionConfig(mvk::SPIRVToMSLConversionConfiguration& s
...
+
+	if ( !_layout->isUsingMetalArgumentBuffers() ) { return false; }
+	if ( getDescriptorCount() == 0 ) { return false; }
+
+	switch (getDescriptorType()) {
+		case VK_DESCRIPTOR_TYPE_UNIFORM_BUFFER:
+		case VK_DESCRIPTOR_TYPE_UNIFORM_BUFFER_DYNAMIC:
+		case VK_DESCRIPTOR_TYPE_STORAGE_BUFFER:
+		case VK_DESCRIPTOR_TYPE_STORAGE_BUFFER_DYNAMIC:
+		case VK_DESCRIPTOR_TYPE_INLINE_UNIFORM_BLOCK_EXT:
```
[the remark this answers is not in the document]
billhollings: This is probably a good idea. For discrete bindings, we could likely also use `setVertexBytes:`, etc., to avoid a separate buffer.

Based on a quick triage, this might be a complex and invasive change. We'd have to track argument buffer offsets as well as current slot indexes, and Metal requires it to be declared as an array, and seems to consume a slot index for each element, which would be odd, if we treat it as a `char` array, as indexes would potentially jump by 4K at a time. And SPIRV-Cross will probably need some mods to accommodate recasting pointers, etc.

I do have plans for a couple of more PRs to come after this, including refactoring how we handle resource usage tracking (incl the new `MTLResidencySet`), so let's leave this out of the current PR, and investigate separately.

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
billhollings: The lack of commas in literals made repeated occurrences of `1000000` hard to visually parse, so decided to use `1e6` to make it easier.

I thought about defining a constant, but the many different settings are only loosely tied together.

MoltenVK/MoltenVK/Utility/MVKLogging.h
billhollings: Thanks for checking, but yes, it's intentional.

Not so much spam, since the config would have to be set to `MVK_CONFIG_LOG_LEVEL_DEBUG` for it to actually show up in the log. 

What the change to `MVK_LOG_LEVEL_DEBUG` does is remove the conditional compilation of the logging call, so the negative impact will be the overhead of a report/log call that will almost never be seen.

I made the change to normalize debug logging. I thought it would be useful to be able to allow a user in the wild to be able to trigger debug logging, by just adjusting an env var, without having to go through a rebuild of MoltenVK as a debug build.

I'm not 100% sure the convenience/performance trade-off is worth it, but felt it might be helpful for debugging in the wild. I'm happy to hear thoughts and suggestions about that trade-off.

MoltenVK/MoltenVK/GPUObjects/MVKDescriptor.mm
```diff
@@ -490,19 +485,50 @@ void mvkPopulateShaderConversionConfig(mvk::SPIRVToMSLConversionConfiguration& s
...
+		case VK_DESCRIPTOR_TYPE_STORAGE_TEXEL_BUFFER:
+			return getMetalFeatures().nativeTextureAtomics ? 1 : 2;
+		default:
+			return 1;
+	}
+}
+
+uint64_t MVKDescriptorSetLayoutBinding::getMetal3ArgumentBufferEncodedSize() {
+	return getMTLResourceCountPerElement() * getDescriptorCount() * kMVKMetal3ArgBuffSlotSizeInBytes;
+}
```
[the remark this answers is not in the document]
billhollings: Multi-plane tracking is problematic, because there does not seem to be any way to know at descriptor set allocation time, that any descriptor might be assigned a multi-plane image, so we'll have to reserve 3 texture slots for all images.

This also affects discrete texture indexing as well, since any of those could be a multi-plane image as well.

So far, we seem to be getting away with this simply because CTS only tests with one image per descriptor set.

So far, I think the best we can do is that when `samplerYcbcrConversion` or `VK_KHR_sampler_ycbcr_conversion` is enabled, allocate 3 argument buffer slots per image descriptor, and otherwise (the more common use case) allocate 1 argument buffer slot per image descriptor.

Anyway, since this is messy, and is not strictly tied to argument buffers, I decided to address this in a separate PR, which I'll start to work on once this PR is finished.

Longer term, another option might be to add a Vulkan extension to add a flag to `VkDescriptorBindingFlagBits` to indicate a multi-plane image might be stored in that descriptor (or maybe the reverse, that it definitely won't).

I'd love to hear any thoughts you might have about handling this.

billhollings:
Thanks for reviewing this large PR so quickly!

I made one suggested change, responded to the others, and asked for your thoughts on a couple of discussion items.

MoltenVK/MoltenVK/Utility/MVKLogging.h
cdavis5e: That makes sense. It's fine. Hopefully, this will make it easier to diagnose issues in the wild.

MoltenVK/MoltenVK/GPUObjects/MVKDescriptor.mm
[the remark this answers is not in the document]
cdavis5e: > So far, I think the best we can do is that when `samplerYcbcrConversion` or `VK_KHR_sampler_ycbcr_conversion` is enabled, allocate 3 argument buffer slots per image descriptor, and otherwise (the more common use case) allocate 1 argument buffer slot per image descriptor.

Yeah, that's what I was thinking.

> Anyway, since this is messy, and is not strictly tied to argument buffers, I decided to address this in a separate PR, which I'll start to work on once this PR is finished.

Very well.

> Longer term, another option might be to add a Vulkan extension to add a flag to `VkDescriptorBindingFlagBits` to indicate a multi-plane image might be stored in that descriptor (or maybe the reverse, that it definitely won't).

That actually sounds like something that should've been part of `VK_KHR_sampler_ycbcr_conversion` in the first place. I wonder how other drivers handle this situation...

As you probably know, I can't push for this in Khronos, since I'm no longer a member.

cdavis5e (approved):

giomasce:
Hi, thanks for working on these changes. I don't know much about MoltenVK, so this might already be known, but after I saw this was merged I tried vkd3d's test on `main` and I am experiencing some problems. For example consider [this MRE](https://paste.debian.net/1322797/): the program creates a descriptor set layout with a single variable binding consisting of two descriptors, then create a descriptor pool with a single descriptor and then tries to allocate the descriptor set layout from the descriptor pool, but requesting just a single descriptor. I think this should work, because the program is effectively allocating a single descriptor, even if the underlying descriptor set layout is variable and has at most two descriptors. But MoltenVK will return `VK_ERROR_OUT_OF_POOL_MEMORY`.

The same problem doesn't happen with #2199.

billhollings:
> Hi, thanks for working on these changes. I don't know much about MoltenVK, so this might already be known, but after I saw this was merged I tried vkd3d's test on `main` and I am experiencing some problems. For example consider [this MRE](https://paste.debian.net/1322797/): the program creates a descriptor set layout with a single variable binding consisting of two descriptors, then create a descriptor pool with a single descriptor and then tries to allocate the descriptor set layout from the descriptor pool, but requesting just a single descriptor. I think this should work, because the program is effectively allocating a single descriptor, even if the underlying descriptor set layout is variable and has at most two descriptors. But MoltenVK will return `VK_ERROR_OUT_OF_POOL_MEMORY`.
> 
> The same problem doesn't happen with #2199.

Thanks for reporting this. This is covered in issue #2246. Working on it.

giomasce:
> Thanks for reporting this. This is covered in issue #2246. Working on it.

Ah, sorry, I hadn't noticed that. Well, thanks for your work anyway! :-)

billhollings:
> > Thanks for reporting this. This is covered in issue #2246. Working on it.
> 
> Ah, sorry, I hadn't noticed that. Well, thanks for your work anyway! :-)

No need to apologize. It's always better to have backup info, and more examples of the issue in the wild.

PR #2268, opened 2024-07-08
Support multiplanar images in Metal argument buffers.

billhollings:
- `MVKDescriptorSetLayoutBinding` add `getMaxPlaneCount()` to determine the number of planes for a combined image-sampler, from the immutable samplers.
- When using `MTLArgumentEncoder`, add encoder arguments for each texture plane.
- Don't build `MTLArgumentEncoder` if descriptor count is zero.
- For discrete multi-plane and multiple descriptors per binding, use max plane count to determine a consistent array size for each plane.
- Some minor formatting fixes found along the way.

Although there was some original concern in [this discussion](https://github.com/KhronosGroup/MoltenVK/pull/2260#discussion_r1663613174) that we might need to hold 3 argument buffer slots for all images if `samplerYcbcrConversion` was enabled, it turned out that this feature only permits multiplanar images as combined image-samplers, and only with samplers that are immutable. So we know at descriptor layout binding definition time which descriptors require more than one slot per image.

### DIFF
--- a/MoltenVK/MoltenVK/GPUObjects/MVKDescriptor.h
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKDescriptor.h
@@ -28,10 +28,6 @@ class MVKCommandEncoder;
 class MVKResourcesCommandEncoderState;
 
 
-// The size of one Metal3 Argument Buffer slot in bytes.
-static const size_t kMVKMetal3ArgBuffSlotSizeInBytes = sizeof(uint64_t);
-
-
 #pragma mark MVKShaderStageResourceBinding
 
 /** Indicates the Metal resource indexes used by a single shader stage in a descriptor. */
@@ -183,8 +179,8 @@ protected:
 										uint32_t dslIndex);
 	bool validate(MVKSampler* mvkSampler);
 	void encodeImmutableSamplersToMetalArgumentBuffer(MVKDescriptorSet* mvkDescSet);
-	uint32_t getMTLResourceCountPerElement();
-	uint64_t getMetal3ArgumentBufferEncodedSize();
+	uint8_t getMaxPlaneCount();
+	uint32_t getMTLResourceCount();
 	bool needsBuffSizeAuxBuffer();
 	std::string getLogDescription();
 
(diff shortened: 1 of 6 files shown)


cdavis5e (approved):
Well, all right, then.

cdavis5e:
There's one possible future optimization that just occurred to me. Since every Y'CbCr sampler knows the number of planes of the image it will be used with, and since they have to be immutable samplers, it should theoretically be possible to allocate only that many slots in the argument buffer for each combined image-sampler, instead of allocating (1+max planes) for every combined image-sampler. But we'd have to know which sampler is to be bound to each descriptor, and I'm not sure we have that information until the descriptor set is actually created.

billhollings:
> There's one possible future optimization that just occurred to me. Since every Y'CbCr sampler knows the number of planes of the image it will be used with, and since they have to be immutable samplers, it should theoretically be possible to allocate only that many slots in the argument buffer for each combined image-sampler, instead of allocating (1+max planes) for every combined image-sampler. But we'd have to know which sampler is to be bound to each descriptor, and I'm not sure we have that information until the descriptor set is actually created.

Good catch! We do indeed know which sampler is bound to each descriptor in a binding, so it is theoretically possible to do this optimization, and I did look into it. Other than the extra optimization effort for what seems like an extreme outlier use case, there are two reasons I didn't try to include it:

1. The spec includes [wording](https://registry.khronos.org/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-VkSamplerYcbcrConversionImageFormatProperties-sType-sType) advising not to:
> All descriptors in a binding use the same maximum `combinedImageSamplerDescriptorCount` descriptors to allow implementations to use a uniform stride for dynamic indexing of the descriptors in the binding.
>
>For example, consider a descriptor set layout binding with two descriptors and immutable samplers for multi-planar formats that have `VkSamplerYcbcrConversionImageFormatProperties::combinedImageSamplerDescriptorCount` values of 2 and 3 respectively. There are two descriptors in the binding and the maximum `combinedImageSamplerDescriptorCount` is 3, so descriptor sets with this layout consume 6 descriptors from the descriptor pool. To create a descriptor pool that allows allocating four descriptor sets with this layout, descriptorCount must be at least 24.

2. SPIRV-Cross necessarily treats the planes as separate `MTLTexture` variables:

```
struct spvDescriptorSetBuffer1
{
    texture2d<float> m_36 [[id(0)]];
    texture2d<float> m_36Plane1 [[id(1)]];
    texture2d<float> m_36Plane2 [[id(2)]];
};
```
AFAICT, there are no CTS that actually test arrays of multi-planar images, but if there were, say, a binding with 5 elements, SPIRV-Cross would probably emit:

```
struct spvDescriptorSetBuffer1
{
    texture2d<float> m_36 [[id(0)]] [5];
    texture2d<float> m_36Plane1 [[id(5)]] [5];
    texture2d<float> m_36Plane2 [[id(10)]] [5];
};
```

Because of this, if any of the binding elements has a 3-plane image, we need the `m_36Plane2` member array, and there's really no way to break this apart to allow, say binding element 3 to consume only 2 slots.

PR #2275, opened 2024-07-16
Add MTLStoreAction mapping for VK_ATTACHMENT_STORE_OP_NONE.

squidbus:
As per https://github.com/KhronosGroup/MoltenVK/issues/1930#issuecomment-2089164911, adds a mapping for `VK_ATTACHMENT_STORE_OP_NONE` to `MTLStoreActionStore`, since it is required by `VK_KHR_dynamic_rendering` and storing is the closest equivalent in behavior provided by Metal. This will help silence log spam from the default case in applications that use this store op.

I assumed that it should mirror the behavior of `VK_ATTACHMENT_STORE_OP_STORE` with regards to multisampling and assigned it the same switch case. Let me know if that's not the case and I'll adjust it.

### DIFF
--- a/MoltenVK/MoltenVK/Vulkan/mvk_datatypes.mm
+++ b/MoltenVK/MoltenVK/Vulkan/mvk_datatypes.mm
@@ -608,8 +608,13 @@ MVK_PUBLIC_SYMBOL MTLStoreAction mvkMTLStoreActionFromVkAttachmentStoreOp(VkAtta
 // If we need to resolve, but the format doesn't support it, we must store the attachment so we can run a post-renderpass compute shader to perform the resolve.
 MTLStoreAction mvkMTLStoreActionFromVkAttachmentStoreOpInObj(VkAttachmentStoreOp vkStoreOp, bool hasResolveAttachment, bool canResolveFormat, MVKBaseObject* mvkObj) {
 	switch (vkStoreOp) {
-		case VK_ATTACHMENT_STORE_OP_STORE:		return hasResolveAttachment && canResolveFormat ? MTLStoreActionStoreAndMultisampleResolve : MTLStoreActionStore;
-		case VK_ATTACHMENT_STORE_OP_DONT_CARE:	return hasResolveAttachment ? (canResolveFormat ? MTLStoreActionMultisampleResolve : MTLStoreActionStore) : MTLStoreActionDontCare;
+		// Metal does not support VK_ATTACHMENT_STORE_OP_NONE. Next best option is store.
+		case VK_ATTACHMENT_STORE_OP_NONE:
+		case VK_ATTACHMENT_STORE_OP_STORE:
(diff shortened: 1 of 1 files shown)


billhollings (approved):
Thanks for submitting this!

LGTM.

MoltenVK/MoltenVK/Vulkan/mvk_datatypes.mm
```diff
@@ -608,8 +608,13 @@ MVK_PUBLIC_SYMBOL MTLStoreAction mvkMTLStoreActionFromVkAttachmentStoreOp(VkAtta
 // If we need to resolve, but the format doesn't support it, we must store the attachment so we can run a post-renderpass compute shader to perform the resolve.
 MTLStoreAction mvkMTLStoreActionFromVkAttachmentStoreOpInObj(VkAttachmentStoreOp vkStoreOp, bool hasResolveAttachment, bool canResolveFormat, MVKBaseObject* mvkObj) {
 	switch (vkStoreOp) {
-		case VK_ATTACHMENT_STORE_OP_STORE:		return hasResolveAttachment && canResolveFormat ? MTLStoreActionStoreAndMultisampleResolve : MTLStoreActionStore;
-		case VK_ATTACHMENT_STORE_OP_DONT_CARE:	return hasResolveAttachment ? (canResolveFormat ? MTLStoreActionMultisampleResolve : MTLStoreActionStore) : MTLStoreActionDontCare;
+		// Metal does not support VK_ATTACHMENT_STORE_OP_NONE. Next best option is store.
```
aitor-lunarg, line 611: Sorry, I just saw this. From the Vulkan spec:
```
VK_ATTACHMENT_STORE_OP_NONE specifies the contents within the render area are not accessed by the store operation as long as no values are written to the attachment during the render pass. If values are written during the render pass, this behaves identically to VK_ATTACHMENT_STORE_OP_DONT_CARE and with matching access semantics.
```

So unsure if storing here is the best option, shouldn't it be the same as `VK_ATTACHMENT_STORE_OP_DONT_CARE`?

billhollings: Have a look at the discussion mentioned above in https://github.com/KhronosGroup/MoltenVK/issues/1930#issuecomment-2089164911, and let me know what your thoughts are after that.

It's a grey area, and the question is about what is the state of the attachment for future render passes.

aitor-lunarg: It does make sense for the canonical render pass approach since the attachment is only supposed to be written at the end, but contents are preserved throughout the subpasses.

For the dynamic rendering approach, it would not be the optimal approach since we would be paying the cost of the actual storage when it cannot be used as input attachment (maybe with `VK_KHR_dynamic_rendering_local_read` extension can). But I guess we can live with it unless there's some CTS test that complains about this I guess (unlikely).

Issue #2284, opened 2024-07-22
Uncaught selector, -[CAMetalDrawable addPresentedHandler:] (first introduced 1.2.6)


WalkerWhite:
I have been working with MoltenVK on top of SDL. I have not seen this error/warning anywhere else online.  But I consistently get the message

`Uncaught selector, -[CAMetalDrawable addPresentedHandler:]
`
in the debug console when running  Vulkan on top of an SDL Vulkan surface.  It occurs EVERY frame, hurting performance and making debugging near impossible.

At first I though this was an SDL issue. I have rolled back multiple SDL versions and I always get this problem.  I even get this issue in projects/tutorials produced by others online -- from people who report no such errors. But searching the SDL code I can see no point where addPresentedHandler would be used.

However, I noticed that these SDL tutorials link to older versions of MoltenVK.  And the source of MoltenVK does call addPresentedHandler. So I tried varying the MoltenVK release.  And this does indeed solve the problem.

Apparently this issue was introduced in v1.2.6-rc1 released on Oct 17, 2023.  The previous 1.2.5 release on August 17, 2023 does not have this issue.

SRSaunders:
This is interesting, as I use MoltenVK with an SDL2 game ([RBDoom3-BFG](https://github.com/RobertBeckebans/RBDOOM-3-BFG)) that uses an SDL2 Vulkan surface but does not experience this issue.  MoltenVK will only add a presentedHandler if Metal can respond to this capability, which is then used to record actual presentation times.  The presentation times are typically used by the application to profile, optimize, and/or control latency which is important for games.  The MoltenVK code looks like this:

```
#if !MVK_OS_SIMULATOR
	if ([mtlDrawable respondsToSelector: @selector(addPresentedHandler:)]) {
		[mtlDrawable addPresentedHandler: ^(id<MTLDrawable> mtlDrwbl) {
			endPresentation(presentInfo, signaler, mtlDrwbl.presentedTime * 1.0e9);
		}];
	} else
#endif
	{
		// If MTLDrawable.presentedTime/addPresentedHandler isn't supported,
		// treat it as if the present happened when requested.
		endPresentation(presentInfo, signaler);
	}
```

[...]

PR #2289, opened 2024-07-25
Add present handler only if VkPresentTimesInfoGOOGLE struct available in VkPresentInfoKHR

SRSaunders:
This PR solves issue #2284.

A presentation handler only needs to be activated if the `VK_GOOGLE_display_timing` extension is enabled and the `VkPresentTimesInfoGOOGLE` struct is included in the pNext chain of the `VkPresentInfoKHR` struct during calls to `vkQueuePresentKHR()`.  Previously the presentation handler would always be added.  This could lead to system-level diagnostic messages when used with `SDL_PollEvent()` as described in the issue noted above.

This change also permits `presentInfo.actualPresentTime` info to be recorded whether `VkPresentTimeGOOGLE* pPresentTimes` is defined or not.  This more closely follows the Vulkan spec which permits `pPresentTimes = NULL`.

I have tested this change against my Optick profiler integration with RBDoom3-BFG and correct `actualPresentTime` and `presentID` information is returned.  No regressions observed running on macOS Ventura 13.6.6 on x86_64.

### DIFF
--- a/MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
@@ -1552,13 +1552,13 @@ VkResult MVKPresentableSwapchainImage::presentCAMetalDrawable(id<MTLCommandBuffe
 	id<CAMetalDrawable> mtlDrwbl = getCAMetalDrawable();
 	MVKSwapchainSignaler signaler = getPresentationSignaler();
 	[mtlCmdBuff addScheduledHandler: ^(id<MTLCommandBuffer> mcb) {
-
-		addPresentedHandler(mtlDrwbl, presentInfo, signaler);
-
 		// Try to do any present mode transitions as late as possible in an attempt
 		// to avoid visual disruptions on any presents already on the queue.
 		if (presentInfo.presentMode != VK_PRESENT_MODE_MAX_ENUM_KHR) {
 			mtlDrwbl.layer.displaySyncEnabledMVK = (presentInfo.presentMode != VK_PRESENT_MODE_IMMEDIATE_KHR);
+		}
+		if (presentInfo.hasPresentTimesInfo) {
+			addPresentedHandler(mtlDrwbl, presentInfo, signaler);
 		}
 		if (presentInfo.desiredPresentTime) {
 			[mtlDrwbl presentAtTime: (double)presentInfo.desiredPresentTime * 1.0e-9];
--- a/MoltenVK/MoltenVK/GPUObjects/MVKImage.h
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKImage.h
@@ -448,7 +448,7 @@ typedef struct  {
 	uint64_t desiredPresentTime;  	// VK_GOOGLE_display_timing desired presentation time in nanoseconds
 	uint32_t presentID;           	// VK_GOOGLE_display_timing presentID
 	VkPresentModeKHR presentMode;	// VK_EXT_swapchain_maintenance1 present mode specialization
-	bool hasPresentTime;      		// Keep track of whether presentation included VK_GOOGLE_display_timing
+	bool hasPresentTimesInfo;      	// Keep track of whether presentation included VK_GOOGLE_display_timing
 } MVKImagePresentInfo;
 
 /** Tracks a semaphore and fence for later signaling. */
--- a/MoltenVK/MoltenVK/GPUObjects/MVKQueue.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKQueue.mm
@@ -799,10 +799,12 @@ MVKQueuePresentSurfaceSubmission::MVKQueuePresentSurfaceSubmission(MVKQueue* que
 		presentInfo.presentableImage = mvkSC->getPresentableImage(pPresentInfo->pImageIndices[scIdx]);
 		presentInfo.presentMode = pPresentModes ? pPresentModes[scIdx] : VK_PRESENT_MODE_MAX_ENUM_KHR;
 		presentInfo.fence = pFences ? (MVKFence*)pFences[scIdx] : nullptr;
-		if (pPresentTimes) {
-			presentInfo.hasPresentTime = true;
-			presentInfo.presentID = pPresentTimes[scIdx].presentID;
-			presentInfo.desiredPresentTime = pPresentTimes[scIdx].desiredPresentTime;
+		if (pPresentTimesInfo) {
+			presentInfo.hasPresentTimesInfo = true;
+			if (pPresentTimes) {
+				presentInfo.presentID = pPresentTimes[scIdx].presentID;
+				presentInfo.desiredPresentTime = pPresentTimes[scIdx].desiredPresentTime;
+			}
 		}
 		mvkSC->setLayerNeedsDisplay(pRegions ? &pRegions[scIdx] : nullptr);
 		_presentInfo.push_back(presentInfo);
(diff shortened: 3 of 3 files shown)


billhollings:
> This PR solves issue #2284.

While the changes in this PR might be useful, and be pulled in, how does this PR fix #2284? 

The changes in this PR would seem to mask the issue for the specific use case in  #2284, but still leave the problem of the `Uncaught selector` unresolved for anyone who is using `VK_GOOGLE_display_timing`.

I'm going to hold off pulling this in until we make some further progress on figuring out #2284.

billhollings (approved):
Thanks for submitting. LGTM as a general improvement.

Adding @cdavis5e as a reviewer.

SRSaunders:
Before you provide final reviews and possibly merge, I would like to raise a point re interpretation of the Vulkan Spec, and encourage "sober second thought".

In MVK 1.2.5 and earlier, the presentation handler was guarded by a condition.  I think this was changed in MVK 1.2.6 based on discussion (with me) in https://github.com/KhronosGroup/MoltenVK/issues/1925#issuecomment-1571317425.  The end result was that calls to [vkGetPastPresentationTimingGOOGLE()](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/vkGetPastPresentationTimingGOOGLE.html) did not require the `VkPresentTimesInfoGOOGLE` struct in the pNext chain during calls to `vkQueuePresentKHR()` for valid data to be returned.  Note the Vulkan spec does not specifically say this is necessary for simply getting past presentation data.  I guess that's why we enabled it all the time.

However, given the more recent discussion this PR now reinstates a guard, but one that is more general and simply looks for the presence of `VkPresentTimesInfoGOOGLE` (and not does require `VkPresentTimeGOOGLE*` to be specified).  This means that `vkGetPastPresentationTimingGOOGLE()` will now require inclusion of `VkPresentTimesInfoGOOGLE` in calls to `vkQueuePresentKHR()` for past presentation data to be collected and available for return.

I am comfortable with this change for a couple of reasons:
1. Performance and footprint: Why collect the information all the time if an application will never use it.  Activating it only when the app really needs it (by including `VkPresentTimesInfoGOOGLE` in `vkQueuePresentKHR()` calls) seems more logical to me.  However, this is potentially a Vulkan Spec issue, and I would not want CTS tests to fail because of it.
2. If there really is a bug in Apple's instrumentation logic with `CaptureMTLDrawable`, then why bother everyone with a superfluous error message by default?  A guard here may help reduce distracting issues caused by a bug coupled with Xcode defaults for debug instrumentation.

billhollings:
> This means that once again `vkGetPastPresentationTimingGOOGLE()` will require inclusion of `VkPresentTimesInfoGOOGLE` in calls to `vkQueuePresentKHR()` for past presentation data to be collected and available for return.

If I understand your concern, with `VK_GOOGLE_display_timing` enabled, requiring `VkPresentTimesInfoGOOGLE` to be present might miss some history timestamps on any presentations where `VkPresentTimesInfoGOOGLE` is not provided?

In that case, what about basing the guard on whether the extension is enabled instead?

```
if (getEnabledExtensions().vk_GOOGLE_display_timing.enabled) {
    addPresentedHandler(mtlDrwbl, presentInfo, signaler);
}
```

Or perhaps move it inside `addPresentedHandler()` and exit the function immediately if the extension is not enabled:

```
void MVKPresentableSwapchainImage::addPresentedHandler(id<CAMetalDrawable> mtlDrawable,
													   MVKImagePresentInfo presentInfo,
													   MVKSwapchainSignaler signaler) {
	if ( !getEnabledExtensions().vk_GOOGLE_display_timing.enabled ) { return; }

	beginPresentation(presentInfo);
	...
```

SRSaunders:
> In that case, what about basing the guard on whether the extension is enabled instead?
> 
> ```
> if (getEnabledExtensions().vk_GOOGLE_display_timing.enabled) {
>     addPresentedHandler(mtlDrwbl, presentInfo, signaler);
> }
> ```

I like this option the best, since it makes it clear that the handler is activated only when the extension is enabled.  I will push this change tomorrow after a bit of testing.

I have also prototyped a minor enhancement to calculate the `VkPastPresentationTimingGOOGLE ::presentMargin` field which up to now has been blank.  I will include that too for your review.

cdavis5e (approved):

MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
```diff
@@ -1552,14 +1552,14 @@ static void signalAndUntrack(const MVKSwapchainSignaler& signaler) {
...
 	[mtlCmdBuff addScheduledHandler: ^(id<MTLCommandBuffer> mcb) {
-
-		addPresentedHandler(mtlDrwbl, presentInfo, signaler);
-
 		// Try to do any present mode transitions as late as possible in an attempt
 		// to avoid visual disruptions on any presents already on the queue.
 		if (presentInfo.presentMode != VK_PRESENT_MODE_MAX_ENUM_KHR) {
 			mtlDrwbl.layer.displaySyncEnabledMVK = (presentInfo.presentMode != VK_PRESENT_MODE_IMMEDIATE_KHR);
 		}
+		if (getEnabledExtensions().vk_GOOGLE_display_timing.enabled) {
```
billhollings, line 1560: In doing another review, I've just realized something that I didn't catch in our earlier discussions here.

`addPresentedHandler(), along with beginPresentation() and endPresentation()` do more than just track `VK_GOOGLE_display_timing`. They also:

- `retain()/release()` the swapchain image, which is important for ensuring the image doesn't get destroyed before the presentation finishes.
- Signal the presentation fence
- Signal presentation semaphores that don't use `MTLEvent`
- Track presentation performance statistics for MoltenVK

See the f0cb31a  commit comments for more discussion.

Removing `hasPresentTime`, and the new calculations for `presentMargin`, are both useful here, but I think the original idea of reintroducing a guard on `addPresentedHandler` may regress several previous sync and race-conditions issue fixes.

billhollings (changes requested):
Per my comment below, I don't think we can reintroduce a guard on `addPresentedHandler()`, after all.

MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
SRSaunders: Thanks for remembering this.  Darn - I should have at least noticed the performance statistics being part of the presentHandler logic.  Glad you caught this.  I guess we will have to live with folks potentially running into the uncaught exception when Xcode GPU frame capture is default on with debug builds.  Unless of course you can find a solution with Apple or some other workaround.

I have reverted the guard, and renamed `startPresentTime` to `beginPresentTime`.  I think this more closely matches the naming nomenclature of the begin/end presentation routines, and differentiates it more clearly from the performance statistics variable `_presentationStartTime`.

billhollings: > I guess we will have to live with folks potentially running into the uncaught exception when Xcode GPU frame capture is default on with debug builds.

TBH...I run GPU capture all the time during dev & testing, and I have never encountered an issue. Nor did the errors make any kind of sense. To me, that original problem feels like a complex environmental issue of some kind (OS, Xcode, app usage of MoltenVK, etc). Until we get some further replaceable data on it, bypassing legit functionality to avoid it seems like dangerous overkill. Let's see what happens from further users.

SRSaunders: I have been able to reproduce it.  Let me see if I can document the conditions to help you with diagnosis.  I will put my comments in the original reported issue.  Hopefully this PR is now refocussed and can proceed.

billhollings (approved):
LGTM now. Thanks!

Issue #2283, opened 2024-07-22
Latest MoltenVK instable on iOS and macOS


aerofly:
Hello,

we tried to use the latest MoltenVK version 1.2.10 in our games. However, this version is quite unstable and crashes a lot compared to the previous release MoltenVK 1.2.9.

We can't pinpoint it down yet on whats causing this. We initially thought it had to do with the introduction of argument buffers, but even setting

`mvkConfig.useMetalArgumentBuffers = false;`

doesn't help. We still get random crashes either in vkQueueSubmit or in mvkUpdateDescriptorSets.

Any idea what new feature or code change could cause this so we can further investigate the possible causes?

aerofly:
We tried to pinpoint the issue with the latest MoltenVK version, but the results are very random. Again MoltenVK 1.2.9 works fine with no crashes in our game, but 1.2.10 just doesn't work.

A few crash cases:

1) One crash happens in 'void MVKSamplerDescriptorMixin::write(MVKDescriptorSetLayoutBinding* mvkDSLBind,...'
We get this error
`_mvkSampler->reportError(VK_ERROR_FEATURE_NOT_PRESENT, "vkUpdateDescriptorSets(): Tried to push an immutable sampler.");
`
2) Then on another occasion we get this:

```
*** Terminating app due to uncaught exception 'NSInvalidArgumentException', reason: '-[GCKeyboardInput gpuAddress]: unrecognized selector sent to instance 0x148108700'
*** First throw call stack:
(
	0   CoreFoundation                      0x00000001856fb2ec __exceptionPreprocess + 176
	1   libobjc.A.dylib                     0x00000001851e2788 objc_exception_throw + 60
	2   CoreFoundation                      0x00000001857ad56c -[NSObject(NSObject) __retain_OA] + 0
	3   CoreFoundation                      0x0000000185664f3c ___forwarding___ + 1580
	4   CoreFoundation                      0x0000000185664850 _CF_forwarding_prep_0 + 96
	5   Aerofly FS 4                        0x0000000101ce8894 _ZN22MVKMetalArgumentBuffer9setBufferEPU19objcproto9MTLBuffer11objc_objectmj + 72
	6   Aerofly FS 4                        0x0000000101d774b4 _ZN19MVKBufferDescriptor5writeEP29MVKDescriptorSetLayoutBindingP16MVKDescriptorSetjjmPKv + 252
[...]

PR #2293, opened 2024-07-26
Add additional defensive code within vkUpdateDescriptorSets().

billhollings:
- ~`MVKDescriptorSet::write()`: Revert to verifying each descriptor has the same type as its descriptor binding, to guard against an app mistakenly running off the end of an array of descriptors.~
- ~`MVKDescriptorSet::hasMetalArgumentBuffer()` ensure _layout is available.~

- MVKDescriptorSet::write() & read(): Refactor descriptor iteration to clarify loop indexing, and to check array limits to guard against an app mistakenly running off the end of an array of descriptors.

### DIFF
--- a/MoltenVK/MoltenVK/GPUObjects/MVKDescriptorSet.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKDescriptorSet.mm
@@ -433,15 +433,17 @@ void MVKDescriptorSet::write(const DescriptorAction* pDescriptorAction,
 	VkDescriptorType descType = mvkDSLBind->getDescriptorType();
 	if (descType == VK_DESCRIPTOR_TYPE_INLINE_UNIFORM_BLOCK_EXT) {
 		// For inline buffers dstArrayElement is a byte offset
-		uint32_t descIdx = _layout->getDescriptorIndex(pDescriptorAction->dstBinding);
-		_descriptors[descIdx]->write(mvkDSLBind, this, pDescriptorAction->dstArrayElement, 0, srcStride, pData);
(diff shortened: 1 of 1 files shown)


cdavis5e (changes requested):
This doesn't seem to have actually helped much with #2283... I'm worried that this could negatively affect the performance of MoltenVK, especially with large descriptor sets. If the bug is in the client, this may not even be necessary.

billhollings:
> This doesn't seem to have actually helped much with #2283... I'm worried that this could negatively affect the performance of MoltenVK, especially with large descriptor sets. If the bug is in the client, this may not even be necessary.

Thanks for pointing that out. Yeah, I agree. I've removed the existing code from here.

I do think there is still some validity in guarding against an app mistakenly running off the end of a descriptor binding array, and into descriptors of the wrong type, during a descriptor set update. So I've modified both write and read to check array limits before looping, which will accomplish the same thing the descriptor type checking, without the cost of doing it on each array element.

In doing so, I also refactored `write()` and `read()` to clarify the logic, and the various indexes being iterated.

cdavis5e (approved):

billhollings:
> I do think there is still some validity in guarding against an app mistakenly running off the end of a descriptor binding array, and into descriptors of the wrong type, during a descriptor set update. So I've modified both write and read to check array limits before looping, which will accomplish the same thing the descriptor type checking, without the cost of doing it on each array element.

Oops. This PR went too far. Vulkan requires the ability to write and copy beyond the size of a descriptor binding and into subsequent bindings of the same type, including for inline buffer bindings. PR #2303 reverts this and adds the same capability for inline buffer bindings.

Issue #2295, opened 2024-07-27
vkQueuePresentKHR often crashes


vaylor27:
About 1 in 10-15 times, when I run my code (even a simple tutorial engine), MoltenVK crashes (upon the first frame) with this error:

```
MVKCrashTest(11214,0x7ff855dd27c0) malloc: Heap corruption detected, free list is damaged at 0x600000124ee0
*** Incorrect guard value: 183628
MVKCrashTest(11214,0x7ff855dd27c0) malloc: *** set a breakpoint in malloc_error_break to debug
```
The error occurs when calling `return disp->QueuePresentKHR(queue, pPresentInfo);`

Using the LLDB Debugger, I have found that `vkQueuePresentKHR` causes this for no discernible reason. I am on the latest release. My hardware is an AMD Radeon 757 Pro 4GB and an Intel(R) Core(TM) i5-7600 CPU @ 3.50GHz. I upgraded the memory to be 32 GB 2400 MHz DDR4. This is a 27-inch 5K iMac from 2017 running macOS ventura 13.6.7

The code works 100% of the time on the same machine when booted into ubuntu and win11. I started up the application 200 times on win11, ubuntu, and macos. On macOS, it failed 23 times. It did not fail on either of the other operating systems. I am using the `VK_LAYER_KHRONOS_validation` layer, and am seeing no errors.

billhollings:
Were you are running with `MVK_CONFIG_SYNCHRONOUS_QUEUE_SUBMITS=0` (disabled)?

If so, PR #2297 may fix this. Please test again with that update, and close this issue if it fixes the problem.

vaylor27:
this issue no longer occurs

lsdh-fei:
@billhollings this bug also exists in the latest  MoltenVK, it crash as i described in https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/8473 ,please check it, thx

billhollings:
> @billhollings this bug also exists in the latest MoltenVK, it crash as i described in [KhronosGroup/Vulkan-ValidationLayers#8473](https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/8473) ,please check it, thx

@aitor-lunarg has created #2321 to cover this new issue.

Issue #2031, opened 2023-10-05
Sometimes crashing during animated layer resize


orion1vi:
Getting crash when addPresentedHandler is called at:

https://github.com/KhronosGroup/MoltenVK/blob/568cc3acc0e2299931fdaecaaa1fc3ec5b4af281/MoltenVK/MoltenVK/GPUObjects/MVKImage.mm#L1321

```
Thread 2 Queue : com.Metal.CompletionQueueDispatch (serial)
#0	0x000000010b2e9678 in unsigned int std::__1::__cxx_atomic_fetch_add<unsigned int>(std::__1::__cxx_atomic_base_impl<unsigned int>*, unsigned int, std::__1::memory_order) at /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX12.1.sdk/usr/include/c++/v1/atomic:1050
#1	0x000000010b2e95d1 in std::__1::__atomic_base<unsigned int, true>::fetch_add(unsigned int, std::__1::memory_order) at /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX12.1.sdk/usr/include/c++/v1/atomic:1719
#2	0x000000010b2e7672 in std::__1::__atomic_base<unsigned int, true>::operator++(int) at /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX12.1.sdk/usr/include/c++/v1/atomic:1748
#3	0x000000010b2e7645 in MVKSwapchain::beginPresentation(MVKImagePresentInfo const&) at /Users/orion/Developer/MoltenVK/MoltenVK/MoltenVK/GPUObjects/MVKSwapchain.mm:248
#4	0x000000010b17f131 in MVKPresentableSwapchainImage::beginPresentation(MVKImagePresentInfo const&) at /Users/orion/Developer/MoltenVK/MoltenVK/MoltenVK/GPUObjects/MVKImage.mm:1403
#5	0x000000010b17ed9c in MVKPresentableSwapchainImage::addPresentedHandler(id<CAMetalDrawable>, MVKImagePresentInfo, MVKSwapchainSignaler) at /Users/orion/Developer/MoltenVK/MoltenVK/MoltenVK/GPUObjects/MVKImage.mm:1384
[...]

PR #2297, opened 2024-07-31
MVKQueue: Make sure waitIdle waits for all commands to be encoded before enqueuing its own waitUntilCompleted.

js6i:
That seems to be causing mysterious crashes with asynchronous queue submissions. I've been debugging one where it'd crash the encoding thread because we destroyed the swap chain right after calling `vkQueueWaitIdle`, and it wasn't done encoding the Present.

For what it's worth in my case the bug bisected to f0cb31a12b59f05177f07ab5a46bc9084ba5fbc9.

Ping #2295, possibly others..

### DIFF
--- a/MoltenVK/MoltenVK/GPUObjects/MVKQueue.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKQueue.mm
@@ -85,7 +85,16 @@ VkResult MVKQueue::submit(MVKQueueSubmission* qSubmit) {
 	// The submissions will ensure a misconfiguration will be safe to execute.
 	VkResult rslt = qSubmit->getConfigurationResult();
 	if (_execQueue) {
-		dispatch_async(_execQueue, ^{ execute(qSubmit); } );
+		std::unique_lock lock(_execQueueMutex);
+		_execQueueJobCount++;
+
+		dispatch_async(_execQueue, ^{
+			execute(qSubmit);
+
+			std::unique_lock execLock(_execQueueMutex);
+			if (!--_execQueueJobCount)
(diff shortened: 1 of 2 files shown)


js6i:
#2031 looks possibly related.

billhollings (approved):
Very nice catch! Thanks for submitting this!

LGTM.

Just going to give @cdavis5e a chance to chime in, before pulling this in.

MoltenVK/MoltenVK/GPUObjects/MVKQueue.mm
```diff
@@ -85,7 +85,16 @@
...
-		dispatch_async(_execQueue, ^{ execute(qSubmit); } );
+		std::unique_lock lock(_execQueueMutex);
+		_execQueueJobCount++;
+
+		dispatch_async(_execQueue, ^{
+			execute(qSubmit);
+
+			std::unique_lock execLock(_execQueueMutex);
+			_execQueueJobCount--;
+			_execQueueConditionVariable.notify_all();
```
cdavis5e, line 96: The wait code below checks that the job count is zero before actually exiting the wait. You could probably avoid spurious wakeups by only notifying when `_execQueueJobCount` becomes zero. (Or did you try that and experience problems?)

cdavis5e (approved):
One small optimization I noticed.

MoltenVK/MoltenVK/GPUObjects/MVKQueue.mm
js6i: No problems, I think that's good.

PR #2298, opened 2024-08-01
iOS: Support storage images in Metal argument buffers.

billhollings:
On iOS GPUs with Tier1 argument buffer support, Metal does not support writable textures in argument buffers. MoltenVK and SPIRV-Cross were addressing this differently.

- Set `CompilerMSL::Options::ArgumentBuffersTier` to align behaviour between MoltenVK and SPIRV-Cross.
- On iOS Tier1 GPUs, if a descriptor set contains a storage image, don't use a Metal argument buffer for that descriptor set.

### DIFF
--- a/MoltenVK/MoltenVK/GPUObjects/MVKDescriptorSet.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKDescriptorSet.mm
@@ -300,7 +300,7 @@ bool MVKDescriptorSetLayout::populateBindingUse(MVKBitArray& bindingUse,
 }
 
 bool MVKDescriptorSetLayout::isUsingMetalArgumentBuffers() {
-	return MVKDeviceTrackingMixin::isUsingMetalArgumentBuffers() && !_isPushDescriptorLayout;
+	return MVKDeviceTrackingMixin::isUsingMetalArgumentBuffers() && _canUseMetalArgumentBuffer;
 };
 
 // Returns an autoreleased MTLArgumentDescriptor suitable for adding an auxiliary buffer to the argument buffer.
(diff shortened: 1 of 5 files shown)


MoltenVK/MoltenVK/GPUObjects/MVKDescriptorSet.mm
```diff
@@ -354,6 +354,7 @@ static void populateAuxBuffer(mvk::SPIRVToMSLConversionConfiguration& shaderConf
                                                const VkDescriptorSetLayoutCreateInfo* pCreateInfo) : MVKVulkanAPIDeviceObject(device) {
 
 	_isPushDescriptorLayout = mvkIsAnyFlagEnabled(pCreateInfo->flags, VK_DESCRIPTOR_SET_LAYOUT_CREATE_PUSH_DESCRIPTOR_BIT_KHR);
+	_canUseMetalArgumentBuffer = !_isPushDescriptorLayout;	// Push constants don't use argument buffers
```
cdavis5e, line 357:
```suggestion
	_canUseMetalArgumentBuffer = !_isPushDescriptorLayout;	// Push constants/descriptors don't use argument buffers
```

Granted, push constants don't either, but this is also for `VK_KHR_push_descriptor`.

cdavis5e (approved):
Just one minor nit.

MoltenVK/MoltenVK/GPUObjects/MVKDescriptorSet.mm
billhollings: Thanks. Good catch. Done.

Issue #2301, opened 2024-08-05
Support more 32-bit floating-point pixel format caps on iOS


lexaknyazev:
It seems that some format caps are not aligned with the Metal docs. However, the docs themselves are not very precise.

- `RG32Float` and `RGBA32Float` must support MSAA on Apple7 GPUs and newer.
- `R32Float`, `RG32Float`, and `RGBA32Float` must be filterable and resolvable on Apple7 GPUs and newer if `mtlDevice.supports32BitFloatFiltering` is true. This configuration includes iPad models with M-series chips, Vision Pro, and iPhone models with A17 Pro. The latter is technically covered by #2107 but it would redundant if the proper detection is implemented.
- `RGBA32Float` must support blending on Apple7 GPUs and newer if `mtlDevice.supports32BitFloatFiltering` is true. This is not mentioned in the docs but has been confirmed with the devices.

billhollings:
PR #2308 should fix these issues. Please retest with that code, and close this issue if the problem is fixed.

Issue #2302, opened 2024-08-05
Enable blending cap for RGB9E5Float


lexaknyazev:
The Vulkan spec has been clarified to disallow different color write masks for RGB color channels of `E5B9G9R9` attachments.

See `VUID-VkGraphicsPipelineCreateInfo-None-09043` and `VUID-*-None-09116`.

Past discussions: https://github.com/KhronosGroup/Vulkan-Portability/issues/28 and https://github.com/KhronosGroup/MoltenVK/pull/1399.

billhollings:
The original change was triggered by avoiding a large number of CTS failures, so we'll have to check whether CTS has adapted to this update.

PR #2308, opened 2024-08-14
Refactor defining MTLPixelFormat and GPU capabilities.

billhollings:
- Add `MVKMTLDeviceCapabilities` to encapsulate Metal GPU family capabilities, and use it in `MVKPhysicalDevice` and `MVKPixelFormats` to identify GPU capabilities.
- Remove all references to `MTLGPUFamily` and `MTLFeatureSet` outside `MVKMTLDeviceCapabilities`.
- Blend or select AppleGPU and MacGPU format capabilities.
- Simplify `MVKPixelFormats::modifyMTLFormatCapabilities()` based on `MVKMTLDeviceCapabilities`.
- Add guard code to ensure `MTLPixelFormatInvalid` & `MTLVertexFormatInvalid` capabilities are not overwritten by other formats that don't exist on some platforms.
- Update `MTLPixelFormat` capabilities based on latest Metal docs.
- Remove some variations tied to very older OS versions, or `MTLFeatureSets` that have the same GPU family, including setting MSL 2.0 as the earliest supported version.
- Enable `VkPhysicalDeviceFeatures::vertexPipelineStoresAndAtomics` & `fragmentStoresAndAtomics` on all platforms.
- During startup, only log the highest GPU family.

Fixes issue #2301.

### DIFF
--- a/MoltenVK/MoltenVK/GPUObjects/MVKPixelFormats.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKPixelFormats.mm
@@ -137,7 +137,6 @@ using namespace std;
 #   define MTLPixelFormatX24_Stencil8               MTLPixelFormatInvalid
 #endif
 
-
 #if MVK_TVOS
 #       define MTLPixelFormatASTC_4x4_HDR           MTLPixelFormatInvalid
 #       define MTLPixelFormatASTC_5x4_HDR           MTLPixelFormatInvalid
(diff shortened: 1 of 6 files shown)


MoltenVK/MoltenVK/GPUObjects/MVKPixelFormats.mm
```diff
@@ -1325,704 +1343,253 @@
...
+
+	setMTLPixFmtCapsIf( iosOnly2, RGB10A2Unorm, RFCMRB );
+	setMTLPixFmtCapsIf( iosOnly2, RGB10A2Uint, RCM );
+	setMTLPixFmtCapsIf( iosOnly2, RG11B10Float, RFCMRB );
+	setMTLPixFmtCapsIf( iosOnly2, RGB9E5Float, RFCMRB );
+
+	// Blending is actually supported for RGB9E5Float, but format channels cannot
+	// be individually write-enabled during blending on macOS. Disabling blending
+	// on macOS is the least-intrusive way to handle this in a Vulkan-friendly way.
+	disableMTLPixFmtCapsIfGPU( Mac1, RGB9E5Float, Blend);
```
cdavis5e, line 1509: Somebody had recently posted that this is now confirmed to be true in Vulkan as well. We should now be able to enable this.

cdavis5e (changes requested):
Sorry for the delay. GitHub went down for a bit yesterday.

Since you're bumping the minimum OS version needed to use MoltenVK, you should probably update the README and User's Guide.

MoltenVK/MoltenVK/GPUObjects/MVKPixelFormats.mm
billhollings: I think you're referring to #2302. 

It doesn't look like CTS has been modified for this yet. With latest CTS, we're still seeing several hundred blend tests failing.

We can track it in #2302 to support this once CTS is updated.

billhollings:
> Sorry for the delay. GitHub went down for a bit yesterday.

No worries. Thanks for having a look, and for your feedback.
 
> Since you're bumping the minimum OS version needed to use MoltenVK, you should probably update the README and User's Guide.

The changes embodied here in terms of older OS's should be covered by the existing minimums in `MoltenVK_Runtime_UserGuide`: macOS 10.15 & iOS/tvOS 13.0.

Do you see anything particular that doesn't get covered by those minimums?

MoltenVK/MoltenVK/GPUObjects/MVKPixelFormats.mm
cdavis5e: Very well.

cdavis5e:
> The changes embodied here in terms of older OS's should be covered by the existing minimums in `MoltenVK_Runtime_UserGuide`: macOS 10.15 & iOS/tvOS 13.0.

Must've missed that...

cdavis5e (approved):

PR #2311, opened 2024-08-19
Fix debug utils messenger leak

aitor-lunarg:
Related https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/8408


billhollings (approved):
Uggh! Good catch! Thanks for the fix.

PR #2317, opened 2024-08-27
Minor debugging enhancements.

billhollings:
- On GPU error, log indicate whether device has been resumed or lost, and report as either warning or error, respectively.
- On `VkDevice` loss, stop device-scope GPU capture.
- In debug mode, include address of Metal object in Metal object label.
- Replace `setLabelIfNotNil()` with `setMetalObjectLabel()` and move inside `MVKVulkanAPIObject` to allow access to config info.
- When logging contents of `MVKDescriptorPool`, only list descriptor types with allocations.
- Add `@autoreleasepool` around complex `NSString` creation.

Expansion and replacement of PR #2305.

### DIFF
--- a/MoltenVK/MoltenVK/GPUObjects/MVKVulkanAPIObject.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKVulkanAPIObject.mm
@@ -33,6 +33,20 @@ VkResult MVKVulkanAPIObject::setDebugName(const char* pObjectName) {
 	return VK_SUCCESS;
 }
 
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wobjc-method-access"
+void MVKVulkanAPIObject::setMetalObjectLabel(id mtlObj, NSString* label) {
+	if ( !label ) { return; }
+
+	// If debug mode is enabled, append the object address.
+	if (getMVKConfig().debugMode) {
+		[mtlObj setLabel: [label stringByAppendingFormat: @" (%p)", mtlObj]];
+	} else {
+		[mtlObj setLabel: label];
+	}
+}
+#pragma clang diagnostic pop
+
 MVKVulkanAPIObject* MVKVulkanAPIObject::getMVKVulkanAPIObject(VkDebugReportObjectTypeEXT objType, uint64_t object) {
 	void* pVkObj = (void*)object;
 	switch (objType) {
--- a/MoltenVKShaderConverter/MoltenVKShaderConverterTool/OSSupport.mm
+++ b/MoltenVKShaderConverter/MoltenVKShaderConverterTool/OSSupport.mm
@@ -108,13 +108,13 @@ bool mvk::compile(const string& mslSourceCode,
 		mslVerEnum = MTLLanguageVersion1_1;
 	}
 
-	if ( !mslVerEnum ) {
-		errMsg = [NSString stringWithFormat: @"%d.%d.%d is not a valid MSL version number on this device",
-				  mslVersionMajor, mslVersionMinor, mslVersionPoint].UTF8String;
-		return false;
-	}
-
 	@autoreleasepool {
+		if ( !mslVerEnum ) {
+			errMsg = [NSString stringWithFormat: @"%d.%d.%d is not a valid MSL version number on this device",
+					  mslVersionMajor, mslVersionMinor, mslVersionPoint].UTF8String;
+			return false;
+		}
+
 		NSArray* mtlDevs = [MTLCopyAllDevices() autorelease];
 		if (mtlDevs.count == 0) {
 			errMsg = "Could not retrieve MTLDevice to compile shader.";
--- a/MoltenVK/MoltenVK/Commands/MVKCmdTransfer.mm
+++ b/MoltenVK/MoltenVK/Commands/MVKCmdTransfer.mm
@@ -615,7 +615,7 @@ void MVKCmdBlitImage<N>::encode(MVKCommandEncoder* cmdEncoder, MVKCommandUse com
                     mtlStencilAttDesc.slice = mvkIBR.region.dstSubresource.baseArrayLayer + layIdx;
                 }
                 id<MTLRenderCommandEncoder> mtlRendEnc = [cmdEncoder->_mtlCmdBuffer renderCommandEncoderWithDescriptor: mtlRPD];
-                setLabelIfNotNil(mtlRendEnc, mvkMTLRenderCommandEncoderLabel(commandUse));
+				cmdEncoder->_cmdBuffer->setMetalObjectLabel(mtlRendEnc, mvkMTLRenderCommandEncoderLabel(commandUse));
 
                 float zIncr;
                 if (blitKey.srcMTLTextureType == MTLTextureType3D) {
@@ -893,7 +893,7 @@ void MVKCmdResolveImage<N>::encode(MVKCommandEncoder* cmdEncoder) {
 			mtlRPD.renderTargetArrayLengthMVK = rslvSlice.dstSubresource.layerCount;
 		}
 		id<MTLRenderCommandEncoder> mtlRendEnc = [cmdEncoder->_mtlCmdBuffer renderCommandEncoderWithDescriptor: mtlRPD];
-		setLabelIfNotNil(mtlRendEnc, mvkMTLRenderCommandEncoderLabel(kMVKCommandUseResolveImage));
+		cmdEncoder->_cmdBuffer->setMetalObjectLabel(mtlRendEnc, mvkMTLRenderCommandEncoderLabel(kMVKCommandUseResolveImage));
 
 		[mtlRendEnc pushDebugGroup: @"vkCmdResolveImage"];
 		[mtlRendEnc popDebugGroup];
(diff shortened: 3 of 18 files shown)


MoltenVK/MoltenVK/GPUObjects/MVKVulkanAPIObject.mm
```diff
@@ -27,12 +27,31 @@
...
 	return VK_SUCCESS;
 }
 
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wobjc-method-access"
+void MVKVulkanAPIObject::setMetalObjectLabel(id mtlObj, NSString* label) {
+	if ( !label ) { return; }
+
+	// If debug mode is enabled, append the object address.
+	// Since doing so might internally create other Metal objects, so catch them in an autorelease pool.
```
cdavis5e, line 44:
```suggestion
	// Doing so might internally create other Metal objects, so catch them in an autorelease pool.
```

Using both "since" and "so" in the same sentence like that feels awkward.

MoltenVK/MoltenVK/GPUObjects/MVKVulkanAPIObject.mm
```diff
@@ -27,12 +27,31 @@
...
+	if (getMVKConfig().debugMode) {
+		@autoreleasepool {
+			[mtlObj setLabel: [label stringByAppendingFormat: @" (%p)", mtlObj]];
+		}
+	} else {
+		[mtlObj setLabel: label];
```
cdavis5e, line 50: Should this call also be wrapped in an autorelease pool? I fail to see how calling `-[NSString stringByAppendingFormat:]` would result in Metal creating more objects, though I understand that the string needs to be released once it's been set as the label. I _could_ see, however, `-setLabel:` triggering the creation of Metal objects, particularly with one of the debug Metal runtimes loaded.

MoltenVKShaderConverter/MoltenVKShaderConverterTool/OSSupport.mm
```diff
@@ -108,13 +108,13 @@
...
-		return false;
-	}
-
 	@autoreleasepool {
+		if ( !mslVerEnum ) {
+			errMsg = [NSString stringWithFormat: @"%d.%d.%d is not a valid MSL version number on this device",
+					  mslVersionMajor, mslVersionMinor, mslVersionPoint].UTF8String;
+			return false;
+		}
+
```
cdavis5e, line 117: This will cause the error message string to be released and destroyed immediately before the function returns, because the autorelease pool will be torn down at that point. (ARC is off in MoltenVK, remember?) You may need to add a `-retain` call here... which would, in turn, entail adding `-release` calls to this function's callers.

cdavis5e (changes requested):

MoltenVKShaderConverter/MoltenVKShaderConverterTool/OSSupport.mm
cdavis5e: Wait, will it? I was operating on the assumption that the `-[NSString string...]` family of methods return autoreleased, which is part of why they need to be called in an autorelease pool in the first place, but I might be mistaken.

auspice-app:
Thanks for pushing this forward and sorry for my lack of response, I was out last week and didn't get to it! I definitely glossed over the setLabel change in the original submission..
It would be convenient to have an autorelease pool assigned to the Metal command buffer, put the strings in there and drain on completion, but I can't find a way to do it..?

js6i:
Thanks for pushing this forward, I was out last week and didn't get to it! I definitely glossed over the label part in the original submission.
It would be convenient to have an autoreleasepool assigned to the Metal command buffer, put the strings there and drain it on completion, but I can't find a way to do it..?

cdavis5e:
> Thanks for pushing this forward, I was out last week and didn't get to it! I definitely glossed over the label part in the original submission. It would be convenient to have an autoreleasepool assigned to the Metal command buffer, put the strings there and drain it on completion, but I can't find a way to do it..?

The `@autoreleasepool` block is just syntactic sugar for the `NSAutoreleasePool` object:

```objc
{
    NSAutoreleasePool *pool = [[NSAutoreleasePool alloc] init];
    // contents of block...
    [pool release];
}
```

For a long time creating an `NSAutoreleasePool` object explicitly was the _only_ way to create an autorelease pool.

`NSAutoreleasePool` has a `-drain` method that releases all objects registered with the pool, without destroying the pool object itself. The intended use case is for a loop, like the main run loop of an application. Instead of creating and destroying an autorelease pool on every iteration, you create an autorelease pool before the loop, call `-drain` at the end of every iteration, and then, when the loop terminates, you release the pool. Theoretically, it shouldn't be difficult to adapt this to a command buffer; the one problem is that autorelease pools exist in a stack, and creating a new pool pushes it onto the stack and causes all calls to `-[NSObject autorelease]` to go to that pool. But, it turns out, `NSAutoreleasePool` also has explicit `-addObject:` and `+addObject:` methods. All `-[NSObject autorelease]` actually does is call `+[NSAutoreleasePool addObject:]`, which in turn, invokes `-[NSAutoreleasePool addObject:]` on the pool on top of the stack. Any objects we create, then, could be explicitly placed in the autorelease pool associated with a command buffer. But it won't help with other frameworks and dylibs that call `-[NSObject autorelease]`. For that, the only thing I could think of is "swizzling"* `+[NSAutoreleasePool addObject:]` so that, if we're currently encoding or executing a particular command buffer, the call to `-[NSAutoreleasePool addObject:]` goes to the pool associated with that command buffer.

Anyway, this all sounds really complicated, and I'm not sure it's worth it, and why did I write all this again?

\* Replacing a method's `IMP`lementation with one of our own. You use `class_getClassMethod()` or `class_getInstanceMethod()` on the class and selector of the method you want to replace; then you optionally save the old `IMP` with `method_getImplementation()` and then insert the new one with `method_setImplementation()`. If you want to use a closure block instead of writing out a function with the signature `void (id, SEL, id)`, you can use `imp_implementationWithBlock()` to produce an `IMP` from the block.

MoltenVKShaderConverter/MoltenVKShaderConverterTool/OSSupport.mm
cdavis5e: Wait, it doesn't even matter, because you get it as a `UTF8String` right away and then copy it into the `errMsg`. Sorry about that.

js6i:
>For that, the only thing I could think of is "swizzling"* `+[NSAutoreleasePool addObject:]` so that, if we're currently encoding or executing a particular command buffer, the call to `-[NSAutoreleasePool addObject:]` goes to the pool associated with that command buffer.
> 
> Anyway, this all sounds really complicated, and I'm not sure it's worth it, and why did I write all this again?
> 
> * Replacing a method's `IMP`lementation with one of our own. You use `class_getClassMethod()` or `class_getInstanceMethod()` on the class and selector of the method you want to replace; then you optionally save the old `IMP` with `method_getImplementation()` and then insert the new one with `method_setImplementation()`. If you want to use a closure block instead of writing out a function with the signature `void (id, SEL, id)`, you can use `imp_implementationWithBlock()` to produce an `IMP` from the block.

I was thinking of just using the pool for these label strings, and maybe some other selected allocations (not sure if there are any that would benefit from it). We could `addObject:` them if we allocate the string with something that doesn't put it in the pool that's currently on top of the stack, or maybe just `retain`, as there's no `pushPool`/`popPool` methods from what I could tell to guide where the autorelease objects go.
I'm assuming the lifetime of these labels should be at most the same as command buffer's, and keeping a pool seems better than e.g. piling callbacks for each allocated string.. not ideal but good enough for a debug path.

MoltenVK/MoltenVK/GPUObjects/MVKVulkanAPIObject.mm
billhollings: My intention is to avoid creating an autorelease pool unless absolutely necessary, since `setLabel:` can be called in some performance-sensitive areas.

I expect that, internally, `setLabel:` simply sets the label property (ivar) and sends `retain` to the label object. I don't see where it would be creating more objects by doing that. Can you elaborate on that?

With `stringByAppendingFormat:`, I expect a bunch of internal allocs to parse the variable input, convert it to strings, and build the larger string. If this work is done in C, not a problem, but if done in Obj-C, it would result in a bunch of temp Obj-C objects being created internally.

MoltenVK/MoltenVK/GPUObjects/MVKVulkanAPIObject.mm
billhollings: Thanks. Fixed.

MoltenVKShaderConverter/MoltenVKShaderConverterTool/OSSupport.mm
billhollings: Yes, the `-[NSString string...]` family returns an autoreleased `NSString` object, so it will be destroyed by the autorelease pool.

However, the subtlety here is that  `errMsg` is a `std::string`, and the assignment will copy the internal `UTF8String` from the `NSString` object into the `std::string`.

billhollings:
The `NSAutoreleasePool` hierarchy is inherently tied to a thread, so it's not clear that we could use that for a particular object, like the command buffers or encoders. If they are accessed only from single thread, we can just wrap the higher calls in `@autoreleasepool`. If they're accessed from multiple threads




I've pushed an update.

I've realized

cdavis5e (approved):

Issue #2206, opened 2024-04-13
Crash with partially bound texture arrays with variable descriptor count


mbechard:
If I have a texture array that has been sized 10, but I'm only providing 8 elements in a call to vkUpdateDescriptorSets(), it doesn't seem like the other elements in the array in MVKDescriptorSet get cleared out. This can cause a crash with deleted resources, since bindings for elements 8 and 9 will still try to get bound in `MVKDescriptorSetLayoutBinding::bind`. That function loops over the entire length of the variableDescriptorCount, and doesn't seem to account for partially bound sizes.
Happy to do a PR, just looking for confirmation my assumptions are correct.
I would think the solution is to call `MVKImageDescriptor::reset()` on every element of the array that isn't set in the Update. Is that correct?
Thanks

mbechard:
Thinking about this more. I think when a MVK resource is destroyed it needs to know all of the partially bound descriptors that reference it, and clears out those references. Otherwise dangling references to that resource can occur when converting the MVK state into Metal calls.
My test case only uses variable sized descriptors, but I'm thinking this crash would affect statically sized descriptors as well that are partially bound.

mbechard:
It works fine on Windows for all GPUs there. There are two crashes so far I've seen that show up.
Here is the call stack from the first crash. The mtlTexture passed into setFragmentTexture is invalid.
```
#0	0x00000001a23e5834 in objc_msgSend ()
#1	0x00000001a2fcd6e8 in -[MTLDebugRenderCommandEncoder setFragmentTexture:atIndex:] ()
#2	0x00000001187fe6f8 in MVKGraphicsResourcesCommandEncoderState::encodeImpl(unsigned int)::$_20::operator()(MVKCommandEncoder*, MVKMTLTextureBinding&) const at /Users/malcolm/dev/devel/MoltenVK/MoltenVK/MoltenVK/Commands/MVKCommandEncoderState.mm:1110
[...]

PR #2320, opened 2024-08-27
Fix race condition when VkImage destroyed while used by descriptor.

billhollings:
- Retain `MVKImage` in `MVKImageView`.

- Fixes issue #2206.

### DIFF
--- a/Docs/Whats_New.md
+++ b/Docs/Whats_New.md
@@ -27,6 +27,7 @@ Released TBD
   end of a descriptor binding count, including inline uniform block descriptors.
 - Update `VkFormat` capabilities based on latest Metal docs.
 - Fix rendering issue with render pass that immediately follows a kernel dispatch.
+- Fix race condition when `VkImage` destroyed while used by descriptor.
 - Ensure all MoltenVK config info set by `VK_EXT_layer_settings` is used.
 - Move primitive-restart-disabled warning from renderpass to pipeline creation, to reduce voluminous log noise.
 - iOS: Support storage images in _Metal_ argument buffers.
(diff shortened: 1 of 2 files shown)


mbechard:
Does the descriptor not retain the MVKImageView, which would mean images won't be cleaned up unless descriptors referencing them are also cleaned up?

billhollings:
> Does the descriptor not retain the MVKImageView, which would mean images won't be cleaned up unless descriptors referencing them are also cleaned up?

Yes. The image view is retained by the descriptor and the image is retained by the image view. That way, when the descriptor binds to the encoder, it has access to the `MTLTexture`.

mbechard:
I agree this is likely the only fix for this problem, but it does result in a spec/compatibility change, since GPU memory isn't clean up anymore unless descriptors are cleaned up. Likely needs to be documented somewhere?

cdavis5e (approved):
We're already using command buffers with unretained references, so that's not it... I can't think of any other way to solve this.

mbechard:
Could a weak_ptr workflow work? When the descriptor is bound it obtains strong references to the resources (if they are still valid), and releases then when the command buffer is done executing?

billhollings:
> Could a `weak_ptr` workflow work? When the descriptor is bound it obtains strong references to the resources (if they are still valid), and releases then when the command buffer is done executing?

Using `weak_ptr` paradigm might work when holding an image-view in a descriptor, but I don't think it would help for this issue. Basically, here we are saying, if an image-view lives on, the associated image must also live on. 

A `shared_ptr` paradigm could have been used from the start, as an alternative of the `retain()/release()` mixin we use.

Issue #2323, opened 2024-09-02
PushDescriptorWithTemplate has broken size computation


HansKristian-Work:
```
// Work out how big the memory block in pData is.
    const VkDescriptorUpdateTemplateEntry* pEntry =
        _descUpdateTemplate->getEntry(_descUpdateTemplate->getNumberOfEntries()-1);
    size_t size = pEntry->offset;
    // If we were given a stride, use that; otherwise, assume only one info
    // struct of the appropriate type.
    if (pEntry->stride)
        size += pEntry->stride * pEntry->descriptorCount;
```

This assumes that the last element in the template update maps to the highest address. This is a flawed assumption, and breaks Granite. Granite sorts the updates by descriptor type, and push descriptors segfaults.

Also, there's a worrying malloc every push, which seems very suboptimal.

billhollings:
PR #2327 should fix both of these issues. Please retest and close this issue if solved.

warmenhoven:
Worked for me. Thanks!

cdavis5e:
I could've sworn that I had read in the spec that the entries had to be sorted by address.

/me goes to read the spec...

I guess not. I must've been mistaken when I implemented that functionality.

billhollings:
> Worked for me. Thanks!

Great. Closing now.

PR #2327, opened 2024-09-05
Fixes to vkCmdPushDescriptorSetWithTemplateKHR().

billhollings:
- Fix crash when entries in `VkDescriptorUpdateTemplateCreateInfo` are not sorted by offset.
- `MVKCmdPushDescriptorSetWithTemplate` allocate new data memory only if existing memory allocation is too small.
- `mvkCopy()` support void pointers, and copy a single value directly.

Fixes issue #2323.

### DIFF
--- a/MoltenVK/MoltenVK/Commands/MVKCmdPipeline.mm
+++ b/MoltenVK/MoltenVK/Commands/MVKCmdPipeline.mm
@@ -478,49 +478,20 @@ VkResult MVKCmdPushDescriptorSetWithTemplate::setContent(MVKCommandBuffer* cmdBu
 														 uint32_t set,
 														 const void* pData) {
 	if (_pipelineLayout) { _pipelineLayout->release(); }
-
-	_descUpdateTemplate = (MVKDescriptorUpdateTemplate*)descUpdateTemplate;
 	_pipelineLayout = (MVKPipelineLayout*)layout;
-	_set = set;
-
 	_pipelineLayout->retain();
+	_set = set;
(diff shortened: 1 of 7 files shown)


MoltenVK/MoltenVK/Commands/MVKCmdPipeline.mm
```diff
@@ -478,49 +478,20 @@
...
-			size += sizeof(VkBufferView);
-			break;
-
-		default:
-			break;
+	size_t oldSize = _dataSize;
+	_dataSize = _descUpdateTemplate->getSize();
+	if (_dataSize > oldSize) {
+		free(_pData);
+		_pData = malloc(_dataSize);
```
cdavis5e, line 490: Since you are now using `malloc(3)` instead of `operator new[]`, you should now be able to use `realloc(3)` to resize the memory block, instead of freeing and `malloc`ing a new one. `realloc(3)` is documented to act like `malloc(3)` when the input heap pointer is `nullptr`, so you don't have to handle that case specially either.

cdavis5e (approved):
Just one small thing.

MoltenVK/MoltenVK/Commands/MVKCmdPipeline.mm
billhollings: TBH...I had forgotten about the option of using `ralloc()`.

However, looking at it now, since we don't need to retain the old contents, `realloc()` will almost certainly be [slower](https://stackoverflow.com/a/39562813/1678054) when it can't grow in place, because of the internal copy.

PR #2132, opened 2024-01-19
Fix unresolvable layered compute resolve

aitor-lunarg:
Implementation did not correctly handle array textures when resolving unresolvable textures like r8uint formats. Add kernel modules to correctly resolve through the compute pass.

Fixes CTS failures in test families:
`dEQP-VK.api.image_clearing.core.clear_color_attachment.*`
`dEQP-VK.renderpass.suballocation.multisample_resolve.*`

PR #2330, opened 2024-09-09
Fixes to enable running on tvOS 13

warmenhoven:
Two parts:

* SPIRV-Cross uses std::stable_sort, and on tvOS using the new linker, this causes a crash: https://forums.developer.apple.com/forums/thread/742470. A workaround is to use the old linker with the flag `-ld_classic`.
* #2132 causes compilation problems on tvOS 13 because `texture2d_ms_array` is not available.

### DIFF
--- a/MoltenVK/MoltenVK.xcodeproj/project.pbxproj
+++ b/MoltenVK/MoltenVK.xcodeproj/project.pbxproj
@@ -2561,6 +2561,7 @@
 				GENERATE_INFOPLIST_FILE = YES;
 				LD_DYLIB_INSTALL_NAME = "@rpath/${PRODUCT_NAME}.framework/${PRODUCT_NAME}";
 				OTHER_LDFLAGS = (
+					"-ld_classic",
 					"-all_load",
 					"-w",
 				);
@@ -2574,6 +2575,7 @@
 				GENERATE_INFOPLIST_FILE = YES;
 				LD_DYLIB_INSTALL_NAME = "@rpath/${PRODUCT_NAME}.framework/${PRODUCT_NAME}";
 				OTHER_LDFLAGS = (
+					"-ld_classic",
 					"-all_load",
 					"-w",
 				);
--- a/MoltenVK/MoltenVK/Commands/MVKCommandPipelineStateFactoryShaderSource.h
+++ b/MoltenVK/MoltenVK/Commands/MVKCommandPipelineStateFactoryShaderSource.h
@@ -163,6 +163,7 @@ kernel void cmdResolveColorImage2DFloat(texture2d<float, access::write> dst [[ t
     dst.write(src.read(pos, 0), pos);                                                                           \n\
 }                                                                                                               \n\
                                                                                                                 \n\
+#if __HAVE_TEXTURE_2D_MS_ARRAY__                                                                                \n\
 kernel void cmdResolveColorImage2DFloatArray(texture2d_array<float, access::write> dst [[ texture(0) ]],        \n\
                                              texture2d_ms_array<float, access::read> src [[ texture(1) ]],      \n\
                                              uint2 pos [[thread_position_in_grid]]) {                           \n\
@@ -170,6 +171,7 @@ kernel void cmdResolveColorImage2DFloatArray(texture2d_array<float, access::writ
         dst.write(src.read(pos, i, 0), pos, i);                                                                 \n\
     }                                                                                                           \n\
 }                                                                                                               \n\
+#endif                                                                                                          \n\
                                                                                                                 \n\
 kernel void cmdResolveColorImage2DUInt(texture2d<uint, access::write> dst [[ texture(0) ]],                     \n\
                                        texture2d_ms<uint, access::read> src [[ texture(1) ]],                   \n\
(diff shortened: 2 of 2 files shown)


MoltenVK/MoltenVK.xcodeproj/project.pbxproj
```diff
@@ -2695,6 +2695,7 @@
 				MACOSX_DEPLOYMENT_TARGET = 10.15;
 				MARKETING_VERSION = "${CURRENT_PROJECT_VERSION}";
 				MTL_ENABLE_DEBUG_INFO = YES;
+				OTHER_LDFLAGS = "-ld_classic";
```
cdavis5e, line 2698: I'm worried that this could cause problems with older versions of Xcode. Unless Bill is willing to drop support for those old versions, there needs to be some sort of version check here.

MoltenVK/MoltenVK.xcodeproj/project.pbxproj
```diff
@@ -2768,6 +2769,7 @@
 				MACOSX_DEPLOYMENT_TARGET = 10.15;
 				MARKETING_VERSION = "${CURRENT_PROJECT_VERSION}";
 				MTL_ENABLE_DEBUG_INFO = NO;
+				OTHER_LDFLAGS = "-ld_classic";
```
cdavis5e, line 2772: Ditto.

MoltenVK/MoltenVK/Commands/MVKCommandResourceFactory.mm
```diff
@@ -551,6 +551,10 @@ static void getSwizzleString(char swizzleStr[4], VkComponentMapping vkMapping) {
 id<MTLComputePipelineState> MVKCommandResourceFactory::newCmdResolveColorImageMTLComputePipelineState(MVKFormatType type,
 																									  MVKVulkanAPIDeviceObject* owner,
 																									  bool isTextureArray) {
+#if MVK_TVOS
+    if (isTextureArray && !mvkOSVersionIsAtLeast(14))
+        isTextureArray = false;
+#endif
```
cdavis5e, line 557: At first, this looks wrong. If a 2D multisampled array texture were passed here in this case, we'd have to resolve each element of the array individually.

Then I realized that this shouldn't even be reached in that case, because the client **must** (in the normative, [RFC 2119][1] sense) not create a 2D multisampled array image if that image type isn't supported, or the ICD is permitted to make demons fly out of the client's nose. You could probably delete this code.

[1]: https://www.ietf.org/rfc/rfc2119.txt

cdavis5e (changes requested):

MoltenVK/MoltenVK.xcodeproj/project.pbxproj
billhollings: @cdavis5e How "old" are you thinking that might be problematic?

warmenhoven: I've pushed a change that moves this to be tvOS-specific as I haven't seen the same problem on iOS, and currently MoltenVK doesn't compile for tvOS using Xcode 14 anyway. If there's something that's preferred I'm happy to change this.

MoltenVK/MoltenVK/Commands/MVKCommandResourceFactory.mm
warmenhoven: Removed, thanks!

MoltenVK/MoltenVK.xcodeproj/project.pbxproj
cdavis5e: > @cdavis5e How "old" are you thinking that might be problematic?

Anything prior to Xcode 14, I think. It was either that or 15 that introduced the new new linker. I think older versions of Clang don't understand `-ld_classic` and will actually produce an error diagnostic for the unknown option.

cdavis5e: > I've pushed a change that moves this to be tvOS-specific as I haven't seen the same problem on iOS, and currently MoltenVK doesn't compile for tvOS using Xcode 14 anyway. If there's something that's preferred I'm happy to change this.

Very well, then.

cdavis5e (approved):

MoltenVK/MoltenVK.xcodeproj/project.pbxproj
```diff
@@ -2561,6 +2561,7 @@
 				GENERATE_INFOPLIST_FILE = YES;
 				LD_DYLIB_INSTALL_NAME = "@rpath/${PRODUCT_NAME}.framework/${PRODUCT_NAME}";
 				OTHER_LDFLAGS = (
+					"-ld_classic",
```
billhollings, line 2564: You've only applied this to the `MoltenVK-tvOS-dynamic` target. Does this issue only affect dynamic linking? Otherwise, what about the `MoltenVK-tvOS-static` target?

Also, the linked [Apple support question](https://forums.developer.apple.com/forums/thread/742470) indicates this was affecting only up to iOS 12. Does the issue being fixed here only affect older versions of tvOS? If so, how far back does the current Xcode config (without `-ld_classic`) work? Presumably, at some point in the future, once the new linker works properly for all versions of tvOS we support (currently vsOS 13+), we'd want to walk this change back.

billhollings (commented):
Thanks for submitting this!

A couple of questions about the linking workaround.

billhollings (approved):
LGTM now.

PR #2331, opened 2024-09-09
Add support for RGB10A2Unorm and BGR10A2Unorm formats on ios/tvos

warmenhoven:
The Apple documentation is ambiguous on exactly when these formats were added to iOS and tvOS, but it was sometime around tvOS 11. They are definitely available in iOS/tvOS 13+.


billhollings (approved):
Thanks for picking up on this and submitting this patch.

Apple's documentation for [`CAMetalLayer.pixelFormat`](https://developer.apple.com/documentation/quartzcore/cametallayer/1478155-pixelformat?language=objc) still lists these formats as macOS-only, but if you've tested it on iOS and tvOS, this should be fine (I assume you've tested this as a surface format on iOS & tvOS...right?).

PR #2332, opened 2024-09-10
WIP: Add partial VK_EXT_image_2d_view_of_3d support

ncesario-lunarg:
Given a 3D texture that is backed by a placement heap in MoltenVK, the
approach taken here is to create a 2D texture that is backed by memory
pointing into a 3D texture's memory.

While ideal compared to alternative implementation solutions for this
extension, this approach is sensitive to how Apple lays out the memory
for 3D textures. The solution here uses
heapTextureSizeAndAlignWithDescriptor to determine the overall size of a
given 3D texture and index into the beginning of each "slice" of the 3D
texture. So far this is good enough for storage images in CTS.

FYI @kocdemir.

### DIFF
--- a/MoltenVK/MoltenVK/GPUObjects/MVKDeviceMemory.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKDeviceMemory.mm
@@ -188,15 +188,11 @@ bool MVKDeviceMemory::ensureMTLHeap() {
 	// Can't create MTLHeaps of zero size.
 	if (_allocationSize == 0) { return true; }
 
-#if MVK_MACOS
-	// MTLHeaps on macOS must use private storage for now.
-	if (_mtlStorageMode != MTLStorageModePrivate) { return true; }
-#endif
-#if MVK_IOS
-	// MTLHeaps on iOS must use private or shared storage for now.
-	if ( !(_mtlStorageMode == MTLStorageModePrivate ||
-		   _mtlStorageMode == MTLStorageModeShared) ) { return true; }
-#endif
+    if (getPhysicalDevice()->getMTLDeviceCapabilities().isAppleGPU) {
+        // MTLHeaps on Apple silicon must use private or shared storage for now.
+        if ( !(_mtlStorageMode == MTLStorageModePrivate ||
+               _mtlStorageMode == MTLStorageModeShared) ) { return true; }
+    }
 
 	MTLHeapDescriptor* heapDesc = [MTLHeapDescriptor new];
 	heapDesc.type = MTLHeapTypePlacement;
--- a/MoltenVK/MoltenVK/GPUObjects/MVKImage.h
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKImage.h
@@ -91,6 +91,8 @@ protected:
     id<MTLTexture> _mtlTexture;
     std::unordered_map<NSUInteger, id<MTLTexture>> _mtlTextureViews;
     MVKSmallVector<MVKImageSubresource, 1> _subresources;
+
+    HeapAllocation _heapAllocation;
 };
 
 
@@ -233,6 +235,7 @@ public:
     /** Populates the specified transfer image descriptor data structure. */
     void getTransferDescriptorData(MVKImageDescriptorData& imgData);
 
+    MTLTextureDescriptor* getTextureDescriptor(uint32_t planeIndex);
 
 #pragma mark Resource memory
 
@@ -334,6 +337,8 @@ public:
 	/** Returns the Metal CPU cache mode used by this image. */
 	MTLCPUCacheMode getMTLCPUCacheMode();
 
+    HeapAllocation* getHeapAllocation(uint32_t planeIndex);
+
 
 #pragma mark Construction
 
@@ -395,6 +400,7 @@ protected:
 	bool _hasMutableFormat;
 	bool _shouldSupportAtomics;
 	bool _isLinearForAtomics;
+    bool _is2DViewOn3DImageCompatible = false;
 };
 
 
--- a/MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
@@ -48,6 +48,10 @@ id<MTLTexture> MVKImagePlane::getMTLTexture() {
         MVKImageMemoryBinding* memoryBinding = getMemoryBinding();
 		MVKDeviceMemory* dvcMem = memoryBinding->_deviceMemory;
 
+        if (_image->_is2DViewOn3DImageCompatible && !dvcMem->ensureMTLHeap()) {
+            MVKAssert(0, "Creating a 2D view of a 3D texture currently requires a placement heap, which is not available.");
+        }
+
         if (_image->_ioSurface) {
             _mtlTexture = [_image->getMTLDevice()
                            newTextureWithDescriptor: mtlTexDesc
@@ -60,6 +64,11 @@ id<MTLTexture> MVKImagePlane::getMTLTexture() {
                            bytesPerRow: _subresources[0].layout.rowPitch];
         } else if (dvcMem && dvcMem->getMTLHeap() && !_image->getIsDepthStencil()) {
             // Metal support for depth/stencil from heaps is flaky
+            _heapAllocation.heap = dvcMem->getMTLHeap();
+            _heapAllocation.offset = memoryBinding->getDeviceMemoryOffset() + _subresources[0].layout.offset;
+            const auto texSizeAlign = [dvcMem->getMTLDevice() heapTextureSizeAndAlignWithDescriptor:mtlTexDesc];
+            _heapAllocation.size = texSizeAlign.size;
+            _heapAllocation.align = texSizeAlign.align;
             _mtlTexture = [dvcMem->getMTLHeap()
                            newTextureWithDescriptor: mtlTexDesc
                            offset: memoryBinding->getDeviceMemoryOffset() + _subresources[0].layout.offset];
@@ -823,6 +832,10 @@ void MVKImage::getTransferDescriptorData(MVKImageDescriptorData& imgData) {
     imgData.usage = getCombinedUsage();
 }
 
+MTLTextureDescriptor* MVKImage::getTextureDescriptor(uint32_t planeIndex) {
+    return _planes[planeIndex]->newMTLTextureDescriptor();
+}
+
 // Returns whether an MVKImageView can have the specified format.
 // If the list of pre-declared view formats is not empty,
 // and the format is not on that list, the view format is not valid.
@@ -1069,6 +1082,11 @@ MTLCPUCacheMode MVKImage::getMTLCPUCacheMode() {
 	return _memoryBindings[0]->_deviceMemory ? _memoryBindings[0]->_deviceMemory->getMTLCPUCacheMode() : MTLCPUCacheModeDefaultCache;
 }
 
+HeapAllocation* MVKImage::getHeapAllocation(uint32_t planeIndex) {
+    auto& heapAllocation = _planes[planeIndex]->_heapAllocation;
+    return (heapAllocation.isValid()) ? &heapAllocation : nullptr;
+}
+
 MTLTextureUsage MVKImage::getMTLTextureUsage(MTLPixelFormat mtlPixFmt) {
 
 	// In the special case of a dedicated aliasable image, we must presume the texture can be used for anything.
@@ -1254,6 +1272,8 @@ MVKImage::MVKImage(MVKDevice* device, const VkImageCreateInfo* pCreateInfo) : MV
 	if (pExportInfo && pExportInfo->exportObjectType == VK_EXPORT_METAL_OBJECT_TYPE_METAL_IOSURFACE_BIT_EXT && !_ioSurface) {
 		setConfigurationResult(useIOSurface(nil));
 	}
+
+    _is2DViewOn3DImageCompatible = mvkIsAnyFlagEnabled(pCreateInfo->flags, VK_IMAGE_CREATE_2D_VIEW_COMPATIBLE_BIT_EXT);
 }
 
 VkSampleCountFlagBits MVKImage::validateSamples(const VkImageCreateInfo* pCreateInfo, bool isAttachment) {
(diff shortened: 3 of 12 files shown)


MoltenVK/MoltenVK/GPUObjects/MVKImage.h
```diff
@@ -566,6 +582,10 @@ class MVKImageViewPlane : public MVKBaseDeviceObject {
     bool _useMTLTextureView;
 	bool _useNativeSwizzle;
 	bool _useShaderSwizzle;
+
+    // For now, this is only used for VK_EXT_image_2d_view_of_3d. Specifically, when the backing image of this view is 3D and the view is 2D,
+    // this texture will be allocated using a placement heap "on top of" the 3D textures backing memory.
+    id<MTLTexture> _mtlShadowTexture;
```
billhollings, line 588: Since `_mtlTexture` end up holding a reference to the same object, does this need to be a separate iVar than `_mtlTexture`? 

If it is needed for some reason, move it up just below the `_mtlTexture` declaration. We try to organize iVars to minimize gaps in the underlying class structure (ie- keep all 8-byte iVars together).

It also feels dangerous that the `MTLTexture` is not retained in both `_mtlTexture` and `_mtlShadowTexture`, and released from both in the destructor. If in the future, we might possibly have different objects in the two iVars, then we'll end up with a hard to discover memory leak.

Finally, maybe it's me, but when I see `_mtlShadowTexture`, I keep thinking "shadow map texture". Maybe rename it to `_mtlAliasTexture` or something.

billhollings (changes requested):
Thanks for getting this started!

MoltenVK/MoltenVK/GPUObjects/MVKDeviceMemory.mm
```diff
@@ -188,10 +188,6 @@
 	// Can't create MTLHeaps of zero size.
 	if (_allocationSize == 0) { return true; }
 
-#if MVK_MACOS
-	// MTLHeaps on macOS must use private storage for now.
-	if (_mtlStorageMode != MTLStorageModePrivate) { return true; }
-#endif
 #if MVK_IOS
```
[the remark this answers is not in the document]
ncesario-lunarg: TBH I'm not sure based on the docs, but I'll add that to be on the safe side. It looked like the usage of heaps in MoltenVK was mainly dependent on [placement heaps](https://developer.apple.com/documentation/metal/mtlheaptype/mtlheaptypeplacement?language=objc), but my knowledge is still pretty sparse on the details here.

MoltenVK/MoltenVK/GPUObjects/MVKDeviceMemory.mm
```diff
@@ -188,10 +188,6 @@
 	// Can't create MTLHeaps of zero size.
 	if (_allocationSize == 0) { return true; }
 
-#if MVK_MACOS
-	// MTLHeaps on macOS must use private storage for now.
-	if (_mtlStorageMode != MTLStorageModePrivate) { return true; }
```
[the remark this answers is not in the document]
ncesario-lunarg: See comment below.

MoltenVK/MoltenVK/GPUObjects/MVKImage.h
```diff
@@ -48,6 +48,16 @@ typedef struct {
 class MVKImagePlane : public MVKBaseObject {
 
 public:
+    struct HeapAllocation {
+        id<MTLHeap> heap = nil; // Reference to the heap containing this allocation
+        size_t offset = 0; // Offset into the heap
+        size_t size = 0; // Total size of this allocation
+        size_t align = 0; // Allocation alignment requirement
+
+        operator bool() const {
```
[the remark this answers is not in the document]
ncesario-lunarg: 👍 Done

MoltenVK/MoltenVK/GPUObjects/MVKImage.h
```diff
@@ -395,6 +410,7 @@ class MVKImage : public MVKVulkanAPIDeviceObject {
 	bool _hasMutableFormat;
 	bool _shouldSupportAtomics;
 	bool _isLinearForAtomics;
+    bool _is2DViewCompatible = false;
```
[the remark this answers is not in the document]
ncesario-lunarg: Yep, makes sense. Done.

MoltenVK/MoltenVK/GPUObjects/MVKImage.h
ncesario-lunarg: Nope, there is no need for the extra MTLTexture in MVKImageViewPlane; I will remove it. I added this from a debug session and never reverted it (and also forgot to clean it up in the destructor when I was using it :/).

MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
```diff
@@ -1066,6 +1079,11 @@ static MTLRegion getMTLRegion(const ImgRgn& imgRgn) {
 	return _memoryBindings[0]->_deviceMemory ? _memoryBindings[0]->_deviceMemory->getMTLCPUCacheMode() : MTLCPUCacheModeDefaultCache;
 }
 
+MVKImagePlane::HeapAllocation* MVKImage::getHeapAllocation(uint32_t planeIndex) {
+    auto& heapAllocation = _planes[planeIndex]->_heapAllocation;
+    return (heapAllocation) ? &heapAllocation : nullptr;
```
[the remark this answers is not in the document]
ncesario-lunarg: Changed to `isValid()`.

MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
```diff
@@ -1251,6 +1269,8 @@ static MTLRegion getMTLRegion(const ImgRgn& imgRgn) {
 	if (pExportInfo && pExportInfo->exportObjectType == VK_EXPORT_METAL_OBJECT_TYPE_METAL_IOSURFACE_BIT_EXT && !_ioSurface) {
 		setConfigurationResult(useIOSurface(nil));
 	}
+
+    _is2DViewCompatible = pCreateInfo->flags & VK_IMAGE_CREATE_2D_VIEW_COMPATIBLE_BIT_EXT;
```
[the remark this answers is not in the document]
ncesario-lunarg: Done.

ncesario-lunarg:
Thanks for the review @billhollings! I've made the changes you've suggested locally, but I'm hitting a crash on `dEQP-VK.api.get_memory_commitment.memory_commitment` at https://github.com/ncesario-lunarg/MoltenVK/blob/d01e7ecf13ef9bd4d9af3dc61b13ce683299d5ba/MoltenVK/MoltenVK/GPUObjects/MVKDeviceMemory.mm#L205. I'm going to resolve that before updating this PR.

squidbus:
Does this PR impact the `imageView2DOn3DImage` feature disabled in the portability subset at all?

ncesario-lunarg:
@squidbus yes it does, and I had mistakenly not enabled that, but it should be enabled now.

MoltenVK/MoltenVK/Utility/MVKEnvironment.cpp
```diff
@@ -26,10 +26,9 @@
 // Return the expected size of MVKConfiguration, based on contents of MVKConfigMembers.def.
 static constexpr uint32_t getExpectedMVKConfigurationSize() {
 #define MVK_CONFIG_MEMBER(member, mbrType, name)         cfgSize += sizeof(mbrType);
-	uint32_t cfgSize = 0;
+	size_t cfgSize = 0;
 #include "MVKConfigMembers.def"
-	cfgSize += kMVKConfigurationInternalPaddingByteCount;
```
ncesario-lunarg, line 31: @billhollings This seemed to be calculating the incorrect size, but perhaps that meant I added the config member incorrectly? I changed this function to align the expected size to the alignment of `MVKConfiguration`; please let me know if something is "off" here.

MoltenVK/MoltenVK/Layers/MVKExtensions.def
```diff
@@ -144,6 +144,7 @@ MVK_EXTENSION(EXT_swapchain_maintenance1,             EXT_SWAPCHAIN_MAINTENANCE_
 MVK_EXTENSION(EXT_texel_buffer_alignment,             EXT_TEXEL_BUFFER_ALIGNMENT,             DEVICE,   10.13, 11.0,  1.0)
 MVK_EXTENSION(EXT_texture_compression_astc_hdr,       EXT_TEXTURE_COMPRESSION_ASTC_HDR,       DEVICE,   11.0,  13.0,  1.0)
 MVK_EXTENSION(EXT_vertex_attribute_divisor,           EXT_VERTEX_ATTRIBUTE_DIVISOR,           DEVICE,   10.11,  8.0,  1.0)
+MVK_EXTENSION(EXT_image_2d_view_of_3d,                EXT_IMAGE_2D_VIEW_OF_3D,                DEVICE,   10.15, 13.0,  1.0)
```
kocdemir, line 147: Do you think we can modify mvkIsSupportedOnPlatform check and not advertise support on this when the config flag is not enabled?

ncesario-lunarg: Yes, I'll wrap that in an `#ifdef`.

MoltenVK/MoltenVK/Utility/MVKEnvironment.cpp
billhollings: Have a look at the description of `kMVKConfigurationInternalPaddingByteCount`. It's sole purpose is to adjust the calculation of:

`static_assert(getExpectedMVKConfigurationSize() == sizeof(MVKConfiguration), "MVKConfigMembers.def does not match the members of MVKConfiguration.");
`

And the purpose of **that** `static_assert()` is to ensure that the members of `MVKConfiguration` are all accounted for in `MVKConfigMembers.def` (ie- a dev mod isn't made to `MVKConfiguration` without adding the corrsponding entry in `MVKConfigMembers.def`), which would be tough to debug.

The reason `kMVKConfigurationInternalPaddingByteCount` is needed is to account for alignment padding that the compiler adds to `sizeof(MVKConfiguration)`.

So, the solution to the issue you have dealt with here is to adjust the value of `kMVKConfigurationInternalPaddingByteCount` accordingly. In this specific case, the value should be set to `0`, instead of removing the padding calculation. The next time an entry is added to `MVKConfiguration`, it will need to be set to a different value, and the padding calculation needs to be available.

MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
```diff
@@ -1801,12 +1821,34 @@ static void signalAndUntrack(const MVKSwapchainSignaler& signaler) {
...
+            mtlTextureType = MTLTextureType3D;
+            sliceRange = NSMakeRange(0, 1);
+        } else {
+            if (!_mtlTexture) {
+                const auto heapAllocation = image->getHeapAllocation(_planeIndex);
+                MVKAssert(heapAllocation, "Attempting to create a 2D view of a 3D texture without a placement heap");
+                // TODO (ncesario-lunarg) untested where _imageView->subresourceRange.layerCount > 1, but VK_EXT_image_2d_view_of_3d
+                //      allows for 2D_ARRAY views of 3D textures.
+                const auto relativeSliceOffset = _imageView->_subresourceRange.baseArrayLayer * (heapAllocation->size / image->_extent.depth);
+                MTLTextureDescriptor* mtlTexDesc = image->getTextureDescriptor(_planeIndex);
```
billhollings, line 1838: `mtlTexDesc` will leak if it isn't released after it is used. See other uses of `newMTLTextureDescriptor()` for guidance on that.

MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
```diff
@@ -1801,12 +1821,34 @@ static void signalAndUntrack(const MVKSwapchainSignaler& signaler) {
...
+    id<MTLTexture> mtlTex = image->getMTLTexture(_planeIndex);
+    if (image->getImageType() == VK_IMAGE_TYPE_3D &&
         (mtlTextureType == MTLTextureType2D || mtlTextureType == MTLTextureType2DArray)) {
-        mtlTextureType = MTLTextureType3D;
-        sliceRange = NSMakeRange(0, 1);
+        if (!image->_is2DViewOn3DImageCompatible) {
+            mtlTextureType = MTLTextureType3D;
+            sliceRange = NSMakeRange(0, 1);
+        } else {
+            if (!_mtlTexture) {
```
billhollings, line 1832: This test is unnecessary, since `newMTLTexture()` is only called if `_mtlTexture` is `nil`.

MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
```diff
@@ -1801,12 +1821,34 @@ static void signalAndUntrack(const MVKSwapchainSignaler& signaler) {
...
+                const auto relativeSliceOffset = _imageView->_subresourceRange.baseArrayLayer * (heapAllocation->size / image->_extent.depth);
+                MTLTextureDescriptor* mtlTexDesc = image->getTextureDescriptor(_planeIndex);
+
+                mtlTexDesc.depth = _imageView->_subresourceRange.layerCount;
+                mtlTexDesc.textureType = mtlTextureType;
+
+                // Create a temporary texture that is backed by the 3D texture's memory
+                _mtlTexture = [heapAllocation->heap
+                                     newTextureWithDescriptor: mtlTexDesc
+                                     offset: heapAllocation->offset + relativeSliceOffset];
```
billhollings, line 1846: Two issues here.

1) This unexpectedly plugs a value into `_mtlTexture`, which will be overridden by the value returned by this function.

2) The new `MTLTexture` created here is retained, and will leak if it is not released at some point.

I suggest:

- Declare a `id<MTLTexture> aliasTex = nil;` at the top of this function.
- Assign `aliasTex` here from the call to `newTextureWithDescriptor()`, and set `mtlTex = aliasTex;`.
- Replace the two `return [mtlTex newTextureView...` below with `texView = [mtlTex newTextureView...`.
- End the function with:
```
[aliasTex release];
return texView;
```

It's important to release `aliasTex` after `texView` has been created, and `aliasTex` will be retained within `texView`, and will be released and deallocated when `texView` is released and deallocated.

billhollings (changes requested):
Thanks for getting this ready for release!

I've made a few change requests to catch a few potential memory leaks and simplify the configuration option.

squidbus:
Is this planned to be finished up in the near future? From the review comments it seems to me this is working but has some changes requested,and I would be interested in trying to address those comments and get this over the line if the author doesn't currently have the time available.

billhollings:
Closing as superseded by #2442.

PR #2339, opened 2024-09-17
Improve fetchDependencies

jeroenbakker-atmind:
fetchDependencies can loop forever without any feedback to the user when leading `-` are missing.

Detected when running `./fetchDependencies macos` which is of course incorrect.

This is fixed by widening the wildcard for unsupported options.


billhollings (approved):
Good catch. Thanks for submitting the fix.

PR #2345, opened 2024-09-21
Fix leak where texel buffer is occasionally accidentally created during image-memory unbinding (#2341)

### LINKED ISSUE #2341: GPU Memory leak on A12Z-based iPad Pro, ATV 4K 1st Gen
Starting with 188a21ad33df4705ea4050f0116939bee097c7fe, on these older devices, I'm seeing a GPU memory leak of about 22MB per second. I don't see it under the same usage scenarios on my M2 Mac or my iPhone 15 Pro.

[ipad.log](https://github.com/user-attachments/files/17038259/ipad.log)

warmenhoven:
Fixes #2341.

### DIFF
--- a/MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
@@ -412,8 +412,10 @@ VkResult MVKImageMemoryBinding::bindDeviceMemory(MVKDeviceMemory* mvkMem, VkDevi
     if (_deviceMemory) { _deviceMemory->removeImageMemoryBinding(this); }
     MVKResource::bindDeviceMemory(mvkMem, memOffset);
 
+    if (!_deviceMemory) { return VK_SUCCESS; }
+
 	auto& mtlFeats = getMetalFeatures();
-    bool usesTexelBuffer = mtlFeats.texelBuffers && _deviceMemory; // Texel buffers available
+    bool usesTexelBuffer = mtlFeats.texelBuffers; // Texel buffers available
     usesTexelBuffer = usesTexelBuffer && (isMemoryHostAccessible() || mtlFeats.placementHeaps) && _image->_isLinear && !_image->getIsCompressed(); // Applicable memory layout
 
     // macOS before 10.15.5 cannot use shared memory for texel buffers.
@@ -425,7 +427,8 @@ VkResult MVKImageMemoryBinding::bindDeviceMemory(MVKDeviceMemory* mvkMem, VkDevi
             _mtlTexelBufferOffset = getDeviceMemoryOffset();
         } else {
             // Create our own buffer for this.
-            if (_deviceMemory && _deviceMemory->_mtlHeap && _image->getMTLStorageMode() == _deviceMemory->_mtlStorageMode) {
+            if (_ownsTexelBuffer) { [_mtlTexelBuffer release]; }
+            if (_deviceMemory->_mtlHeap && _image->getMTLStorageMode() == _deviceMemory->_mtlStorageMode) {
                 _mtlTexelBuffer = [_deviceMemory->_mtlHeap newBufferWithLength: _byteCount options: _deviceMemory->getMTLResourceOptions() offset: getDeviceMemoryOffset()];
                 if (_image->_isAliasable) { [_mtlTexelBuffer makeAliasable]; }
             } else {
@@ -443,7 +446,7 @@ VkResult MVKImageMemoryBinding::bindDeviceMemory(MVKDeviceMemory* mvkMem, VkDevi
     }
 
     flushToDevice(getDeviceMemoryOffset(), getByteCount());
-    return _deviceMemory ? _deviceMemory->addImageMemoryBinding(this) : VK_SUCCESS;
+    return _deviceMemory->addImageMemoryBinding(this);
 }
 
 void MVKImageMemoryBinding::applyMemoryBarrier(MVKPipelineBarrier& barrier,


MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
```diff
@@ -2308,7 +2308,10 @@ static void signalAndUntrack(const MVKSwapchainSignaler& signaler) {
 // Memory detached in destructor too, as a fail-safe.
 MVKImageView::~MVKImageView() {
 	detachMemory();
-	_image->release();
+    if (_image) {
+        _image->release();
+        _image = NULL;
```
cdavis5e, line 2313: It shouldn't be necessary to set the `_image` to `nullptr` here, since the object will cease to exist after the destructor is run.

MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
```diff
@@ -2317,6 +2320,10 @@ static void signalAndUntrack(const MVKSwapchainSignaler& signaler) {
 void MVKImageView::destroy() {
 	detachMemory();
 	MVKVulkanAPIDeviceObject::destroy();
+    if (_image) {
+        _image->release();
+        _image = NULL;
+    }
```
cdavis5e, line 2326: So, the reason that `MVKBaseObject::destroy()` exists in the first place is that Metal could still indirectly reference one of these objects even after the Vulkan spec says it's OK to destroy them, through a completion block. These objects are `retain()`ed so that this case doesn't crash. The objects should still be destroyed after the completion blocks have finished running.

Releasing the image here could cause a crash later on. Why is this here? Are you seeing a case where the `MVKImage` and `MVKImageView` objects are getting leaked?

cdavis5e (changes requested):

MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
warmenhoven: > Are you seeing a case where the MVKImage and MVKImageView objects are getting leaked?

I am, but only on some older hardware, and not in all cases. I only notice the leak after/as a result of 188a21ad33df4705ea4050f0116939bee097c7fe, and if I revert that I no longer see it. It's been sort of hard for me to narrow down the specific case that's causing the leak.

I understand what you're saying though. I'll see if I can go through and figure out which objects are being destroy()'d but not having their destructors called, and where that destroy() is coming from.

warmenhoven: I've updated my PR based on my investigation that I've documented in #2341 ; basically `MVKImageMemoryBinding::bindMemoryAddress()` ends up getting called from the `MVKDeviceMemory` destructor with a `nullptr`, but because `_image->_isLinearForAtomics` is true it allocates a buffer that never gets freed.

MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
```diff
@@ -419,7 +419,7 @@
     // macOS before 10.15.5 cannot use shared memory for texel buffers.
     usesTexelBuffer = usesTexelBuffer && (mtlFeats.sharedLinearTextures || !isMemoryHostCoherent());
 
-    if (_image->_isLinearForAtomics || (usesTexelBuffer && mtlFeats.placementHeaps)) {
+    if ((_deviceMemory && _image->_isLinearForAtomics) || (usesTexelBuffer && mtlFeats.placementHeaps)) {
```
cdavis5e, line 422: I apologize for the extra churn, but it occurs to me that the entire `if`-`else` chain ought to be guarded by `if (_deviceMemory)`. Then the null check on line 428 could go away.

cdavis5e (changes requested):
Looks good. Just one small thing.

MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
warmenhoven: Not a problem at all. I had originally put together a change that bailed out early and decided it would be safer to do something more targeted; but because `_deviceMemory` effectively gets checked in so many places, the entire function ends up being a no-op after the `MVKResource::bindDeviceMemory` call.

cdavis5e (approved):

MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
```diff
@@ -412,8 +412,12 @@
     if (_deviceMemory) { _deviceMemory->removeImageMemoryBinding(this); }
     MVKResource::bindDeviceMemory(mvkMem, memOffset);
 
+    if (!_deviceMemory) {
+        return VK_SUCCESS;
+    }
```
billhollings, line 417: MoltenVK style typically has this kind of trivial `if` statement on a single line. See for instance line 412 immediately above. Can you compress it please?

MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
```diff
@@ -425,7 +429,7 @@
             _mtlTexelBufferOffset = getDeviceMemoryOffset();
         } else {
             // Create our own buffer for this.
```
billhollings, line 431: Technically, the Vulkan spec mandates a memory-binding activity should only occur once for an image, meaning this function should only ever be called once by the app. But just to make the leak protection a little bit more defensive in case an app makes a mistake, it might be a good to include the following line in between this comment and  the `if()` statement:

```
if (_ownsTexelBuffer) { [_mtlTexelBuffer release]; }
```

billhollings (changes requested):
Thanks for submitting this PR. 

And thanks for all your incredible detective work! I appreciate you digging into this so thoroughly! Great catch!

LGTM. Just a couple of very minor style and robustness requests.

And when amending this, can you modify the commit message to match the updated fix, please? I was able to change the PR description, but I can't change the commit message itself. 

Thanks.

warmenhoven:
I've updated the PR for the wording, the style, and the extra defensive check, and it's all running well for me. Thanks very much!

> Out of curiosity, under what conditions was vkFreeMemory() being called so frequently on every frame?

I wish I could say I understood [Granite](https://github.com/Themaister/Granite) well enough to explain what it's doing [here](https://github.com/Themaister/Granite/blob/533740fe9a443ad48c81df22548736c10faa88b4/vulkan/device.cpp#L2816). My exact test case is just running the [Nintendo 64 RetroArch core](https://github.com/libretro/mupen64plus-libretro-nx/tree/develop) using its [Vulkan renderer](https://github.com/libretro/mupen64plus-libretro-nx/tree/develop/mupen64plus-video-paraLLEl).

billhollings (approved):
LGTM now. Thanks!

Issue #2271, opened 2024-07-10
SPIR-V to MSL conversion error: Argument buffer resource base type could not be determined


patrick-han:
```
[mvk-error] SPIR-V to MSL conversion error: Argument buffer resource base type could not be determined. When padding argument buffer elements, all descriptor set resources must be supplied with a base type by the app.
[mvk-error] VK_ERROR_INVALID_SHADER_NV: Fragment shader function could not be compiled into pipeline. See previous logged error.

validation layer: VK_ERROR_INVALID_SHADER_NV: Fragment shader function could not be compiled into pipeline. See previous logged error.
```

This seems to have cropped up at various points before, but unsure if it's related. I have the latest VulkanSDK as of yesterday (1.3.283.0) so in theory I have the fix for these issues:

https://github.com/KhronosGroup/MoltenVK/issues/2216
https://github.com/KhronosGroup/MoltenVK/issues/2016

After enabling `MVK_CONFIG_USE_METAL_ARGUMENT_BUFFERS` programmatically I also disabled shader validation at the advice of one of the recent Vulkanised 2024 talks.

The specific descriptor indexing features I enabled are:
`descriptorBindingSampledImageUpdateAfterBind`, `descriptorBindingPartiallyBound`, `runtimeDescriptorArray` with the appropriate pool/binding/layout flags:
`VK_DESCRIPTOR_POOL_CREATE_UPDATE_AFTER_BIND_BIT_EXT`, `VK_DESCRIPTOR_BINDING_PARTIALLY_BOUND_BIT_EXT`, `VK_DESCRIPTOR_BINDING_UPDATE_AFTER_BIND_BIT_EXT`, `VK_DESCRIPTOR_SET_LAYOUT_CREATE_UPDATE_AFTER_BIND_POOL_BIT_EXT`

Although in my fragment shader I'm still just declaring my texture normally:

```glsl
layout (set = 0, binding = 0) uniform texture2D texture0;
layout (set = 0, binding = 1) uniform sampler linearSampler;
```

billhollings:
It's likely that PR #2260 and followups may have fixed this.

Please build from the latest MoltenVK, and try again. Or you can wait for the next SDK, which should be released in a couple of weeks.

patrick-han:
I just tried building the latest now, building the macOS library. Unfortunately I'm running into another error now with a crash + artifacts:

```
[...]

PR #2349, opened 2024-09-23
Separately track bitarray capacity

etang-cw:
Fixes a bug where when resizing from > 64 bits to ≤ 64 bits, BitArray wouldn't realize that it was going to go from interpreting _data as a pointer to interpreting it as bits, and would fail to both free the pointer and copy its contents into _data, resulting in a memory leak, loss of the data in the BitArray, and a desync with the `_lowestNeverClearedBitIndex` and `_clearedSectionCount` variables.  In the worst case scenario, such a resize when `_lowestNeverClearedBitIndex` was 0 would make future calls to `setAllBits` not set all bits, even though many bits were not set.

### DIFF
--- a/MoltenVK/MoltenVK/Utility/MVKBitArray.h
+++ b/MoltenVK/MoltenVK/Utility/MVKBitArray.h
@@ -181,50 +181,53 @@ public:
 	void resize(size_t size, bool val = false) {
 		if (size == _bitCount) { return; }
 
-		size_t oldBitCnt = _bitCount;
-		size_t oldSecCnt = getSectionCount();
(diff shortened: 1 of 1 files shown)


billhollings:
Thanks for much for submitting this. And good catch and research!

This issue was causing the crash in #2271.

I've been sitting on an overhaul of `BitArray` that fixes this issue, and a few other things, including performance, because I didn't want to include a hefty change right before an SDK release, in case it unexpectedly affected other things.

So it's my bad that it's not out there yet. Sorry about that. I'll put that out in the few days.

billhollings:
Superseded by PR #2355.

PR #2370, opened 2024-10-09
Fix set, but never read variable within MVKDevice::getFirstViewIndexInMetalPass.

nlutsenko:
A super basic cleanup, caught by enabling `-Wunused-but-set-variable` in clang.


billhollings:
Good catch! Thanks for submitting this!

PR #2372, opened 2024-10-09
Fix header namespace pollution.

nlutsenko:
This removes `using namespace ...;` from headers, as such reducing header namespace pollution.
Whilst not critical for MoltenVK itself, when incorporated in a larger project makes sure that symbols/functionality is addressed appropriately and items within `mvk` namespace aren't considered to be part of the global namespace.

Also enabled the warning within xcodeproj, so it's visible when this is added again.

### DIFF
--- a/MoltenVK/MoltenVK.xcodeproj/project.pbxproj
+++ b/MoltenVK/MoltenVK.xcodeproj/project.pbxproj
@@ -2702,7 +2702,10 @@
 				PRODUCT_NAME = MoltenVK;
 				SKIP_INSTALL = YES;
 				TVOS_DEPLOYMENT_TARGET = 13.0;
-				WARNING_CFLAGS = "-Wreorder";
+				WARNING_CFLAGS = (
+					"-Wreorder",
+					"-Wheader-hygiene",
+				);
 			};
 			name = Debug;
 		};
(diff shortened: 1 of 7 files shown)


billhollings (changes requested):
LGTM.

Since you are modifying `WARNING_CFLAGS`, would you mind also adding the `-Wunused-but-set-variable` that you identified in PR #2370, please?

This will trigger a warning in the `MVKSampler` constructor. To fix that, can you also remove the two lines there referencing `pSampReductInfo`, please? Please leave the empty `case` statement though, to highlight that option exists and may be needed at a later time.

After your PR #2370, I was going to issue a follow up PR with those changes, but you might as well close the loop here.

billhollings (approved):

billhollings:
> Since you are modifying `WARNING_CFLAGS`, would you mind also adding the `-Wunused-but-set-variable` that you identified in PR #2370, please?
> 
> This will trigger a warning in the `MVKSampler` constructor. To fix that, can you also remove the two lines there referencing `pSampReductInfo`, please? Please leave the empty `case` statement though, to highlight that option exists and may be needed at a later time.

I've included these additional items in PR #2382.

PR #2388, opened 2024-11-04
Fix Random Crashes when Using Pool with Freeable Descriptor Sets

WDog367:
Some small changes to fix crashes and random descriptor set issues I was running into when using a pool with `FREE_DESCRIPTOR_SET` enabled

First issue was caused by `_nextMetalArgumentBufferOffset` in the pool becoming out-of-date if the allocated set in the highest index is freed and re-used. Argument buffers of any size would be allowed to use that slot, but the offset wasn't adjusted, resulting in some argument buffer ranges overlapping

Eventually crashing with the error: `[mvk-error] VK_ERROR_OUT_OF_DEVICE_MEMORY: Lost VkDevice after MTLCommandBuffer "vkQueueSubmit MTLCommandBuffer on Queue 0-0 (0x406e75600)" execution failed (code 3): Caused GPU Address Fault Error (0000000b:kIOGPUCommandBufferCallbackErrorPageFault)`

Second issue is caused by a bug in `enumerateEnabledBits()` where it is enumerating bits that were disabled.
Specifically, in the case where `startIndex` is higher than the first enabled bit, but the next enabled bit is in a different section than `startIndex`, it looks like it just return the first index in the next section, instead of the next enabled bit.

I just added a quick fix so that the enumerate func checks whether the returned bit is enabled before calling the functor. More changes to the MVKBitArray functions might be called for, but I hope this is helpful 😄

### DIFF
--- a/MoltenVK/MoltenVK/GPUObjects/MVKDescriptorSet.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKDescriptorSet.mm
@@ -761,7 +761,6 @@ VkResult MVKDescriptorPool::allocateDescriptorSet(MVKDescriptorSetLayout* mvkDSL
 			// on a reset pool), set the offset and update the next available offset value.
 			if ( !mtlArgBuffOffset && (dsIdx || !_nextMetalArgumentBufferOffset)) {
 				mtlArgBuffOffset = _nextMetalArgumentBufferOffset;
-				_nextMetalArgumentBufferOffset += mtlArgBuffEncAlignedSize;
 			}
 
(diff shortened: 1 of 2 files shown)


billhollings:
Replaced by PR #2392.

billhollings:
@WDog367  Thanks for submitting this, and for pointing out the issues.

Since this PR is mostly a workaround, I've replaced it with #2392, which should fix the same issues without workarounds.

Please retest with PR #2392, and confirm that it solves the same issues this PR does. Thanks.

WDog367:
Yes PR #2392 fixes the issue

Thank you!

PR #2390, opened 2024-11-07
Clang-format proposal

GOOG-sergiu:
In Android, we're depending on MoltenVK as one of our external dependencies.

We occasionally make small modifications for our purposes that would not make sense to be merged back into the upstream.

However, ensuring consistent formatting seems to be a challenge since we can get merge conflicts from the upstream.  I wanted to propose a simple clang-format (we're not attached to the formatter in any which way - we would just prefer if there were a formatter). 

* First commit presents a simple clang-format.
* In my second commit I present the changes to GPUObjects/ * 

I am happy to remove the formatting example and for one of the main contributors to apply the formatting changes or if desired I am happy to reformat the whole project if we agree on the format. 

I am a bit surprised there isn't a global formatter already - so if I missed that, please just let me know and we will use that.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,49 @@
+# Indentation
+IndentWidth: 4
+UseTab: Never
+ContinuationIndentWidth: 4
+
+# Braces
+BreakBeforeBraces: Attach  # Braces attach to control statements (if, while, for)
+
+# Spacing
+SpaceBeforeParens: ControlStatements
+SpacesInParentheses: false
+SpacesInSquareBrackets: false
+SpaceBeforeAssignmentOperators: true
+
+# Line Breaking
+ColumnLimit: 80  # Larger column limit to prevent premature breaks
+PenaltyBreakBeforeFirstCallParameter: 1000  # Prevent breaking at the first call parameter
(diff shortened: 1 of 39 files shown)


etang-cw:
I don't think we currently have a global formatter.

Personal formatting preferences:
- UseTabs: AlignWithSpaces
- ColumnLimit: 0 (Otherwise clang-format reflows everything to hit the column limit, which makes everything really hard to read.  Let the programmer pick where to wrap.  This should also reduce the amount of reflowing the initial formatting commit does, which might make it easier to compare the before and after.)
- AlignTrailingComments: Leave (I've never seen auto-align consistently work.  There are too many cases of unrelated comments being next to each other and getting aligned when they shouldn't, or related comments being too far apart and getting not aligned.  If we have to support a clang-format before 16, I guess I'd prefer false, but it would be really nice to allow us to align comments ourselves without leaving it up to an auto-aligner.)
- AlignEscapedNewlines: DontAlign (If you edit on a small window with soft wrap, aligning the backslashes frequently results in them wasting space on their own line.  It's nice to be able to use two-column view on a laptop screen and have that be usable.)
- BraceWrapping: AfterControlStatement: Multiline

Issue #2399, opened 2024-12-01
memory management issue in MVKSwapchain::setHDRMetadataEXT


NekoAsakura:
MVK version 1.2.12, supporting Vulkan version 1.2.296
Build from https://github.com/KhronosGroup/MoltenVK/commit/26c7b0234dd093e6153237c4a8d94eedd07f9614

-----

mpv community has reported an issue causing program to crash, after investigation, was found to be caused by a memory management issue in `MVKSwapchain::setHDRMetadataEXT`. https://github.com/mpv-player/mpv/issues/12666#issuecomment-2509030237

According to [README](https://github.com/KhronosGroup/MoltenVK/blob/main/README.md#Memory-Management):

> As a contributor to **MoltenVK**, avoid the use of the *Metal* `autorelease` method, or any object *Metal* creation
methods that imply internal use of `autorelease`, (eg- `[NSString stringWithFormat: ]`, etc). 

https://github.com/KhronosGroup/MoltenVK/blob/26c7b0234dd093e6153237c4a8d94eedd07f9614/MoltenVK/MoltenVK/GPUObjects/MVKSwapchain.mm#L359-L360

However in `MVKSwapchain::setHDRMetadataEXT` NSData was assigned with type `dataWithBytes`, which is implied with internal use of autorelease:

https://github.com/apportable/Foundation/blob/master/System/Foundation/src/NSData.m#L744-L747

```
+ (id)dataWithBytes:(const void *)bytes length:(NSUInteger)length

{
    return [[[self alloc] initWithBytes:(void *)bytes length:length copy:YES freeWhenDone:NO bytesAreVM:NO] autorelease];
}
```

For manually track and release, it should be 

```
NSData* colorVolData = [[NSData alloc] initWithBytes: &colorVol length: sizeof(colorVol)];
```

cdavis5e:
This should, of course, be fixed by your change #2400. Normally, this is where we'd ask you to test this, but I would presume you did that when you wrote the change.

PR #2400, opened 2024-12-01
fix memory management issue in MVKSwapchain

NekoAsakura:
https://github.com/KhronosGroup/MoltenVK/issues/2399
Fix memory management issue in MVKSwapchain

### DIFF
--- a/MoltenVK/MoltenVK/GPUObjects/MVKSwapchain.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKSwapchain.mm
@@ -356,8 +356,8 @@ void MVKSwapchain::setHDRMetadataEXT(const VkHdrMetadataEXT& metadata) {
 	colorVol.min_display_mastering_luminance = OSSwapHostToBigInt32((uint32_t)(metadata.minLuminance * 10000));
 	lightLevel.max_content_light_level = OSSwapHostToBigInt16((uint16_t)metadata.maxContentLightLevel);
 	lightLevel.max_pic_average_light_level = OSSwapHostToBigInt16((uint16_t)metadata.maxFrameAverageLightLevel);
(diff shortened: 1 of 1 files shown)


cdavis5e (approved):

gnattu:
I'm afraid this change alone is not enough to fix the problem described in #2399. It appears that `CAEDRMetadata` is also a managed objc object which the autorelease pool might trying to release and still cause the double free crash. [Chromium](https://chromium.googlesource.com/chromium/src/+/refs/heads/main/components/metal_util/hdr_copier_layer.mm#351) does not do explicit release on this object and I *think* we should do the same.

This is already reproducible by using mpv with gpu-next and `target-colorspace-hint=yes`, play an HDR file, enter the full screen then quit the full screen, then close the player, you would see a segfault. This segfault disappears after removing the explicit release of `CAEDRMetadata`

m154k1:
> This is already reproducible by using mpv with gpu-next and `target-colorspace-hint=yes`, play an HDR file, enter the full screen then quit the full screen, then close the player, you would see a segfault.

I can confirm this.

> This segfault disappears after removing the explicit release of `CAEDRMetadata`

@gnattu could you please upload the patch?

gnattu:
To test this for mpv it is quite trivial, you just remove this line:
https://github.com/KhronosGroup/MoltenVK/blob/138415d56890aacd1bcdc2477f2219da9f2acbad/MoltenVK/MoltenVK/GPUObjects/MVKSwapchain.mm#L367

But since the autorelease pool might not exist everywhere, I think the more proper way of doing this is to wrap the usage of `CAEDRMetadata` in `@autoreleasepool {...}` block

billhollings:
> I think the more proper way of doing this is to wrap the usage of `CAEDRMetadata` in `@autoreleasepool {...}` block

Fixed in #2436.

PR #2404, opened 2024-12-05
Work around 10.15 AMD driver's broken blit encoder timestamp sampling

etang-cw:
macOS 10.15's AMD driver is broken and crashes if you try to encode a `sampleCountersInBuffer:` to a blit command encoder before encoding any actual work.


billhollings (commented):
Nice catch! And thanks for submitting a fix!

PR #2383, opened 2024-10-19
MVKPixelFormats: Add support for B5G6R5_UNORM_PACK16 and B5G5R5A1_UNORM_PACK16 using swizzle.

squidbus:
Adds support for `VK_FORMAT_B5G6R5_UNORM_PACK16` and `VK_FORMAT_B5G5R5A1_UNORM_PACK16` using swizzles, same as how 4444 format variations are supported.

These formats are pretty broadly supported elsewhere, so it's helpful to have support for them out-of-the-box on Apple systems as well.

PR #2405, opened 2024-12-08
MVKPixelFormats: Add support for more B8G8R8A8 formats using swizzle.

squidbus:
Similar to https://github.com/KhronosGroup/MoltenVK/pull/2383, adds support for some more B8G8R8A8 formats using swizzle.

I've tested this with B8G8R8A8 storage images in [shadPS4](https://github.com/shadps4-emu/shadPS4). We don't have proper support for emulating arbitrary storage image swizzles used by the emulated game yet, but we have a simple path for emulating RGBA -> BGRA storage image swizzles that changes out the Vulkan format. Adding these formats helps us with that.


cdavis5e (approved):
I had to make sure you got the swizzle right. They can be quite tricky at times, but in this case, this should be pretty straightforward.

PR #2407, opened 2024-12-09
GHActions: Update legacy build to macOS 13 / Xcode 14

etang-cw:
macOS 12 runners are gone: https://github.com/actions/runner-images/issues/10721


billhollings (approved):
Thanks for submitting this!

I had noticed the same, and was going to fix this myself, but you beat me to it!

PR #2411, opened 2024-12-13
Fix compile with MVK_USE_CEREAL=0

squidbus:
Ran into this while testing with a local compile. In `MVKPipeline.mm`, it seems that with `MVK_USE_CEREAL=0`, `sstream` is not included, but `stringstream` is used.


billhollings (approved):
Thanks for catching this, and for the fix.

Issue #2414, opened 2024-12-29
This device does not support ulong texture atomics, so MTLPixelFormatRG32Uint cannot be used with texture atomics


squidbus:
Have some users getting this error on M1 family GPUs:
```
-[MTLTextureDescriptorInternal validateWithDevice:]:1405: failed assertion `Texture Descriptor Validation
This device does not support ulong texture atomics, so MTLPixelFormatRG32Uint cannot be used with texture atomics
```

`MTLPixelFormatRG32Uint` with atomics does indeed seem to be unsupported on those GPUs according to the spec, but there are two concerns I have here:
* The shader in question does not appear to actually be using image atomics, so I'm unclear why this error is appearing. Something to do with how MoltenVK applies `MTLTextureUsageShaderAtomic`?
* Regardless, a support flag for this doesn't appear to be part of the Vulkan spec, so I feel this should be falling back to atomics emulation if it was actually needed. Although, I don't think reading RG32 at 64-bit atomic is part of the Vulkan spec at all so I'm not sure why this would ever happen?

squidbus:
Should be fixed by https://github.com/KhronosGroup/MoltenVK/pull/2415

PR #2415, opened 2024-12-29
Remove atomic usage for RG32Uint.

squidbus:
Looking into https://github.com/KhronosGroup/MoltenVK/issues/2414, it appears that 64-bit atomics for RG32 texture is not supported by Vulkan at this time, and code for it was only added to prepare for if it is in the future.

However, marking every texture with this format as atomic not only could have performance implications, it is also causing issues on devices like M1 family GPUs that support native texture atomics but not 64-bit atomics. Thus it is better to remove it and revisit if Vulkan ever does decide to support this.

I left the part enabling MTLPixFmtCaps as it already has a flag to force-disable it since it isn't supported by Vulkan along with a comment explaining, and thus it seems pretty harmless. Although I removed enabling it on `RG32Sint` as according to the spec Metal only supports it for `RG32Uint`.

### DIFF
--- a/MoltenVK/MoltenVK/GPUObjects/MVKPixelFormats.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKPixelFormats.mm
@@ -748,7 +748,7 @@ MTLTextureUsage MVKPixelFormats::getMTLTextureUsage(VkImageUsageFlags vkImageUsa
 	}
 
 #if MVK_XCODE_15
-	if (supportAtomics && (mtlFormat == MTLPixelFormatR32Uint || mtlFormat == MTLPixelFormatR32Sint || mtlFormat == MTLPixelFormatRG32Uint)) {
+	if (supportAtomics && (mtlFormat == MTLPixelFormatR32Uint || mtlFormat == MTLPixelFormatR32Sint)) {
 		mvkEnableFlags(mtlUsage, MTLTextureUsageShaderAtomic);
 	}
 #endif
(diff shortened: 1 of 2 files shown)


MoltenVK/MoltenVK/GPUObjects/MVKPixelFormats.mm
```diff
@@ -1542,7 +1542,6 @@
 	// Including this here so we remember to update this if support is added to Vulkan in the future.
 	bool atomic64 = noVulkanSupport && (gpuCaps.supportsApple9 || (gpuCaps.supportsApple8 && gpuCaps.supportsMac2));
 	enableMTLPixFmtCapsIf( atomic64, RG32Uint, Atomic );
-	enableMTLPixFmtCapsIf( atomic64, RG32Sint, Atomic );
```
cdavis5e, line 1545: Why did you delete one line, but not the other?

cdavis5e (changes requested):

MoltenVK/MoltenVK/GPUObjects/MVKPixelFormats.mm
squidbus: Mentioned this in the description, this part being left here seems benign since it's disabled by a bool and has a comment that its left as a reminder, but according to the Metal spec 64-bit texture atomics are only supported for RG32Uint, not RG32Sint, so that had to be removed regardless.

squidbus: I can remove the whole section if you want though, since it's not being used.

cdavis5e: Ah, OK. Silly me. You can leave it in.

cdavis5e (approved):

PR #2419, opened 2025-01-13
MVKCmdDraw: Fix indirect index for triangle fan topology

### LINKED ISSUE #2418: Rendering artifact when using ANGLE's vulkan backend with MoltenVK
I'm using qemu+virgl for 3d-accelerated linux vm with akihikodaki's patches (details available [here](https://gist.github.com/akihikodaki/87df4149e7ca87f18dc56807ec5a1bc5)). The host-side virgl renderer is backed by ANGLE (which itself can use different backends). I recently tried to play with different ANGLE backends but found if I use the vulkan backend it will result in heavy rendering artifact with moltenvk (while it renders fine with ANGEL's default OpenGL backend). Here's some more details.

## Environment
* macOS 15.2 on Apple M1 Max
* Vulkan SDK 1.3.296
* ANGLE: https://chromium.googlesource.com/angle/angle/+/81a41e4f10f960a290d89147c7a152fcd48f09a4
* Other components are the same as described [here](https://gist.github.com/akihikodaki/87df4149e7ca87f18dc56807ec5a1bc5).

## Issue description

First, to provide a little bit more background, the patched qemu on macos provides 3d acceleration to guest in the following way:
```
Guest GL --(virgl)-> Host GLES on EGL --(ANGLE)-> Backends (OpenGL / Metal / Vulkan)
```

When I choose ANGLE's backend to vulkan (using configuration when initializing EGL), the rendering will contain heavy artifact when the VM boots into operating system. For example, here is a screenshot at the gdm login screen:

gdm-artifact

while a correct rendering should look like the following:

gdm-correct

## Preliminary analysis

I'm aware that a chain with multiple steps is involved in the rendering process and mistake in any step can make result incorrect. Specifically, there are the following possibilities:
1. virgl generates incorrect gles commands on host (unlikely, as discussed below)
2. ANGLE's vulkan backend translates gles input incorrectly
3. mvk translates vulkan to metal incorrectly

I'm not 100% sure for now but I'm sharing my data to provide some clarification and they may serve as pointers to further diagnose the issue.

I used two capture tools to record the command stream at different stages of the chain. First, I recorded the OpenGL ES command stream at the input of ANGLE (with a heavily adapted version of apitrace). Meanwhile, I used gfxreconstruct to record the vulkan command stream generated by ANGLE. I attached both files recorded from a single vm launch at the same time:
* [GLES+EGL trace from apitrace](https://drive.google.com/file/d/1v6kmtDxngIFn2dDy615nyzKNxssC8DLe/view?usp=sharing), can be replayed on linux
* [Vulkan trace from gfxreconstruct](https://drive.google.com/file/d/1Cd0mgOg7_fCN9lsKTQgrV4gO4JLr-vLV/view?usp=drive_link)

Actually the first trace can be correctly rendered when replayed directly on Linux. In fact, the second screenshot above is taken from there. This means that step 1 (virgl) should be most likely correct. However, the vulkan trace cannot help to a equal degree because it is apparently not portable across platforms (and the desktop doesn't even render when replayed on macos, not very sure if the issue is with recording or rendering replay). So now I cannot tell if this is ANGLE's or mvk's fault.

Further I can try to build ANGLE with vulkan on linux and try to see the result from native vulkan rendering from the same gles input, but I cannot promise to get it done very soon. But please let me know if I can provide any clarification or other necessary information.

dboyan:
When we call vkCmdDraw with vertexCount=N and firstVertex=X, we only need an array with N element, X, X+1, ..., X+N-1, to generate the indirect indices. However, the current logic generates all the way from 0 to X+N-1 (which is X+N elements) while only reserving N spaces. This causes issue when the unreserved part is overwritten, for example, when issuing consecutive draw calls with triangle fan primitives.

Fixes #2418

### DIFF
--- a/MoltenVK/MoltenVK/Commands/MVKCmdDraw.mm
+++ b/MoltenVK/MoltenVK/Commands/MVKCmdDraw.mm
@@ -112,19 +112,20 @@ void MVKCmdDraw::encodeIndexedIndirect(MVKCommandEncoder* cmdEncoder) {
 	auto* indirectIdxBuff = cmdEncoder->getTempMTLBuffer(indirectIdxBuffStride);
 	auto* pIndArg = (MTLDrawIndexedPrimitivesIndirectArguments*)indirectIdxBuff->getContents();
 	pIndArg->indexCount = _vertexCount;
-	pIndArg->indexStart = _firstVertex;
+	// let the indirect index point to the beginning of vertex index buffer below
+	pIndArg->indexStart = 0;
 	pIndArg->baseVertex = 0;
 	pIndArg->instanceCount = _instanceCount;
 	pIndArg->baseInstance = _firstInstance;
 
 	// Create an index buffer populated with synthetic indexes.
-	// Start populating indexes below _firstVertex so that indexes align with their corresponding vertexes
+	// Start populating indexes directly from the beginning and align with corresponding vertexes by adding _firstVertex
 	MTLIndexType mtlIdxType = MTLIndexTypeUInt32;
 	auto* vtxIdxBuff = cmdEncoder->getTempMTLBuffer(mvkMTLIndexTypeSizeInBytes(mtlIdxType) * _vertexCount);
 	auto* pIdxBuff = (uint32_t*)vtxIdxBuff->getContents();
-	uint32_t idxCnt = _firstVertex + _vertexCount;
-	for (uint32_t idx = 0; idx < idxCnt; idx++) {
-		pIdxBuff[idx] = idx;
+
+	for (uint32_t idx = 0; idx < _vertexCount; idx++) {
+		pIdxBuff[idx] = _firstVertex + idx;
 	}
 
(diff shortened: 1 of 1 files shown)


cdavis5e (approved):
At first, this looked wrong, but, studying it for a while, I'm more and more convinced that this is correct.

cdavis5e:
Oh, and, uh, thank you for your contribution. :)

cdavis5e:
Actually, thinking about this some more... it's occurred to me that a similar change might need to be made for `MVKCmdDrawIndirect::encodeIndirectIndexed()`.

/me checks...

...

Yes, a similar change needs to be made for the shader function, `cmdDrawIndirectPopulateIndexes()`. I could make an analogous change to fix this, or do you want to do the honors?

cdavis5e:
> I could make an analogous change to fix this,

I went ahead and made #2420.

PR #2420, opened 2025-01-13
MVKCmdDrawIndirect: Fix indirect index for triangle fans.

cdavis5e:
Analogously to #2419, this fixes generation to add `vertexStart` to the _generated_ indices, and start from zero in the array.

### DIFF
--- a/MoltenVK/MoltenVK/Commands/MVKCommandPipelineStateFactoryShaderSource.h
+++ b/MoltenVK/MoltenVK/Commands/MVKCommandPipelineStateFactoryShaderSource.h
@@ -334,14 +334,14 @@ kernel void cmdDrawIndirectPopulateIndexes(const device char* srcBuff [[buffer(0
     const device auto& src = *reinterpret_cast<const device MTLDrawPrimitivesIndirectArguments*>(srcBuff + idx * srcStride);\n\
 	device auto& dst = destBuff[idx];                                                                           \n\
     dst.indexCount = src.vertexCount;                                                                           \n\
-	dst.indexStart = src.vertexStart;                                                                           \n\
+	dst.indexStart = 0;                                                                                         \n\
 	dst.baseVertex = 0;                                                                                         \n\
 	dst.instanceCount = src.instanceCount;                                                                      \n\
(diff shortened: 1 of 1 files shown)


dboyan:
I'm not 100% sure yet if this is correct. The case with `drawCount`>1 might be interesting, but currently I haven't tested it. I tested `drawCount`=1 case with my own minimal program (written in gles and hooked to angle's vulkan backend because I'm not a vulkan expert), via `glDrawArraysIndirect`. It rendered correctly with or without the change (I guess separate calls don't suffer from buffer overwriting problem here). But I couldn't verify the more generic case as I didn't have the extension`GL_EXT_multi_draw_indirect` with angle. I'll see if I can obtain a vulkan version of test program.

dboyan:
I was able to get a working test program and now I can confirm the change isn't in the right direction. Actually the original code makes sense somewhat.

Let's consider the situation where we have 16 vertices, and we call vkCmdDrawIndirect with `drawCount=2`, using triangle fan primitive topology. The indirect buffer contains
```
1. { vertexCount=4, firstVertex=0 }
2. { vertexCount=4, firstVertex=8 }
```
Notice that the compute kernel `cmdDrawIndirectPopulateIndexes` handles both draws with 2 threads, and the resulting index are populated into a single buffer `idxBuff`. With the original logic, `idxBuff` will contain
```
[0, 1, 2, 3, X, X, X, X, 4, 5, 6, 7, X, X ...]    (X means uninitialized element)
```
This is the intended behavior because indices from multiple draws are put "in place" and will not interfere with each other. They will be further processed to generate indices for triangles according to the `destBuff` output of the current stage. It seems a waste of space but it's logically correct.

On the other hand, after the change, all the indices from different draw all start from the beginning of `idxBuff`, and that creates trouble. My #2420 was correct because vkCmdDraw can only issue a single draw, so one doesn't need to worry about the issue above.

In fact, even with the original code, the rendering is not 100% correct. The triangle fan translation seems a little off. I'll open a issue when I get time.

cdavis5e:
Yeah, I realized a bit after the fact that it was already adding the base before my change. Thanks for confirming.

Issue #2416, opened 2025-01-06
vkCmdFillBuffer crashes with Metal Shader Validation enabled


squidbus:
With `MTL_SHADER_VALIDATION=1` in my environment, I get the following crash after some number of `vkCmdFillBuffer` calls:

```
Crashed Thread:        2  Dispatch queue: com.apple.root.default-qos

...

Thread 2 Crashed::  Dispatch queue: com.apple.root.default-qos
0   <translation info unavailable>	      0x7001999b28 ???
1   shadps4                       	     0x200003eea8b 0x20000000000 + 4123275
2   libsystem_platform.dylib      	    0x7ff80affae1d _sigtramp + 29
3   ???                           	    0x600003cb7d80 ???
4   libobjc.A.dylib               	    0x7ff80abf63ba objc_autoreleasePoolPop + 235
5   libdispatch.dylib             	    0x7ff80ae1b455 _dispatch_call_block_and_release + 12
6   libdispatch.dylib             	    0x7ff80ae1c7e2 _dispatch_client_callout + 8
7   libdispatch.dylib             	    0x7ff80ae2c033 _dispatch_root_queue_drain + 872
8   libdispatch.dylib             	    0x7ff80ae2c5b2 _dispatch_worker_thread2 + 147
9   libsystem_pthread.dylib       	    0x7ff80afc0c3e _pthread_wqthread + 261
10  libsystem_pthread.dylib       	    0x7ff80afbfbdb start_wqthread + 15

...

Thread 35:: GpuCommandProcessor
0   ???                           	    0x7ff89b47aa84 ???
1   libsystem_kernel.dylib        	    0x7ff80af85aaa __psynch_cvwait + 10
2   libsystem_pthread.dylib       	    0x7ff80afc47a8 _pthread_cond_wait + 1193
3   libc++.1.dylib                	    0x7ff80af08a8d std::__1::condition_variable::__do_timed_wait(std::__1::unique_lock<std::__1::mutex>&, std::__1::chrono::time_point<std::__1::chrono::system_clock, std::__1::chrono::duration<long long, std::__1::ratio<1l, 1000000000l>>>) + 93
[...]

PR #2421, opened 2025-01-14
Fix crash when shader validation is enabled.

squidbus:
I didn't get any feedback on https://github.com/KhronosGroup/MoltenVK/issues/2416 whether this would be an acceptable fix, so I'm opening it as a PR in case it's wanted.

Basically it seems that for some reason when Metal shader validation is enabled, there may be a crash due to auto-release pool double-release on an `MTLComputePipelineDescriptorInternal`, but only when using `newComputePipelineStateWithFunction`. I'm not sure why this is happening or whether it's an internal Metal bug with the validation layer or MoltenVK's fault, but changing it to create a temporary `MTLComputePipelineDescriptor` and call `newComputePipelineStateWithDescriptor` resolves the issue for me.


cdavis5e (approved):
My only concern is that this wouldn't work on iOS 8, but I suspect there aren't many people who are going to care about that.

squidbus:
> My only concern is that this wouldn't work on iOS 8, but I suspect there aren't many people who are going to care about that.

If you're referring to the minimum version for `newComputePipelineStateWithDescriptor`, it is used for other pipelines already so there's no difference for compatibility using it here as well.

PR #2424, opened 2025-01-20
Advertise VK_KHR_shader_terminate_invocation.

squidbus:
`OpTerminateInvocation` is already handled in SPIRV-Cross identically to `OpKill`. Advertise support for the extension.

### DIFF
--- a/MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
@@ -473,6 +473,11 @@ void MVKPhysicalDevice::getFeatures(VkPhysicalDeviceFeatures2* features) {
 				formatFeatures->formatA4B4G4R4 = canSupport4444;
 				break;
 			}
+			case VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_DEPTH_CLIP_CONTROL_FEATURES_EXT: {
+				auto* depthFeatures = (VkPhysicalDeviceDepthClipControlFeaturesEXT*)next;
+				depthFeatures->depthClipControl = true;
+				break;
+			}
 			case VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_EXTENDED_DYNAMIC_STATE_FEATURES_EXT: {
 				auto* extDynState = (VkPhysicalDeviceExtendedDynamicStateFeaturesEXT*)next;
 				extDynState->extendedDynamicState = true;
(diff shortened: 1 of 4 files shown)


MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
```diff
@@ -582,6 +582,11 @@
 				shaderIntFuncsFeatures->shaderIntegerFunctions2 = true;
 				break;
 			}
+			case VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_SHADER_TERMINATE_INVOCATION_FEATURES_KHR: {
+				auto* terminateFeatures = (VkPhysicalDeviceShaderTerminateInvocationFeaturesKHR*)next;
+				terminateFeatures->shaderTerminateInvocation = true;
+				break;
+			}
```
billhollings, line 589: `VkPhysicalDeviceShaderTerminateInvocationFeaturesKHR` also needs to be added to `MVKDeviceFeatureStructs.def`, between these two:

https://github.com/KhronosGroup/MoltenVK/blob/138415d56890aacd1bcdc2477f2219da9f2acbad/MoltenVK/MoltenVK/GPUObjects/MVKDeviceFeatureStructs.def#L66-L67

Also, pay attention to ordering, both as described in `MVKDeviceFeatureStructs.def`, and in this section, as described in the comment above here:

https://github.com/KhronosGroup/MoltenVK/blob/138415d56890aacd1bcdc2477f2219da9f2acbad/MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm#L265

billhollings (changes requested):
Thanks for submitting this!

Please see notes below about also updating `MVKDeviceFeatureStructs.def`.

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
squidbus: Fixed, thanks for pointing this out. I also added `DepthClipControl` in here as it was missed in that PR, not sure if there's any other extension inconsistencies in this file that might need addressing separately.

MoltenVK/MoltenVK/GPUObjects/MVKDeviceFeatureStructs.def
```diff
@@ -64,7 +64,9 @@ MVK_DEVICE_FEATURE(VariablePointer,                   VARIABLE_POINTER,
 MVK_DEVICE_FEATURE(VulkanMemoryModel,                 VULKAN_MEMORY_MODEL,                    3)
 MVK_DEVICE_FEATURE_EXTN(FragmentShaderBarycentric,    FRAGMENT_SHADER_BARYCENTRIC,     KHR,   1)
 MVK_DEVICE_FEATURE_EXTN(PortabilitySubset,            PORTABILITY_SUBSET,              KHR,  15)
+MVK_DEVICE_FEATURE_EXTN(ShaderTerminateInvocation,    SHADER_TERMINATE_INVOCATION,     KHR,   1)
 MVK_DEVICE_FEATURE_EXTN(4444Formats,                  4444_FORMATS,                    EXT,   2)
+MVK_DEVICE_FEATURE_EXTN(DepthClipControl,             DEPTH_CLIP_CONTROL,              EXT,   1)
```
cdavis5e, line 69: Bah. Maybe I shouldn't have the right to merge changes...

cdavis5e (approved):

MoltenVK/MoltenVK/GPUObjects/MVKDeviceFeatureStructs.def
billhollings: It would be nice to figure out a compile-time assertion that could compare that case statements in `getFeatures()` to `MVKDeviceFeatureStructs.def`, but it would be a challenge at best, and maybe not even possible.

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
billhollings: > I also added `DepthClipControl` in here

Thanks. Good catch!

billhollings (approved):
Thanks. LGTM now.

PR #2425, opened 2025-01-21
Improved estimation of vertex attribute buffer count when reserving for implicit buffer

dboyan:
Currently, the vertex attribute buffer count estimation (to reserve binding for implicit buffers) is too generous. This merge request improves the estimation by introducing one case from the actual allocation logic, `addVertexInputToPipeline`, i.e., synthetic binding buffer if the vertex attribute offset is 0. This helps in reducing the required binding slots the majority of pipelines generated by ANGLE and allows vertex shaders that uses more than 10 attribute inputs to work.

In the meantime, the change brings bindings of implicit buffers and vertex attributes into close vicinity, and exposes a bug where implicit buffer interferes with vertex attribute at the same binding index across draw calls in the same render pass. This is masked by the previous over-conservative estimation. The second change fix the bug by adding dirty marks.

The only other (first) change is trivial style fix, and does not affect the visible behavior of the program.

### DIFF
--- a/MoltenVK/MoltenVK/Commands/MVKCommandEncoderState.mm
+++ b/MoltenVK/MoltenVK/Commands/MVKCommandEncoderState.mm
@@ -942,6 +942,7 @@ void MVKGraphicsResourcesCommandEncoderState::encodeImpl(uint32_t stage) {
                        });
 
 	} else if (!forTessellation && stage == kMVKGraphicsStageRasterization) {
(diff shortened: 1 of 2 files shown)


cdavis5e (commented):
I should run this against the CTS and see if anything breaks...

billhollings:
> I should run this against the CTS and see if anything breaks...

CTS looks good. Pulling this in.

Issue #2403, opened 2024-12-04
Enabling Metal argument buffers in 1.2.10+ breaks rendering in Dolphin


OatmealDome:
It appears that using Metal argument buffers, which is enabled by default in 1.2.10+, breaks rendering in Dolphin. Output is either a black screen or corrupted:

Expected Output

![image](https://github.com/user-attachments/assets/cdd2a6bf-e978-4790-8d9e-2c9ebe818940)

Apple M1

![image](https://github.com/user-attachments/assets/d1c2a8c7-8a90-438f-8f20-51a00b2f33ba)

Intel UHD 630

![image](https://github.com/user-attachments/assets/69a3c0db-0e03-4fde-87f6-4640c306d1d0)

Here's a GPU trace from the M1: [LineWidthTest.gputrace.zip](https://github.com/user-attachments/files/18013831/LineWidthTest.gputrace.zip)

If you would like to run the test for yourself, you can download a build of Dolphin with the latest MoltenVK commit (`1d2ec638532268d076a550262df2e5e5e4fd2c9e`) here: [pr-13205-dolphin-latest-universal.dmg](https://dl.dolphin-emu.org/prs/17/91/pr-13205-dolphin-latest-universal.dmg). Switch the video backend in Graphics button -> General tab to Vulkan, and then drag and drop this file on the Dolphin window to start rendering the test image: [LineWidthTest.dff](https://fifo.ci/media/dff/LineWidthTest.dff).

Turning off Metal argument buffers with `MVK_CONFIG_USE_METAL_ARGUMENT_BUFFERS=0` resolves the issue.

I was able to reproduce the issue on:

* M1 Pro / Sequoia 15.2 beta 4 (black screen)
* M1 / Sonoma 14.6.1 (black screen)
* Intel UHD 630 / Sequoia 15.1 (corruption)

I was unable to reproduce the issue on:

* AMD Radeon Pro 5500M / Sequoia 15.1

billhollings:
Sorry for the delay in getting to this.

Unfortunately, it seems to run fine on my M3 Pro. I don't have an M1 to verify your specific condition.

I was not able to start your app from Xcode, so I was not able to either set environment variables, or review the sys err logs to confirm state such as argument buffer use. If you can let me know how to do those, I'll have another shot at it.

squidbus:
For the Intel corruption specifically, I've been seeing the same in multiple apps on Intel GPUs.

dboyan:
[...]

PR #2427, opened 2025-01-24
WIP: MVKDescriptor: Prevent batched write combined image-sampler from overwriting each other in argument buffer

dboyan:
Currently samplerXXArray is translated to one textureXX_array and a single sampler in MSL, consuming 2 bindings in Metal regardless of the array size. This is already not an accurate translation. In the meantime, the descriptor set update does not consider the difference in numbers between the approaches. Despite the inaccuracy, the current solution "works" in some cases when not using argument buffers. They are broken when argument buffer is enabled.

The underlying reason is that image (texture) and sampler are in the same index space when used in argument buffers, and when updated in a batch (as sampler array) using combined image-sampler descriptor type, the sampler is overwritten by the next texture. This change avoids this kind of overwriting.

The ultimate solution would be properly implement texture/sampler arrays and account for the semantic differences between Vulkan and Metal in descriptor sets. But let's make it less broken first as the appropriate solution would require considerable effort.

Not a full fix, but addresses the symptom of #2403 .

### DIFF
--- a/MoltenVK/MoltenVK/GPUObjects/MVKDescriptor.h
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKDescriptor.h
@@ -449,6 +449,14 @@ public:
 			   size_t srcStride,
 			   const void* pData) override;
 
+	void writeCombined(MVKDescriptorSetLayoutBinding* mvkDSLBind,
+					   MVKDescriptorSet* mvkDescSet,
+					   uint32_t dstIdx,
+					   uint32_t srcIdx,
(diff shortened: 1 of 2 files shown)


dboyan:
Now I think what I said above is not fully correct. The issue I referenced might also have the incorrect application itself as a contributing factor. However, I think we still need some sort of fix because the current argument buffer logic makes the behavior too broken.

V380-Ori:
This fixes another emulator — **Ryujinx**. Thank you!

Issue #2428, opened 2025-01-26
Dynamic vertex stride does not seem to be working with tessellation


squidbus:
It appears that when doing a draw with tessellation, `VK_DYNAMIC_STATE_VERTEX_INPUT_BINDING_STRIDE` does not function properly. Instead, every vertex appears to be getting the exact same attributes in. Removing the dynamic state and using static stride fixes the issue with no other changes to my code.

squidbus:
Fixed in https://github.com/KhronosGroup/MoltenVK/pull/2429

PR #2429, opened 2025-01-26
Fix dynamic vertex stride with tessellation.

squidbus:
Updates the compute-based vertex shader path for tessellation to take into account the dynamic vertex binding stride, when enabled for the pipeline, in the same fashion as regular vertex bindings.

Fixes https://github.com/KhronosGroup/MoltenVK/issues/2428


italomandara:
Quick question about ShadPS4:
are those changes already included in a custom version of MVK bundled in ShadPS4 pre-releases or are you waiting those to be merged?

squidbus:
> Quick question about ShadPS4: are those changes already included in a custom version of MVK bundled in ShadPS4 pre-releases or are you waiting those to be merged?

I'd rather keep discussion here specifically about MoltenVK, but no, I don't include anything that hasn't landed upstream here yet.

cdavis5e (approved):
Nice catch.

PR #2431, opened 2025-01-29
Use concurrent compute command encoders.

js6i:
May improve performance in some cases although it didn't do much to the game I was writing this for.


billhollings (approved):
LGTM. Thanks for submitting.

dboyan:
Seems that this change breaks some compute workload on some AMD RX6000 systems. See #2458. If this is caused by a driver bug, we should probably make concurrent encoder conditional.

PR #2430, opened 2025-01-28
Add support for VK_EXT_depth_clip_control

squidbus:
SPIRV-Cross already has a flag for converting from `[-w, w]` depth to `[0, w]`. Take advantage of that to implement `VK_EXT_depth_clip_control`.

Ran CTS tests `dEQP-VK.pipeline.monolithic.depth.depth_clip_control.*`, results:

```
Test run totals:
  Passed:        36/72 (50.0%)
  Failed:        0/72 (0.0%)
  Not supported: 36/72 (50.0%)
  Warnings:      0/72 (0.0%)
  Waived:        0/72 (0.0%)
```

With the general case for `Not supported` just being tests using unsupported depth formats.

PR #2432, opened 2025-01-29
Fix not saving shouldFixupClipSpace when serializing a pipeline.

squidbus:
Small fix pointed out in https://github.com/KhronosGroup/MoltenVK/pull/2430#issuecomment-2621122639; need to add `shouldFixupClipSpace` option to pipeline serialization.


cdavis5e (approved):
/me smacks forehead

This is as much my fault as it is yours, since I should've caught this before recklessly approving your change.

Issue #2423, opened 2025-01-17
MoltenVK bug people are discussing re: ollama, llama.cpp


FellowTraveler:
https://github.com/ollama/ollama/issues/1016#issuecomment-2572959904

![Image](https://github.com/user-attachments/assets/7b0c8c52-77fa-4760-a99b-f9627f9f68a1)

dboyan:
I'm not an expert but I guess people should first figure out Vulkan code is *correct*. Especially when seeing this kind of messages (from the second issue link):

> VUID-VkWriteDescriptorSet-descriptorType-00328(ERROR / SPEC): msgNum: -368569266 - Validation Error: [ VUID-VkWriteDescriptorSet-descriptorType-00328 ] | MessageID = 0xea08144e | vkUpdateDescriptorSets(): pDescriptorWrites[0].pBufferInfo[0].offset (31457794) must be a multiple of device limit minStorageBufferOffsetAlignment 16 when descriptor type is VK_DESCRIPTOR_TYPE_STORAGE_BUFFER. The Vulkan spec states: If descriptorType is VK_DESCRIPTOR_TYPE_STORAGE_BUFFER or VK_DESCRIPTOR_TYPE_STORAGE_BUFFER_DYNAMIC, the offset member of each element of pBufferInfo must be a multiple of VkPhysicalDeviceLimits::minStorageBufferOffsetAlignment (https://vulkan.lunarg.com/doc/view/1.3.296.0/mac/1.3-extensions/vkspec.html#VUID-VkWriteDescriptorSet-descriptorType-00328)

I would regard it as pretty serious issue because offset alignment requirements are there for a reason (hardware limit), and if the user doesn't obey it, it *will* cause errors (e.g., the kernels cannot read/write data correctly).

One potential way to verify the issue is to run the same program on native vulkan implementation on another OS. If it is also producing incorrect result there, then the vulkan code itself is problematic (and I think vulkan on other mainstream GPUs usually pose more restrictions on alignment). If not, then we can start suspecting MoltenVK. In that case, it would be ideal if there is some minimal failed tests and try to [capture metal gpu trace](https://github.com/KhronosGroup/MoltenVK/blob/main/Docs/MoltenVK_Runtime_UserGuide.md#trace_tool) and analyze the behavior.

soerenkampschroer:
[...]

PR #2434, opened 2025-02-02
PoC: MVKShaderLibrary: Handle specializtion with macros

dboyan:
The converted MSL may use macro instead of function constants to realize spirv specialization constant for various reasons (e.g. when the constant is used as array size).

In this case, we should define the macros at shader compilation stage and generate different variants of the metal shader library depending on macro-value mapping to make specializtion work properly when we cannot rely on metal's specialization.

The current version is hacky as it scans the shader for macro each time we create the function, and guesses the macro based on spirv-cross's habit. Ideally, we should obtain the information from spirv-cross instead.

Although the current form is not ready to merge but I would like to throw it in to see how people think about the approach. I would appreciate insights and suggestions on API design and coding as I'm not an objc guru or equipped with a good knowledge of how the internal API should be.

Fixes: #2423


dboyan:
Closing this in favor of #2441, as the required SPIRV-Cross API has been mainlined.

PR #2435, opened 2025-02-04
Implement basic support for VK_EXT_tooling_info.

squidbus:
Implements basic support for `VK_EXT_tooling_info`.

Everything is wired up for the `MVKPhysicalDevice` to provide tool properties, but it currently just returns that there are no tools, since (as far as I can tell) Metal does not have a standard way of reporting any attached tools on the Metal side of things. However, with support now advertised, applications can now properly detect support and get information on Vulkan-side tools from the Vulkan loader in the middle, such as the Vulkan Validation Layers if attached.

Passes both available CTS tests `dEQP-VK.api.tooling_info.validate_getter` and `dEQP-VK.api.tooling_info.validate_tools_properties`.


billhollings (approved):
Thanks! LGTM.

PR #2436, opened 2025-02-05
Autorelease CAEDRMetadata in MVKSwapchain

gnattu:
This is a followup of #2400. `CAEDRMetadata` is a type that has internal use of `autorelease`. As there is no other alternatives, wrap its usage in an `@autoreleasepool {...}` block instead of doing explict release, which would cause segfaults from double release.


cdavis5e (approved):
I should've known. Most class factory methods return autoreleased. (AFAIK only `+[NSObject new]` doesn't.)

PR #2441, opened 2025-02-10
MVKShaderLibrary: Handle specializtion with macros (v2)

dboyan:
The converted MSL may use macro instead of function constants to realize spirv specialization constant for various reasons (e.g. when the constant is used as array size).

In this case, we should define the macros at shader compilation stage and generate different variants of the metal shader library depending on macro-value mapping to make specialization work properly when we cannot rely on metal's specialization.

There have been multiple use cases that require the proper handling we are missing, including llama.cpp and Vulkan CTS tests on `VK_KHR_zero_initialize_workgroup_memory` (we haven't supported it yet but I'm working on it).

This PR supersedes #2434, and is intended to be a version that can be mainlined. It is almost functionally equivalent to the previous PR (with some correctness fixes), but uses a new API from SPIRV-Cross to obtain the information of macros instead of unreliable heuristics to guess from the translated shader code. The new SPIRV-Cross API has been merged (https://github.com/KhronosGroup/SPIRV-Cross/pull/2442).

~~There is one caveat though. Currently I assume the specialization constants (that are turned into macros) are 32-bit integers. It works on all uses I have seen because they can only be sizes of arrays, which *are* integers and usually represented with 32-bit values. I'm not very sure if my handling is enough or the most appropriate approach here.~~
The current version is taking types into account.

Fixes: #2423


billhollings (approved):
Thanks for all this work, both here and in SPIRV-Cross!

LGTM. Is this ready to be pulled in now?

dboyan:
Yes. I think so.

PR #2448, opened 2025-02-14
Store the residency set in MVKPhysicalDevice.

### LINKED ISSUE #2444: Failed assertion 'IOGPUMetalCommandQueue: command queue residency set limit of 32 exceeded' in VK CTS
Recent version of MoltenVK will crash some vulkan CTS test with the failed assertion as shown in the title. An example test to trigger the failure is `dEQP-VK.api.object_management.multithreaded_shared_resources.device`.

The usage of residency set is introduced in #2402, seems that this is problematic. Reverting the merge resolves the problem.

js6i:
@billhollings Attaching once to the MTLQueue seems like the more efficient thing to do, but on the other hand Metal presumably caches things and does not do big intersections repeatedly when the same set is attached to all command buffers. If we care about having separate sets for each device that's something to consider.

Fixes #2444

### DIFF
--- a/MoltenVK/MoltenVK/GPUObjects/MVKDevice.h
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKDevice.h
@@ -395,6 +395,12 @@ public:
 	/** Returns the MTLDevice capabilities. */
 	const MVKMTLDeviceCapabilities getMTLDeviceCapabilities() { return _gpuCapabilities; }
 
+	/** Adds the global residency set to a Metal queue. */
+	void addResidencySet(id<MTLCommandQueue> queue) {
+#if MVK_XCODE_16
+		if (_mtlResidencySet) [queue addResidencySet:_mtlResidencySet];
+#endif
+	}
 
 #pragma mark Construction
 
@@ -464,6 +470,9 @@ protected:
 	VkPhysicalDeviceTexelBufferAlignmentPropertiesEXT _texelBuffAlignProperties;
 	VkPhysicalDeviceMemoryProperties _memoryProperties;
 	MVKSmallVector<MVKQueueFamily*, kMVKQueueFamilyCount> _queueFamilies;
+#if MVK_XCODE_16
+	id<MTLResidencySet> _mtlResidencySet = nil;
+#endif
 	VkExternalMemoryProperties _hostPointerExternalMemoryProperties;
 	VkExternalMemoryProperties _mtlBufferExternalMemoryProperties;
 	VkExternalMemoryProperties _mtlTextureExternalMemoryProperties;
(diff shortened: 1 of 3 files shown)


MoltenVK/MoltenVK/GPUObjects/MVKDevice.h
```diff
@@ -830,9 +839,9 @@ class MVKDevice : public MVKDispatchableVulkanAPIObject {
 	void makeResident(id allocation) {}
 #else
 	void makeResident(id<MTLAllocation> allocation) {
```
billhollings, line 841: Since we're moving the residency set to `MVKPhysicalDevice`, these functions will be more consistent if moved there too.

Then elsewhere, just call `getPhysicalDevice()->makeResident()` where you now call `getDevice()->makeResident()`.

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
```diff
@@ -1952,6 +1952,23 @@
 #endif
 
 #endif  // MVK_MACOS
+
+#if MVK_XCODE_16
+	if (_isUsingMetalArgumentBuffers && _metalFeatures.residencySets) {
+		MTLResidencySetDescriptor *setDescriptor;
+		setDescriptor = [MTLResidencySetDescriptor new];
```
billhollings, line 1959: No biggie, but I'm never a fan of uninitialized vars. It would read nicer if these two lines were combined.

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
```diff
@@ -1952,6 +1952,23 @@
...
+		setDescriptor.label = @"Primary residency set";
+		setDescriptor.initialCapacity = 256;
+
+		NSError *error;
+		_mtlResidencySet = [_mtlDevice newResidencySetWithDescriptor:setDescriptor
+															   error:&error];
+		if (error) {
+			reportMessage(MVK_CONFIG_LOG_LEVEL_ERROR, "Error allocating residency set: %s", error.description.UTF8String);
```
billhollings, line 1967: Prefer to use `reportError(VK_ERROR_INITIALIZATION_FAILED, ...)` instead.

billhollings (changes requested):
> Attaching once to the MTLQueue seems like the more efficient thing to do, but on the other hand Metal presumably caches things and does not do big intersections repeatedly when the same set is attached to all command buffers. If we care about having separate sets for each device that's something to consider.

I assume another way of fixing this would be to stick with your existing design around `MVKDevice`, and simply call `removeResidencySet:` in the `MVKQueue` destructor? That would be a good RAII complement to the existing `addResidencySet:` in the `MVKQueue` constructor, and would seem to better track to the idea of tracking separate sets per device.

If you want to stick with this design, around physical device, I've added some minor review recommendations.

Also, I'm not sure why it's asking you to renew your CLA, but please do so, before we can merge any changes.

billhollings:
Superseded and replaced by #2452.

PR #2452, opened 2025-02-18
MVKQueue: Remove residency set from MTLCommandQueue on destruction.

js6i:
MTLCommandQueues are shared, we need to clean up so that we don't leak and hit the limit of residency sets per queue.

Supersedes #2448 

Fixes issue #2444.


billhollings (approved):
LGTM. Thanks!

PR #2462, opened 2025-03-04
Revert "Use concurrent compute command encoders."

### LINKED ISSUE #2458: Concurrent command encoder breaks compute workload on AMD RX 6000
This is discovered when debugging another issue with specialization constant (See https://github.com/KhronosGroup/MoltenVK/issues/2423#issuecomment-2676234155). A user discovered that [llama.cpp](https://github.com/ggml-org/llama.cpp) still doesn't work properly on the AMD RX 6000 system with the latest fix (#2441), which is known to work on apple silicon and AMD RX 5000 systems. The problem is later located to the introduction of concurrent command encoder for compute pipeline (#2431). Reverting the change helps fixing issue on the AMD RX 6000 system.

I'm not very sure if the problem is with #2431 or the driver. Actually I can't find a problem with #2431 myself. If it is a driver bug, we should properly consider enabling concurrent command encoder for compute either with environment variable or by checking the device id.

SRSaunders:
This reverts commit 835f85ecfed65bf660fd2d15a543e2c4ae2258f6.

This addresses concerns raised in #2458 regarding breakage on AMD GPUs, until such time as a replacement PR can be developed that works on both M-series GPUs and AMD GPUs.

This PR tested on my AMD 6600 GPU as working properly again using Sascha Willems' _computecloth_ example.

Fixes #2458.


billhollings (approved):
Thanks for submitting this!

Issue #2261, opened 2024-07-03
ImGui flickering when use mutable swapchain format for two dynamic renderings.


stripe2933:
I'm not sure if this error is related to ImGui or MoltenVK. However, since this flickering is not observed when testing the same code in a Windows environment with NVIDIA drivers, I am raising this issue with MoltenVK.

ImGui does not support proper gamma correction for the `B8G8R8A8_SRGB` format. Therefore, I am using the `VK_KHR_mutable_swapchain_format` extension and specifying the two formats, `B8G8R8A8_SRGB` and `B8G8R8A8_UNORM`, in `VkImageFormatListCreateInfo` in the `pNext` of `VkSwapchainCreateInfoKHR`. During ImGui rendering, I render to the swapchain image view using the `B8G8R8A8_UNORM` format. Here is some of the code I am using:

- Swapchain creation

```cpp
auto createSwapchain(
    vk::SwapchainKHR oldSwapchain
) -> vk::raii::SwapchainKHR {
    const vk::SurfaceCapabilitiesKHR surfaceCapabilities = gpu.physicalDevice.getSurfaceCapabilitiesKHR(surface);
    return { gpu.device, vk::StructureChain {
        vk::SwapchainCreateInfoKHR {
            vk::SwapchainCreateFlagBitsKHR::eMutableFormat,
            surface,
            std::min(surfaceCapabilities.minImageCount + 1, surfaceCapabilities.maxImageCount),
            vk::Format::eB8G8R8A8Srgb,
            vk::ColorSpaceKHR::eSrgbNonlinear,
            (swapchainImageExtent = surfaceCapabilities.currentExtent),
            1,
            vk::ImageUsageFlagBits::eColorAttachment,
            vk::SharingMode::eExclusive, {},
            surfaceCapabilities.currentTransform,
            vk::CompositeAlphaFlagBitsKHR::eOpaque,
            vk::PresentModeKHR::eFifo,
            {},
            oldSwapchain,
        },
        vk::ImageFormatListCreateInfo {
            unsafeProxy({
                vk::Format::eB8G8R8A8Srgb,
                vk::Format::eB8G8R8A8Unorm,
            })
        },
    }.get() };
}
```

- ImGui initialization
```cpp
    IMGUI_CHECKVERSION();
    ImGui::CreateContext();

    ImGuiIO &io = ImGui::GetIO();
    int framebufferWidth, framebufferHeight;
[...]

PR #2465, opened 2025-03-13
Fix Metal API violation when using VK_KHR_swapchain_mutable_format.

stripe2933:
Texture view creation is not allowed from `framebufferOnly` textures. The flag should be disabled if texture view will be created from it.

Related: https://github.com/KhronosGroup/MoltenVK/issues/2261

### DIFF
--- a/MoltenVK/MoltenVK/GPUObjects/MVKSwapchain.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKSwapchain.mm
@@ -483,17 +483,8 @@ void MVKSwapchain::initCAMetalLayer(const VkSwapchainCreateInfoKHR* pCreateInfo,
 	mtlLayer.framebufferOnly = !mvkIsAnyFlagEnabled(pCreateInfo->imageUsage, (VK_IMAGE_USAGE_TRANSFER_SRC_BIT |
 																			  VK_IMAGE_USAGE_TRANSFER_DST_BIT |
 																			  VK_IMAGE_USAGE_SAMPLED_BIT |
-																			  VK_IMAGE_USAGE_STORAGE_BIT));
-
-	// Texture view creation is not allowed from framebufferOnly textures.
-	if (mtlLayer.framebufferOnly) {
-		for (const auto* next = (const VkBaseInStructure*)pCreateInfo->pNext; next; next = next->pNext) {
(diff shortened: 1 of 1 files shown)


MoltenVK/MoltenVK/GPUObjects/MVKSwapchain.mm
```diff
@@ -485,6 +485,16 @@ static CALayerContentsGravity getCALayerContentsGravity(VkSwapchainPresentScalin
...
 
+	// Texture view creation is not allowed from framebufferOnly textures.
+	if (mtlLayer.framebufferOnly) {
+		for (const auto* next = (const VkBaseInStructure*)pCreateInfo->pNext; next; next = next->pNext) {
+			if (next->sType == VK_STRUCTURE_TYPE_IMAGE_FORMAT_LIST_CREATE_INFO) {
+				mtlLayer.framebufferOnly = false;
+				break;
+			}
+		}
+	}
```
billhollings, line 496:
```suggestion
																			  VK_IMAGE_USAGE_STORAGE_BIT)) &&
								!mvkAreAllFlagsEnabled(pCreateInfo->flags, VK_SWAPCHAIN_CREATE_MUTABLE_FORMAT_BIT_KHR);
```

I think you can accomplish the same thing directly in the initial setting of `framebufferOnly`, by testing for `VK_SWAPCHAIN_CREATE_MUTABLE_FORMAT_BIT_KHR`. 

If so, it makes for a cleaner logic flow.

Does that makes sense?

billhollings (changes requested):
Good catch. Thanks for submitting this.

I've made a suggestion for streamlining the logic.

MoltenVK/MoltenVK/GPUObjects/MVKSwapchain.mm
stripe2933: Good suggestion!

stripe2933:
One thing I'm concerned about is [setting `framebufferOnly` to `false` may disable the optimization](https://developer.apple.com/documentation/quartzcore/cametallayer/framebufferonly). I'm not familiar with Metal, and I'm wondering is there a Metal specific way to treat the color attachment's format different from `CAMetalLayer`'s format, without creating an alias texture for it.

billhollings (approved):
LGTM now. Thanks.

Issue #2247, opened 2024-06-03
MVK Access to Indirect Command Buffers


EricAtPlanGrid:
My application (CAD visualization) is heavily draw-call bound. As a result we find that on larger models we spend upwards of 80% of frametime simply encoding these calls. I was able to prototype a simple Metal renderer which encodes the drawcalls upfront using ICBs, which was able to reduce the CPU usage from 100% during rendering to about 10%. 

My question here is in two parts:

1. is there a reason that Vulkan secondary command buffers cannot be treated as ICBs?
2. if the answer to one is "yes", have you considered / is there a known way to work around this somehow so that performance critical applications could still benefit from ahead-of-time encoding?

Thank you!

cdavis5e:
> 1. is there a reason that Vulkan secondary command buffers cannot be treated as ICBs?

Yes, because not everything that can be encoded to a secondary command buffer can be represented in an ICB. In particular, IMR GPUs on Intel Macs don't support compute from ICBs. (Admittedly, this particular instance should become less of a problem as Intel Macs become rarer and rarer.)

> 2. if the answer to one is "yes", have you considered / is there a known way to work around this somehow so that performance critical applications could still benefit from ahead-of-time encoding?

It should actually be mostly straightforward--I wouldn't quite say "trivial"--to implement secondary command buffers with ICBs. The big sticking point I think would be checking if all commands could be encoded to the ICB, and, if not, falling back to the old path.

PR #2481, opened 2025-04-11
[WIP] An attempt to allow Vulkan secondary command buffers to be encoded as Metal IndirectCommandBuffers

EricAtPlanGrid:
Following on from #2247 I finally took a very naive stab at this. Things appear to be more or less working up until the point where I have a fragment shader that takes an array of textures (or perhaps even a single texture) as input, which fails with this log under Metal API Validation:

```
[mvk-error] VK_ERROR_INITIALIZATION_FAILED: Render pipeline compile failed (Error code 3):
Fragment shader cannot be used with indirect command buffers.
-[MTLDebugRenderCommandEncoder setRenderPipelineState:]:1604: failed assertion `Set Render Pipeline State Validation
renderPipelineState must not be nil.
```

There are several deficiencies with the code I'm submitting here -- my eyes are wide open, I assure you -- and that's discounting all the problems I don't even know that I have yet ;) 

I'm opening this mostly as a plea for help, since I would love to get to a point where I could selectively enable ICBs for specific pipelines, so that any of the incompatible shaders would not be opted in to `supportIndirectCommandBuffers`. My use case is very specific, which should in theory mostly work by means of `inheritBuffers`, but a true implementation would need to be much more robust -- in theory if I could avoid requesting `supportIndirectCommandBuffers` except for one pipeline I believe this would actually be sufficient for me (but not anyone else :)

What else should I be aware of? Obviously there are some definite code-quality improvements to be made here :)

### DIFF
--- a/MoltenVK/MoltenVK/Commands/MVKCommand.h
+++ b/MoltenVK/MoltenVK/Commands/MVKCommand.h
@@ -67,6 +67,10 @@ public:
 	/** Encodes this command on the specified command encoder. */
 	virtual void encode(MVKCommandEncoder* cmdEncoder) = 0;
 
+    virtual void encodeToICB(id<MTLIndirectCommandBuffer> icb) {
+        throw "Kablooie!";
+    }
+
 protected:
 	friend MVKCommandBuffer;
 
--- a/MoltenVK/MoltenVK/Commands/MVKCommandResourceFactory.mm
+++ b/MoltenVK/MoltenVK/Commands/MVKCommandResourceFactory.mm
@@ -658,6 +658,7 @@ id<MTLFunction> MVKCommandResourceFactory::newMTLFunction(NSString* mslSrcCode, 
 id<MTLRenderPipelineState> MVKCommandResourceFactory::newMTLRenderPipelineState(MTLRenderPipelineDescriptor* plDesc,
 																				MVKVulkanAPIDeviceObject* owner) {
 	MVKRenderPipelineCompiler* plc = new MVKRenderPipelineCompiler(owner);
+    plDesc.supportIndirectCommandBuffers = YES;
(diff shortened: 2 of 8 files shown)


MoltenVK/MoltenVK/Commands/MVKCommand.h
```diff
@@ -67,6 +67,10 @@ class MVKCommand : public MVKBaseObject, public MVKLinkableMixin<MVKCommand> {
 	/** Encodes this command on the specified command encoder. */
 	virtual void encode(MVKCommandEncoder* cmdEncoder) = 0;
 
+    virtual void encodeToICB(id<MTLIndirectCommandBuffer> icb) {
```
EricAtPlanGrid, line 70: Something like this could be useful, but I believe we'd need to have an additional context object provided with it so individual commands could look up relevant index/vertex/other buffers.

MoltenVK/MoltenVK/Commands/MVKCommandResourceFactory.mm
```diff
@@ -658,6 +658,7 @@ static void getSwizzleString(char swizzleStr[4], VkComponentMapping vkMapping) {
 id<MTLRenderPipelineState> MVKCommandResourceFactory::newMTLRenderPipelineState(MTLRenderPipelineDescriptor* plDesc,
 																				MVKVulkanAPIDeviceObject* owner) {
 	MVKRenderPipelineCompiler* plc = new MVKRenderPipelineCompiler(owner);
+    plDesc.supportIndirectCommandBuffers = YES;
```
EricAtPlanGrid, line 661: there was a lot of spray-and-pray of this :)

EricAtPlanGrid:
Hello! I finally signed the CLA :) I am hoping to get some assistance in figuring out how to get the appropriate buffers registered in the argbuffers rather than using `setVertexBuffer` etc

Issue #2480, opened 2025-04-09
Application immediately GPU page faults since macOS 15.4 on M1/M2


squidbus:
I'm working on a macOS port of [UnleashedRecomp](https://github.com/hedge-dev/UnleashedRecomp). Since the macOS 15.4 update, it seems that on M1 and M2 systems the game now immediately GPU faults. On 15.3.2 and earlier, or 15.4 on M3 and M4 systems, it works fine.

I took a GPU capture on M4 vs M1 and it seems that on M1, the textures are complete garbage, as well as the texture and sampler heaps being used by fragment shaders, likely leading to the fault. You can also see in a full capture that early on when the app starts, some copies from buffer to texture occur where on M4 they produce the correct textures, where as on M1 they produce garbage textures. So at least from the capture, it seems that buffers and textures are garbage and only push constants are correct.

This also happens before the change to MTLHeap by default, so that should not be the cause.

A working build can be found [here](https://github.com/squidbus/UnleashedRecomp/actions/runs/14296082511). All you need to do is open the app to reproduce immediately; no additional setup is required.

squidbus:
Thanks for the insights, do you think that this should be an issue on SPIRV-Cross then?

squidbus:
I can confirm from some testing that hacking the variable descriptor array size to a large enough number makes the game work again, which should confirm it as the source of the issue:
```
diff --git a/MoltenVK/MoltenVK/GPUObjects/MVKDescriptor.mm b/MoltenVK/MoltenVK/GPUObjects/MVKDescriptor.mm
index e1802201..ff325e27 100644
--- a/MoltenVK/MoltenVK/GPUObjects/MVKDescriptor.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKDescriptor.mm
@@ -533,7 +533,7 @@ void mvkPopulateShaderConversionConfig(mvk::SPIRVToMSLConversionConfiguration& s
                                                                    uint32_t dslIndex) {
        // For argument buffers, set variable length arrays to length 1 in shader.
        bool isUsingMtlArgBuff = _layout->isUsingMetalArgumentBuffers();
-       uint32_t descCnt = isUsingMtlArgBuff ? getDescriptorCount(1) : getDescriptorCount();
+       uint32_t descCnt = isUsingMtlArgBuff ? getDescriptorCount(128) : getDescriptorCount();
 
        // Establish the resource indices to use, by combining the offsets of the DSL and this DSL binding.
[...]

Issue #2453, opened 2025-02-19
Regression in Vulkan CTS with descriptor set


dboyan:
Recent versions of moltenvk fails on case `dEQP-VK.binding_model.descriptorset_random.sets4.unifindexed.ubolimitlow.nosbo.sampledimglow.outimgtexlow.noiub.nouab.frag.ialimitlow.0` on my M1 system. The error is as follows:

> [mvk-error] VK_ERROR_OUT_OF_DEVICE_MEMORY: MTLCommandBuffer "vkQueueSubmit MTLCommandBuffer on Queue 0-0" execution failed (code 3): Caused GPU Address Fault Error (0000000b:kIOGPUCommandBufferCallbackErrorPageFault)
  DeviceLost (vk.waitForFences(device, 1u, &fence, VK_TRUE, timeoutNanos): VK_ERROR_DEVICE_LOST at vkCmdUtil.cpp:292)

After bisection, the first bad commit is `3262113df` from #2273.

```
3262113dfc8a3d13701eca8e7adcaf62c2da418f is the first bad commit
commit 3262113dfc8a3d13701eca8e7adcaf62c2da418f
Author: Bill Hollings <bill.hollings@brenwill.com>
Date:   Sat Jul 13 14:49:09 2024 -0400

    Descriptor set only consumes the variable number of descriptors from the pool.
```

SRSaunders:
@dboyan I have run into a similar symptom (same error code) on my M1 Air when developing for [RBDOOM-3-BFG](https://github.com/RobertBeckebans/RBDOOM-3-BFG).

In that case I was able to temporarily work around the problem by setting `MVK_CONFIG_USE_METAL_ARGUMENT_BUFFERS=0`.  I am curious whether the CTS test case passes with this setting. While this may not be a solution, if it allows the CTS case to pass it may provide a clue for a proper fix.

dboyan:
Indeed, the CTS test can pass by disabling argument buffers. I guess that's because descriptor set log use different code paths with argument buffer enabled/disabled.

SRSaunders:
> Indeed, the CTS test can pass by disabling argument buffers.

Thanks @dboyan for verifying this.  Hopefully @billhollings can find the issue with the argument buffers code path.

billhollings:
> Interestingly this error is not apparent and the setting is not needed on my x86/AMD GPU based Mac - only on my M1.

[...]

PR #2482, opened 2025-04-12
Revert use of size 1 for variable length arrays in shader.

squidbus:
Reverts part of https://github.com/KhronosGroup/MoltenVK/pull/2273 that set variable length arrays in argument buffers to size 1 in shaders.

This seems to be the cause behind various recent issues with argument buffers (https://github.com/KhronosGroup/MoltenVK/issues/2480, https://github.com/KhronosGroup/MoltenVK/issues/2453, probably https://github.com/KhronosGroup/MoltenVK/issues/2300). It seems that using an array of size 1 for variable length arrays may have been partially broken before, and is now completely broken as of macOS 15.4 on M1 and M2 systems. I verified on M1 myself that with this partial revert in place, my application works perfectly instead of immediately hitting a GPU page fault, and that it still works fine on M4.

As I understand, this may trigger some Metal validation errors from the shader size being larger than the contents, but does actually work across GPU families, unlike the current solution. I'm open to more discussion on this but felt that it makes sense to fix the issue the easy way first to get things actually functioning again, and then it can be further improved and optimized down the line in a way that continues to work everywhere.

### DIFF
--- a/MoltenVK/MoltenVK/GPUObjects/MVKDescriptor.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKDescriptor.mm
@@ -531,9 +531,10 @@ void MVKDescriptorSetLayoutBinding::encodeImmutableSamplersToMetalArgumentBuffer
 void MVKDescriptorSetLayoutBinding::populateShaderConversionConfig(mvk::SPIRVToMSLConversionConfiguration& shaderConfig,
                                                                    MVKShaderResourceBinding& dslMTLRezIdxOffsets,
                                                                    uint32_t dslIndex) {
-	// For argument buffers, set variable length arrays to length 1 in shader.
 	bool isUsingMtlArgBuff = _layout->isUsingMetalArgumentBuffers();
-	uint32_t descCnt = isUsingMtlArgBuff ? getDescriptorCount(1) : getDescriptorCount();
+	// Previously this would set variable length array sizes to 1 for argument buffers.
+	// However, this no longer works on M1 and M2 GPUs, so use the full length.
+	uint32_t descCnt = getDescriptorCount();
 
 	// Establish the resource indices to use, by combining the offsets of the DSL and this DSL binding.
     MVKShaderResourceBinding mtlIdxs = _mtlResourceIndexOffsets + dslMTLRezIdxOffsets;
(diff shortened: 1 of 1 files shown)


MoltenVK/MoltenVK/GPUObjects/MVKDescriptor.mm
```diff
@@ -531,9 +531,10 @@ void mvkPopulateShaderConversionConfig(mvk::SPIRVToMSLConversionConfiguration& s
 void MVKDescriptorSetLayoutBinding::populateShaderConversionConfig(mvk::SPIRVToMSLConversionConfiguration& shaderConfig,
                                                                    MVKShaderResourceBinding& dslMTLRezIdxOffsets,
                                                                    uint32_t dslIndex) {
-	// For argument buffers, set variable length arrays to length 1 in shader.
 	bool isUsingMtlArgBuff = _layout->isUsingMetalArgumentBuffers();
-	uint32_t descCnt = isUsingMtlArgBuff ? getDescriptorCount(1) : getDescriptorCount();
+	// Previously this would set variable length array sizes to 1 for argument buffers.
+	// However, this no longer works on M1 and M2 GPUs, so use the full length.
+	uint32_t descCnt = getDescriptorCount();
```
billhollings, line 537:
```suggestion
	// GPUs prior to M3 & A17 cannot support true runtime arrays, so use full binding array size.
	// However, in doing so, it may be necessary to disable Apple shader validation errors on those systems,
	// when descriptors contain smaller runtime arrays than declared in the binding.
	bool useRuntimeArray = isUsingMtlArgBuff && getPhysicalDevice()->getMTLDeviceCapabilities().supportsApple9;
	uint32_t descCnt = useRuntimeArray ? getDescriptorCount(1) : getDescriptorCount();
```

I'd like to modify to capture two things here:
1. Don't apply this fix to M3/A17 and above, where true runtime arrays seem to be supported.
2. Document that shader validation errors may occur.

billhollings (changes requested):
Thanks very much for figuring this out, and submitting a fix!

I've made a mod request, to allow the original code to work on M3/A17+, and to document the possibility of Apple shader validation errors on older systems.

MoltenVK/MoltenVK/GPUObjects/MVKDescriptor.mm
squidbus: That seems reasonable. I worry that this might be undefined behavior that could also break on newer GPUs in the future, but I suppose that bridge can be crossed later.

billhollings: Agreed. Fingers crossed that it is an intentional improvement in Apple9 and future GPUs, and not some temporary fluke. If the latter, we'll have to revert my patch out in the future.

billhollings (approved):
Thanks. LGTM now.

PR #2484, opened 2025-04-14
Add support for VK_KHR_maintenance4.

billhollings:


MoltenVK/MoltenVK/Vulkan/vulkan.mm
```diff
@@ -2823,9 +2823,41 @@ MVK_PUBLIC_VULKAN_SYMBOL void vkDestroyPrivateDataSlot(
...
-MVK_PUBLIC_VULKAN_STUB(vkGetDeviceImageSparseMemoryRequirements, void, VkDevice, const VkDeviceImageMemoryRequirements*, uint32_t*, VkSparseImageMemoryRequirements2*)
+MVK_PUBLIC_VULKAN_SYMBOL void vkGetDeviceBufferMemoryRequirements(
+	VkDevice                                    device,
+	const VkDeviceBufferMemoryRequirements*     pInfo,
+	VkMemoryRequirements2*                      pMemoryRequirements) {
+	
+	MVKTraceVulkanCallStart();
+	auto* mvkDev = MVKDevice::getMVKDevice(device);
+	MVKBuffer mvkBuff(mvkDev, pInfo->pCreateInfo);
+	mvkBuff.getMemoryRequirements(pMemoryRequirements);
```
cdavis5e, line 2834: I'd prefer to have a method on `MVKDevice` that fills out the requirements, instead of having to create a temporary resource and use that--which is exactly what this functionality is supposed to enable avoiding in the first place. But, I suppose that could wait for a future refactoring.

MoltenVKShaderConverter/MoltenVKShaderConverter/SPIRVToMSLConverter.cpp
```diff
@@ -502,6 +502,21 @@ void SPIRVToMSLConverter::logSource(string& log, string& src, const char* srcLan
     log += "\n\n";
 }
 
+// Extracts the workgroup dimension from either the LocalSizeId, LocalSize, or WorkgroupSize Builtin.
+// Although LocalSizeId is the modern mechanism, the Builtin takes precedence if it is presidence.
```
cdavis5e, line 506:
```suggestion
// Although LocalSizeId is the modern mechanism, the Builtin takes precedence if it is present.
```

fetchDependencies
```diff
@@ -336,7 +336,7 @@ if [ ! "$SPIRV_CROSS_ROOT" = "" ]; then
 	rm -rf ${REPO_NAME}
 	ln -sfn ${SPIRV_CROSS_ROOT} ${REPO_NAME}
 else
-	REPO_URL="https://github.com/KhronosGroup/${REPO_NAME}.git"
+	REPO_URL="https://github.com/billhollings/${REPO_NAME}.git"
```
cdavis5e, line 339: Is this necessary? I thought the changes needed for `VK_KHR_maintenance4` were already upstream.

cdavis5e (changes requested):

fetchDependencies
squidbus: This does not appear to be yet: https://github.com/KhronosGroup/SPIRV-Cross/pull/2464

MoltenVK/MoltenVK/Vulkan/vulkan.mm
billhollings: I hear you, so I dug into it a bit.

Unfortunately, this looks like it would be a particularly messy refactoring, especially for `MVKImage` and its planes.

And AFAICT, through the magic of lazy init, we don't create any Metal resources during `MVKImage` or `MVKBuffer` construction. So the cost is mostly limited to `pCreateInfo` processing. For `MVKBuffer`, that cost is particularly trivial. For `MVKImage`, that processing is fairly complicated, but it's not clear how much could be skipped by refactoring.

Previously, I did modify the original code to instantiate `MVKBuffer` & `MVKImage` inline, in an attempt to reduce allocs.

So, while I agree in theory, AFAICT this would be a bunch of messy work for limited benefit, on a maintenance extension. 😉

fetchDependencies
billhollings: What @squidbus said. 

This is a temporary bypass, so we can get this in soon, since it's the last hold up before 1.3. Will back it out as soon as the final support is available.

MoltenVKShaderConverter/MoltenVKShaderConverter/SPIRVToMSLConverter.cpp
billhollings: Eagle eyes! Thanks. Good catch.

cdavis5e (approved):

PR #2486, opened 2025-04-17
Fix type in MoltenVKShaderCoverter description in README

zfergus:
Fix description of MoltenVKShaderCoverter based on the tool's command line help message.

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,8 +63,9 @@ The **MoltenVK** runtime package contains two products:
 - **MoltenVK** is a implementation of an almost-complete subset of the
   [*Vulkan 1.2*](https://www.khronos.org/vulkan) graphics and compute API.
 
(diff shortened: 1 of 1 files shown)


dboyan:
Not really. The description in the readme is correct and [explicitly changed into the current wording](https://github.com/KhronosGroup/MoltenVK/commit/ab774be98f5fd38cf7ad3f0029c3e0705df252fe#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5) to reflect the removal of `NV_glsl_shader` support. I guess an appropriate change is to remove the spirv output option from the converter tool and let it exclusively output msl instead.

billhollings:
Thanks for submitting. Although your replacement is not applicable here, you've highlighted that the wording that you had removed is duplicated. I'll fix that part at least as part of another PR.

PR #2489, opened 2025-04-19
Add support for Vulkan 1.3.

billhollings:
- Advertise Vulkan 1.3 support.
- Enable mandatory `VkPhysicalDeviceVulkanMemoryModelFeatures::vulkanMemoryModel` and `vulkanMemoryModelDeviceScope`
- Advertise support for `VK_KHR_vulkan_memory_model` extension.
- Log a warning if requested API is Vulkan 1.3 and buffer device address is not supported.
- Support `VkPhysicalDeviceVulkan13Features` & `VkPhysicalDeviceVulkan13Properties`.
- Remove enablement checks for promoted extensions.
- Promote extension structs.
- `MVK_CONFIG_API_VERSION_TO_ADVERTISE` supports shorthand version values.
- Populate CTS conformance version.
- Update documentation.

Completes #1930.


MoltenVK/MoltenVK/Utility/MVKEnvironment.cpp
```diff
@@ -44,17 +44,26 @@ static constexpr uint32_t getExpectedMVKConfigurationStringCount() {
...
 
+// Expand the shortcut versions, Add the VK_HEADER_VERSION.
+static uint32_t expandAPIVersion(uint32_t apiVer) {
+	switch (apiVer) {
+		case  0:  return VK_API_VERSION_1_0;
+		case 10:  return VK_API_VERSION_1_0;
+		case 11:  return VK_API_VERSION_1_1;
+		case 12:  return VK_API_VERSION_1_2;
+		case 13:  return VK_API_VERSION_1_3;
+		default:  return std::min(apiVer, VK_API_VERSION_1_3);
```
spnda, line 55: Shouldn't this be `MVK_VULKAN_API_VERSION` instead of hardcoding 1.3, as it was before? Also, these shorthands feel a little weird to me tbh...

MoltenVK/MoltenVK/Utility/MVKEnvironment.h
```diff
@@ -39,10 +39,13 @@
 #	define Vk_PLATFORM_SurfaceCreateInfoMVK		VkMacOSSurfaceCreateInfoMVK
 #endif
 
+/** Macro to adjust the specified Vulkan to include the VK_HEADER_VERSION patch value. */
+#define MVK_VULKAN_API_VERSION_HEADER(api_ver)		VK_MAKE_VERSION(VK_VERSION_MAJOR(api_ver),	\
+                                                                    VK_VERSION_MINOR(api_ver),	\
+                                                                    VK_HEADER_VERSION)
```
spnda, line 45: Technically speaking `VK_MAKE_VERSION` is deprecated, and `VK_MAKE_API_VERSION` should be used instead.

MoltenVK/MoltenVK/Utility/MVKEnvironment.cpp
billhollings: Fixed. Thanks.

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
```diff
@@ -645,17 +673,16 @@
 	supportedProps11.maxPerSetDescriptors = getMaxPerSetDescriptorCount();
 	supportedProps11.maxMemoryAllocationSize = _metalFeatures.maxMTLBufferSize;
 
+	static VkConformanceVersion testedCTSVer = { 1, 3, 8, 0};	// Latest version of CTS used to test
```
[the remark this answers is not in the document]
billhollings: Fixed. Thanks.

MoltenVK/MoltenVK/Utility/MVKEnvironment.h
billhollings: Fixed. Thanks.

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
```diff
@@ -707,6 +734,56 @@
...
+	supportedProps13.integerDotProductAccumulatingSaturating32BitSignedAccelerated = false;
+	supportedProps13.integerDotProductAccumulatingSaturating32BitMixedSignednessAccelerated = false;
+	supportedProps13.integerDotProductAccumulatingSaturating64BitUnsignedAccelerated = false;
+	supportedProps13.integerDotProductAccumulatingSaturating64BitSignedAccelerated = false;
+	supportedProps13.integerDotProductAccumulatingSaturating64BitMixedSignednessAccelerated = false;
+	supportedProps13.storageTexelBufferOffsetAlignmentBytes = _texelBuffAlignProperties.storageTexelBufferOffsetAlignmentBytes;
+	supportedProps13.storageTexelBufferOffsetSingleTexelAlignment = _texelBuffAlignProperties.storageTexelBufferOffsetSingleTexelAlignment;
+	supportedProps13.uniformTexelBufferOffsetAlignmentBytes = _texelBuffAlignProperties.uniformTexelBufferOffsetAlignmentBytes;
+	supportedProps13.uniformTexelBufferOffsetSingleTexelAlignment = _texelBuffAlignProperties.uniformTexelBufferOffsetSingleTexelAlignment;
+	supportedProps13.maxBufferSize = _metalFeatures.maxMTLBufferSize;
```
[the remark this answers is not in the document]
billhollings: This population technique is consistent with how MoltenVK populate `VkPhysicalDeviceVulkan11Properties` and `VkPhysicalDeviceVulkan12Properties`. The difference in technique relative to setting features is because properties can sometimes be more complicated values, including strings.

MoltenVK/MoltenVK/Commands/MVKCmdPipeline.mm
```diff
@@ -450,24 +449,22 @@ static uint64_t mvkPipelineStageFlagsToBarrierStages(VkPipelineStageFlags2 flags
...
-            if (pInlineUniformBlock) {
-                auto *pNewInlineUniformBlock = new VkWriteDescriptorSetInlineUniformBlockEXT(*pInlineUniformBlock);
-                pNewInlineUniformBlock->pNext = nullptr; // clear pNext just in case, no other extensions are supported at this time
-                descWrite.pNext = pNewInlineUniformBlock;
-            }
-        }
+		const VkWriteDescriptorSetInlineUniformBlockEXT* pInlineUniformBlock = nullptr;
+		for (const auto* next = (VkBaseInStructure*)descWrite.pNext; next; next = next->pNext) {
+			switch (next->sType) {
+				case VK_STRUCTURE_TYPE_WRITE_DESCRIPTOR_SET_INLINE_UNIFORM_BLOCK_EXT: {
```
cdavis5e, line 455: Should we then use the unsuffixed names here?

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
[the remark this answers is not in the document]
spnda: I don't see how that makes a difference to be honest, since the assignment behavior is identical? And the "more complicated values" are initialized before the struct is declared anyway. And then the old code could be changed, to be more inline with modern C++ programming practices.

cdavis5e (approved):
_Finally_, we can use DXVK 2.x.

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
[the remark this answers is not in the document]
billhollings: > since the assignment behavior is identical? And the "more complicated values" are initialized before the struct is declared anyway

Not in all cases. For instance, strings and UUIDs are inline in the property structs. See the two `strcpy()` calls when populating `VkPhysicalDeviceVulkan12Properties`, and for `VkPhysicalDeviceVulkan11Properties`, there is `populateDeviceIDProperties()` which sets up and copies UUIDs, and `populateSubgroupProperties()` which consolidates a bunch of processing around subgroup properties that would otherwise be a bunch of separate functions each returning a single value.

MoltenVK/MoltenVK/Commands/MVKCmdPipeline.mm
billhollings: Damn. Good catch. Fixed now.

spnda:
what's the point of `DXMT` when `D3DMetal` is a thing?

MrGcGamer:
One thing to consider with DXVK over DXMT is the support for Intel Macs, as Intel support for DXMT is quite a bit off afaik.

MrGcGamer:
@Gcenx 
> but current macOS AMD drivers are rather broken

Out of interest: does that apply for all supported models, or are there models which are particularly bad?
Are there resources / forum / other github issues where one can idea an idea of the current state of things?

Issue #2490, opened 2025-04-19
Vulkan 1.4 requirements


billhollings:
[Appendix D of the Vulkan spec](https://docs.vulkan.org/spec/latest/appendices/versions.html#versions-1.4) summarizes the list of extensions promoted to Vulkan 1.4 as below:

Extension | Optional<br/> in 1.4 | MoltenVK<br/>Support | Status Notes |
| :--| :-: | :-: |:-- 
| `VK_KHR_dynamic_rendering_local_read`|partial support<br/>required (see below)|✅| Added (#2579) |
|`VK_KHR_global_priority`||✅| Added (#2516) |
|`VK_KHR_index_type_uint8`||✅| Added (#2504) |
|`VK_KHR_line_rasterization`||✅| Added (#2556) |
|`VK_KHR_load_store_op_none`||✅| Added (#2499) |
|`VK_KHR_maintenance5`||✅| Added (#2508) |
|`VK_KHR_maintenance6`||✅| Added (#2501)|
|`VK_KHR_map_memory2`||✅||
|`VK_KHR_push_descriptor`||✅||
|`VK_KHR_shader_expect_assume`||✅| Added (#2493)|
|`VK_KHR_shader_float_controls2`||✅| Added (#2564)|
|`VK_KHR_shader_subgroup_rotate`||✅| Added (#2494)|
|`VK_KHR_vertex_attribute_divisor`||✅||
|`VK_EXT_host_image_copy`|☑️|✅||
|`VK_EXT_pipeline_protected_access`|☑️<br/>mandatory<br/>only if<br/>`protectedMemory`<br/>is enabled|||
|`VK_EXT_pipeline_robustness`||✅| Added (#2498) |

- Differences Relative to `VK_KHR_dynamic_rendering_local_read`
  - ✅ If the `VK_KHR_dynamic_rendering_local_read` extension is not supported, Vulkan 1.4 implementations must support local read only for storage resources and single sampled color attachments.

- Differences Relative to `VK_EXT_host_image_copy`
  - ✅ If the `VK_EXT_host_image_copy` extension is not supported, support for it is optional in Vulkan 1.4.
  - ✅ An implementation that has a `VK_QUEUE_GRAPHICS_BIT` queue must support either:
the `hostImageCopy` feature; or an additional queue that supports `VK_QUEUE_TRANSFER_BIT``.

- Differences Relative to `VK_KHR_push_descriptor`
[...]

PR #2498, opened 2025-04-25
Add support for VK_EXT_pipeline_robustness

squidbus:
Adds `VK_EXT_pipeline_robustness`, required for Vulkan 1.4 (https://github.com/KhronosGroup/MoltenVK/issues/2490).

The properties here advertise the following defaults, matching the behavior of MoltenVK today:
* Storage buffer: Disabled
* Uniform buffer: Disabled
* Vertex input: Disabled
* Image: Robustness2

For buffers, I have added warnings for both this extension's controls and the existing `robustBufferAccess` and `robustBufferAccess2` features, whenever any of these are enabled by the application, to indicate that Metal does not support buffer robustness. We do advertise `robustBufferAccess` since it is required, so this is technically not completely up to spec, but for `robustBufferAccess2` applications should never request it anyway since it is correctly advertised as unsupported.

For images, my understanding of the spec is that as long as we indicate that the device default is the Robustness2 behavior, we do not actually have to honor disabling robustness in pipelines; instead, applications simply are promising to not produce out-of-bounds accesses if they mark it as disabled, and the driver is free to optimize based on that.

All of the relevant CTS tests seem to be unsupported due to other reasons (portability subset, missing pipeline libraries, ...), so I was not able to get any results out of that.


billhollings (changes requested):
Thanks for the submission! 

LGTM. A couple of conflicts arose due to other PRs. Please fix them, then I'll pull this in.

squidbus:
Ran into a little trouble resolving conflicts on mobile but should be good now.

billhollings (commented):
Thanks. LGTM.

PR #2504, opened 2025-04-26
Add support for VK_EXT_index_type_uint8

squidbus:
Managed to figure this extension out, so... bonus PR I guess. Another extension for 1.4 (https://github.com/KhronosGroup/MoltenVK/issues/2490).

Support is implemented by noting that the Vulkan index type was `VK_INDEX_TYPE_UINT8`, and then at encode time copying the 8-bit indices to 16-bit indices in a temporary buffer. If the index buffer is private, compute is used to do this; otherwise, a CPU copy is used to avoid breaking the render pass.

CTS results for `dEQP-VK.pipeline.*.index_type_uint8.*`:
```
Test run totals:
  Passed:        15/389 (3.9%)
  Failed:        3/389 (0.8%)
  Not supported: 371/389 (95.4%)
  Warnings:      0/389 (0.0%)
  Waived:        0/389 (0.0%)
```

Most CTS skips here are due to missing `VK_EXT_primitive_topology_list_restart`, `VK_EXT_shader_object`, or `geometryShader`.

3 test failures are related to disabled primitive restart; this behavior is carried over the same as UINT16 and UINT32 index types, and indeed the same 3 test variants fail for those index types as well. Technically we could probably utilize the extra space of a UINT16 buffer to actually support disabling primitive restart, by only translating to the larger restart index if it is enabled, but I left that as a possible future exercise due to the extra state management needed; only supporting restart disable for specific index types seems a bit weird to me anyway.


squidbus:
Rebased on top of maintenance6

MoltenVK/MoltenVK/Commands/MVKCmdDraw.mm
```diff
@@ -90,6 +97,34 @@
...
+            }
+        } else {
+            // Index buffer is private, so we must do the conversion on the GPU.
+            cmdEncoder->encodeStoreActions(true);
+            id<MTLComputeCommandEncoder> mtlComputeEnc = cmdEncoder->getMTLComputeEncoder(kMVKCommandConvertUint8Indices, true);
+            [mtlComputeEnc setComputePipelineState: cmdEncoder->getCommandEncodingPool()->getConvertUint8IndicesMTLComputePipelineState()];
+            [mtlComputeEnc setBuffer:_binding.mtlBuffer offset: _binding.offset atIndex: 0];
+            [mtlComputeEnc setBuffer:uint16Buf->_mtlBuffer offset: 0 atIndex: 1];
+            [mtlComputeEnc setBytes: &numIndices length: sizeof(numIndices) atIndex: 2];
+            [mtlComputeEnc dispatchThreadgroups: MTLSizeMake(1, 1, 1) threadsPerThreadgroup: MTLSizeMake(1, 1, 1)];
```
billhollings, line 121: See my comments below about using `thread_position_in_grid` to let the GPU do what it does best.

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
```diff
@@ -609,6 +609,11 @@
 				}
 				break;
 			}
+			case VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_INDEX_TYPE_UINT8_FEATURES_EXT: {
+				auto* indexTypeUint8Features = (VkPhysicalDeviceIndexTypeUint8FeaturesEXT*)next;
+				indexTypeUint8Features->indexTypeUint8 = true;
+				break;
+			}
```
billhollings, line 616: To support both `KHR`, `EXT`, and future 1.4, suggest you remove `EXT` suffixes and move this to the appropriate sorted position in the main non-extension area above.

MoltenVK/MoltenVK/GPUObjects/MVKDeviceFeatureStructs.def
```diff
@@ -80,6 +80,7 @@ MVK_DEVICE_FEATURE_EXTN(ExtendedDynamicState3,        EXTENDED_DYNAMIC_STATE_3,
 MVK_DEVICE_FEATURE_EXTN(FragmentShaderInterlock,      FRAGMENT_SHADER_INTERLOCK,       EXT,   3)
 MVK_DEVICE_FEATURE_EXTN(HostImageCopy,                HOST_IMAGE_COPY,                 EXT,   1)
 MVK_DEVICE_FEATURE_EXTN(Image2DViewOf3D,              IMAGE_2D_VIEW_OF_3D,             EXT,   2)
+MVK_DEVICE_FEATURE_EXTN(IndexTypeUint8,               INDEX_TYPE_UINT8,                EXT,   1)
```
billhollings, line 83: Ditto re `KHR`, `EXT` & 1.4.

MoltenVK/MoltenVK/Commands/MVKCommandPipelineStateFactoryShaderSource.h
```diff
@@ -547,5 +547,14 @@ kernel void accumulateOcclusionQueryResults(device VisibilityBuffer& dest [[buff
     if (dest.count < max(oldDestCount, src.count)) { dest.countHigh++; }                                        \n\
 }                                                                                                               \n\
                                                                                                                 \n\
+kernel void convertUint8Indices(device uint8_t* src [[ buffer(0) ]],                                            \n\
+                                device uint16_t* dst [[ buffer(1) ]],                                           \n\
+                                constant uint32_t& numIndices [[ buffer(2) ]]) {                                \n\
+    for (uint32_t i = 0; i < numIndices; i++) {                                                                 \n\
```
billhollings, line 553: This would be better structured to use GPU parallelism, with `thread_position_in_grid`. Then you could skip the single-thread loop entirely, and let the GPU do what it does best. 

See for example `cmdFillBuffer()` in this same file.

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
```diff
@@ -609,6 +609,11 @@
 				}
 				break;
 			}
+			case VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_INDEX_TYPE_UINT8_FEATURES_EXT: {
+				auto* indexTypeUint8Features = (VkPhysicalDeviceIndexTypeUint8FeaturesEXT*)next;
```
billhollings, line 613: To support both `KHR`, `EXT`, and future 1.4, suggest you remove `EXT` suffixes and move this to the appropriate sorted position in the main non-extension area above.

squidbus:
Thanks for the suggestions, I've rebased and updated to reflect the suggested changes around KHR/EXT, style, and utilizing thread groups better.

billhollings (approved):
Thanks. LGTM now.

K0bin:
How are GPU writes synchronized in the CPU copy path? I'm probably missing something here but to me it looks like this just copies the buffer data at cmd buffer recording/submission time (depending on the MVK configuration).

Do buffers with `VK_BUFFER_USAGE_STORAGE_BUFFER_BIT` or `VK_BUFFER_USAGE_COPY_DST_BIT` usage always use the private storage mode? (Although buffers without either of those are probably rare enough that having a special case for those isn't worth it.)

Issue #2471, opened 2025-03-15
Triangle fans break renderpass even in direct draw case


squidbus:
From what I understand looking at the code, it seems currently all types of triangle fan draws funnel into using an indirect draw to emulate, I assume to make the code simpler. However this path uses compute and requires a render pass restart, which can negatively impact performance.

I think in the case of direct draws it should be possible to generate indices and perform the draws without compute and thus without restarting the render pass. It may require more code to cover this additional path, but it would be more ideal for performance to avoid breaking the render pass where possible.

squidbus:
I mention indirect draws because, to my understanding, that is where compute is actually necessary to support fans. However it is also currently used for cases of direct draws from the application, which should be able to generate indices for this without compute, in turn avoiding the need to split the render pass.

This came up in [UnleashedRecomp](https://github.com/hedge-dev/UnleashedRecomp), where the recompiled game is the one using triangle fans, but I don’t see it as being an application specific optimization at all. It’s simply the case today that, if an application does many triangle fan draw calls, even just simple direct ones, every single one will split the render pass.

dboyan:
I get your point. I think the usage indirect is only meant to share code to a single implementation. No matter if one uses indirect or not, one *has to* generate the indices anyway. The question is if you want to use GPU to do the generation, which will inevitably break the renderpass. The usage of indirect plays a secondary role here in my opinion.

I just looked at the ANGLE code a little more. Turns out it defaults to CPU generation path for apple silicon in its [generateTriFanBufferFromElementsArray](https://chromium.googlesource.com/angle/angle/+/7ee4f499d6f283f3f9c5148e4a775049ab6e16c5/src/libANGLE/renderer/metal/mtl_render_utils.mm#1756) function because it sets `hasCheapRenderPass` feature [to false](https://chromium.googlesource.com/angle/angle/+/7ee4f499d6f283f3f9c5148e4a775049ab6e16c5/src/libANGLE/renderer/metal/DisplayMtl.mm#1233).

squidbus:
Extending this to cover Uint8 buffers as well, see: https://github.com/KhronosGroup/MoltenVK/pull/2505

The branch there has CPU copy paths for both triangle fans and Uint8 index buffers, but they do not account for potentially GPU modified buffers through shader write or copies. We likely need a good way to know if the buffer is a trivial, CPU-side buffer or not to take the CPU path without costly command flushes.

PR #2505, opened 2025-04-26
Support triangle fans without breaking render pass where possible.

squidbus:
* Support reindexing triangle fans on the CPU for direct draws, avoiding breaking render pass for compute. Exception is that when the index buffer is in GPU-private memory, we need to do it in compute anyway as CPU does not have access.
* Fix some mistakes with the recent temporary index buffer related changes, where the offset of the temporary buffer was not propagated.

Fixes https://github.com/KhronosGroup/MoltenVK/issues/2471

### DIFF
--- a/MoltenVK/MoltenVK/Commands/MVKCmdDraw.h
+++ b/MoltenVK/MoltenVK/Commands/MVKCmdDraw.h
@@ -92,7 +92,7 @@ public:
 						uint32_t firstInstance);
 
     void encode(MVKCommandEncoder* cmdEncoder) override;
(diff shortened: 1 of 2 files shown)


K0bin:
Basically the same question as in #2504. How does the CPU copy path deal with GPU writes? What if I write/copy to the index buffer in the same not-yet-submitted command buffer just before the draw?

squidbus:
Hmm, you're right. Storage buffer seems simple to check but I'm not sure how to go about knowing if there's a pending copy or not. That's annoying, I might just have to pull back on the CPU paths for now.

PR #2508, opened 2025-04-27
Add support for VK_KHR_maintenance5

squidbus:
Adds `VK_KHR_maintenance5`, required for Vulkan 1.4 (https://github.com/KhronosGroup/MoltenVK/issues/2490). This was a bit more far-reaching than some of the other maintenance extensions I've done recently since it's a lot larger and broader, so want to go ahead and get this up for some review.

List of changes:
* A new VK_FORMAT_A1B5G5R5_UNORM_PACK16_KHR format
  * Added this format, passes all CTS.

* A new VK_FORMAT_A8_UNORM_KHR format
  * Added this format, passes all CTS *except for border color opaque black*.

* A property to indicate that multisample coverage operations are performed after sample counting in EarlyFragmentTests mode
  * `true`, based on CTS validation.

* Relax VkBufferView creation requirements by allowing subsets of the associated VkBuffer usage using VkBufferUsageFlags2CreateInfoKHR
  * Added handling for this, narrowing down the parent buffer's usage for the purpose of creating the `MTLTexture`.

* A new command [vkCmdBindIndexBuffer2KHR](https://registry.khronos.org/vulkan/specs/latest/man/html/vkCmdBindIndexBuffer2KHR.html), allowing a range of memory to be bound as an index buffer
  * Added, along with the `size` field for index buffer bindings.

* [vkGetDeviceProcAddr](https://registry.khronos.org/vulkan/specs/latest/man/html/vkGetDeviceProcAddr.html) must return NULL for supported core functions beyond the version requested by the application.
  * Checked and this appears to already be the case.

* A property to indicate that the sample mask test is performed after sample counting in EarlyFragmentTests mode
  * `false`, based on CTS validation.

* vkCmdBindVertexBuffers2 now supports using VK_WHOLE_SIZE in the pSizes parameter.
  * Added handling for this case.

* A default size of 1.0 is used if PointSize is not written
  * Added an option to SPIRV-Cross to default the PointSize to 1.0 if not written.

* Shader modules are deprecated - applications can now pass [VkShaderModuleCreateInfo](https://registry.khronos.org/vulkan/specs/latest/man/html/VkShaderModuleCreateInfo.html) as a chained structure to pipeline creation via [VkPipelineShaderStageCreateInfo](https://registry.khronos.org/vulkan/specs/latest/man/html/VkPipelineShaderStageCreateInfo.html)
  * Added fields to pipelines for holding shader modules, along with the concept of the pipeline owning and needing to release the shader if it was created in this manner.

* A function [vkGetRenderingAreaGranularityKHR](https://registry.khronos.org/vulkan/specs/latest/man/html/vkGetRenderingAreaGranularityKHR.html) to query the optimal render area for a dynamic rendering instance.
  * Added based on the existing granularity function in render passes.

* A property to indicate that depth/stencil texturing operations with VK_COMPONENT_SWIZZLE_ONE have defined behavior
  * `true`, based on CTS validation.

* Add [vkGetImageSubresourceLayout2KHR](https://registry.khronos.org/vulkan/specs/latest/man/html/vkGetImageSubresourceLayout2KHR.html) and a new function [vkGetDeviceImageSubresourceLayoutKHR](https://registry.khronos.org/vulkan/specs/latest/man/html/vkGetDeviceImageSubresourceLayoutKHR.html) to allow the application to query the image memory layout without having to create an image object and query it.
  * `vkGetImageSubresourceLayout2EXT` already exists from `VK_EXT_host_image_copy` and just needed to be aliased.
  * `vkGetDeviceImageSubresourceLayoutKHR` is added similar to how `vkGetDeviceImageMemoryRequirements` works, creating a temporary lazy-init MVKImage and querying the info.

* Allow VK_REMAINING_ARRAY_LAYERS as the layerCount member of [VkImageSubresourceLayers](https://registry.khronos.org/vulkan/specs/latest/man/html/VkImageSubresourceLayers.html)
  * Added handling for this.

* Adds stronger guarantees for propagation of VK_ERROR_DEVICE_LOST return values
  * As far as I can tell we sufficiently handle this already.

* A property to indicate whether PointSize controls the final rasterization of polygons if [polygon mode](https://registry.khronos.org/vulkan/specs/latest/html/vkspec.html#primsrast-polygonmode) is VK_POLYGON_MODE_POINT
  * `true`, as we emulate `VK_POLYGON_MODE_POINT` via point topology.

* Two properties to indicate the non-strict line rasterization algorithm used
  * Set these to parallelogram, which I believe Metal uses (correct me if I'm wrong).

* Two new flags words [VkPipelineCreateFlagBits2KHR](https://registry.khronos.org/vulkan/specs/latest/man/html/VkPipelineCreateFlagBits2KHR.html) and [VkBufferUsageFlagBits2KHR](https://registry.khronos.org/vulkan/specs/latest/man/html/VkBufferUsageFlagBits2KHR.html)
  * Added handling for these new flags. All uses of the original types have been changed to their `2` equivalents, and when fetching the flags, the originals are OR-ed with the new versions if present to get the full flag set.

* Physical-device-level functions can now be called with any value in the valid range for a type beyond the defined enumerants, such that applications can avoid checking individual features, extensions, or versions before querying supported properties of a particular enumerant.
  * Made some fixes to image format properties handling of usage for this, otherwise CTS already passed.

* Clarification that copies between images of any type are allowed, treating 1D images as 2D images with a height of 1.
  * As far as I can tell this is already the case, CTS copy tests seem to indicate so.

CTS results for this batch of tests I identified as interacting with `maintenance5`: [test.txt](https://github.com/user-attachments/files/19931451/test.txt)
```
Test run totals:
  Passed:        5085/19184 (26.5%)
  Failed:        17/19184 (0.1%)
  Not supported: 14082/19184 (73.4%)
  Warnings:      0/19184 (0.0%)
  Waived:        0/19184 (0.0%)
```

The 17 failing tests are the issue with `A8_UNORM` and opaque black border color, as mentioned above. It seems that while Vulkan expects the alpha to be 1, Metal provides 0. I'm not sure how to handle this without something like complex logic to create extra variants of each sampler and correct match to `A8_UNORM` textures in the shader, since samplers are generally independent from images. I'm not sure that would even be viable for bindless either. Since it seems the opaque black case already does not work for many formats that require internal swizzling, I left this as a 'known limitation' that we may need a portability bit or something for in the future.


cdavis5e:
> * Set these to parallelogram, which I believe Metal uses (correct me if I'm wrong).

This is true in all cases where `strictLines` is false, which it is on Apple Silicon and on AMD. On Intel and (I believe) NV, `strictLines` is true, because they use rectangles, so the property doesn't apply there.

squidbus:
> > * Set these to parallelogram, which I believe Metal uses (correct me if I'm wrong).
> 
> This is true in all cases where `strictLines` is false, which it is on Apple Silicon and on AMD. On Intel and (I believe) NV, `strictLines` is true, because they use rectangles, so the property doesn't apply there.

Okay, so I assume then since this property is for non-strict lines we should set it to false whenever strictLines is true?

cdavis5e:
> The 17 failing tests are the issue with `A8_UNORM` and opaque black border color, as mentioned above. It seems that while Vulkan expects the alpha to be 1, Metal provides 0. I'm not sure how to handle this without something like complex logic to create extra variants of each sampler and correct match to `A8_UNORM` textures in the shader, since samplers are generally independent from images.

Have you considered working around this with a swizzle?

cdavis5e:
> > > * Set these to parallelogram, which I believe Metal uses (correct me if I'm wrong).
> > 
> > 
> > This is true in all cases where `strictLines` is false, which it is on Apple Silicon and on AMD. On Intel and (I believe) NV, `strictLines` is true, because they use rectangles, so the property doesn't apply there.
> 
> Okay, so I assume then since this property is for non-strict lines we should set it to false whenever strictLines is true?

It doesn't really matter either way, since the properties only apply to non-strict lines.

squidbus:
> Have you considered working around this with a swizzle?

I don’t think this will work as CTS indicates that other swizzled formats are not functioning properly with opaque black either. Plus we would have to match it up to only swizzle when used with the right sampler.

cdavis5e (approved):
I think this'll be good to go once your SPIRV-Cross change lands.

cdavis5e:
> > Have you considered working around this with a swizzle?
> 
> I don’t think this will work as CTS indicates that other swizzled formats where alpha is swizzled are not functioning properly with opaque black either. Plus we would have to match it up to only swizzle when used with the right sampler.

Yeah, I realized it wouldn't work after the fact--especially because the swizzle would also break sampling the texture in bounds.

I think I remember now hitting this when I ran the CTS on Apple Silicon. I filed a feedback with Apple, but they've so far refused to fix it. I can't remember what the feedback ID is, because this was when I was still with CW.

billhollings (approved):
Wow! You accomplished adding quite a lot of diverse features here, in not a lot of time! Very impressive!

LGTM.

As @cdavis5e already mentioned, once the SPIRV-Cross update lands, we can pull this in.

cdavis5e:
> I've changed the non-strict lines parallelogram properties to false, to indicate they are Bresenham lines not parallelograms as per spec.

I'm not sure about this. As you saw, it depends on whether or not the sample count is greater than 1. Does the CTS still pass with this change?

squidbus:
> I'm not sure about this. As you saw, it depends on whether or not the sample count is greater than 1. Does the CTS still pass with this change?

Yes, it does. I can change it back if you want but my rationale is:
* Neither of the styles of line match parallelogram lines which is what the flag is named for.
* Implementing line rasterization extension will likely involve fixing this for the multisample case as well.

squidbus:
SPIRV-Cross is now pointed back upstream.

PR #2512, opened 2025-04-29
Add support for VK_KHR_maintenance8

squidbus:
Adds `VK_KHR_maintenance8`, which is the last of the current maintenance extensions not supported.

* Allow copies between depth/stencil and “matching” color attachments
  * Updated the copy image validation and logic to account for combined depth-stencil formats.
  * Passes all CTS tests for this.

* Allow dstCache in vkMergePipelineCaches to be implicitly synchronized.
  * Updated pipeline cache to account for this flag and synchronize `mergePipelineCaches` accordingly.
  * It seems `mergePipelineCaches` was already checking `_isExternallySynchronized` but I don't see any indication this flag is supposed to effect that command, only cases where the cache is impacted by a create pipeline call. Thus I removed that check and replaced it with the new one.
  * A CTS test for this fails but it seems that is only because the test expects the cache hit to be at the pipeline level, but we currently only do cache hits at the shader stage level.

* Require src/dst sync scopes to work when doing queue family ownership transfers
  * I don't see any indication we do anything related to queue family ownership in MoltenVK, so ignored this. Related CTS for multiple queues using `maintenance8` seems to pass.

* Support Offset (as an alternative to ConstOffset) image operand in texture sampling and fetch operations
  * Already works, CTS tests are passing.

* Use the SPIR-V definition of OpSRem and OpSMod, making these operations produce well-defined results for negative operands
  * Fixed in SPIRV-Cross.

* Loosen layer restrictions when blitting from 3D images to other image types
  * Updated the blit logic to account for relaxed restrictions on 3D to/from non-3D copies, with the logic for determining layer count and z increments.

* Add space for an additional 64 access flags for use with VkMemoryBarrier2, VkBufferMemoryBarrier2, and VkImageMemoryBarrier2
  * The extended flags are separate from the existing flags rather than including them, and there are no new flags yet to look at, so we can ignore this.

CTS results for tests which interact with `maintenance8`: [test_m8.txt](https://github.com/user-attachments/files/20026011/test_m8.txt)
```
Test run totals:
  Passed:        5064/6966 (72.7%)
  Failed:        1/6966 (0.0%)
  Not supported: 1901/6966 (27.3%)
  Warnings:      0/6966 (0.0%)
  Waived:        0/6966 (0.0%)
```

Single test failure here is the previously mentioned unrelated pipeline cache issue, where our implementation of pipeline cache control does not record a hit for the pipeline as a whole, only individual shader stages.


squidbus:
Ready for review, the last of the requirements is fulfilled now.

MoltenVK/MoltenVK/Commands/MVKCmdTransfer.mm
```diff
@@ -102,15 +102,26 @@ static inline MTLSize mvkClampMTLSize(MTLSize size, MTLOrigin origin, MTLSize ma
     return VK_SUCCESS;
 }
 
+static inline MTLPixelFormat getDepthStencilAspectFormat(const MTLPixelFormat format, const VkImageAspectFlags aspectMask) {
+    if (format == MTLPixelFormatDepth32Float_Stencil8) {
```
cdavis5e, line 106: What about `MTLPixelFormatDepth24Unorm_Stencil8`?

MoltenVK/MoltenVK/Commands/MVKCmdTransfer.mm
```diff
@@ -133,17 +144,20 @@ static inline MTLSize mvkClampMTLSize(MTLSize size, MTLOrigin origin, MTLSize ma
         
         MTLPixelFormat srcMTLPixFmt = _srcImage->getMTLPixelFormat(srcPlaneIndex);
         bool isSrcCompressed = _srcImage->getIsCompressed();
+        bool isSrcCombinedDepthStencil = srcMTLPixFmt == MTLPixelFormatDepth32Float_Stencil8;
```
cdavis5e, line 147: Here, too.

MoltenVK/MoltenVK/Commands/MVKCmdTransfer.mm
```diff
@@ -133,17 +144,20 @@ static inline MTLSize mvkClampMTLSize(MTLSize size, MTLOrigin origin, MTLSize ma
...
         bool canReinterpretSrc = _srcImage->hasPixelFormatView(srcPlaneIndex);
 
         MTLPixelFormat dstMTLPixFmt = _dstImage->getMTLPixelFormat(dstPlaneIndex);
         bool isDstCompressed = _dstImage->getIsCompressed();
+        bool isDstCombinedDepthStencil = dstMTLPixFmt == MTLPixelFormatDepth32Float_Stencil8;
```
cdavis5e, line 152: Ditto.

MoltenVK/MoltenVK/GPUObjects/MVKPipeline.mm
```diff
@@ -2645,7 +2645,7 @@ static MTLVertexFormat mvkAdjustFormatVectorToSize(MTLVertexFormat format, uint3
 }
 
 VkResult MVKPipelineCache::mergePipelineCaches(uint32_t srcCacheCount, const VkPipelineCache* pSrcCaches) {
-	if (_isExternallySynchronized) {
+	if (!_isMergeInternallySynchronized) {
```
cdavis5e, line 2648: Should this be:
```suggestion
	if (_isExternallySynchronized || !_isMergeInternallySynchronized) {
```

?

cdavis5e (changes requested):

MoltenVK/MoltenVK/GPUObjects/MVKPipeline.mm
squidbus: I don't think so, the spec says:
```
VUID-vkMergePipelineCaches-dstCache-10202
Host access to dstCache must be externally synchronized if it was not created with
VK_PIPELINE_CACHE_CREATE_INTERNALLY_SYNCHRONIZED_MERGE_BIT_KHR
```

MoltenVK/MoltenVK/Commands/MVKCmdTransfer.mm
squidbus: I can add stencil handling for this but I don't think there's any equivalent supported format for the 24-bit depth.

MoltenVK/MoltenVK/GPUObjects/MVKPipeline.mm
squidbus: Also another note, the spec says these bits are mutually exclusive, so if `_isMergeInternallySynchronized` is true, `_isExternallySynchronized` will be false.

cdavis5e: OK, thanks for clearing that up.

cdavis5e (approved):

PR #2513, opened 2025-04-29
WIP: Download External Dependencies and Build using CMake

zfergus:
I need to build using CMake to integrate MoltenVK into a downstream CMake project. Seeing that https://github.com/KhronosGroup/MoltenVK/pull/1947 has not been worked on in over a year, I decided to set up a CMake build myself.

Compared to https://github.com/KhronosGroup/MoltenVK/pull/1947, my version downloads the external dependencies automatically through CMake using [CPM](https://github.com/cpm-cmake/CPM.cmake).

This PR is still a work-in-progress as it only supports building a `MoltenVK.dylib` file for macOS and does not neatly package the built files as the current Xcode build does. I'd love to hear your feedback and see if this is a useful addition. I know there is a lot of discussion on https://github.com/KhronosGroup/MoltenVK/pull/1947, so I will review that next to make sure my changes align with the previous thinking.


.github/workflows/CI.yml
```diff
@@ -113,6 +113,56 @@ jobs:
...
+
+    runs-on: ${{ matrix.os }}
+
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Dependencies
+        if: runner.os == 'macOS'
+        run: |
+          brew install ccache boost
```
zfergus, line 135: Boost should be downloaded using CPM.

MoltenVKShaderConverter/MoltenVKShaderConverterTool/CMakeLists.txt
```diff
@@ -0,0 +1,19 @@
+################################################################################
+# MoltenVKShaderConverter Executable
+################################################################################
+
+# Find the source files in this directory. We prefer to use GLOB because it is
+# easier to maintain in conjunction with the Xcode project file.
+file(GLOB_RECURSE SOURCES "*.c" "*.cpp" "*.m" "*.mm")
+
+add_executable(MoltenVKShaderConverter_bin ${SOURCES})
```
zfergus, line 9: I had to add a `_bin` suffix to avoid name clash with the `MoltenVKShaderConverter` folder. I'll look into how to avoid this without changing the name. Maybe it's okay because we can change the name on install or copy to `Package`.

billhollings:
Thanks for working on this. Since you're actively contributing to this PR, I've marked it as WIP in the title. Please remove that marker once you are ready for review and merge.

zfergus:
> Thanks for working on this. Since you're actively contributing to this PR, I've marked it as WIP in the title. Please remove that marker once you are ready for review and merge.

This PR is mostly finished and ready for review. Things I know I need to add:
* Build instructions in the README
* Other config variables; only `MVK_CONFIG_LOG_LEVEL` is supported now

rdrsss:
@zfergus question, does this PR allow for the user to define location of other dependencies if I have already included them. For example if I already have Spir-V as part of my project can I point this config to use that one? The reason I ask is currently with my project I use cpm to pull source archives and then vendor (and commit) them, so I'd like to do the same with this.

Thanks for taking up the mantle btw, this is looking awesome!

zfergus:
@rdrsss there are a few ways you can override the dependencies to use your version.

## Option 1

Every dependency recipe checks if the target already exists before proceeding. For example, at the top of `cmake/recipes/SPIRV-Cross.cmake` is 
```cmake
if(TARGET SPRIV-Cross::SPRIV-Cross)
    return()
endif()
```
This checks if the target already exists before downloading and adding the dependency. You can use this to override the default dependency. That is, before adding MoltenVK as a subproject, you define the correct target aliases such that MoltenVK picks up on them and triggers the above if.

## Option 2

If you have a local clone of a dependency you want to use (e.g., it has changes you are actively developing), use the following to tell CPM to use your local clone.
```cmake
set(CPM_SPIRV-Cross_SOURCE "/path/to/SPIRV-Cross" CACHE PATH "" FORCE)
```

Hopefully, one of these options will work for your scenario.

rdrsss:
@zfergus yes, this works perfectly! Thank you for the explanation! Looking forward to seeing this merged.

rdrsss:
@zfergus @billhollings is this still active?

zfergus:
> @zfergus @billhollings is this still active?

I felt like the PR was at a finalized state, but didn't get much feedback from the maintainers.

squidbus:
This looks like it would be pretty useful for me. I think the CI file conflict would need to be resolved, let me tag for review and see if we can move forward on this.

.gitignore
```diff
@@ -35,3 +35,4 @@ Backup[ ]of[ ]*.pages/
 Backup[ ]of[ ]*.key/
 Backup[ ]of[ ]*.numbers/
 
+MoltenVKOptions.cmake
```
cdavis5e: This file doesn't end in a `\n`. Several other files are also like this.

Wait a minute... are you on Windows?

CMakeLists.txt
```diff
@@ -0,0 +1,121 @@
...
+	cmake_policy(SET CMP0135 NEW) # Set the timestamps of all extracted contents to the time of the extraction.
+endif()
+cmake_policy(SET CMP0114 NEW) # Support the Xcode "new build system"
+
+################################################################################
+
+project(MoltenVK
+	DESCRIPTION "MoltenVK is a Vulkan Portability implementation. It layers a subset of the high-performance, industry-standard Vulkan graphics and compute API over Apple's Metal graphics framework, enabling Vulkan applications to run on macOS, iOS and tvOS."
+	LANGUAGES C CXX OBJC OBJCXX
+	VERSION "1.3.1")
```
cdavis5e: Should be updated to:
```suggestion
	VERSION "1.4.0")
```

MoltenVK/CMakeLists.txt
```diff
@@ -0,0 +1,60 @@
...
+		ERROR_QUIET
+		OUTPUT_STRIP_TRAILING_WHITESPACE
+	)
+	configure_file(
+		"${CMAKE_CURRENT_SOURCE_DIR}/../Templates/cmake/mvkGitRevDerived.h.in"
+		"${CMAKE_CURRENT_BINARY_DIR}/mvkGitRevDerived.h"
+	)
+else()
+	message(FATAL_ERROR "GIT not found")
+endif()
```
cdavis5e, line 41: This will cause builds from distributed source tarballs to fail if the user does not have Git installed.

Templates/cmake/MoltenVKOptions.cmake.sample
```diff
@@ -0,0 +1,40 @@
...
+# Use a specific C/C++ compiler, e.g. llvm-clang on macOS (so we can use clangd)
+# set(CMAKE_C_COMPILER   "/usr/bin/cc"  CACHE STRING "C compiler")
+# set(CMAKE_CXX_COMPILER "/usr/bin/c++" CACHE STRING "C++ compiler")
+
+# Use a specific Objective-C/C++ compiler, e.g. llvm-clang on macOS (so we can use clangd)
+# set(CMAKE_OBJC_COMPILER   "/usr/bin/cc"  CACHE STRING "C compiler")
+# set(CMAKE_OBJCXX_COMPILER "/usr/bin/c++" CACHE STRING "C++ compiler")
+
+# Set deployment platform for macOS
+# set(CMAKE_OSX_DEPLOYMENT_TARGET 10.12 CACHE STRING "macOS deployment target")
```
cdavis5e:
```suggestion
# set(CMAKE_OSX_DEPLOYMENT_TARGET 11.0 CACHE STRING "macOS deployment target")
```
The current minimum is 11.0 for upstream.

.gitignore
zfergus: No, I am using VSCode. Trailing newline is not strictly necessary, right?

cdavis5e: No. It's just that Windows editors tend to be less scrupulous about making sure the file ends in a newline. There are some Unix editors that don't add it though--Emacs and VSCode are the two big ones that come to mind.

MoltenVK/CMakeLists.txt
zfergus: Users will need git to clone the external dependencies regardless, so it's safe to assume it will be available.

cdavis5e: Oh right. Apparently, I did not think this all the way through. 😅

cdavis5e (approved):

Issue #2510, opened 2025-04-28
Setting scissor for render-target-less rasterization violates Metal API Validation.


stripe2933:
When I use render-target-less rasterization pipeline, `MTLRenderPassDescriptor` render target width and height are set to `1`, and setting scissor rect causes Metal API validation error.

```
-[MTLDebugRenderCommandEncoder setScissorRect:]:4053: failed assertion `Set Scissor Rect Validation
(rect.x(346) + rect.width(1))(347) must be <= render pass width(1)
(rect.y(172) + rect.height(1))(173) must be <= render pass height(1)
'
```

It looks like the following code is problematic:

https://github.com/KhronosGroup/MoltenVK/blob/a3ea65e66ef6fa5ce5677d773216d7b0f2d6c962/MoltenVK/MoltenVK/Commands/MVKCommandBuffer.mm#L722-L724

If there's no attachment for the render pass, `getFramebufferExtent()` returns `(0,0)`, and `mtlRPDesc.renderTargetWidthMVK` and `mtlRPDesc.renderTargetHeightMVK` being set to `1`. Is there reason for add the bound-checking code instead of just using

```objectivec
mtlRPDesc.renderTargetWidthMVK = _renderArea.offset.x + _renderArea.extent.width;
mtlRPDesc.renderTargetHeightMVK = _renderArea.offset.y + _renderArea.extent.height;
```

?

---

Not related to this issue, but anybody know how to debug the render-target-less fragment shader in Xcode?

billhollings:
An earlier version of this code had the following comment, explaining the logic. That comment must have been mistakenly removed at some point, hence some confusion about the design.

	// Metal uses MTLRenderPassDescriptor properties renderTargetWidth, renderTargetHeight,
	// and renderTargetArrayLength to preallocate tile memory storage on machines using tiled
	// rendering. This memory preallocation is not necessary if we are not rendering to
	// attachments, and some apps actively define extremely oversized framebuffers when they
	// know they are not rendering to actual attachments, making this internal tile memory
	// allocation even more wasteful, occasionally to the point of triggering OOM crashes.

So. The odd-looking bounds checking is to avoid some errant apps exploding memory. The lesser of two evils is to live with a mistaken Metal validation (Apple shouldn't be validating this when there are no render attachments).

I'll add the comment back, to clarify why the code is as it is.

stripe2933:
[...]

PR #2514, opened 2025-04-29
Fix scissor for render-target-less rasterization

stripe2933:
Completes #2510.


billhollings (changes requested):
See my response and explanation in #2510 about why the bounds checks are there.

I don't think the change in this PR is going to work.

billhollings:
Closing as a zombie PR that hasn't been worked on for a while and might not actually work anyway.

PR #2516, opened 2025-04-30
Add support for VK_KHR_global_priority

squidbus:
Adds `VK_KHR_global_priority`, required for Vulkan 1.4 (https://github.com/KhronosGroup/MoltenVK/issues/2490). Along with that, we also get support for `VK_EXT_global_priority` and `VK_EXT_global_priority_query` for free.

The implementation maps the queue priorities to the dispatch queue QoS classes described as available in the documentation for [dispatch_queue_attr_make_with_qos_class](https://developer.apple.com/documentation/dispatch/1453028-dispatch_queue_attr_make_with_qo), which we currently use when setting up queue dispatches. Medium priority, which is described in spec as the default, is mapped to `QOS_CLASS_USER_INITIATED`, as was previously used by MoltenVK for all queues. Working from there, low is mapped one step down at `QOS_CLASS_UTILITY`, and high is mapped one step up at `QOS_CLASS_USER_INTERACTIVE`. We have no higher class to map to real-time priority, so it is excluded from the query properties as not supported.

Ran the implementation through relevant CTS: [test_gp.txt](https://github.com/user-attachments/files/19971795/test_gp.txt)
```
Test run totals:
  Passed:        60/354 (16.9%)
  Failed:        0/354 (0.0%)
  Not supported: 294/354 (83.1%)
  Warnings:      0/354 (0.0%)
  Waived:        0/354 (0.0%)
```


billhollings (approved):
Thanks for the submission. LGTM.

PR #2519, opened 2025-05-03
Implement VK_KHR_present_id and VK_KHR_present_wait

squidbus:
Adds `VK_KHR_present_wait` and its dependency `VK_KHR_present_id`.

* When a present is queued with a `VkPresentIdKHR`, the IDs contained within are now propagated to the corresponding `MVKImagePresentInfo`.
* When a swapchain is notified that one of its images finished presenting, the current present ID is updated to the maximum of the current ID and the ID of the presented image, and a condition variable is notified.
* As described in the spec, `vkWaitForPresentKHR` will wait on a condition variable for the swapchain's present ID to be greater than or equal to the provided ID to wait on, up to either the timeout or if the swapchain becomes out of date.
* Misc: Fixed some mis-ordered extension definitions. Used the official extension registry as a reference.

Passes `dEQP-VK.wsi.metal.present_id_wait.*`:
```
Test run totals:
  Passed:        9/9 (100.0%)
  Failed:        0/9 (0.0%)
  Not supported: 0/9 (0.0%)
  Warnings:      0/9 (0.0%)
  Waived:        0/9 (0.0%)
```

Passes `dEQP-VK.wsi.headless.present_id_wait.*`:
```
Test run totals:
  Passed:        9/9 (100.0%)
  Failed:        0/9 (0.0%)
  Not supported: 0/9 (0.0%)
  Warnings:      0/9 (0.0%)
  Waived:        0/9 (0.0%)
```


MoltenVK/MoltenVK/GPUObjects/MVKInstance.mm
```diff
@@ -761,6 +764,7 @@
 	ADD_DVC_EXT_ENTRY_POINT(vkGetSwapchainImagesKHR, KHR_SWAPCHAIN);
 	ADD_DVC_EXT_ENTRY_POINT(vkAcquireNextImageKHR, KHR_SWAPCHAIN);
 	ADD_DVC_EXT_ENTRY_POINT(vkQueuePresentKHR, KHR_SWAPCHAIN);
+	ADD_DVC_EXT3_ENTRY_POINT(vkWaitForPresentKHR, KHR_SWAPCHAIN, KHR_PRESENT_ID, KHR_PRESENT_WAIT);
```
billhollings, line 767: There might be some confusion here. 

What is it that you are trying to accomplish with `ADD_DVC_EXT3_ENTRY_POINT()`?

If I'm reading this correctly, you seem to be using this here to indicate that multiple extensions must be enabled for the function to be available. If so, that's not what the logic of `MVKEntryPoint::isExtnSupported()` does. We don't check whether prerequisite extensions are enabled, only the extension that defines the function. The expectation is that the validation layers would catch any extension prerequisite inconstancies. So, `KHR_SWAPCHAIN` and `KHR_PRESENT_ID` are not needed here.

The original addition of an `EXT2` was to cope with a very specific case with Vulkan 1.1 and a few device group entry points. In that case, some functions were defined in more than one extension.

I just realized that a bunch of your `KHR_maintenance6` code, such as `_PROMOTED2_` and `_PROMOTED3_` seems to have been added on a similar assumption. I apologize that I didn't catch and discuss this then.

If this is the case, please remove this here. I can take care of removing the maintenance6 logic around this, since that PR has already been pulled in.

billhollings (changes requested):
Thanks for submitting. 

I have a question about enabling the proc addresses. See my inline comment.

MoltenVK/MoltenVK/GPUObjects/MVKInstance.mm
squidbus: Hmm... I will make that change then, thanks. Although it seems to me that since https://github.com/KhronosGroup/MoltenVK/commit/e09ca0186f7d600125008431652c82046288e3ca the function *is* requiring the presence of all extensions listed, or the corresponding core version.

cdavis5e (approved):

MoltenVK/MoltenVK/GPUObjects/MVKInstance.mm
billhollings:
```
bool isAPISupported = MVK_VULKAN_API_VERSION_CONFORM(enabledVersion) >= apiVersion;
auto isExtnSupported = [this, isAPISupported](const MVKExtensionList& extList) {
	return extList.isEnabled(this->ext1Name) && (isAPISupported ||
			((!this->ext2Name || extList.isEnabled(this->ext2Name)) && (!this->ext3Name || extList.isEnabled(this->ext3Name))));
};
```

Admittedly, the logic here is odd, but it's saying "ext1, _**plus**_ either the api ver, or ext2".

and

```
#define ADD_DVC_1_4_PROMOTED2_ENTRY_POINT(func, suffix, EXT1, EXT2) \
	ADD_DVC_1_4_ENTRY_POINT(func); \
	ADD_ENTRY_POINT_MAP(func##suffix, func, 0, VK_##EXT1##_EXTENSION_NAME, VK_##EXT2##_EXTENSION_NAME, nullptr, true)
```

the `apiVersion` is always set to 0 for every extension function (except those couple I mentioned earlier), so for an extension function, it will always trigger `isAPISupported`, and the additional extensions will never be checked.

Basically, all that tortuous logic is just to handle the odd-ball case I mentioned. I guess I should document that in a comment.

billhollings (approved):
Thanks. Looks great now.

PR #2521, opened 2025-05-04
MVKDevice refactor enabling promoted physical device features.

billhollings:
- Add and use macro to enable extension features from promoted features structs.
- Rename Vulkan 1.2 feature structs to allow macro to work on them.

### DIFF
--- a/MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
@@ -194,8 +194,8 @@ void MVKPhysicalDevice::getFeatures(VkPhysicalDeviceFeatures2* features) {
 	VkPhysicalDeviceVulkan12Features supportedFeats12 = {
 		.sType = VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_VULKAN_1_2_FEATURES,
 		.pNext = nullptr,
-		.samplerMirrorClampToEdge = _vulkan12FeaturesNoExt.samplerMirrorClampToEdge,
-		.drawIndirectCount = _vulkan12FeaturesNoExt.drawIndirectCount,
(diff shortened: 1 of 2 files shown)


MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
```diff
@@ -5324,164 +5324,65 @@ static uint32_t mvkGetEntryProperty(io_registry_entry_t entry, CFStringRef prope
...
-				enableFeatures(requestedFeatures,
-							   &_enabledVulkanMemoryModelFeatures.vulkanMemoryModel,
-							   &requestedFeatures->vulkanMemoryModel,
-							   &pdVulkanMemoryModelFeatures.vulkanMemoryModel, 3);
-				enableFeatures(requestedFeatures,
-							   &_enabledVulkan12FeaturesNoExt.shaderOutputViewportIndex,
-							   &requestedFeatures->shaderOutputViewportIndex,
-							   &pdvulkan12FeaturesNoExt.shaderOutputViewportIndex, 3);
+				enablePromotedFeatures(Vulkan12NoExt, samplerMirrorClampToEdge, 2);
+				enablePromotedFeatures(8BitStorage, storageBuffer8BitAccess, 3);
```
cdavis5e, line 5349: I'm amazed this works, given that the lexical rules for C++ say that an identifier can't start with a numeral. I guess the preprocessor is OK with that as long as it gets pasted into an identifier that _is_ legal.

cdavis5e (approved):

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
billhollings: Yeah. I had a brief "oh damn this won't work" moment when I got to `8BitStorage`, but then I remembered that we had used numeric starts elsewhere already. 

It actually serves as a reminder that the preprocessor really is just a text-substitution mechanism.

PR #2522, opened 2025-05-04
Add support for a few miscellaneous small extensions

squidbus:
Adds support for a few miscellaneous small extensions that require little to no change to advertise:
* `VK_KHR_surface_protected_capabilities`: This extension simply exists to inform the user whether a surface supports protected mode. Since we don't support protected at all, this can simply set the capability to false when asked.
* `VK_KHR_shader_subgroup_uniform_control_flow`: Passes tests on Apple GPUs with no modifications.
  * This one required me to increase the alignment spacing for the extension definitions since it was too wide (again...)
* `VK_KHR_shader_maximal_reconvergence`: Passes tests on Apple GPUs with no modifications.

Ran these CTS tests:
```
dEQP-VK.wsi.*.surface.query_protected_capabilities
dEQP-VK.subgroups.subgroup_uniform_control_flow.*
dEQP-VK.reconvergence.maximal.*
```

All passed except `dEQP-VK.reconvergence.maximal.fragment.terminate_invocation` on Apple GPUs (tested M1 and M4), because we do not have terminate vs demote semantics implemented in MSL, only one depending on the Metal version. You can see with the `dEQP-VK.spirv_assembly.instruction.terminate_invocation.*` tests for that instruction as well that some of the subgroup related cases fail.

On Intel and AMD GPUs only about 15 random tests fail out of the roughly 1500, so I erred on the side of not exposing the extension for those.


billhollings (approved):
LGTM. Thanks.

PR #2526, opened 2025-05-06
Cleanup MVKInstance proc-addr function mapping.

billhollings:
Multi-extension configuration on proc-addrs really wasn't doing what it was being used for. It really only had once specialized purpose, so it's been removed it altogether. The specialized device group functions support is now handled more explicitly with a clean test for existence of a function either within a version or extension.

- Remove multi-extension testing when determine function support.
- Ensure entry point API version is set before comparing it.
- Add the ability to test the special case of an extension function in core, or an extension, under the same function name (promoted but not renamed).


squidbus (approved):
Seems more straightforward now, thanks.

cdavis5e:
Does this pass the CTS?

billhollings:
> Does this pass the CTS?

Yes in the sense that it doesn't add anything new by itself AFAICT.

Issue #2529, opened 2025-05-07
Possible regression in `OpSMod`?


billhollings:
There are about 75 previously-passing SPIR-V `OpSMod` CTS tests that are now failing:

`dEQP-VK.spirv_assembly.type.*.mod_*`

including for example:

`dEQP-VK.spirv_assembly.type.scalar.i32.mod_vert`

Could this possibly be related to [this](https://github.com/KhronosGroup/SPIRV-Cross/pull/2475) recent `OpSMod` fix in SPIRV-Cross?

@squidbus

squidbus:
Hmm, I will take a look. The related `maintenance8` test cases for positive and negative numbers are still passing.

squidbus:
Fixing in SPIRV-Cross here: https://github.com/KhronosGroup/SPIRV-Cross/pull/2480

squidbus:
Fixed by SPIRV-Cross update here: https://github.com/KhronosGroup/MoltenVK/pull/2533

billhollings:
Thanks!

Issue #2478, opened 2025-04-08
Support for `VK_KHR_shader_subgroup_uniform_control_flow` or even `VK_KHR_shader_maximal_reconvergence`?


tareksander:
I don't know anything about the uniformity model of Metal. Would it be possible to implement these extensions? Given that subgroup operations are supported, it wouldn't seem far-fetched that MSL supports reconvergence, too.

squidbus:
The main two asks here pass tests on Apple GPUs and have been added here: https://github.com/KhronosGroup/MoltenVK/pull/2522

For `VK_KHR_shader_quad_control` I've been taking a look and have some of the SPIRV-Cross instruction wiring done but currently failing `dEQP-VK.subgroups.shader_quad_control.divergent_condition`.

squidbus:
> For `VK_KHR_shader_quad_control` I've been taking a look and have some of the SPIRV-Cross instruction wiring done but currently failing `dEQP-VK.subgroups.shader_quad_control.divergent_condition`.

This seems likely to be a Metal bug as opening a trace in a debugger shows the shader outputting the correct values but the actual render target containing incorrect values.

squidbus:
It seems actually that with some fixes for `OpSMod` regressions all tests are now passing, so it may have been my fault. Not sure what was up with the debugger view I was seeing though.

squidbus:
Quad control is added here: https://github.com/KhronosGroup/MoltenVK/pull/2533

For `VK_KHR_shader_clock` it does not seem to be possible to implement currently in MSL.

PR #2533, opened 2025-05-08
Add support for VK_KHR_shader_quad_control

squidbus:
* Updates SPIRV-Cross to latest revision, including fix for `OpSMod` regression and support for `SPV_KHR_quad_control`.
* Advertises support for `VK_KHR_shader_quad_control` for Apple GPUs with quad operations.
  * Other GPUs do not have maximal reconvergence which this depends on, plus fail the test for requiring full quads.

CTS results:

`dEQP-VK.spirv_assembly.type.*.mod_*`:
```
Test run totals:
  Passed:        200/240 (83.3%)
  Failed:        0/240 (0.0%)
  Not supported: 40/240 (16.7%)
  Warnings:      0/240 (0.0%)
  Waived:        0/240 (0.0%)
```

`dEQP-VK.spirv_assembly.*.maintenance8.*`:
```
Test run totals:
  Passed:        24/28 (85.7%)
  Failed:        0/28 (0.0%)
  Not supported: 4/28 (14.3%)
  Warnings:      0/28 (0.0%)
  Waived:        0/28 (0.0%)
```

`dEQP-VK.subgroups.shader_quad_control.*`:
```
Test run totals:
  Passed:        3/3 (100.0%)
  Failed:        0/3 (0.0%)
  Not supported: 0/3 (0.0%)
  Warnings:      0/3 (0.0%)
  Waived:        0/3 (0.0%)
```

Completes https://github.com/KhronosGroup/MoltenVK/issues/2529 and https://github.com/KhronosGroup/MoltenVK/issues/2478. I believe this also gets us the last extension for roadmap 2024 other than those remaining required for Vulkan 1.4.


cdavis5e (approved):

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
```diff
@@ -618,6 +618,11 @@
 				shaderReconvergenceFeatures->shaderMaximalReconvergence = _gpuCapabilities.isAppleGPU;
 				break;
 			}
+			case VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_SHADER_QUAD_CONTROL_FEATURES_KHR: {
+				auto* shaderQuadControlFeatures = (VkPhysicalDeviceShaderQuadControlFeaturesKHR*)next;
+				shaderQuadControlFeatures->shaderQuadControl = _gpuCapabilities.isAppleGPU && _metalFeatures.quadPermute;
+				break;
```
billhollings, line 624: Given Apple's GPU support path, this is probably a moot academic point to make, but:

1) Is the dependency on Apple GPU, or maximal reconvergence? If a future update supports maximal reconvergence on a non-Apple GPU, or a future Apple GPU doesn't support maximal reconvergence, this might not get updated.

2) The `_gpuCapabilities.isAppleGPU && _metalFeatures.quadPermute` test appears twice, once here and once below for the extension (negated). There is a chance a future change would update one but forget to update the other. I always like a SSOT.

As I say, this is probably entirely academic in practice, but I'm wondering if a new `MVKPhysicalDeviceMetalFeatures::maximalReconvergence` would help, making the equivalent test `_metalFeatures.maximalReconvergence && _metalFeatures.quadPermute` instead?

I'm not stuck on this, so it's entirely fine to tell me I'm being unnecessarily pedantic and it's not worth the trouble. 😉

billhollings (changes requested):
Thanks for doing this.

LGTM. I just have a minor nit question about how the test for this capability is made.

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
squidbus: I touched on this briefly but for other GPUs it seems we do not have the guarantee required by `RequireFullQuadsKHR`: `Helper invocations must be spawned such that all quad groups start with four active invocations. Only valid with the Fragment Execution Model.`

Given this, would you want to see two metal features, one for `maximalReconvergence` and one for something like `spawnsFullQuads` (let me know if you prefer a different name for this)?

squidbus: I've updated with some Metal feature flags I think make sense for this, let me know if it looks okay or you would like further changes.

billhollings: > I've updated with some Metal feature flags I think make sense for this, let me know if it looks okay or you would like further changes.

Very nice! I like it! Thanks!

billhollings: > I've updated with some Metal feature flags I think make sense for this, let me know if it looks okay or you would like further changes.

Very nice! I like it! Thanks.

billhollings (approved):
LGTM now. Thanks.

Issue #2532, opened 2025-05-08
vkGetPerformanceStatisticsMVK() is no longer visible for dynamic proc addr lookups


SRSaunders:
It seems that `vkGetPerformanceStatisticsMVK()` is no longer visible for dynamic lookups as follows:

`vkGetPerformanceStatisticsMVK = ( PFN_vkGetPerformanceStatisticsMVK )vkGetInstanceProcAddr( m_VulkanInstance, "vkGetPerformanceStatisticsMVK" );`

This operation now returns a null pointer.  I suspect the recent PR #2526 has introduced a regression here.

I can reference `vkGetPerformanceStatisticsMVK()` via linking at build time, but this is not what I want.

@billhollings would it be possible to return dynamic lookup functionality for this private API function?

Note that `vkGetPhysicalDeviceMetalFeaturesMVK()` may suffer from the same problem, but I have not tested.

SRSaunders:
Looking at the code, a possible solution is:

```
typedef struct MVKEntryPoint {
	PFN_vkVoidFunction functionPointer;
	const char* extName;
	uint32_t apiVersion;
	bool isDevice;

	bool isCore() { return apiVersion >= 0; }
                                           ^
```

This is since you currently define:

```
#define ADD_INST_OPEN_ENTRY_POINT(func)		ADD_ENTRY_POINT(func, 0, nullptr, false)
                                                                      ^

// MoltenVK-specific instannce functions, not tied to a Vulkan API version or an extension.
ADD_INST_OPEN_ENTRY_POINT(vkGetPhysicalDeviceMetalFeaturesMVK);
ADD_INST_OPEN_ENTRY_POINT(vkGetPerformanceStatisticsMVK);
```

However, I am not sure about the implications of allowing `"0"` for permitted apiVersion in `isCore()`, or whether you could simply change the  `ADD_INST_OPEN_ENTRY_POINT` definition to arbitrarily be `VK_API_VERSION_1_0`.

Note that previous `isCore()` logic worked since you were testing for `!ext1Name && !ext2Name && !ext3Name` which passes in the case of these MoltenVK private API functions.

billhollings:
Uggh. You're right. In that change, I neglected the `ADD_INST_OPEN_ENTRY_POINT` functions.

Thanks for the detective work, I'll sort this out and push a patch.

billhollings:
[...]

PR #2538, opened 2025-05-11
Re-expose proc addr for VK_MVK_moltenvk extension functions.

billhollings:
Several proc addr for `VK_MVK_moltenvk` extension functions were erroneously hidden by commit 70db7651.

- Handle the proc addrs for functions `vkGetPhysicalDeviceMetalFeaturesMVK()`, `vkGetPerformanceStatisticsMVK()`, and `vkSetMoltenVKConfigurationMVK()` consistently with all other functions, by replacing these functions under the `VK_MVK_moltenvk` extension, and requiring that extension to be enabled to use the functions.
- Remove `ADD_INST_OPEN_ENTRY_POINT()` and `ADD_DVC_OPEN_ENTRY_POINT()` macros.
- Improve warning message output when extension `VK_MVK_moltenvk` is enabled.

Completes #2532.

@SRSaunders  I put the previously un-categorized functions under `VK_MVK_moltenvk`, where they started out. This makes them like any other extension functions, to avoid the special-case situation that caused them to fall through the cracks in #2532.

It does mean that someone using these functions now needs to enable the `VK_MVK_moltenvk`, which again is consistent with all other Vulkan behaviour. But if this will cause legacy issues, what we can do is simply enable `VK_MVK_moltenvk` by default in MoltenVK, so the user doesn't need to take that step.

### DIFF
--- a/Docs/Whats_New.md
+++ b/Docs/Whats_New.md
@@ -29,6 +29,9 @@ Released TBD
 	- `VK_KHR_shader_relaxed_extended_instruction`
 	- `VK_KHR_shader_subgroup_uniform_control_flow`
 	- `VK_KHR_surface_protected_capabilities`
+- For behaviouraly consistency, require deprecated unofficial extension `VK_MVK_moltenvk` be enabled to access 
+  the proc addrs for functions `vkSetMoltenVKConfigurationMVK()` and `vkGetPhysicalDeviceMetalFeaturesMVK()`. 
+- Move `vkGetPhysicalDeviceMetalFeaturesMVK()` from `mvk_private_api.h` to `mvk_deprecated_api.h`.
 - Update to latest SPIRV-Cross:
   - MSL: Add support for `DebugPrintf`.
   - MSL: Fix crash due to regression caused by recent changes to location calculations.
--- a/MoltenVK/MoltenVK/API/mvk_deprecated_api.h
+++ b/MoltenVK/MoltenVK/API/mvk_deprecated_api.h
@@ -46,11 +46,128 @@ extern "C" {
  * The VK_EXT_metal_objects extension is supported by the Vulkan Loader and Layers.
  */
 
+#pragma mark -
+#pragma mark VkPhysicalDevice Metal capabilities
+
+/** Identifies the type of rounding Metal uses for float to integer conversions in particular calculatons. */
+typedef enum MVKFloatRounding {
+	MVK_FLOAT_ROUNDING_NEAREST     = 0,	 /**< Metal rounds to nearest. */
+	MVK_FLOAT_ROUNDING_UP          = 1,	 /**< Metal rounds towards positive infinity. */
+	MVK_FLOAT_ROUNDING_DOWN        = 2,	 /**< Metal rounds towards negative infinity. */
+	MVK_FLOAT_ROUNDING_UP_MAX_ENUM = 0x7FFFFFFF
+} MVKFloatRounding;
+
+/** Identifies the pipeline points where GPU counter sampling can occur. Maps to MTLCounterSamplingPoint. */
+typedef enum MVKCounterSamplingBits {
+	MVK_COUNTER_SAMPLING_AT_DRAW           = 0x00000001,
+	MVK_COUNTER_SAMPLING_AT_DISPATCH       = 0x00000002,
+	MVK_COUNTER_SAMPLING_AT_BLIT           = 0x00000004,
+	MVK_COUNTER_SAMPLING_AT_PIPELINE_STAGE = 0x00000008,
+	MVK_COUNTER_SAMPLING_MAX_ENUM          = 0X7FFFFFFF
+} MVKCounterSamplingBits;
+typedef VkFlags MVKCounterSamplingFlags;
+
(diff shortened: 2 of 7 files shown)


SRSaunders:
@billhollings thanks for looking into this, and while it works as designed, I have some concerns with the solution:

1. While `vkGetPhysicalDeviceMetalFeaturesMVK()` and `vkSetMoltenVKConfigurationMVK()` are indeed deprecated and have other officially supported extensions, the one I was concerned about `vkGetPerformanceStatisticsMVK()` is not deprecated.  Your design forces it down the deprecated path with a few negative implications:  a) a user has to include  `<MoltenVK/mvk_deprecated_api.h>` for access to the `VK_MVK_MOLTENVK_EXTENSION_NAME` symbol, and b) you print a `[mvk-warn]` message that the user is doing something unsupported, which is incorrect for the performance API.
2. I do have legacy concerns about mandating enablement of the `VK_MVK_moltenvk` extension for access to the `vkGetPerformanceStatisticsMVK()` API.  In my case I have put a lot of work into using this API for the RBDoom3BFG project and it is currently being prepped for an official 1.6 release and I don't want breakage at the last second.
3. Placing `vkSetMoltenVKConfigurationMVK()` (even though it's deprecated) behind this checking logic is not strictly correct.  This function was originally designed to work with a NULL instance parameter, and for some config variables must be called ahead of `vkCreateInstance()`.  To meet that requirement it really should be part of the decision tree in `vkGetInstanceProcAddr()` and not be forced down the `mvkInst->getProcAddr()` path.

Is there any way to revisit your design so that users are properly warned about use of deprecated private API functions, but not falsely warned for the performance API?  And if you still want to protect `vkGetPerformanceStatisticsMVK()` behind the `VK_MVK_moltenvk` extension, then your suggestion of enabling it by default would be fine, just as long as it does not print out a false positive deprecated API warning.

squidbus:
Just my two cents: MVK performance statistics probably *should* be deprecated and replaced with something like VK_KHR_performance_query, but perhaps that’s something that would take longer to do.

SRSaunders:
I am all for advancement and standardization, but until such time as an alternative performance stats implementation is available, I would hope that the current solution can remain supported and not be marked as deprecated.  The current API is a very useful capability, and in fact was the tool I used to determine that Metal encoding for constant buffers is very inefficient on macOS MoltenVK, and the reason I migrated to using push constants in the RBDoom3BFG application.  That is a different problem and not for this thread, but without the perf API I would not have been able to clearly measure this.

In any case, and to avoid any confusion, the changes I am suggesting for this PR are the following:

```
	// For deprecated MoltenVK-specific functions, suppress compiler deprecation warning.
#pragma clang diagnostic push
#pragma clang diagnostic ignored "-Wdeprecated-declarations"
	ADD_INST_EXT_ENTRY_POINT(vkGetPhysicalDeviceMetalFeaturesMVK, MVK_MOLTENVK);
del -->	ADD_INST_EXT_ENTRY_POINT(vkGetPerformanceStatisticsMVK, MVK_MOLTENVK);
	ADD_INST_EXT_ENTRY_POINT(vkSetMoltenVKConfigurationMVK, MVK_MOLTENVK);
	ADD_INST_EXT_ENTRY_POINT(vkGetVersionStringsMVK, MVK_MOLTENVK);
	...
```


```
MVK_PUBLIC_SYMBOL PFN_vkVoidFunction vkGetInstanceProcAddr(
    VkInstance                                  instance,
    const char*                                 pName) {

	// Handle the special platform functions where the instance parameter may be NULL.
	PFN_vkVoidFunction func = nullptr;
	MVKTraceVulkanCallStart();
	if (mvkStringsAreEqual(pName, "vkGetInstanceProcAddr")) {
		func = (PFN_vkVoidFunction)vkGetInstanceProcAddr;
	...
	} else if (mvkStringsAreEqual(pName, "vkGetMoltenVKConfigurationMVK")) {
		func = (PFN_vkVoidFunction)vkGetMoltenVKConfigurationMVK;
add -->	} else if (mvkStringsAreEqual(pName, "vkGetPerformanceStatisticsMVK")) {
add -->		func = (PFN_vkVoidFunction)vkGetPerformanceStatisticsMVK;
	} else if (instance) {
		MVKInstance* mvkInst = MVKInstance::getMVKInstance(instance);
		func = mvkInst->getProcAddr(pName);
	}
```

billhollings:
@SRSaunders I mades some more changes that should work for you now.

Aligned with an earlier suggestion of yours, `vkGetPerformanceStatisticsMVK()` is now marked with `VK_API_VERSION_1_0`, with the plan being to deprecate it once `VK_KHR_performance_query` is supported.

`vkGetPhysicalDeviceMetalFeaturesMVK()` now moved to `mvk_deprecated_api.h`. Thanks for noticing that and making the suggestion.

SRSaunders:
Thanks very much @billhollings.

This solution works well and addresses my concerns.  I have tested against the newly released RBDoom3BFG 1.6.0 and all is good.  I appreciate you solving this without breakage until something new can be put in place around `VK_KHR_performance_query`.

Please feel free to merge this PR and close my original issue.

PR #2543, opened 2025-05-16
Fix high latency on present wait.

squidbus:
Encountered an issue using present wait in practice in [Unleashed Recompiled](https://github.com/hedge-dev/UnleashedRecomp), where when v-sync is enabled the present wait would take too long and severely disrupt frame pacing.

While looking into it I noticed a section near the present handler logic where we signal the present fence in a separate completed handler, because the present handler callbacks can possibly stall. Indeed, moving the logic to notify the swapchain to update the present ID seems to have solved my issues; I went from having wild fluctuations around 9ms for present waits for `currentPresentId - 1` to around a microsecond. I also verified with timestamps that the completed handler is paced correctly, rather than just seeing the results I expect due to it calling very quickly or something.

So, in this PR I've split out the present ID update to a separate swapchain function `notifyPresentComplete`, and call that from the end of the completed handler.

### DIFF
--- a/MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
@@ -1629,8 +1629,9 @@ VkResult MVKPresentableSwapchainImage::presentCAMetalDrawable(id<MTLCommandBuffe
 	// Ensure this image, the drawable, and the present fence are not destroyed while
 	// awaiting MTLCommandBuffer completion. We retain the drawable separately because
 	// a new drawable might be acquired by this image by then.
-	// Signal the fence from this callback, because the last one or two presentation
-	// completion callbacks can occasionally stall.
+	// Signal the fence and notify the swapchain that the present has completed
+	// from this callback, because the last one or two presentation completion
+	// callbacks can occasionally stall.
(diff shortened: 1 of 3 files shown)


MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
```diff
@@ -1629,8 +1629,9 @@ static void signalAndUntrack(const MVKSwapchainSignaler& signaler) {
 	// Ensure this image, the drawable, and the present fence are not destroyed while
 	// awaiting MTLCommandBuffer completion. We retain the drawable separately because
 	// a new drawable might be acquired by this image by then.
-	// Signal the fence from this callback, because the last one or two presentation
-	// completion callbacks can occasionally stall.
+	// Signal the fence and notify the swapchain that the present has completed
+	// from this callback, because the last one or two presentation completion
+	// callbacks can occasionally stall.
```
billhollings, line 1634: Nice. Great catch!

I really struggled with these lost/stalled presentation callbacks at the time I added this. Thanks for figuring out some further requirements!

billhollings (approved):
Nice. Great catch! Thanks for submitting this!

LGTM.

Issue #2547, opened 2025-05-22
CTS stalls with MoltenVK 1.3.1


billhollings:
I'm trying to get CTS 1.4.2 to make a full run of MoltenVK 1.3.1, however it is stalling to the point of running **_very_** slowly in two significant areas, and probably will not complete within the lifetime of the universe. MoltenVK 1.3.0 runs fine.

1) Running `dEQP-VK.synchronization2.*` illustrates that `vkGetPipelineCacheData()` and `vkCreatePipelineCache()` are slowing down by at least 2 orders of magnitude or so due to a huge increase in the number of shaders being cached. I assume this has something to do with how shaders are hashed in the cache, but so far no clear indication why. Version 1.3.1 doesn't seem to have made any changes that would affect this.

2) `dEQP-VK.wsi.headless.*` tests are retaining a **_lot_** GPU memory from the headless textures, as evidenced by:

`[mvk-error] Destroyed VkPhysicalDevice for GPU Apple M3 Pro with 11753 MB of GPU memory still allocated.`

This memory accumulates and is never destroyed. As checked by logging, the textures themselves seem to be destroyed properly, and _Instruments_ shows no leaks (which would be the case if a`MTLTexture` was leaked).

This headless problem probably predates 1.3.1, but it affects both memory and completion time.

I am currently investigating both of these problems, but thought I'd document them here in case anyone else has any insights.

@squidbus @cdavis5e

billhollings:
> For what it's worth I don't think I've ever done a full run as I always hang at `dEQP-VK.api.device_init.create_instance_device_intentional_alloc_fail.basic` (unless I'm just not waiting long enough or something).

Yes. `dEQP-VK.api.device_init.create_instance_device_intentional_alloc_fail.basic`, takes forever, but eventually finishes in 5+ minutes or so. The delays are actually not in MoltenVK, but in the loader layers. The tests make many, many `vkCreateInstance()` calls with bogus allocators, that never make it to MoltenVK.

A bunch of the `dEQP-VK.wsi.metal.*` tests also take 5+ minutes each. And as I mention above, headless has its issues with time and size.

> For the synchronization tests, some of the maintenance extensions involve pipeline caches and synchronization so I could see the number of supported tests having increased since 1.3.0.

Other than internal merge sync, I don't see anything obvious with pipeline caches, but there is a definite problem with shaders that should match cached shaders being added over and over again. I'm continuing to research it.

squidbus:
[...]

PR #2550, opened 2025-05-24
Fix shader cache misses within pipeline cache.

billhollings:
- Update Cereal Archive for `CompilerMSL::Options` to fix cache misses.
- Add validation functions to report if Cereal Archive definitions deviate from struct definitions in the future.
- Replace unused variable warning suppression with `[[maybe_unused]]` compilation attribute (unrelated).

Fixes part of #2547.

### DIFF
--- a/MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
@@ -52,25 +52,19 @@ using namespace std;
 #	define MVKViewClass		NSView
 #endif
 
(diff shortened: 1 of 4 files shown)


squidbus (approved):
Makes sense. I guess the missing point size stuff from maintenance5 probably contributed to the recent issues.

Would have liked to avoid needing to manually specify the padding, but I can’t think of a good way to do that either. I thought about if we could write then read back the data to compare instead of size. But then you still have an issue with how to put data into the struct without manually writing all the fields (defeating the purpose) or writing over padding which will make the read struct unequal.

Issue #2541, opened 2025-05-14
vkDeviceWaitIdle does nothing if MVKDevice is in lost state


TheMrButcher:
vkDeviceWaitIdle must wait till all the work is done (even if device is lost), cause it's used to ensure that it is safe to delete logical device. Right now vkDeviceWaitIdle doesn't do anything if device is lost: [MVKDevice::waitIdle](https://github.com/KhronosGroup/MoltenVK/blob/147fbeaf7a7851327f8af20078e6e5a346c2ec05/MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm#L4021) uses [MVKQueue::waitIdle](https://github.com/KhronosGroup/MoltenVK/blob/147fbeaf7a7851327f8af20078e6e5a346c2ec05/MoltenVK/MoltenVK/GPUObjects/MVKQueue.mm#L152C10-L152C28) which checks the state of device and if it is not VK_SUCCES, it returns immediately. This leads to [this problem](https://github.com/KhronosGroup/MoltenVK/issues/2270) that I reported before. Also see answer of Nicol Bolas in [this stackoverflow discussion](https://stackoverflow.com/questions/79614545/deleting-vkdevice-after-vk-error-device-lost), he also says that this is bug.

Issue #2270, opened 2024-07-10
Crash when deleting lost device


TheMrButcher:
I am using MoltenVK on iOS, and I get a lot of VK_ERROR_DEVICE_LOST during work of application. I am trying to add support of this error code and recreate logical device when this happens. The problem is that I can't use vkDeviceWaitIdle to wait for stop of all work. Due to specs and according to code of MoltenVK vkDeviceWaitIdle can and will return VK_ERROR_DEVICE_LOST if device is lost.

So it seems to me that I don't have any way to wait until all command buffer submissions are finished. Without this wait I get a lot of errors with this trace in thread that is managed by MoltenVK just after deletion of device:
`SIGABRT 0x0000000000000000
Pure virtual function called!
abort() called
SWIFT TASK CONTINUATION MISUSE: navigate(to:) leaked its continuation!

Crashed: com.Metal.CompletionQueueDispatch
0  libsystem_kernel.dylib   __pthread_kill + 8
1  libsystem_pthread.dylib  pthread_kill + 208
2  libsystem_c.dylib        abort + 124
3  libc++abi.dylib          __cxxabiv1::__aligned_malloc_with_fallback(unsigned long) + 0
4  libc++abi.dylib          __cxa_deleted_virtual + 0
5  MyApp                    ___ZN13MVKBaseObject13reportMessageEPS_17MVKConfigLogLevelPKcPc + 100
6  MyApp                    ___ZN13MVKBaseObject11reportErrorEPS_8VkResultPKcPc + 180
7  MyApp                    ___ZN13MVKBaseObject11reportErrorE8VkResultPKcz + 36
8  MyApp                    ___ZN31MVKQueueCommandBufferSubmission28commitActiveMTLCommandBufferEb_block_invoke + 196
9  Metal                    MTLDispatchListApply + 44
10 Metal                    -[_MTLCommandBuffer didCompleteWithStartTime:endTime:error:] + 596
11 IOGPU                    -[IOGPUMetalCommandBuffer didCompleteWithStartTime:endTime:error:] + 216
12 Metal                    -[_MTLCommandQueue commandBufferDidComplete:startTime:completionTime:error:] + 132
13 IOGPU                    __54-[IOGPUMetalCommandQueue _submitCommandBuffers:count:]_block_invoke.22 + 164
14 IOGPU                    __IOGPUNotificationQueueSetDispatchQueue_block_invoke + 156`

TheMrButcher:
[...]

PR #2553, opened 2025-05-30
vkDeviceWaitIdle() & vkQueueWaitIdle() wait even on VK_ERROR_DEVICE_LOST.

billhollings:
Fixes issues #2541 and #2270.


cdavis5e (approved):
Didn't someone point out in an issue that we're supposed to perform the wait even when the device is lost?

Now, where was it...? Ah, here it is: #2541.

cdavis5e (approved):

Issue #2518, opened 2025-05-01
Support VK_KHR_line_rasterization


squidbus:
This is required for Vulkan 1.4. As an extension, it requires only that you support at least one line style, but for 1.4 it requires specifically that you support Bresenham lines.

CTS tests seem to produce mixed results on which is used by Metal on Apple Silicon GPUs. The Bresenham tests are the only to pass with single sample, and the Rectangular tests are the only to pass with multi sample.

squidbus:
It seems the expectation of CTS is that Bresenham lines have no AA: https://github.com/KhronosGroup/VK-GL-CTS/blob/main/external/vulkancts/modules/vulkan/rasterization/vktRasterizationTests.cpp#L2147-L2150

Perhaps we can somehow 'turn off' multisample for line rasterization?

squidbus:
Took a stab at it here, using the sample positions feature to fix all to center when drawing lines: https://github.com/squidbus/MoltenVK/tree/line_rasterization

It does work in tests but I'm having trouble figuring out how to wire everything up to take care of dynamic state instead of just doing it within a pipeline's state, if anyone wants to pitch in there.

squidbus:
These are the results I get from the rasterization line tests, which are admittedly rather confusing:

Bresenham (`dEQP-VK.rasterization.*.bresenham*`):
* Apple Silicon/AMD: Single-sample tests pass, multi-sample tests pass with sample positions forced to middle of pixel. Other drivers like Mesa also seem to take the strategy of forcing multisampling off for these lines.
* Intel: Single-sample tests pass, all multi-sample tests fail even with sample positions forced to middle.

Rectangular (`dEQP-VK.rasterization.*.rectangular*`):
* Apple Silicon: Single-sample tests fail, multi-sample tests pass.
* Intel/AMD: All tests pass with some quality warnings.

Maintenance5 non-strict lines (`dEQP-VK.rasterization.*.non_strict*`), with the relevant parallelogram properties set to either `false` or `true`:
* Apple Silicon/Intel/AMD: Pass

Also worth noting that some of the strict lines tests fail on Intel.

alyssarosenzweig:
> It does work in tests but I'm having trouble figuring out how to wire everything up to take care of dynamic state instead of just doing it within a pipeline's state, if anyone wants to pitch in there.

[...]

PR #2556, opened 2025-06-03
Support extensions VK_KHR_line_rasterization & VK_EXT_line_rasterization.

billhollings:
- Add `MVKCmdSetLineRasterizationMode`.
- Add `MVKCommandEncoder::getSampleCount()`.
- `MVKRenderingCommandEncoderState` when rendering multisample Bresenham lines, override sample positions and force them to the center.
- `MVKGraphicsPipeline` set `PrimitiveTopology` & `LineRasterizationMode` into rendering state at pipeline bind time, as they can affect sample position at renderpass begin.
- Enable Bresenham lines on all platforms in preparation for Vulkan 1.4.
- Disable other line rasterization modes that fail CTS.

Tested for CTS on M3 Apple Silicon. Other GPU's may have different results.

Completes #2518.


MoltenVK/MoltenVK/Commands/MVKCommandEncoderState.mm
```diff
@@ -433,6 +440,14 @@
 	}
 }
 
+bool MVKRenderingCommandEncoderState::isDrawingLines() {
+	switch (getPrimitiveType()) {
+		case MTLPrimitiveTypeLine:      return true;
+		case MTLPrimitiveTypeLineStrip: return true;
+		default:                        return false;
+	}
```
squidbus, line 458: Do we need to account for line polygon mode here as well?

MoltenVK/MoltenVK/Commands/MVKCommandEncoderState.mm
```diff
@@ -476,7 +491,36 @@
...
+// Multisample Bresenham lines require sampling from the pixel center.
+// Additional sample position overrides based on other rendering requirements could be added here too.
+void MVKRenderingCommandEncoderState::checkSamplePositionsOverride() {
+	auto oldSamplePositionsOverride = _samplePositionsOverride;
+
+	_samplePositionsOverride = None;
+
+	// Multisample Bresenham lines require sampling from the pixel center.
+	if (isDrawingLines() &&
+		getVkContent(LineRasterizationMode) == VK_LINE_RASTERIZATION_MODE_BRESENHAM &&
```
squidbus, line 517: The definition of `VK_LINE_RASTERIZATION_MODE_DEFAULT` when rectangular lines is not supported is still not fully clear to me, but I assume only applying this to explicit `VK_LINE_RASTERIZATION_MODE_BRESENHAM` passes CTS?

If we're not defaulting to correct Bresenham lines for all sample counts then I assume this might affect the values of `nonStrictSinglePixelWideLinesUseParallelogram` and `nonStrictWideLinesUseParallelogram` in the maintenance5/1.4 properties, as the spec indicates `false` means that Bresenham lines are drawn, which wouldn't match up in the multisampled `VK_LINE_RASTERIZATION_MODE_DEFAULT` case here. Those might need to be flipped to `true` then, although I'm not sure how strict the spec treats the definition of parallelogram lines either.

MoltenVK/MoltenVK/Commands/MVKCommandEncoderState.mm
billhollings: Can you clarify what you're thinking of here?

MoltenVK/MoltenVK/Commands/MVKCommandEncoderState.mm
billhollings: When it comes to line primitives, Metal is gonna render what it does. None of the Vulkan line rasterization settings (with the exception of the sample points that you found) affect what is actually rendered. So it's really just about does the Metal line match what CTS is expecting, based on CTS's interpretation of the Vulkan line rasterization settings.

Since we are limited to single-width lines, there is less variation on the line structure. Rectangular lines fail on a single pixel at the start of the line. So far, with a value of `VK_LINE_RASTERIZATION_MODE_DEFAULT` or `VK_LINE_RASTERIZATION_MODE_BRESENHAM`, the Metal line matches what CTS is expecting, even with multisample:

```
Test case 'dEQP-VK.rasterization.primitives_multisample_4_bit.no_stipple.non_strict_lines'..
Using line rasterization mode 0.
Using line rasterization mode 0.
Using line rasterization mode 0.
  Pass (Pass)

Test case 'dEQP-VK.rasterization.primitives_multisample_4_bit.no_stipple.bresenham_lines'..
Using line rasterization mode 2.
Using line rasterization mode 2.
Using line rasterization mode 2.
  Pass (Pass)
```

(The `Using line...` output is my logging during testing).

If we enable `MVK_CONFIG_USE_METAL_PRIVATE_API` in MoltenVK config, and expand out to wider lines, we will no doubt run into differences, but I didn't attempt to deal with that on this PR. That would take a bunch more research into other line settings available in Apple's private APIs, for us to make any headway with wide lines on these Vulkan extensions.

MoltenVK/MoltenVK/Commands/MVKCommandEncoderState.mm
squidbus: If we need to enable the logic this gates for other primitive types when the polygon mode is VK_POLYGON_MODE_LINE

billhollings: Ah! Good point. And good catch. Should be fixed now.

squidbus (approved):

cdavis5e (approved):
I'll look into the interaction with wide lines.

PR #2561, opened 2025-06-10
Support compiling using Xcode 26

squidbus:
* Adds `MVK_XCODE_26` for Xcode 26.
* Adds necessary code to use `MTLLanguageVersion4_0` when supported.
* Fixes a missing include in `MVKCodec.h` for using `std::unique_ptr`.
* Minor fixes to whitespace consistency around these changes.


billhollings (approved):
Thanks!

Issue #2557, opened 2025-06-04
Add support for VK_KHR_shader_float_controls2


billhollings:
Vulkan 1.4 requires (#2490) support for `VK_KHR_shader_float_controls2`.

This is the final extension still outstanding to support Vulkan 1.4.

An [early analysis](https://github.com/KhronosGroup/MoltenVK/issues/2490#issuecomment-2898919635) by @squidbus:

> Need to wire up all the new fast math stuff to be exposed through SPIRV-Cross interface. When any type of fast math is disabled in the shader we need to disable fast math entirely in MoltenVK when compiling the shader. Possibly could make use of the new Metal fast math precise/relaxed/fast modes, when supported, to use relaxed instead of precise when only certain conditions are needed.

squidbus:
I didn't touch upon it in my previous analysis but it is worth noting that Metal also has pragmas that you can put in-source to enable/disable certain fast math properties as well, which could help with implementing decorations on specific instructions. Although I'm not sure how they interact with the global compile properties, and it may be more complexity than its worth versus bubbling it up to configuring fast math globally for the shader if any disables are detected.

billhollings:
PR #2564 completes this.

PR #2564, opened 2025-06-13
Add Support for VK_KHR_shader_float_controls2 extension.

billhollings:
- Add `SPIRVEntryPoint::fpFastMathMode` and set from shader reflection query.
- Add support for `MTLCompileOptions` `mathMode` & `mathFloatingPointFunctions`,
  and match to SPIR-V `FPFastMathModeMask` flags as closely as possible.
- Set default value of configuration parameter `MVK_CONFIG_FAST_MATH_ENABLED` to `MVK_CONFIG_FAST_MATH_ON_DEMAND`.

CTS results for `--deqp-case=dEQP-VK.*float_controls2*`:

```
Test run totals:
  Passed:        4058/5933 (68.4%)
  Failed:        222/5933 (3.7%)
  Not supported: 1653/5933 (27.9%)
  Warnings:      0/5933 (0.0%)
  Waived:        0/5933 (0.0%)
```

The remaining failures are mostly differences in how MSL functions handle `INF` & `NAN`, relative to Vulkan, and will require dedicated function overrides to be defined in SPRIV-Cross. Since handling `INF` & `NAN` in a specific way are by definition marginal cases, I didn't feel it was worth the niggly effort to work through that for this PR.

Before this PR can be merged, [SPIRV-Cross 2495](https://github.com/KhronosGroup/SPIRV-Cross/pull/2495) must be merged.

Completes #2557.


squidbus:
The spec for this states:
> If an operation is decorated with FPFastMathMode then the flags from that decoration apply. Otherwise, if the current entry point sets any FPFastMathDefault execution mode then all flags specified for any operand type or for the result type of the operation apply. If the operation is not decorated with FPFastMathMode and the entry point sets no FPFastMathDefault execution modes then the flags to be applied are determined by the client API and not by SPIR-V.

I see the SPIRV-Cross side handles the default fast math mode, but what about per-operation mode decorations? Presumably we would have to either combine all operation modes to produce the minimum set of allowed fast math operations or handle it in shader code with pragmas.

I think in cases where the shader default mode is more strict than an operation it would still be fine to not handle individual operation modes. However, the problem case here I think is what if a specific operation is decorated more strictly than the overall shader default? It seems to me from the spec description that its an either-or rather than the flags being combined together.

MoltenVK/MoltenVK/GPUObjects/MVKDevice.h
```diff
@@ -765,12 +766,14 @@ class MVKDevice : public MVKDispatchableVulkanAPIObject {
 
 	/**
 	 * Returns an autoreleased options object to be used when compiling MSL shaders.
+	 * The fpFastMathDefault indicates
```
cdavis5e: Unfinished comment?

cdavis5e (changes requested):
This, of course, cannot go in until H-K accepts the change to SPIRV-Cross. Other than @squidbus's concern, looks good otherwise.

MoltenVK/MoltenVK/GPUObjects/MVKDevice.h
billhollings: Uggh. Thanks. Good catch. Fixed now.

billhollings:
> I see the SPIRV-Cross side handles the default fast math mode, but what about per-operation mode decorations? Presumably we would have to either combine all operation modes to produce the minimum set of allowed fast math operations or handle it in shader code with pragmas.
> 
> I think in cases where the shader default mode is more strict than an operation it would still be fine to not handle individual operation modes. However, the problem case here I think is what if a specific operation is decorated more strictly than the overall shader default? It seems to me from the spec description that its an either-or rather than the flags being combined together.

Because these apply at the individual operation level, but Metal applies them as `pragmas`, this would become a nightmare of unwinding expressions and inserting fake temporary variables so that we could wrap a pragma around one part of the expression. So, I'm unlikely to try that.

However, your point about an individual operation being stricter than the default is valid. I'm not sure it will be worth tracking all that in SPIRV-Cross, and then rolling it all up into the minimum value, but I'll have a look.

billhollings:
> However, your point about an individual operation being stricter than the default is valid. I'm not sure it will be worth tracking all that in SPIRV-Cross, and then rolling it all up into the minimum value, but I'll have a look.

@squidbus Your insight was helpful, and it reduced the failures by 24 tests. The changes also caused some different failures, but that seems to be because CTS is expecting inaccuracies based on math mode, whereas Metal ends up computing correctly due to the improved accuracy that results from taking into consideration the lowest-common-denominator fast-math approach.

With that, I think we're as close as we can get to passing CTS tests, while ensuring the most correct behaviour, without getting into the work and messiness of overriding MSL function calls.

squidbus (approved):
Nice, LGTM then when the SPIRV-Cross PR is in.

billhollings:
SPIRV-Cross patch is pulled in. This is good to go now.

PR #2570, opened 2025-06-25
MVKDevice: Apple, AMD, and Intel devices really do use parallelograms…

cdavis5e:
… for lines.

I know this is so because the lines are antialiased with multisampled rendering. This means these devices render wide lines with polygons. In the Bresenham rendering tests, the reference image does not have antialiasing, even when the sample count is greater than 1.

I've also learned that Intel devices now draw non-strict lines. On the Mac, this has been so since at least macOS 10.15. This is not unique to Metal: Sascha's GPUInfo database shows that only the earliest versions of Intel's Vulkan drivers draw strict lines; later drivers, around 2017 and implementing Vulkan 1.1, draw non-strict lines. I suspect the change has to do with the `VK_EXT_line_rasterization` extension; it provides rectangular (i.e. strict) lines as an option. Intel thus changed to non-strict lines to allow clients to select between parallelograms and rectangles.

Other than correcting Intel's capabilities for macOS 10.15+, this doesn't make any difference in passing the tests, which suggests they're inadequate.

### DIFF
--- a/MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
@@ -911,8 +911,8 @@ void MVKPhysicalDevice::getProperties(VkPhysicalDeviceProperties2* properties) {
 	supportedProps14.earlyFragmentSampleMaskTestBeforeSampleCounting = false;
 	supportedProps14.depthStencilSwizzleOneSupport = true;
 	supportedProps14.polygonModePointSize = true;
-	supportedProps14.nonStrictSinglePixelWideLinesUseParallelogram = false;
-	supportedProps14.nonStrictWideLinesUseParallelogram = false;
+	supportedProps14.nonStrictSinglePixelWideLinesUseParallelogram = !_properties.limits.strictLines;
+	supportedProps14.nonStrictWideLinesUseParallelogram = !_properties.limits.strictLines;
 	supportedProps14.blockTexelViewCompatibleMultipleLayers = false;
 	supportedProps14.maxCombinedImageSamplerDescriptorCount = 3;
 	supportedProps14.fragmentShadingRateClampCombinerInputs = false;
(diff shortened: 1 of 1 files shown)


MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
```diff
@@ -3373,7 +3373,7 @@
     _properties.limits.lineWidthGranularity = _features.wideLines ? 0.125f : 0;
 
     _properties.limits.standardSampleLocations = VK_TRUE;
-    _properties.limits.strictLines = _properties.vendorID == kIntelVendorId || _properties.vendorID == kNVVendorId;
+    _properties.limits.strictLines = (_properties.vendorID == kIntelVendorId && !mvkOSVersionIsAtLeast(10.15)) || _properties.vendorID == kNVVendorId;
```
squidbus: Do we actually need to even consider pre-10.15? I don’t think it’s supported for building MoltenVK anymore.

Also maybe not directly related to this PR, but do we need to consider NVIDIA anymore either? As far as I know their driver isn’t officially supported on anything since before Mojave.

vaylor27: the sdk installer doesn't even open on pre 10.15, so most people can't even use mvk on it.

cdavis5e: Now unconditionally setting it to false.

TellowKrinkle: Nvidia GPUs (e.g. GT 750M) are supported to macOS 11 officially, and further if you want to support OCLP

> the sdk installer doesn't even open on pre 10.15, so most people can't even use mvk on it.

The most common use of MoltenVK I see is it getting shipped with other applications, in which case the SDK installer will not be run on the machine that's running MoltenVK.

cdavis5e: All right, I've added the Nvidia check back.

squidbus: I looked into it a bit further and it seems NVIDIA’s drivers are not supported on Big Sur, and then the drivers included by Apple which support Kepler GPUs are dropped from Monterey on.

So NVIDIA is supported on Big Sur but only to a limited capacity. It’s probably something that we can consider outside of this PR though, along with the consideration of dropping pre-Metal 3.

squidbus (approved):

Issue #2567, opened 2025-06-15
Add support for VK_KHR_dynamic_rendering_local_read extension


billhollings:
Vulkan 1.4 (see #2490) promotes part of extension `VK_KHR_dynamic_rendering_local_read`. 

Full support for `VK_KHR_dynamic_rendering_local_read` is preferred, but support for at least local read for storage resources and single sampled color attachments, is required for Vulkan 1.4.

I'm working on this.

billhollings:
PR #2579 completes this issue.

PR #2579, opened 2025-07-20
Add support for extension VK_KHR_dynamic_rendering_local_read

billhollings:
Trigger Metal renderpass restarts when necessary, as determined by subpass attachment location and input attachment changes.

- Add support for `vkCmdSetRenderingAttachmentLocations()` and `vkCmdSetRenderingInputAttachmentIndices()`.
- `MVKCommandEncoder` if color attachment locations or input attachments change, restart Metal renderpass.
- `MVKCommandBuffer` support secondary inheritance of attachment location and input attachment changes.
- `MVKRenderSubpass` support updating attachment locations and input attachments, and track if dynamic rendering is in use.
- `MVKCmdClearAttachments` accommodate attachment location changes that can occur before encoding. Defer attachment assignments and derivation of clear attachment pipeline until encoding.
- `MVKCmdPipelineBarrier` trigger Metal renderpass restart when dependency region and input attachment is marked as changing.
- `MVKFramebuffer` actually track attachment image views for use when attachment locations change.
- `MVKGraphicsPipeline` track color attachment locations and update attachment locations in MVKCommandEncoder when pipeline is bound.
- Focus `MVKPixelFormats` clear value functions use appropriate clear values instead of being passed parent structure.
- Set Metal renderpass extent correctly when no attachments are used.

Fix misalignment between pipeline pixel formats and dynamic rendering attachments.

- `MVKRenderPass` track all attachments, even those without an image view.
- `MVKRenderingAttachmentIterator` mark attachments with no image view as `VK_ATTACHMENT_UNUSED`.
- `MVKPipeline` set missing Metal stencil format from depth format, and vice versa, when combined depth/stencil format is used.

This was a messy, iterative one, getting the dynamic attachment mappings working correctly in all testing scenarios.

Completes issue #2567.


squidbus (approved):
Checked on my system and all supported tests in `dEQP-VK.renderpass.*` and `dEQP-VK.dynamic_rendering.*` are passing except:
```
dEQP-VK.dynamic_rendering.primary_cmd_buff.basic.partial_binding_depth_stencil
dEQP-VK.renderpass.suballocation.subpass_dependencies.late_fragment_tests.*
```

...which are failing on current main build as well so I assume they're unrelated problems.

Docs/MoltenVK_Runtime_UserGuide.md
```diff
@@ -356,6 +368,7 @@ In addition to core *Vulkan* functionality, **MoltenVK**  also supports the foll
 - `VK_EXT_private_data `
 - `VK_EXT_robustness2`
 - `VK_EXT_sample_locations`
+- `VK_EXT_sampler_filter_minmax`
```
cdavis5e: I don't recall _that_ extension being implemented... Last I checked, it would require the use of undocumented/private Metal SPIs. I noticed that, for whatever reason, there's an entry for it in `MVKExtensions.def`, but it's listed as not supported on any version. It was promoted in Vulkan 1.2, but IIRC as optional functionality.

MoltenVK/MoltenVK/Commands/MVKCommandBuffer.mm
```diff
@@ -426,14 +481,16 @@
 								  ? VK_SUBPASS_CONTENTS_SECONDARY_COMMAND_BUFFERS
 								  : VK_SUBPASS_CONTENTS_INLINE);
 
+	// Track both rendering and resolve attachments, alternating a rendering attachment, then it's corresponding resolve attachment.
```
cdavis5e: "then _it is_ corresponding resolve attachment"??
```suggestion
	// Track both rendering and resolve attachments, alternating a rendering attachment, then its corresponding resolve attachment.
```

MoltenVK/MoltenVK/Commands/MVKCommandBuffer.mm
```diff
@@ -426,14 +481,16 @@
...
 	uint32_t maxAttCnt = (pRenderingInfo->colorAttachmentCount + 2) * 2;
 	MVKImageView* imageViews[maxAttCnt];
 	VkClearValue clearValues[maxAttCnt];
 
+	// Assemble the list of attachments. This must be done identically to how the dynamic renderpass assembles attachments.
```
cdavis5e: That's fragile. There must be a way to keep SPOT without sacrificing performance or turning the code into an eldritch horror... well, any more than it already is.

There should at least be a similar comment in the dynamic renderpass code so future contributors will remember to update both.

MoltenVK/MoltenVK/Commands/MVKCommandBuffer.mm
```diff
@@ -887,6 +953,33 @@ static MVKBarrierStage commandUseToBarrierStage(MVKCommandUse use) {
 	return mvkMTLScissorRectFromVkRect2D(clipToRenderArea(mvkVkRect2DFromMTLScissorRect(scissor)));
 }
 
+// Attachment locations are changing. Neeed to store attachments and beginning another Metal renderpass.
+void MVKCommandEncoder::updateColorAttachmentLocations(const MVKArrayRef<uint32_t> colorAttLocs) {
+	if (_mtlRenderEncoder && isDynamicRendering()) {
+		auto* mvkSP = getSubpass();
```
cdavis5e: Could this ever be `nullptr` at this point? The change you made above implies it could be...

MoltenVK/MoltenVK/Commands/MVKCommandBuffer.mm
```diff
@@ -887,6 +953,33 @@ static MVKBarrierStage commandUseToBarrierStage(MVKCommandUse use) {
...
+		auto atts = _pEncodingContext->getFramebuffer()->getAttachments();
+		if (mvkSP->isChangingColorAttachmentLocations(colorAttLocs, atts)) {
+			encodeStoreActions(true);
+			endMetalRenderEncoding();
+			mvkSP->updateColorAttachmentLocations(colorAttLocs, atts);
+		}
+	}
+}
+
+// Input attachments are changing. Neeed to store attachments and beginning another Metal renderpass.
```
cdavis5e:
```suggestion
// Input attachments are changing. Need to store attachments and begin another Metal renderpass.
```

MoltenVK/MoltenVK/Commands/MVKCommandBuffer.mm
```diff
@@ -887,6 +953,33 @@ static MVKBarrierStage commandUseToBarrierStage(MVKCommandUse use) {
 	return mvkMTLScissorRectFromVkRect2D(clipToRenderArea(mvkVkRect2DFromMTLScissorRect(scissor)));
 }
 
+// Attachment locations are changing. Neeed to store attachments and beginning another Metal renderpass.
```
cdavis5e:
```suggestion
// Attachment locations are changing. Need to store attachments and begin another Metal renderpass.
```

billhollings:
> Checked on my system and all supported tests in `dEQP-VK.renderpass.*` and `dEQP-VK.dynamic_rendering.*` are passing except:
> 
> ```
> dEQP-VK.dynamic_rendering.primary_cmd_buff.basic.partial_binding_depth_stencil
> dEQP-VK.renderpass.suballocation.subpass_dependencies.late_fragment_tests.*
> ```
> 
> ...which are failing on current main build as well so I assume they're unrelated problems.

Yes. Sorry. I should have mentioned in the OG post above that I tested against:

```
dEQP-VK.dynamic_renderpass.*
dEQP-VK.dynamic_renderpass2.*
dEQP-VK.dynamic_rendering.*
```

Same results as yours. Failing tests are the same as before this PR.

Once 1.4 support is released, that will be a good time to start carving into the outstanding CTS failures to try to reduce them. It would be great to work towards as close to Vulkan 1.4 conformance as possible.

MoltenVK/MoltenVK/GPUObjects/MVKRenderPass.mm
```diff
@@ -1031,28 +1148,34 @@
...
-	}
+	linkAttachments();
 }
 
 MVKRenderPass::MVKRenderPass(MVKDevice* device, const VkRenderingInfo* pRenderingInfo) : MVKVulkanAPIDeviceObject(device) {
 
 	_renderingFlags = pRenderingInfo->flags;
 
 	// Add attachments first so subpasses can access them during creation
+	// Attachments alternate between color and resolve, even if resolved are unused, then with depth, depth resolve, stencil, and stencil resolve at the end.
```
cdavis5e, line 1159: This is probably where the comment I mentioned above referring to `MVKFramebuffer` and `MVKCommandEncoder` should go.

cdavis5e (approved):
Looks good; just a few minor things.

Docs/MoltenVK_Runtime_UserGuide.md
billhollings: Ah. Good catch. Thanks.

I don't know why I didn't notice it was fully deselected. And yes, the features are disabled in the Vulkan 1.2 struct.

MoltenVK/MoltenVK/Commands/MVKCommandBuffer.mm
billhollings: Thanks.

MoltenVK/MoltenVK/Commands/MVKCommandBuffer.mm
billhollings: Technically, the `isDynamicRendering()` takes care of handling that. But in the interests of robustness, in case that ever changes, I've added a check for it, since it's trivial to do so here.

MoltenVK/MoltenVK/Commands/MVKCommandBuffer.mm
billhollings: Thanks.

MoltenVK/MoltenVK/Commands/MVKCommandBuffer.mm
billhollings: Thanks.

MoltenVK/MoltenVK/GPUObjects/MVKRenderPass.mm
billhollings: Thanks. Fixed.

MoltenVK/MoltenVK/Commands/MVKCommandBuffer.mm
billhollings: Yeah. I agree. 

`MVKRenderingAttachmentIterator` takes us most of the way there. Although perhaps it is designed a little too flexible to force the same layouts without manual definition. A redesign of it might make it more a SSOT.

In the meantime, I've added the extra comments.

etang-cw:
Is there a reason remapping is handled by restarting render passes instead of by editing shaders?  The required data is all there at pipeline creation time, so it seems like it would be good to not have to restart render passes for this.

cdavis5e:
> Is there a reason remapping is handled by restarting render passes instead of by editing shaders? The required data is all there at pipeline creation time, so it seems like it would be good to not have to restart render passes for this.

I don't know, but if I had to guess, it's because there's infrastructure already for restarting render passes, which makes it easier to do.

billhollings:
> > Is there a reason remapping is handled by restarting render passes instead of by editing shaders? The required data is all there at pipeline creation time, so it seems like it would be good to not have to restart render passes for this.
> 
> I don't know, but if I had to guess, it's because there's infrastructure already for restarting render passes, which makes it easier to do.

Hmmm. This is an interesting idea.

However, my interpretation of the spec is that `vkCmdSetRenderingAttachmentLocations()` can be invoked without a change to the bound pipeline. Meaning, the app can bind a single pipeline that uses one or more `color(m)` declarations, and then the app can use calls to `vkCmdSetRenderingAttachmentLocations()` to flip the attachment locations around willy-nilly so that what the pipeline treats as a `color(m)` gets mapped to different physical attachments on consecutive draw calls.

If this is the case, if we don't use Metal renderpasses to make this remapping happen, and want to use pipelines instead, I think we'd have to create and cache alternate pipelines that differ in what `m` values they use for their `color(m)` declarations.

etang-cw:
> However, my interpretation of the spec is that vkCmdSetRenderingAttachmentLocations() can be invoked without a change to the bound pipeline.

[This command sets the attachment location mappings for subsequent drawing commands, and must match the mappings provided to the bound pipeline, if one is bound](https://registry.khronos.org/vulkan/specs/latest/man/html/vkCmdSetRenderingAttachmentLocations.html)

Funnily enough, I tried removing the pipeline code altogether and relying solely on vkCmdSetRenderingAttachmentLocations, but it turns out [you can let them disagree as long as you don't call vkCmdSetRenderingAttachmentLocations, in which case the pipeline's mapping wins](https://github.com/KhronosGroup/Vulkan-Docs/issues/2565).  (So the pipeline mapping is always correct, but the one set by vkCmdSetRenderingAttachmentLocations may be wrong.  This also means that we can completely ignore vkCmdSetRenderingAttachmentLocations as long as we don't implement EXT_shader_object.)

cdavis5e:
> However, my interpretation of the spec is that `vkCmdSetRenderingAttachmentLocations()` can be invoked without a change to the bound pipeline. Meaning, the app can bind a single pipeline that uses one or more `color(m)` declarations, and then the app can use calls to `vkCmdSetRenderingAttachmentLocations()` to flip the attachment locations around willy-nilly so that what the pipeline treats as a `color(m)` gets mapped to different physical attachments on consecutive draw calls.

Metal 4 has support for remapping color render targets like this. We really need to get that implemented.

etang-cw:
> Metal 4 has support for remapping color render targets like this. We really need to get that implemented.

If it wasn't clear, I don't think it's the case that `vkCmdSetRenderingAttachmentLocations` can be invoked without a change to the bound pipeline.

squidbus:
> If it wasn't clear, I don't think it's the case that `vkCmdSetRenderingAttachmentLocations` can be invoked without a change to the bound pipeline.

I'm not sure I understand what the point of `vkCmdSetRenderingAttachmentLocations` is then, if it the locations must equal what are already in the bound pipeline anyway?

etang-cw:
It's needed for EXT_shader_object, since that doesn't use pipelines.

etang-cw:
(Side note, if they *could* disagree, this code would likely be incorrect, since both `vkCmdSetRenderingAttachmentLocations` and binding a pipeline would fight over who gets to supply the mapping on a last-one-wins basis, which is at least not how any other Vulkan state binding API works.)

(As in `vkCmdSetRenderingAttachmentLocations` followed by binding a pipeline would result in the pipeline's mapping, while binding a pipeline followed by `vkCmdSetRenderingAttachmentLocations` would result in `vkCmdSetRenderingAttachmentLocations`'s mapping.)

PR #2582, opened 2025-07-22
Add support for Vulkan 1.4.

billhollings:
- Advertise Vulkan 1.4 support.
- Update ICD API version to 1.4.
- Update CTS conformance version to 1.4.
- Update platform limits to Vulkan 1.4 levels.
- Update documentation.

Completes issue #2490.


cdavis5e (approved):
I think this is the fastest turnaround we've had on supporting Vulkan API levels.

billhollings:
> I think this is the fastest turnaround we've had on supporting Vulkan API levels.

It definitely is! Just one MoltenVK release iteration between Vulkan 1.3 and 1.4!

Most of the credit goes to @squidbus, for their very impressive (and manic in a good way) development of most of the outstanding extensions in such a short amount of time!

PR #2583, opened 2025-07-22
Add support for VK_KHR_present_id2 and VK_KHR_present_wait2

squidbus:
Adds support for both `VK_KHR_present_id2` and `VK_KHR_present_wait2`. These are basically the same as their original counterparts, except they advertise per-surface capabilities for each and the parameters to `vkWaitForPresent2KHR` are contained in a structure. They also share the same space for present IDs, so for example if you specify present IDs with `VK_KHR_present_id2` they will work with `VK_KHR_present_wait`, and vice-versa.


billhollings:
I guess this must run against a very new version of CTS. Did you get a chance to run CTS on this?

squidbus:
I updated to the latest CTS commit and did not see any tests for it, except the basic ones to check the feature struct (which do pass).

billhollings:
> I updated to the latest CTS commit and did not see any tests for it, except the basic ones to check the feature struct (which do pass).

Ah. Interesting. I guess the CTS tests must still be internal to Khronos then, and haven't made it to the public repos yet.

PR #2584, opened 2025-07-23
MVKDevice: Also apply zombie counter set workaround to AMD RDNA devices.

cdavis5e:
Luckily, I know this one was actually fixed; this doesn't happen after 10.15.

While I'm at it, put all the known RDNA device IDs into a single function, and call it from both the subgroup size code and the zombie counter set hack. This should also correct the minimum subgroup size for those GPUs.

### DIFF
--- a/MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
@@ -2820,19 +2820,7 @@ void MVKPhysicalDevice::initMetalFeatures() {
                 _metalFeatures.minSubgroupSize = 8;
                 break;
             case kAMDVendorId:
-                switch (_properties.deviceID) {
-                    case kAMDRadeonRX5700DeviceId:
-                    case kAMDRadeonRX5500DeviceId:
-                    case kAMDRadeonPROW6800XDeviceId:
-                    case kAMDRadeonRX6800DeviceId:
-                    case kAMDRadeonRX6700DeviceId:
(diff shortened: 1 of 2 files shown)


MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
```diff
@@ -3574,6 +3562,48 @@ static uint32_t mvkGetEntryProperty(io_registry_entry_t entry, CFStringRef prope
 	}
 }
 
+// Sourced from https://admin.pci-ids.ucw.cz/read/PC/1002
+bool MVKPhysicalDevice::isAMDRDNAGPU() {
+	assert(_properties.vendorID == kAMDVendorId);
+	switch (_properties.deviceID) {
+		// RDNA 1
+		case 0x7310:	// Radeon PRO W5700X
```
squidbus, line 3570: Do you think it would be better to just do a check if `deviceID >= 0x7310`? It seems from this code and the reference it would hold true that all AMD GPUs RDNA and newer have at least that ID, and we don't have to worry about that changing since Apple isn't using AMD GPUs anymore.

cdavis5e: Would it matter? As you said, no new Macs use AMD GPUs.

squidbus: Fair enough

squidbus (approved):

billhollings (approved):
Thanks for submitting this.

LGTM.

Issue #2592, opened 2025-08-12
-[MTLDebugResidencySet validateResource:]:108: failed assertion `residency sets do not support memoryless resources.'


stripe2933:
The error is occurred when I enabled the Metal Argument Buffer with Metal Debug Layer (via launching the application with `MTL_DEBUG_LAYER=1` environment variable).

Here's the stacktrace captured by Xcode:

<img width="547" height="627" alt="Image" src="https://github.com/user-attachments/assets/968183e6-e335-49a2-9764-ad08cc33ca71" />

<img width="547" height="619" alt="Image" src="https://github.com/user-attachments/assets/5cde0132-b1c3-425c-a937-abec92c6eab4" />

No validation layer complains and the application runs in the other platform well.

cdavis5e:
As I recall, MoltenVK puts all resources in a single residency set, to avoid the need to juggle multiple residency sets. I guess this doesn't make sense for memoryless resources, because they have no memory to manage in the first place. Clearly, then, we shouldn't be putting images with the `VK_IMAGE_USAGE_TRANSIENT_ATTACHMENT_BIT` flag set into any residency set.

squidbus:
Should be fixed by https://github.com/KhronosGroup/MoltenVK/pull/2593

PR #2593, opened 2025-08-15
Fix attempting to add memoryless images to residency sets.

squidbus:
* Do not attempt to make memoryless images resident.
* Switch two cases in render pass code to directly check the storage mode from `MVKImage` instead of needing to create the `MTLTexture`.

Should fix https://github.com/KhronosGroup/MoltenVK/issues/2592


billhollings (approved):
Thanks! LGTM.

PR #2598, opened 2025-08-30
MVKInstance: Make vkGetPhysicalDeviceToolProperties() instance again.

cdavis5e:
The Vulkan spec [states][1] that `vkGetInstanceProcAddr()` returns all dispatchable functions of a `VkInstance` or its direct children. It further [states][2] that `vkGetDeviceProcAddr()` does _not_ return any dispatchable function of a `VkPhysicalDevice`. This means that, despite that `VK_EXT_tooling_info` is a device extension, the `vkGetPhysicalDeviceToolProperties()` function is an _instance_ function. This case falls under the "available device extension dispatchable command for `instance`" case in Table 1, and footnote 3 in Table 2.

Fixes the `dEQP-VK.api.version_check.entry_points` test.

[1]: https://docs.vulkan.org/spec/latest/chapters/initialization.html#vkGetInstanceProcAddr-behavior
[2]: https://docs.vulkan.org/spec/latest/chapters/initialization.html#vkGetDeviceProcAddr-behavior

### DIFF
--- a/MoltenVK/MoltenVK/GPUObjects/MVKInstance.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKInstance.mm
@@ -428,9 +428,9 @@ void MVKInstance::initMVKConfig(const VkInstanceCreateInfo* pCreateInfo) {
 	ADD_DVC_1_2_ENTRY_POINT(func); \
 	ADD_ENTRY_POINT_MAP(func##extSuffix, func, 0, VK_##EXT##_EXTENSION_NAME, true)
 
-#define ADD_INST_1_3_PROMOTED_ENTRY_POINT(func, EXT)	\
+#define ADD_INST_1_3_PROMOTED_ENTRY_POINT(func, extSuffix, EXT)	\
 	ADD_INST_1_3_ENTRY_POINT(func);	\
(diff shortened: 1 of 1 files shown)


cdavis5e:
@billhollings There's something up with the iOS builder. It looks like there are no iOS configurations available.

squidbus:
Just to help my understanding - it still works to get the function via `vkGetInstanceProcAddr` with this change correct? As that is what the Vulkan-Hpp headers do.

cdavis5e:
> Just to help my understanding - it still works to get the function via `vkGetInstanceProcAddr` with this change correct? As that is what the Vulkan-Hpp headers do.

Yes, it should still work. It actually worked before this change anyway, because `vkGetInstanceProcAddr()` is allowed to return device-dispatchable entry points as well. `vkGetDeviceProcAddr()` is an optimization that short-circuits this dispatch directly to the ICD.

cdavis5e:
> @billhollings There's something up with the iOS builder. It looks like there are no iOS configurations available.

It's only failing sporadically, and it's happening specifically because the simulator for iOS 18 for some reason isn't installed on the builder image--but only sometimes.

billhollings:
> @billhollings There's something up with the iOS builder. It looks like there are no iOS configurations available.

Yeah. Maybe a misconfigured runner update at GitHub? Looks like it's working now at least. We can see if that sticks before looking into how to fix it from our end.

PR #2601, opened 2025-09-01
MVKDescriptorPool: Return `VK_ERROR_OUT_OF_POOL_MEMORY` when running …

cdavis5e:
…out of sets.

As required by `VK_KHR_maintenance1`.

Also keep track of available argument buffer space, and return `VK_ERROR_OUT_OF_POOL_MEMORY` if there is not enough total space left to accommodate the encoded argument buffer for a descriptor set.

Fixes the `dEQP-VK.api.descriptor_pool.out_of_pool_memory` test.


squidbus (approved):

billhollings (approved):
LGTM. Thanks.

PR #2602, opened 2025-09-01
MVKPhysicalDevice: Fix fencepost error.

cdavis5e:
If the client requests exactly as many variable descriptors as would fit, the check would return "not supported." The CTS explicitly tests this case.

Fixes 10 tests under
`dEQP-VK.api.maintenance3_check.support_count_*_extra_bindings`.


billhollings (approved):
LGTM. Thanks!

PR #2572, opened 2025-06-26
New state tracker and descriptor set implementation

etang-cw:
CTS run, no regressions.
<details>
<summary>Newly passing tests</summary>

```sh
dEQP-VK.api.descriptor_pool.out_of_pool_memory
dEQP-VK.multiview{.dynamic_rendering,.renderpass2,}.view_mask_iteration.*
dEQP-VK.pipeline.monolithic.extended_dynamic_state.before_good_static.stencil_state_extra_pipeline_face_*
```
</details>

Sorry for the massive PR.  I could try to split it between the state tracker and descriptor set changes, but due to the design differences between the new state tracker and the old descriptor set implementation, the code sticking them together is very temporary and probably pretty slow.  (Also push descriptors aren't supported at that point, though they could be made to work, once again very slowly.)  I could also move the initial commits that just add helper classes and methods to MVKFoundation to a separate PR if that's preferred.

### Highlights

- Much faster
  - O(1) descriptor pool reset for pools with freeing disabled
  - Descriptor set copy is memcpy, no refcounting
  - Allocating descriptor sets will never allocate MTLBuffers / other memory
- More resilient to desyncing between our view of the Metal command encoder and what's actually there
- Should be much easier to add alternate descriptor layouts if we e.g. wanted to support supplying external Metal shaders using the descriptor layout used by Apple's DXIL shader converter

### Implementation Details

[...]

PR #2604, opened 2025-09-02
Occlusion Query Improvements

etang-cw:
This is based off #2572 so leaving it a draft until that gets merged

- Fixes occlusion queries across multiple render encoders (fixes [these tests](https://github.com/KhronosGroup/VK-GL-CTS/pull/539))
- Accumulates query results in parallel, so games that do a large number of queries don't end up with >1ms result accumulation times
- Uses untracked buffers for query results, avoiding serialization of all render encoders when queries are enabled.


V380-Ori:
I just want to showcase what @etang-cw fixed 😄

Ryujinx (Nintendo Switch emulator) had this issue since #1842.

https://github.com/user-attachments/assets/578f03c8-b6ec-47ee-9a3f-812e8b57cb2a


https://github.com/user-attachments/assets/6ee9a183-131e-4c37-827a-996eb6801c0e


**Thanks to all contributors who make this project better everyday!**

billhollings:
Is this PR at a point to be pulled in, once #2572 is pulled in? Or does it need to stay in Draft mode beyond that?

etang-cw:
> Is this PR at a point to be pulled in, once https://github.com/KhronosGroup/MoltenVK/pull/2572 is pulled in? Or does it need to stay in Draft mode beyond that?

Yes, it's done other than the fact that it's using code from #2572

billhollings:
@etang-cw  Now that #2572 is finally in, can you clean up the conflicts here, and move it out of Draft, please? Then we can pull this in too.

MoltenVK/MoltenVK/Commands/MVKCommandEncoderState.h
```diff
@@ -523,36 +523,56 @@ class MVKCommandEncoderState {
...
 private:
-	void beginOcclusionQuery(MVKCommandEncoder* cmdEncoder, MVKOcclusionQueryPool* pQueryPool, uint32_t query, MTLVisibilityResultMode mode);
+	/** Advance to the next index in the Metal visibility buffer */
+	void nextMetalQuery(MVKCommandEncoder* cmdEncoder);
 
 	typedef struct OcclusionQueryLocation {
 		MVKOcclusionQueryPool* queryPool = nullptr;
 		uint32_t query = 0;
-		NSUInteger visibilityBufferOffset = 0;
+		uint32_t visibilityBufferOffset = 0;
```
billhollings, line 545: Is there a reason for this downgrade? Anytime address offsets are involved, it's a good time to use 64 bits!

MoltenVK/MoltenVK/Commands/MVKCommandPipelineStateFactoryShaderSource.h
```diff
@@ -23,536 +23,549 @@
...
-    uint32_t depth;                                                                                             \n\
-} __attribute__((packed)) VkExtent3D;                                                                           \n\
-                                                                                                                \n\
-typedef struct {                                                                                                \n\
-    int32_t x;                                                                                                  \n\
-    int32_t y;                                                                                                  \n\
-    int32_t z;                                                                                                  \n\
-} __attribute__((packed)) VkOffset3D;                                                                           \n\
-                                                                                                                \n"
+static NSString* _MVKStaticCmdShaderSource = @R"(
```
billhollings, line 26: Nice catch! Much easier now! 🚀

MoltenVK/MoltenVK/Utility/MVKEnvironment.h
```diff
@@ -158,8 +158,8 @@ void mvkSetConfig(MVKConfiguration& dstMVKConfig, const MVKConfiguration& srcMVK
 #   define MVK_CONFIG_MAX_ACTIVE_METAL_COMMAND_BUFFERS_PER_QUEUE    64
 #endif
 
-/** Support more than 8192 or 32768 occlusion queries per device. Enabled by default. */
-#ifndef MVK_CONFIG_SUPPORT_LARGE_QUERY_POOLS
+/** Support more than 8192 or 32768 occlusion queries per device. Always enabled. */
+#ifndef MVK_CONFIG_SUPPORT_LARGE_QUERY_POOLS       // Deprecated
 #   define MVK_CONFIG_SUPPORT_LARGE_QUERY_POOLS    1
 #endif
```
billhollings, line 164: Nice. 

Can you also remove the `MVK_CONFIG_SUPPORT_LARGE_QUERY_POOLS` entry in `MoltenVK_Configuration_Parameters.md`, please?

billhollings (changes requested):
Thanks for submitting this!

LGTM. I just have one request, and a question.

MoltenVK/MoltenVK/Commands/MVKCommandPipelineStateFactoryShaderSource.h
cdavis5e: YOU CAN DO THAT??!

I did not know that would work.

MoltenVK/MoltenVK/Commands/MVKCommandEncoderState.h
cdavis5e: There really is no reason not to, given that all currently supported Macs have CPUs with 64-bit registers.

MoltenVK/MoltenVK/Commands/MVKCommandEncoderState.h
```diff
@@ -523,36 +523,56 @@ class MVKCommandEncoderState {
...
+	/** Ends an occlusion query. */
+	void endOcclusionQuery(MVKCommandEncoder* cmdEncoder, MVKOcclusionQueryPool* pQueryPool, uint32_t query);
 
 	void encode(id<MTLRenderCommandEncoder> encoder, MVKCommandEncoder* mvkEncoder);
 
 	void prepareHelperDraw(id<MTLRenderCommandEncoder> encoder, MVKCommandEncoder* mvkEncoder);
 
 private:
-	void beginOcclusionQuery(MVKCommandEncoder* cmdEncoder, MVKOcclusionQueryPool* pQueryPool, uint32_t query, MTLVisibilityResultMode mode);
+	/** Advance to the next index in the Metal visibility buffer */
```
cdavis5e:
```suggestion
	/** Advances to the next index in the Metal visibility buffer. */
```

MoltenVK/MoltenVK/Commands/MVKCommandEncoderState.h
```diff
@@ -523,36 +523,56 @@ class MVKCommandEncoderState {
...
+	MVKOcclusionQueryPool* _currentPool = nullptr;
+	/// The index of the current active query
+	uint32_t _currentQueryIndex = 0;
+	/// The visibility result mode of the current active query
+	uint8_t _currentVisibilityResultMode = MTLVisibilityResultModeDisabled;
+	/// The visibility result mode of the current Metal render pass
+	uint8_t _metalVisibilityResultMode = MTLVisibilityResultModeDisabled;
+	/// The number of fences that need to be waited on by the result copy shader
+	uint8_t _numCopyFences = 0;
+	/// If true, accumulation will be run at the end of the next render pass
```
cdavis5e:
```suggestion
	/// The pool of the current active query.
	MVKOcclusionQueryPool* _currentPool = nullptr;
	/// The index of the current active query.
	uint32_t _currentQueryIndex = 0;
	/// The visibility result mode of the current active query.
	uint8_t _currentVisibilityResultMode = MTLVisibilityResultModeDisabled;
	/// The visibility result mode of the current Metal render pass.
	uint8_t _metalVisibilityResultMode = MTLVisibilityResultModeDisabled;
	/// The number of fences that need to be waited on by the result copy shader.
	uint8_t _numCopyFences = 0;
	/// If true, accumulation will be run at the end of the next render pass.
```

MoltenVK/MoltenVK/Commands/MVKCommandEncoderState.h
cdavis5e: Well, I guess, saving memory might be one good reason--particularly since the `visibilityBufferOffset` required padding before. With this change, the struct fits into 16 bytes--without, it's 24.

MoltenVK/MoltenVK/Commands/MVKCommandEncoderState.mm
```diff
@@ -1802,31 +1802,87 @@ static void invalidateImplicitBuffer(MVKCommandEncoderState& state, VkPipelineBi
 #pragma mark -
 #pragma mark MVKOcclusionQueryCommandEncoderState
 
-void MVKOcclusionQueryCommandEncoderState::beginMetalRenderPass(MVKCommandEncoder* cmdEncoder) {
-	if (_mtlVisibilityResultMode != MTLVisibilityResultModeDisabled)
-		_dirty = true;
-}
+/// Used to indicate that the current metal visibility result mode needs to be set regardless of what mode is wanted
+/// (When visibility queries are currently active but on the wrong index)
```
cdavis5e:
```suggestion
/// Used to indicate that the current metal visibility result mode needs to be set regardless of what mode is wanted
/// (when visibility queries are currently active but on the wrong index).
```

MoltenVK/MoltenVK/Commands/MVKCommandPipelineStateFactoryShaderSource.h
MrGcGamer: didn't know that either.. looks like it isn't a objc specific thing, but just being able to convert a cpp raw string into an NSString.. didn't find any apple docs about that, but sounds about right

MoltenVK/MoltenVK/Commands/MVKCommandEncoderState.mm
```diff
@@ -1802,31 +1802,87 @@ static void invalidateImplicitBuffer(MVKCommandEncoderState& state, VkPipelineBi
...
+				uint32_t rest = 0;
+				if (count >= executionWidth) {
+					uint32_t first = count / executionWidth;
+					rest = first * executionWidth;
+					count -= rest;
+					state.bindBytes(mtlAccumEncoder, offsets, rest * sizeof(QueryResultOffsets), 0);
+					[mtlAccumEncoder dispatchThreadgroups:MTLSizeMake(first, 1, 1) threadsPerThreadgroup:tgsize];
+				}
+				if (count) {
+					state.bindBytes(mtlAccumEncoder, offsets + rest, count * sizeof(QueryResultOffsets), 0);
```
cdavis5e, line 1879: Why did you name the variable `rest` when it's the number of queries processed in pass 1? I suppose it could be taken as "add this much to get to the rest," but that's not quite obvious looking at the code--it took a while to realize what it really was.

MoltenVK/MoltenVK/Commands/MVKCommandEncoderState.mm
```diff
@@ -1835,52 +1891,83 @@ static void invalidateImplicitBuffer(MVKCommandEncoderState& state, VkPipelineBi
...
-	_dirty = true;
-}
-
 void MVKOcclusionQueryCommandEncoderState::beginOcclusionQuery(MVKCommandEncoder* cmdEncoder, MVKOcclusionQueryPool* pQueryPool, uint32_t query, VkQueryControlFlags flags) {
-	bool shouldCount = cmdEncoder->getEnabledFeatures().occlusionQueryPrecise && mvkAreAllFlagsEnabled(flags, VK_QUERY_CONTROL_PRECISE_BIT);
-	beginOcclusionQuery(cmdEncoder, pQueryPool, query, shouldCount ? MTLVisibilityResultModeCounting : MTLVisibilityResultModeBoolean);
+	if (_currentPool) [[unlikely]] {
+		assert(0 && "Shouldn't have active query when beginning a new one!");
+		nextMetalQuery(cmdEncoder);
+	}
```
cdavis5e: When I worked on LLVM, I learned that `if (cond) assert(0);` and `assert(cond); if (!cond) doStuff();` are antipatterns.
```suggestion
	assert(!_currentPool && "Shouldn't have active query when beginning a new one!");
```
Since we asserted that the `_currentPool` is null, the `if` body (theoretically) can't be entered--and with assertions enabled, the compiler would probably optimize it away.

MoltenVK/MoltenVK/Commands/MVKCommandEncoderState.mm
```diff
@@ -1835,52 +1891,83 @@ static void invalidateImplicitBuffer(MVKCommandEncoderState& state, VkPipelineBi
...
 
 void MVKOcclusionQueryCommandEncoderState::endOcclusionQuery(MVKCommandEncoder* cmdEncoder, MVKOcclusionQueryPool* pQueryPool, uint32_t query) {
-	_mtlVisibilityResultMode = MTLVisibilityResultModeDisabled;
-	cmdEncoder->_pEncodingContext->mtlVisibilityResultOffset += kMVKQuerySlotSizeInBytes;
-	_hasRasterized = true;	// Handle begin and end query with no rasterizing before end of renderpass.
-	_dirty = true;
+	if (_currentPool != pQueryPool || _currentQueryIndex != query) [[unlikely]] {
+		assert(0 && "Ended query that wasn't active!");
+		return;
+	}
```
cdavis5e:
```suggestion
	assert(_currentPool == pQueryPool && _currentQueryIndex == query && "Ended query that wasn't active!");
```

MoltenVK/MoltenVK/GPUObjects/MVKDevice.h
```diff
@@ -597,6 +597,54 @@ struct MVKLiveResourceSet {
...
+	MVKVisibilityBuffer(id<MTLDevice> device, NSUInteger size, uint32_t name_idx);
+	MVKVisibilityBuffer(const MVKVisibilityBuffer&) = delete;
+	MVKVisibilityBuffer(MVKVisibilityBuffer&& other) {
+		move(buffer_, other.buffer_);
+		current_offset = other.current_offset;
+		half_size = other.half_size;
+		for (uint32_t i = 0; i < std::size(fences); i++)
+			for (uint32_t j = 0; j < std::size(fences[0]); j++)
+				move(fences[i][j], other.fences[i][j]);
+	}
```
cdavis5e:
```suggestion
	id<MTLBuffer> _buffer = nullptr;
	uint32_t _currentOffset = 0;
	uint32_t _halfSize;
	id<MTLFence> _fences[3][2] = {};
public:
	id<MTLBuffer> buffer() const { return _buffer; }
	uint32_t offset() const { return _currentOffset; }
	uint32_t size() const { return _halfSize * 2; }
	bool isFirstWithCurrentFence() const { return _currentOffset == _halfSize || _currentOffset == 0; }
	FenceSet currentFence() const {
		uint32_t idx = _currentOffset > _halfSize ? 1 : 0;
		return { _fences[0][idx], _fences[1][idx], _fences[2][idx] };
	}
	MVKVisibilityBuffer() = default;
	MVKVisibilityBuffer(id<MTLDevice> device, NSUInteger size, uint32_t nameIdx);
	MVKVisibilityBuffer(const MVKVisibilityBuffer&) = delete;
	MVKVisibilityBuffer(MVKVisibilityBuffer&& other) {
		_buffer = std::exchange(other._buffer, nil);
		_currentOffset = other._currentOffset;
		_halfSize = other._halfSize;
		for (uint32_t i = 0; i < std::size(_fences); i++)
			for (uint32_t j = 0; j < std::size(_fences[0]); j++)
				_fences[i][j] = std::exchange(other._fences[i][j], nil);
	}
```
I'm kind of amazed you didn't know about `std::exchange()`.

MoltenVK/MoltenVK/GPUObjects/MVKDevice.h
```diff
@@ -597,6 +597,54 @@ struct MVKLiveResourceSet {
...
+	~MVKVisibilityBuffer();
+	/// Advances `current_offset`, swaps fences if needed, and returns the new offset
```
cdavis5e:
```suggestion
	/// Advances `_currentOffset`, swaps fences if needed, and returns the new offset.
```

MoltenVK/MoltenVK/GPUObjects/MVKDevice.h
```diff
@@ -869,20 +917,11 @@ class MVKDevice : public MVKDispatchableVulkanAPIObject {
 	/** Returns the memory alignment required for the format when used in a texel buffer. */
 	VkDeviceSize getVkFormatTexelBufferAlignment(VkFormat format, MVKBaseObject* mvkObj);
 
-    /** 
-     * Returns the MTLBuffer used to hold occlusion query results, 
-     * when all query pools use the same MTLBuffer.
-     */
-    id<MTLBuffer> getGlobalVisibilityResultMTLBuffer();
+	/** Fetch a visibility buffer from the shared pool */
```
cdavis5e:
```suggestion
	/** Fetches a visibility buffer from the shared pool. */
```

MoltenVK/MoltenVK/GPUObjects/MVKDevice.h
```diff
@@ -869,20 +917,11 @@ class MVKDevice : public MVKDispatchableVulkanAPIObject {
...
 
-    /**
-     * Expands the visibility results buffer, used for occlusion queries, by replacing the
-     * existing buffer with a new MTLBuffer that is large enough to accommodate all occlusion
-     * queries created to date, including those defined in the specified query pool.
-     * Returns the previous query count, before the new queries were added, which can
-     * be used by the new query pool to locate its queries within the single large buffer.
-     */
-    uint32_t expandVisibilityResultMTLBuffer(uint32_t queryCount);
+	/** Return a visibility buffer to the shared pool */
```
cdavis5e:
```suggestion
	/** Returns a visibility buffer to the shared pool. */
```
I do prefer clarity and precision in my writing, so I'm worried that "Return" might be taken as "fetch" above. But I'm trying not to be so overbearing this time around. 😉

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
```diff
@@ -4209,6 +4209,51 @@ static uint32_t mvkGetEntryProperty(io_registry_entry_t entry, CFStringRef prope
...
+			// Rotate fences so that the new prevRead is the previous read
+			id<MTLFence> read = fences[0][lohi];
+			for (int i = 0; i < 2; i++)
+				fences[i][lohi] = fences[i + 1][lohi];
+			fences[2][lohi] = read;
+		}
+		return 0;
+	}
+	return current_offset;
+}
```
cdavis5e:
```suggestion
MVKVisibilityBuffer::MVKVisibilityBuffer(id<MTLDevice> device, NSUInteger size, uint32_t nameIdx) {
	_halfSize = static_cast<uint32_t>(size / 2);
	assert(_halfSize % kMVKQuerySlotSizeInBytes == 0);
	@autoreleasepool {
		MTLResourceOptions options = MTLResourceHazardTrackingModeUntracked | MTLResourceStorageModePrivate;
		_buffer = [device newBufferWithLength:_halfSize * 2 options:options];
		if (_buffer) {
			[_buffer setLabel:[NSString stringWithFormat:@"Metal Visibility Result Buffer %d", nameIdx]];
			for (uint32_t gid = 0; gid < std::size(_fences); gid++) {
				for (uint32_t lohi = 0; lohi < std::size(_fences[0]); lohi++) {
					_fences[gid][lohi] = [device newFence];
					[_fences[gid][lohi] setLabel:[NSString stringWithFormat:@"Visibility Fence %d %c%d", nameIdx, lohi ? 'H' : 'L', gid]];
				}
			}
		}
	}
}

MVKVisibilityBuffer::~MVKVisibilityBuffer() {
	if (_buffer) {
		[_buffer release];
		for (auto& groups : _fences)
			for (auto& fence : groups)
				[fence release];
	}
}

uint32_t MVKVisibilityBuffer::advanceOffset() {
	_currentOffset += kMVKQuerySlotSizeInBytes;
	if (_currentOffset >= _halfSize * 2) {
		_currentOffset = 0;
		for (int lohi = 0; lohi < 2; lohi++) {
			// Rotate fences so that the new prevRead is the previous read
			id<MTLFence> read = _fences[0][lohi];
			for (int i = 0; i < 2; i++)
				_fences[i][lohi] = _fences[i + 1][lohi];
			_fences[2][lohi] = read;
		}
		return 0;
	}
	return _currentOffset;
}
```

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
```diff
@@ -5259,31 +5306,21 @@ static uint32_t mvkGetEntryProperty(io_registry_entry_t entry, CFStringRef prope
 	return deviceAlignment ? deviceAlignment : _physicalDevice->_properties.limits.minTexelBufferOffsetAlignment;
 }
 
-id<MTLBuffer> MVKDevice::getGlobalVisibilityResultMTLBuffer() {
-    lock_guard<mutex> lock(_vizLock);
-    return _globalVisibilityResultMTLBuffer;
+/** Fetch a visibility buffer from the shared pool */
```
cdavis5e:
```suggestion
```
It usually isn't necessary to repeat the doc comment on the definition. 😄

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
```diff
@@ -5259,31 +5306,21 @@ static uint32_t mvkGetEntryProperty(io_registry_entry_t entry, CFStringRef prope
...
-        reportError(VK_ERROR_OUT_OF_DEVICE_MEMORY, "vkCreateQueryPool(): A maximum of %d total queries are available on this device in its current configuration. See the API notes for the MVKConfiguration.supportLargeQueryPools configuration parameter for more info.", _globalVisibilityQueryCount);
-    }
-
-    NSUInteger mtlBuffLen = mvkAlignByteCount(newBuffLen, _physicalDevice->_metalFeatures.mtlBufferAlignment);
-    MTLResourceOptions mtlBuffOpts = MTLResourceStorageModeShared | MTLResourceCPUCacheModeDefaultCache;
-    [_globalVisibilityResultMTLBuffer release];
-    _globalVisibilityResultMTLBuffer = [_physicalDevice->_mtlDevice newBufferWithLength: mtlBuffLen options: mtlBuffOpts];     // retained
-
-    return _globalVisibilityQueryCount - queryCount;     // Might be lower than requested if an overflow occurred
+/** Return a visibility buffer to the shared pool */
```
cdavis5e:
```suggestion
```

cdavis5e (approved):
LGTM. Just some more bikeshedding over style. Luckily for both of us, this was a much smaller patch.

MoltenVK/MoltenVK/Utility/MVKEnvironment.h
billhollings: I'll add this in a separate documentation-update PR.

billhollings (approved):
I've pulled in the syntax and format suggestions from @cdavis5e . And I'll update the `MVK_CONFIG_SUPPORT_LARGE_QUERY_POOLS` entry in `MoltenVK_Configuration_Parameters.md` in a subsequent documentation PR.

MoltenVK/MoltenVK/Commands/MVKCommandEncoderState.h
etang-cw: Reduces padding, and we know the max offset is 256k, since that's the entire reason we're doing all this.

MoltenVK/MoltenVK/Commands/MVKCommandEncoderState.mm
etang-cw: > When I worked on LLVM, I learned that `if (cond) assert(0);` and `assert(cond); if (!cond) doStuff();` are antipatterns.

What's the reason?  Because if you made a mistake it's better to crash or leak memory than try to clean it up?  Or because you should have designed your interface to make the compiler enforce the invariants you have?  If it's the latter, this doesn't help it.  If it's the former, can you give some reasoning as to why that's better?

MoltenVK/MoltenVK/Commands/MVKCommandPipelineStateFactoryShaderSource.h
etang-cw: I assume they used the same system in clang for `@"` strings that they use for handling `L"` and `u8"` strings, which have to support raw strings by spec.

MoltenVK/MoltenVK/Commands/MVKCommandEncoderState.mm
etang-cw: Yeah it was meant to be the starting index of the remaining unprocessed elements.  Maybe it would be easier to make `rest` a pointer, start it out with `rest = offsets`, and then do `rest += first * executionWidth; count -= first * executionWidth;`.

MoltenVK/MoltenVK/Commands/MVKCommandEncoderState.mm
cdavis5e: `assert(3)` is for checking invariants, that is, conditions that _must_ be true for the program to function correctly. If the assertion _ever_ fails, that's a bug--especially because, in a build with assertions enabled, the program will immediately abort if the assertion fails. With assertions enabled, it will crash _anyway_.

cdavis5e: Also, Vulkan is not OpenGL.

In OpenGL, implementations were required to validate every single state change. As a result, drivers became relatively bloated and slow. Vulkan is, in part, a reaction to this. The ICD is no longer required to validate that the client is using Vulkan correctly—that's the job of the validation layer. More specifically, the behavior of Vulkan when the client strays from "Valid Usage" is _undefined_—which, as the compiler writers say, means it is legal for the ICD to make demons fly out of your nose in that case.

In the spirit of that, I believe that validation in the driver should be minimized in order to maximize performance. I understand that other people feel differently.

etang-cw: > If the assertion ever fails, that's a bug.

If the assertion fails, that's a bug, but not necessarily one in our code, since we're a library used by other applications.  If assertions are enabled, you're probably working on MoltenVK, in which case crashing is fine.

> I believe that validation in the driver should be minimized in order to maximize performance.

I believe that validation in the driver should be minimized where it affects performance, like with liveness checks of descriptors.  This will not.

Issue #2412, opened 2024-12-14
Metal validation error in dynamic rendering with depth attachment and no stencil attachment, but depth-stencil format


squidbus:
In my code, I have pipelines and dynamic rendering passes where a depth attachment is provided with depth-stencil format, and the stencil attachment is set to `nullptr`/`UNDEFINED`, which as far as I can tell is perfectly valid and triggers no Vulkan validation errors:

https://github.com/shadps4-emu/shadPS4/blob/876445faf1b0ef63ddb9d0111e35b74bd31b4a42/src/video_core/renderer_vulkan/vk_scheduler.cpp#L49-L50

https://github.com/shadps4-emu/shadPS4/blob/876445faf1b0ef63ddb9d0111e35b74bd31b4a42/src/video_core/renderer_vulkan/vk_graphics_pipeline.cpp#L256-L257

In a local branch with some other state fixes, I've checked that these are in sync and that when we pass `UNDEFINED` as as the stencil format, we pass `nullptr` as the stencil attachment.

However, if our depth attachment happens to be a depth-stencil format, we get the following validation error from Metal:
```
-[MTLDebugRenderCommandEncoder setRenderPipelineState:]:1616: failed assertion `Set Render Pipeline State Validation
For stencil attachment, the render pipeline's pixelFormat (MTLPixelFormatInvalid) does not match the framebuffer's pixelFormat (MTLPixelFormatDepth32Float_Stencil8).
'
```

As far as I can tell this is because of this code: https://github.com/KhronosGroup/MoltenVK/blob/5ad3ee5d2f84342950c3fe93dec97719574d1932/MoltenVK/MoltenVK/GPUObjects/MVKRenderPass.mm#L1090-L1107

My depth attachment is being forced into the stencil attachment due to its format, but since I didn't ask for a stencil attachment and provided `UNDEFINED` format it is raising an error.

squidbus:
Note that this is a similar problem space to https://github.com/KhronosGroup/MoltenVK/issues/2337, which I previously reported, in that it is a problem with how MoltenVK's handling of dynamic rendering attachments does not seem to always line up with Metal expectations.

However in this case the issue is that I'm actually doing what Metal prefers (stencil = `nullptr` and stencilFormat = `UNDEFINED` match each other) and MoltenVK is forcing it to be otherwise, where as in that case it was behavior that Vulkan allows but Metal does not (Vulkan allows format in pipeline with null attachment, Metal expects them to always be matching).

[...]

PR #2607, opened 2025-09-03
Do not attempt to use dynamic depth/stencil attachment for alternate purpose.

squidbus:
Fixes:
```
dEQP-VK.pipeline.monolithic.stencil.no_stencil_att.dynamic_rendering.static_enable.d32_sfloat_s8_uint
dEQP-VK.pipeline.monolithic.stencil.no_stencil_att.dynamic_rendering.dynamic_enable.d32_sfloat_s8_uint
```

And should fix https://github.com/KhronosGroup/MoltenVK/issues/2412

The depth attachment is not supposed to automatically be used for stencil, and vice-versa. I'm not sure what exactly this logic was intended to handle originally but it is invalid behavior for Vulkan. Plus it was actually causing Metal validation errors; if you have a depth-stencil formatted image for the depth attachment, a null stencil attachment, and the corresponding depth format and `UNDEFINED` stencil format in the pipeline, you would get this:
```
-[MTLDebugRenderCommandEncoder setRenderPipelineState:]:1616: failed assertion `Set Render Pipeline State Validation
For stencil attachment, the render pipeline's pixelFormat (MTLPixelFormatInvalid) does not match the framebuffer's pixelFormat (MTLPixelFormatDepth32Float_Stencil8).
'
```

### DIFF
--- a/MoltenVK/MoltenVK/GPUObjects/MVKRenderPass.h
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKRenderPass.h
@@ -408,9 +408,6 @@ protected:
 	void handleAttachment(const VkRenderingAttachmentInfo* pAttInfo,
 						  VkImageAspectFlagBits aspect,
 						  MVKRenderingAttachmentInfoOperation attOperation);
-	const VkRenderingAttachmentInfo* getAttachmentInfo(const VkRenderingAttachmentInfo* pAtt,
-													   const VkRenderingAttachmentInfo* pAltAtt,
-													   bool isStencil);
 
 	VkRenderingInfo _renderingInfo;
 };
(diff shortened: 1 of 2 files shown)


cdavis5e (approved):
Ironically, the code (per the comments) was intended to _prevent_ a Metal validation error.

As I recall, there was a period for Metal on certain devices--I think it was actually Apple GPUs on iOS--where, if either the depth or stencil attachment were a packed depth/stencil texture, then _both_ attachments had to be specified as that same texture. I think that's where that code came from, but the corresponding code to set up the render pass has been lost. I don't think it's necessary on current OS versions.

billhollings (approved):

JaclynCodes (commented):
Ok

Issue #2611, opened 2025-09-12
Failing to build on latest macOS 26 / Xcode 26 with MVK_USE_METAL_PRIVATE_API


squidbus:
Noticed this in a CI run today, I assume because GitHub updated their runners:
```
/Users/runner/work/shadPS4/shadPS4/externals/MoltenVK/MoltenVK/MoltenVK/MoltenVK/OS/MTLSamplerDescriptor+MoltenVK.m:28:1: error: illegal redeclaration of property in class extension 
'MTLSamplerDescriptor' (attribute must be 'readwrite', while its primary must be 'readonly')
   28 | @property(nonatomic, readwrite) float lodBias;
      | ^
[ 64%] Building OBJC object externals/MoltenVK/CMakeFiles/MoltenVK.dir/MoltenVK/MoltenVK/MoltenVK/OS/MTLTextureDescriptor+MoltenVK.m.o
/Applications/Xcode_26_Release_Candidate.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/System/Library/Frameworks/Metal.framework/Headers/MTLSampler.h:204:29: note: property declared here
  204 | @property (nonatomic) float lodBias API_AVAILABLE(macos(26.0), ios(26.0));
      |                             ^
1 error generated.
```

It seems `lodBias` is available now and the private API definition here is conflicting: https://developer.apple.com/documentation/metal/mtlsamplerdescriptor/lodbias

squidbus:
This is fixed by https://github.com/KhronosGroup/MoltenVK/pull/2612

PR #2612, opened 2025-09-12
Enable sampler lod bias for Xcode 26.

squidbus:
Fixes https://github.com/KhronosGroup/MoltenVK/issues/2611

Enables sampler lod bias regardless of private API availability and removes the custom definition of the descriptor field when using Xcode 26, as this is now a public API. This fixes a build failure as the definitions were conflicting.

I also adjusted the max sampler lod bias limit to reflect the new Metal docs here: https://developer.apple.com/documentation/metal/mtlsamplerdescriptor/lodbias

This will likely break building with Xcode 26 betas and require using the RC, but I don't see a way to detect the betas (and it might not even be worth supporting them).

### DIFF
--- a/MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
@@ -614,7 +614,7 @@ void MVKPhysicalDevice::getFeatures(VkPhysicalDeviceFeatures2* features) {
 				portabilityFeatures->multisampleArrayImage = _metalFeatures.multisampleArrayTextures;
 				portabilityFeatures->mutableComparisonSamplers = _metalFeatures.depthSampleCompare;
 				portabilityFeatures->pointPolygons = false;
-				portabilityFeatures->samplerMipLodBias = getMVKConfig().useMetalPrivateAPI;
+				portabilityFeatures->samplerMipLodBias = _metalFeatures.samplerMipLodBias;
 				portabilityFeatures->separateStencilMaskRef = true;
 				portabilityFeatures->shaderSampleRateInterpolationFunctions = _metalFeatures.pullModelInterpolation;
 				portabilityFeatures->tessellationIsolines = false;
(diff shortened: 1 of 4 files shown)


billhollings (approved):
LGTM. Nice work.

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
```diff
@@ -614,7 +614,7 @@
 				portabilityFeatures->multisampleArrayImage = _metalFeatures.multisampleArrayTextures;
 				portabilityFeatures->mutableComparisonSamplers = _metalFeatures.depthSampleCompare;
 				portabilityFeatures->pointPolygons = false;
-				portabilityFeatures->samplerMipLodBias = getMVKConfig().useMetalPrivateAPI;
+				portabilityFeatures->samplerMipLodBias = _metalFeatures.samplerMipLodBias;
```
billhollings, line 617: Unfortunately, this change has resulted in several thousand CTS tests of the form:

    dEQP-VK.pipeline.monolithic.sampler.view_type.1d.format.r8g8b8a8_unorm.mipmap.nearest.lod.select_bias_2_1

to move from `NotSupported` to `Fail`.

Perhaps we'll have to look at how the LoD bias is applied in Metal vs Vulkan?

squidbus: These tests are passing for me. Perhaps we need to investigate exactly which hardware family this starts working as expected on? I have an Apple9 GPU (M4 family).

```
./deqp-vk --deqp-case="dEQP-VK.pipeline.monolithic.sampler.*.lod.*"

Test run totals:
  Passed:        4865/6244 (77.9%)
  Failed:        0/6244 (0.0%)
  Not supported: 1379/6244 (22.1%)
  Warnings:      0/6244 (0.0%)
  Waived:        0/6244 (0.0%)
```

PR #2614, opened 2025-09-15
Added working demo target for VisionOS in Cube Demo

rdotrifork:
Added Cube demo for VisionOS.
Tested to work with Xcode 16.4 and Xcode 26 RC. 
VisionOS 2.5 and VisionOS 26 supported on both simulator and device. 

Known limitations:

Vulkan-Tools has been changed to commit 3e9d7a008869360dccbd8c812f4795afa7f8b416, which is the last revision where Cube supports Volk. (This fixes the linked issue about not running on iOS as well)

When using Xcode 26 RC, target 'maccat' from makefile cannot compile. This is deemed out of scope for this PR. 
For Xcode 26 / VisionOS 26 please use 'make visionos visionossim' or temporarily comment out 'maccat' from makefile if using 'make all'.


ShaobinTao:
Locally, it happens to me as well, 

xcode: 16.4
iphone: 16e, ios version: 18.6.2

Thank you for finding and fixing the issues, @rdotrifork

Makefile
```diff
@@ -31,8 +31,8 @@ all:
 	@$(MAKE) maccat
 	@$(MAKE) tvos
 	@$(MAKE) tvossim
-#	@$(MAKE) visionos       # Requires Xcode 15+
-#	@$(MAKE) visionossim    # Requires Xcode 15+
+	@$(MAKE) visionos       # Requires Xcode 15+
+	@$(MAKE) visionossim    # Requires Xcode 15+
```
billhollings, line 33: Also remove the last two lines of the comment at the top of this list that runs: 

`# visionos...earlier Xcode versions`

as it no longer applies.

billhollings (changes requested):
Thanks for submitting. LGTM.

Just a minor nit about comments.

billhollings (approved):
Thanks for making the review suggestion change.

I've adjusted this PR to fix conflicts that arose from other recent commits.

LGTM now.

PR #2615, opened 2025-09-15
Add support for VK_IMAGE_CREATE_BLOCK_TEXEL_VIEW_COMPATIBLE_BIT

squidbus:
Adds support for `VK_IMAGE_CREATE_BLOCK_TEXEL_VIEW_COMPATIBLE_BIT`, using the same technique as `VK_EXT_image_2d_view_of_3d` of aliasing memory with placement heaps.

Passes all supported `dEQP-VK.image.texel_view_compatible.*` tests:
```
Test run totals:
  Passed:        848/8640 (9.8%)
  Failed:        0/8640 (0.0%)
  Not supported: 7792/8640 (90.2%)
  Warnings:      0/8640 (0.0%)
  Waived:        0/8640 (0.0%)
```

Since this code touches the logic related to 2D view of 3D, I also re-ran the CTS tests for that and they are still passing.

I also ran into some trouble with the CTS tests crashing on some sample positions related tests, due to an assert in Metal now that sample positions are unsupported with sample count 1. Made a tweak to supported sample positions sample counts to resolve this.


squidbus:
Hit on I think the general solution but having some trouble trying to use offsets from `MVKImageSubresource`, are those known to be correct?

squidbus:
Now passing all of the tests except for ASTC 8x5 / 8x6 / 8x8 formats.

squidbus:
This is now passing all of its tests and ready to be reviewed. Unfortunately to properly get offsets for the mip levels I had to do some logic specific to Apple GPUs, so I've marked this as only supported for those.

MoltenVK/MoltenVK/API/mvk_deprecated_api.h
```diff
@@ -101,7 +101,8 @@ typedef struct {
     VkDeviceSize mtlBufferAlignment;            	/**< The alignment used when allocating memory for MTLBuffers. Must be PoT. */
     VkDeviceSize maxQueryBufferSize;            	/**< The maximum size of an occlusion query buffer (in bytes). */
 	VkDeviceSize mtlCopyBufferAlignment;        	/**< The alignment required during buffer copy operations (in bytes). */
-    VkSampleCountFlags supportedSampleCounts;   	/**< A bitmask identifying the sample counts supported by the device. */
+	VkSampleCountFlags supportedSampleCounts;   	/**< A bitmask identifying the sample counts supported by the device. */
+	VkSampleCountFlags supportedSamplePosCounts;    /**< A bitmask identifying the sample counts for which the device supports sample positions. */
```
billhollings: Anything new should be placed at the bottom of the structure, to accommodate the legacy query API. It allows the struct to be passed safely even if MoltenVK and the app are compiled using different versions of the header. See the instruction comments at the top of this structure.

billhollings (changes requested):
Thanks for submitting! 

LGTM. Just a minor member declaration location change required.

Before pulling this in, I'm going to wait until #2572 is finally pulled in, in case this code conflict with it.

MoltenVK/MoltenVK/API/mvk_deprecated_api.h
squidbus: Ah okay, fixed.

squidbus:
Rebased this after merge of the mentioned PR.

billhollings (approved):
LGTM now. Thanks!

Issue #2560, opened 2025-06-09
New features available in Metal 4


squidbus:
Metal 4 was just announced at WWDC 2025 with some new features that may be of interest to MoltenVK.

Updated feature set table: https://developer.apple.com/metal/Metal-Feature-Set-Tables.pdf

Of particular interest:
* Placement sparse textures and buffers
* Command barriers

spnda:
This site seems to give a good overview of all new Metal 4 features and API changes: https://developer.apple.com/documentation/metal/understanding-the-metal-4-core-api
The opening sentence of that website makes me happy somehow, since Apple acknowledges the existence of DX and Vulkan explicitly:
> Metal 4 improves runtime performance and memory efficiency through its underlying implementations, while making it easier to adapt your apps and games from other platforms, such as DirectX and Vulkan.

- Metal 4 is based on C++17, and allows lambdas
- Most new types have been moved into a `MTL4` namespace, and some newer versions of existing types have been moved here, e.g. `MTL4RenderPass`, or `MTL4Pipeline`.
- Asynchronous compilation using `MTL4Compiler`
- `MTL4Archive` looks like a replacement for `MTLBinaryArchive`, however this still seems to take in a URL instead of bytes, sadly.
- Tensors, which are used with the `MTL4MachineLearningPipeline`
- All of these types seem to make it easier to model the Vulkan command model on top of Metal, in every way shape and form. It gets rid of specialized encoders such as `MTLBlitCommandEncoder` and `MTLAccelerationStructureCommandEncoder` and fuses them together into one `MTL4ComputeCommandEncoder`.
- The command buffer model also now more closely follows the Vulkan model, with a `MTL4CommandBuffer`, `MTL4CommandQueue`, and `MTL4CommandAllocator`, with the last serving effectively as a `VkCommandPool`, while the other two are `VkCommandBuffer` and `VkQueue`, respectively.
- The concept of descriptor sets now exists with `MTL4ArgumentTable`, which avoids having to set the bindings every time you encode a command buffer
- Automatic hazard tracking has been removed for all the `MTL4` pipelines/command encoders. People now need to do perform manual synchronization using the [`barrier`](https://developer.apple.com/documentation/metal/mtl4commandencoder/barrier(afterqueuestages:beforestages:visibilityoptions:)) methods
- Actual texture views using [`MTLTextureViewPool`](https://developer.apple.com/documentation/metal/mtltextureviewpool)
[...]

PR #2623, opened 2025-10-02
Raise minimum target to macOS 11.0 / iOS 14.0 / tvOS 14.0

squidbus:
The latest Xcode raised the minimum target to macOS 11.0 / iOS 15.0 / tvOS 15.0 (which also causes linker target version warnings, since we configure for 10.15 but it builds for 11.0).

As such, bumping the minimum target versions to macOS 11.0 / iOS 14.0 / tvOS 14.0. Technically iOS/tvOS could go up to 15.0, but I decided to just match it to the corresponding macOS min target for now since a lot of compatibility checks will still need to exist for that anyway.

Along with this, taking the opportunity to clean up any OS or Xcode specific checks and limitations around older, unsupported versions.


galad87:
It doesn't. It recommends setting the deployment target to macOS 11.0 / iOS 15.0 / tvOS 15.0, but it still compile happily for macOS 10.13 for example.

squidbus:
> It doesn't. It recommends setting the deployment target to macOS 11.0 / iOS 15.0 / tvOS 15.0, but it still compile happily for macOS 10.13 for example.

* The support document [here](https://developer.apple.com/support/xcode/) lists the minimum deployment target as macOS 11.0
* Building on Xcode 26 with the MoltenVK deployment target set to 10.15 results in a lot of this:
```
ld: warning: object file (~/Library/Developer/Xcode/DerivedData/MoltenVKPackaging-bazkacdecrkcgoealjsqnxegqneu/Build/Products/Release/libMoltenVKShaderConverter.a(spirv_glsl.o)) was built for newer macOS version (11.0) than being linked (10.15)
```

galad87:
I guess someone at Apple didn't get the right memo, because I have some apps compiled with Xcode 26 running just fine on macOS 10.13, furthermore Xcode 26 gives this warning when trying with an older deployment target: "The macOS deployment target 'MACOSX_DEPLOYMENT_TARGET' is set to 10.12, but the range of supported deployment target versions is 10.13 to 26.0.99."

However, they could be setting a macOS 11 minimum for Apple Silicon, it didn't ever make much sense to target something lower there.

jcm93:
> ```
> ld: warning: object file (~/Library/Developer/Xcode/DerivedData/MoltenVKPackaging-bazkacdecrkcgoealjsqnxegqneu/Build/Products/Release/libMoltenVKShaderConverter.a(spirv_glsl.o)) was built for newer macOS version (11.0) than being linked (10.15)
> ```

I think the linker warning you're seeing here applies to the arm64 slice; the Intel slice should still reflect the 10.15 target. Whether things work on 10.15 is another matter.

squidbus:
> I think the linker warning you're seeing here applies to the arm64 slice; the Intel slice should still reflect the 10.15 target. Whether things work on 10.15 is another matter.

Not sure but I wasn't getting the messages before Xcode 26.

Regardless, historically MoltenVK has targeted the minimum stated deployment target of the latest Xcode version, hence the changes proposed here. There was even discussion in https://github.com/KhronosGroup/MoltenVK/issues/2560 about sticking to Metal 3 and up; this is a bit more conservative than that by just bumping to Metal 2.3 (macOS 11 / iOS 14)

billhollings (approved):
Yup. I'm good with it. 5 years back is far enough!

Thanks! A heck of a cleanup!

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
```diff
@@ -1421,9 +1419,7 @@
 	if (_features.tessellationShader) {
 		pVk11Props->subgroupSupportedStages |= VK_SHADER_STAGE_TESSELLATION_CONTROL_BIT;
 	}
-	if (mvkOSVersionIsAtLeast(10.15, 13.0, 1.0)) {
-		pVk11Props->subgroupSupportedStages |= VK_SHADER_STAGE_FRAGMENT_BIT;
-	}
+	pVk11Props->subgroupSupportedStages |= VK_SHADER_STAGE_FRAGMENT_BIT;
```
roadrunnerto100: This can be combined with the VK_SHADER_STAGE_COMPUTE_BIT above.

squidbus: Done

mrjasonn:
Wouldn't it be a better idea to raise it to macOS 12 / iOS 15 / tvOS 15 instead? Apps that most people need working, such as Google Chrome and WhatsApp, already require this. That way you guys can have more Metal features without breaking too much compatibility.

squidbus:
> Wouldn't it be a better idea to raise it to macOS 12 / iOS 15 / tvOS 15 instead? Apps that most people need working, such as Google Chrome and WhatsApp, already require this. That way you guys can have more Metal features without breaking too much compatibility.

As far as I can tell looking over version checks, 12.0 doesn't seem to be a big deal for MoltenVK. 13.0 would give us guaranteed Metal 3 but that might be a bit too far. Something to maybe discuss further post this PR, for now I focused on just the Apple-supported Xcode minimum targets.

mrjasonn:
> > Wouldn't it be a better idea to raise it to macOS 12 / iOS 15 / tvOS 15 instead? Apps that most people need working, such as Google Chrome and WhatsApp, already require this. That way you guys can have more Metal features without breaking too much compatibility.
> 
> As far as I can tell looking over version checks, 12.0 doesn't seem to be a big deal for MoltenVK. 13.0 would give us guaranteed Metal 3 but that might be a bit too far. Something to maybe discuss further post this PR, for now I focused on just the Apple-supported Xcode minimum targets.

I guess that would make sense. Metal 3 could be something to discuss later on, and to be honest I do not think that it would be too far. I am also aware that there is more ray tracing support in Metal 2.4, so if this project covers that, maybe 12.0 would be a better bet.

PR #2624, opened 2025-10-02
Fix depth bounds test and logic op implementations.

squidbus:
* Update depth bounds test to use the new public API. According to the feature set table, this is only available on the new `Apple10` GPU family; on my `Apple9` GPU the implementation is just a trap instruction.
  * Removed `setDepthBoundsTestAMD` as it appears to no longer exist (not sure since when).
* Fix logic operation private APIs. Even some of the original issues on this (https://github.com/KhronosGroup/MoltenVK/issues/1657, https://github.com/KhronosGroup/MoltenVK/issues/1670) note the correct methods for this, not sure how it ended up the way it did but tested with CTS and they work now.
  * Confirmed as well that the values are equal to D3D11 and updated the related code comment.
* Added note that `lodBias` does not seem to need `Apple10` despite what the feature set table says.


oscarbg:
Nice work!
Can comment if any hope for shaderCullDistance being emulated eventually as shared on the request issue:
"shaderCullDistance could be implemented by transforming the shaders. In fact, asahi driver within mesa did it that way. However, MoltenVK lacks an framework to perform IR-level manipulation. I really hope someone could take it up."

squidbus:
> Nice work! Can comment if any hope for shaderCullDistance being emulated eventually as shared on the request issue: "shaderCullDistance could be implemented by transforming the shaders. In fact, asahi driver within mesa did it that way. However, MoltenVK lacks an framework to perform IR-level manipulation. I really hope someone could take it up."

Not in my plans at least, no.

MoltenVK/MoltenVK/Commands/MVKCommandEncoderState.mm
```diff
@@ -1328,26 +1322,21 @@ static uint32_t getSampleCount(VkSampleCountFlags vk) {
 				[encoder setDepthBias:0 slopeScale:0 clamp:0];
 			}
 		}
-#if MVK_USE_METAL_PRIVATE_API
+#if MVK_XCODE_26
 		if (anyStateNeeded.hasAny({ MVKRenderStateFlag::DepthBounds, MVKRenderStateFlag::DepthBoundsTestEnable }) &&
-		    mvkEncoder.getEnabledFeatures().depthBounds && mvkEncoder.getMVKConfig().useMetalPrivateAPI)
+		    mvkEncoder.getEnabledFeatures().depthBounds && mvkEncoder.getMetalFeatures().depthBoundsTest)
```
billhollings: Since `depthBounds` is set directly from `depthBoundsTest` now, you don't need to test both now. Just use `depthBoundsTest`.

billhollings (approved):
Thanks! LGTM. One minor nit suggestion.

MoltenVK/MoltenVK/Commands/MVKCommandEncoderState.mm
squidbus: Good point, updated.

cdavis5e:
> * Removed `setDepthBoundsTestAMD` as it appears to no longer exist (not sure since when).

That may be because it's defined only in AMD's drivers.

PR #2627, opened 2025-10-04
Additional clean up of old OS version support.

squidbus:
Second round of removing legacy support code paths. This focuses on a few things I missed from the previous PR, as well as deprecating various feature flags that were always set to true.

Additionally, relaxes some `MVK_APPLE_SILICON` checks to simulator/GPU checks to allow features to work under Rosetta, and improves feature parity between tvOS and other platforms.

### DIFF
--- a/MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
@@ -3362,7 +3362,7 @@ bool MVKPhysicalDevice::isAMDRDNAGPU() {
 
 #endif	//MVK_MACOS
 
-#if !MVK_MACOS
+#if MVK_IOS_OR_TVOS_OR_VISIONOS
 // For Apple Silicon, the Device ID is determined by the highest
 // GPU capability, which is a combination of OS version and GPU type.
 void MVKPhysicalDevice::initGPUInfoProperties() {
@@ -3371,7 +3371,7 @@ void MVKPhysicalDevice::initGPUInfoProperties() {
 	_properties.deviceType = VK_PHYSICAL_DEVICE_TYPE_INTEGRATED_GPU;
 	strlcpy(_properties.deviceName, _mtlDevice.name.UTF8String, VK_MAX_PHYSICAL_DEVICE_NAME_SIZE);
 }
-#endif	//!MVK_MACOS
+#endif	//MVK_IOS_OR_TVOS_OR_VISIONOS
 
 // Since this is a uint8_t array, use Big-Endian byte ordering,
 // so a hex dump of the array is human readable in its parts.


MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
```diff
@@ -769,7 +770,7 @@
 			}
 			case VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_TEXEL_BUFFER_ALIGNMENT_FEATURES_EXT: {
 				auto* texelBuffAlignFeatures = (VkPhysicalDeviceTexelBufferAlignmentFeaturesEXT*)next;
-				texelBuffAlignFeatures->texelBufferAlignment = _metalFeatures.texelBuffers && [_mtlDevice respondsToSelector: @selector(minimumLinearTextureAlignmentForPixelFormat:)];
+				texelBuffAlignFeatures->texelBufferAlignment = [_mtlDevice respondsToSelector: @selector(minimumLinearTextureAlignmentForPixelFormat:)];
```
cdavis5e: That method is available starting in macOS 10.13 and iOS 11. We could probably get away with:
```suggestion
				texelBuffAlignFeatures->texelBufferAlignment = true;
```

cdavis5e (changes requested):
One little thing...

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
squidbus: Thanks, I've gone through all the selector checks and eliminated anything that is supported by all of our minimum targets, as well as handling some compile-time disabled APIs on tvOS that have since been made available.

MoltenVK/MoltenVK/Commands/MVKCmdDraw.mm
```diff
@@ -271,10 +271,10 @@
...
 				id<MTLComputePipelineState> vtxState = pipeline->getTessVertexStageState();
 				if (mtlFeats.nonUniformThreadgroups) {
-#if MVK_MACOS_OR_IOS
-					[mtlTessCtlEncoder dispatchThreads: MTLSizeMake(_vertexCount, _instanceCount, 1)
-                                 threadsPerThreadgroup: MTLSizeMake(vtxState.threadExecutionWidth, 1, 1)];
-#endif
+					if ( [mtlTessCtlEncoder respondsToSelector: @selector(dispatchThreads:threadsPerThreadgroup:)] ) {
+						[mtlTessCtlEncoder dispatchThreads: MTLSizeMake(_vertexCount, _instanceCount, 1)
+						             threadsPerThreadgroup: MTLSizeMake(vtxState.threadExecutionWidth, 1, 1)];
+					}
```
cdavis5e: Do we even need the extra `if`? Presumably, any system where `nonUniformThreadgroups` is true is also one where the `-[MTLComputeCommandEncoder dispatchThreads:threadsPerThreadgroup:]` method would be available. Or are you afraid that `nonUniformThreadgroups` would get set on tvOS for an Apple TV model where it isn't actually supported?
```suggestion
					[mtlTessCtlEncoder dispatchThreads: MTLSizeMake(_vertexCount, _instanceCount, 1)
					             threadsPerThreadgroup: MTLSizeMake(vtxState.threadExecutionWidth, 1, 1)];
```

MoltenVK/MoltenVK/Commands/MVKCmdDraw.mm
```diff
@@ -316,10 +316,10 @@
...
 				}
 				if (mtlFeats.nonUniformThreadgroups) {
-#if MVK_MACOS_OR_IOS
-					[mtlTessCtlEncoder dispatchThreads: MTLSizeMake(tessParams.patchCount * outControlPointCount, 1, 1)
-								 threadsPerThreadgroup: MTLSizeMake(wgSize, 1, 1)];
-#endif
+					if ( [mtlTessCtlEncoder respondsToSelector: @selector(dispatchThreads:threadsPerThreadgroup:)] ) {
+						[mtlTessCtlEncoder dispatchThreads: MTLSizeMake(tessParams.patchCount * outControlPointCount, 1, 1)
+									 threadsPerThreadgroup: MTLSizeMake(wgSize, 1, 1)];
+					}
```
cdavis5e:
```suggestion
					[mtlTessCtlEncoder dispatchThreads: MTLSizeMake(tessParams.patchCount * outControlPointCount, 1, 1)
								 threadsPerThreadgroup: MTLSizeMake(wgSize, 1, 1)];
```

MoltenVK/MoltenVK/Commands/MVKCmdDraw.mm
```diff
@@ -492,10 +492,10 @@
...
 				id<MTLComputePipelineState> vtxState = ibb.mtlIndexType == MTLIndexTypeUInt16 ? pipeline->getTessVertexStageIndex16State() : pipeline->getTessVertexStageIndex32State();
 				if (mtlFeats.nonUniformThreadgroups) {
-#if MVK_MACOS_OR_IOS
-					[mtlTessCtlEncoder dispatchThreads: MTLSizeMake(_indexCount, _instanceCount, 1)
-                                 threadsPerThreadgroup: MTLSizeMake(vtxState.threadExecutionWidth, 1, 1)];
-#endif
+					if ( [mtlTessCtlEncoder respondsToSelector: @selector(dispatchThreads:threadsPerThreadgroup:)] ) {
+						[mtlTessCtlEncoder dispatchThreads: MTLSizeMake(_indexCount, _instanceCount, 1)
+						             threadsPerThreadgroup: MTLSizeMake(vtxState.threadExecutionWidth, 1, 1)];
+					}
```
cdavis5e:
```suggestion
					[mtlTessCtlEncoder dispatchThreads: MTLSizeMake(_indexCount, _instanceCount, 1)
					             threadsPerThreadgroup: MTLSizeMake(vtxState.threadExecutionWidth, 1, 1)];
```

MoltenVK/MoltenVK/Commands/MVKCmdDraw.mm
```diff
@@ -538,10 +538,10 @@
...
 				}
 				if (mtlFeats.nonUniformThreadgroups) {
-#if MVK_MACOS_OR_IOS
-					[mtlTessCtlEncoder dispatchThreads: MTLSizeMake(tessParams.patchCount * outControlPointCount, 1, 1)
-								 threadsPerThreadgroup: MTLSizeMake(wgSize, 1, 1)];
-#endif
+					if ( [mtlTessCtlEncoder respondsToSelector: @selector(dispatchThreads:threadsPerThreadgroup:)] ) {
+						[mtlTessCtlEncoder dispatchThreads: MTLSizeMake(tessParams.patchCount * outControlPointCount, 1, 1)
+									 threadsPerThreadgroup: MTLSizeMake(wgSize, 1, 1)];
+					}
```
cdavis5e:
```suggestion
					[mtlTessCtlEncoder dispatchThreads: MTLSizeMake(tessParams.patchCount * outControlPointCount, 1, 1)
								 threadsPerThreadgroup: MTLSizeMake(wgSize, 1, 1)];
```

MoltenVK/MoltenVK/Commands/MVKCmdDraw.mm
```diff
@@ -667,10 +667,10 @@
...
 	state.bindBuffer(mtlConvertEncoder, ibb.mtlBuffer, ibb.offset, 4);
 	if (cmdEncoder->getMetalFeatures().nonUniformThreadgroups) {
-#if MVK_MACOS_OR_IOS
-		[mtlConvertEncoder dispatchThreads: MTLSizeMake(_drawCount, 1, 1)
-					 threadsPerThreadgroup: MTLSizeMake(mtlConvertState.threadExecutionWidth, 1, 1)];
-#endif
+		if ( [mtlConvertEncoder respondsToSelector: @selector(dispatchThreads:threadsPerThreadgroup:)] ) {
+			[mtlConvertEncoder dispatchThreads: MTLSizeMake(_drawCount, 1, 1)
+						 threadsPerThreadgroup: MTLSizeMake(mtlConvertState.threadExecutionWidth, 1, 1)];
+		}
```
cdavis5e:
```suggestion
		[mtlConvertEncoder dispatchThreads: MTLSizeMake(_drawCount, 1, 1)
					 threadsPerThreadgroup: MTLSizeMake(mtlConvertState.threadExecutionWidth, 1, 1)];
```

MoltenVK/MoltenVK/Commands/MVKCmdDraw.mm
```diff
@@ -796,10 +796,10 @@
...
 				state.bindStructBytes(mtlTessCtlEncoder, &tcWorkgroupSize,          8);
 				if (mtlFeats.nonUniformThreadgroups) {
-#if MVK_MACOS_OR_IOS
-					[mtlTessCtlEncoder dispatchThreads: MTLSizeMake(_drawCount, 1, 1)
-								 threadsPerThreadgroup: MTLSizeMake(mtlConvertState.threadExecutionWidth, 1, 1)];
-#endif
+					if ( [mtlTessCtlEncoder respondsToSelector: @selector(dispatchThreads:threadsPerThreadgroup:)] ) {
+						[mtlTessCtlEncoder dispatchThreads: MTLSizeMake(_drawCount, 1, 1)
+									 threadsPerThreadgroup: MTLSizeMake(mtlConvertState.threadExecutionWidth, 1, 1)];
+					}
```
cdavis5e:
```suggestion
					[mtlTessCtlEncoder dispatchThreads: MTLSizeMake(_drawCount, 1, 1)
								 threadsPerThreadgroup: MTLSizeMake(mtlConvertState.threadExecutionWidth, 1, 1)];
```

MoltenVK/MoltenVK/Commands/MVKCmdDraw.mm
```diff
@@ -820,10 +820,10 @@
...
 				state.bindStructBytes(mtlConvertEncoder, &viewCount,                4);
                 if (mtlFeats.nonUniformThreadgroups) {
-#if MVK_MACOS_OR_IOS
-                    [mtlConvertEncoder dispatchThreads: MTLSizeMake(_drawCount, 1, 1)
-                                 threadsPerThreadgroup: MTLSizeMake(mtlConvertState.threadExecutionWidth, 1, 1)];
-#endif
+					if ( [mtlConvertEncoder respondsToSelector: @selector(dispatchThreads:threadsPerThreadgroup:)] ) {
+						[mtlConvertEncoder dispatchThreads: MTLSizeMake(_drawCount, 1, 1)
+						             threadsPerThreadgroup: MTLSizeMake(mtlConvertState.threadExecutionWidth, 1, 1)];
+					}
```
cdavis5e:
```suggestion
					[mtlConvertEncoder dispatchThreads: MTLSizeMake(_drawCount, 1, 1)
					             threadsPerThreadgroup: MTLSizeMake(mtlConvertState.threadExecutionWidth, 1, 1)];
```

MoltenVK/MoltenVK/Commands/MVKCmdDraw.mm
```diff
@@ -1135,10 +1133,10 @@
...
 				state.bindBuffer(mtlConvertEncoder, ibbTriFan.mtlBuffer, ibbTriFan.offset, 7);
 				if (mtlFeats.nonUniformThreadgroups) {
-#if MVK_MACOS_OR_IOS
-					[mtlConvertEncoder dispatchThreads: MTLSizeMake(_drawCount, 1, 1)
-								 threadsPerThreadgroup: MTLSizeMake(mtlConvertState.threadExecutionWidth, 1, 1)];
-#endif
+					if ( [mtlConvertEncoder respondsToSelector: @selector(dispatchThreads:threadsPerThreadgroup:)] ) {
+						[mtlConvertEncoder dispatchThreads: MTLSizeMake(_drawCount, 1, 1)
+									 threadsPerThreadgroup: MTLSizeMake(mtlConvertState.threadExecutionWidth, 1, 1)];
+					}
```
cdavis5e:
```suggestion
					[mtlConvertEncoder dispatchThreads: MTLSizeMake(_drawCount, 1, 1)
								 threadsPerThreadgroup: MTLSizeMake(mtlConvertState.threadExecutionWidth, 1, 1)];
```

MoltenVK/MoltenVK/Commands/MVKCmdDraw.mm
```diff
@@ -907,13 +905,13 @@
...
-														indirectBuffer: mtlIndBuff
-												  indirectBufferOffset: mtlIndBuffOfst];
-#endif
+							if ( [cmdEncoder->_mtlRenderEncoder respondsToSelector: @selector(drawPatches:patchIndexBuffer:patchIndexBufferOffset:indirectBuffer:indirectBufferOffset:)] ) {
+								[cmdEncoder->_mtlRenderEncoder drawPatches: outControlPointCount
+														  patchIndexBuffer: nil
+													patchIndexBufferOffset: 0
+															indirectBuffer: mtlIndBuff
+													  indirectBufferOffset: mtlIndBuffOfst];
+							}
```
cdavis5e: Again, we probably don't need the `if`. If clients call `vkCmdDraw*()` with the primitive topology set to `PATCH_LIST` when `tessellationShader` isn't supported, that's on them.
```suggestion
							[cmdEncoder->_mtlRenderEncoder drawPatches: outControlPointCount
													  patchIndexBuffer: nil
												patchIndexBufferOffset: 0
														indirectBuffer: mtlIndBuff
												  indirectBufferOffset: mtlIndBuffOfst];
```

MoltenVK/MoltenVK/Commands/MVKCmdDraw.mm
```diff
@@ -1227,13 +1223,13 @@
...
-														indirectBuffer: mtlIndBuff
-												  indirectBufferOffset: mtlTempIndBuffOfst];
-#endif
+							if ( [cmdEncoder->_mtlRenderEncoder respondsToSelector: @selector(drawPatches:patchIndexBuffer:patchIndexBufferOffset:indirectBuffer:indirectBufferOffset:)] ) {
+								[cmdEncoder->_mtlRenderEncoder drawPatches: outControlPointCount
+														  patchIndexBuffer: nil
+													patchIndexBufferOffset: 0
+															indirectBuffer: mtlIndBuff
+													  indirectBufferOffset: mtlTempIndBuffOfst];
+							}
```
cdavis5e:
```suggestion
							[cmdEncoder->_mtlRenderEncoder drawPatches: outControlPointCount
													  patchIndexBuffer: nil
												patchIndexBufferOffset: 0
														indirectBuffer: mtlIndBuff
												  indirectBufferOffset: mtlTempIndBuffOfst];
```

MoltenVK/MoltenVK/Commands/MVKCommandEncoderState.mm
```diff
@@ -1144,10 +1144,8 @@ static uint32_t getSampleCount(VkSampleCountFlags vk) {
 			}
 			if (numViewports == 1) {
 				[encoder setViewport:mtlViewports[0]];
-			} else {
-#if MVK_MACOS_OR_IOS
+			} else if ([encoder respondsToSelector: @selector(setViewports:count:)]) {
```
cdavis5e: And here, if clients try to set multiple viewports and/or scissor rects when `multiViewport` isn't supported, they may have summoned nasal demons.
```suggestion
			} else {
```

MoltenVK/MoltenVK/Commands/MVKCommandEncoderState.mm
```diff
@@ -1366,10 +1364,8 @@ static uint32_t getSampleCount(VkSampleCountFlags vk) {
 		}
 		if (numScissors == 1) {
 			[encoder setScissorRect:mtlScissors[0]];
-		} else {
-#if MVK_MACOS_OR_IOS
+		} else if ([encoder respondsToSelector: @selector(setScissorRects:count:)]) {
```
cdavis5e:
```suggestion
		} else {
```

MoltenVK/MoltenVK/Commands/MVKCmdDraw.mm
squidbus: The issue here is specifically indirect tessellation draw, which is not available on tvOS until 14.5. Should we just disable tessellation entirely for previous versions...?

EDIT: Ah, I see that this is also gated by a feature flag as well though, so I can remove the selector check.

MoltenVK/MoltenVK/Commands/MVKCmdDraw.mm
squidbus: You're right, missed that there's already a feature flag check. Going to update the feature flags for tvOS and remove these.

MoltenVK/MoltenVK/Commands/MVKCmdDraw.mm
squidbus: Alternatively, I could bump the iOS/tvOS requirement to 15.0 and get rid of some tvOS missing features, since that's the actual Xcode 26 minimum, even though that doesn't quite correspond to macOS 11.0.

squidbus: I opted to bump just tvOS to 14.5 here, takes care of most APIs that were missing only on tvOS.

squidbus:
There's probably plenty more opportunities to unify the non-macOS paths here but I think I've done enough for this PR.

MoltenVK/MoltenVK/GPUObjects/MVKPixelFormats.mm
```diff
@@ -156,7 +155,6 @@
 		case MTLPixelFormatPVRTC_RGB_4BPP:
 		case MTLPixelFormatPVRTC_RGB_4BPP_sRGB:
 			return true;
-#endif
```
billhollings, line 159: I must admit, the removal of `MVK_APPLE_SILICON` shocks me. Are there no conditions where building for an Intel/AMD/NVIDIA device behaves differently at compile time? 

Here is an example. These formats are not available on Intel/AMD/NVIDIA devices, but we don't check anywhere for that now, either with `MVK_APPLE_SILICON` or `isAppleGPU`. I admit `isAppleGPU` a better test anyway, but we don't seem to use even that. Are we relying on the fact that we don't seem to ever populate `macGPUCaps` for these formats?

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
```diff
@@ -2880,36 +2781,30 @@
 #if MVK_OS_SIMULATOR
 	_metalFeatures.mtlBufferAlignment = 256;	// Even on Apple Silicon
 	_metalFeatures.nativeTextureSwizzle = false;
+	_metalFeatures.renderLinearTextures = false;
```
billhollings, line 2784: Even on Apple Silicon Macs?

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
```diff
@@ -1658,7 +1657,6 @@
 				return VK_ERROR_FORMAT_NOT_SUPPORTED;
 			}
 #endif
-#if MVK_APPLE_SILICON
 			// ETC2 and EAC formats aren't supported for 3D textures.
```
billhollings, line 1660: What about Intel/AMD/NVIDIA devices?

Common/MVKCommonEnvironment.h
```diff
@@ -87,25 +87,15 @@ extern "C" {
 #endif
 
 /** Building for iOS or visionOS. */
-#ifndef MVK_IOS_OR_VISIONOS
-#    define MVK_IOS_OR_VISIONOS    (MVK_IOS || MVK_VISIONOS)
+#ifndef MVK_IOS_OR_TVOS_OR_VISIONOS
+#    define MVK_IOS_OR_TVOS_OR_VISIONOS    (MVK_IOS || MVK_TVOS || MVK_VISIONOS)
 #endif
```
billhollings, line 92: Hmmm. 

Can `MVK_IOS_OR_TVOS_OR_VISIONOS` now be replaced with `!MVK_MACOS`? 

(I think even with the hassle of `TARGET_OS_MACCATALYST` in the mix, the logic works).

But should it? Or is it better to explicitly always target platforms positively, in case any future platforms appear?

If so, can you replace the one existing instance of `!MVK_MACOS` with `MVK_IOS_OR_TVOS_OR_VISIONOS`, please?

MoltenVK/MoltenVK/GPUObjects/MVKDeviceMemory.mm
```diff
@@ -118,7 +117,6 @@
 	if ( !_mtlHeap ) {
 		lock_guard<mutex> lock(_rezLock);
         for (auto& img : _imageMemoryBindings) { img->pullFromDevice(offset, memSize); }
-        for (auto& buf : _buffers) { buf->pullFromDevice(offset, memSize); }
```
billhollings, line 121: Why don't we need to do these buffer memory syncs anymore?

MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
```diff
@@ -2203,7 +2177,7 @@ static void signalAndUntrack(const MVKSwapchainSignaler& signaler) {
 
 		switch (_mtlPixFmt) {
 			case MTLPixelFormatR8Unorm:
-#if MVK_APPLE_SILICON
+#if !MVK_OS_SIMULATOR
 			case MTLPixelFormatR8Unorm_sRGB:
```
billhollings, line 2181: Is this the same test? What about builds on macOS Apple Silicon that target iOS/tvOS/visionOS simulator? Won't that build now exclude this format, even though it's a usable format for that simulator to use?

And similarly, won't Intel/AMD/NVIDIA builds that are not targeting the simulator now include this format, but it isn't supported?

Hmmm. Or is this format is now also supported on Intel/AMD/NVIDIA (as of macOS 11 per API header)? And if so, do we even need the `MVK_OS_SIMULATOR` guard?

This was one of the reasons why I mentioned elsewhere that I was shocked to see `MVK_APPLE_SILICON` go.

billhollings (commented):
Wow! What an epic cleanup! Thanks so much for taking care of all this! It is a tremendous simplification for the future. 🚀🚀

I don't have any specific changes, but I do have some concerns that I've tried to capture as questions. Perhaps they can be answered sufficiently, or perhaps they might trigger requested changes.

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
squidbus: According to the feature set table these formats are not supported at all outside of Apple GPUs, hence a caller should never be using them to begin with.

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
squidbus: Not 100% certain about the simulator capabilities, other than that from past comments around here they seem to follow the same restrictions as on Intel.

MoltenVK/MoltenVK/GPUObjects/MVKPixelFormats.mm
squidbus: This does rely on them not being used outside of Apple GPUs, yes, as they are not supported. Any caller of this function will already fail because they are using an unsupported pixel format.

Let me know if there's a component missing for blocking off PVRTC support for non-Apple GPUs and I can take a look at that.

MoltenVK/MoltenVK/GPUObjects/MVKDeviceMemory.mm
squidbus: All of the buffer syncs relied on `shouldFlushHostMemory()`, which in turn relies on `_isHostCoherentTexelBuffer`, which *in turn* relies on `!getMetalFeatures().sharedLinearTextures`. Since `sharedLinearTextures` is now always true, it is no longer needed.

MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
squidbus: `MTLPixelFormatR8Unorm_sRGB` is set to `MTLPixelFormatInvalid` in `MVKPixelFormats`, if the build is for simulator. If it's safe to enable here still I can do so.

Regarding Intel/AMD/NVIDIA, it's another case of, if the format is unsupported it should be advertised as such, and the user should not be using it. Let me know again if there's some component of this I'm missing here.

Common/MVKCommonEnvironment.h
squidbus: Replaced it. I do want to try and just use `!MVK_MACOS` everywhere since presumably with their unified platform it should be the only special case (up until Intel Macs get dropped anyway), but the PR was getting rather large already, so maybe for a follow up.

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
billhollings: D'oh! You're correct! 🤦🏻‍♂️

billhollings (approved):
Thanks for responding to my questions. I'm convinced! LGTM now.

PR #2628, opened 2025-10-05
#include MVKRenderPass.h typo

ShaobinTao:
#Summary
- MVKRenderPass.h file name has "**P**"
- But in vulkan.mm, the include file name is #include "MVKRender**p**ass.h"
- in some build env, it is causing issues.


ShaobinTao:
maybe because different compiler versions?

PR #2631, opened 2025-10-06
Add base GPU family for samplerMipLodBias support.

squidbus:
Adds a base GPU family of `Apple9` for `samplerMipLodBias` support.

Testing has revealed that it does not function correctly on `Apple7` (e.g. M1) GPUs. I have confirmed it works on `Apple9`, but do not have an `Apple8` device to test, so I have left that as a TODO item. If anyone can test CTS for this on `Apple8`, I can update the support requirement.

On M4:
```
./deqp-vk --deqp-case="dEQP-VK.pipeline.monolithic.sampler.*.lod.*"

Test run totals:
  Passed:        4865/6244 (77.9%)
  Failed:        0/6244 (0.0%)
  Not supported: 1379/6244 (22.1%)
  Warnings:      0/6244 (0.0%)
  Waived:        0/6244 (0.0%)
```

### DIFF
--- a/MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
@@ -2437,8 +2437,9 @@ void MVKPhysicalDevice::initMetalFeatures() {
 #endif
 
 #if MVK_XCODE_26
-	// NOTE: The Metal feature set table claims this requires Apple10, but it seems to work fine on older GPUs as well.
-	_metalFeatures.samplerMipLodBias = mvkOSVersionIsAtLeast(26.0);
+	// NOTE: The Metal feature set table claims this requires Apple10, but it seems to work fine on at least Apple9 as well.
+	// TODO: Test on Apple8 (M2).
+	_metalFeatures.samplerMipLodBias = mvkOSVersionIsAtLeast(26.0) && supportsMTLGPUFamily(Apple9);
 	_metalFeatures.depthBoundsTest = mvkOSVersionIsAtLeast(26.0) && supportsMTLGPUFamily(Apple10);
 #endif
 
--- a/MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
@@ -2437,9 +2437,7 @@ void MVKPhysicalDevice::initMetalFeatures() {
 #endif
 
 #if MVK_XCODE_26
-	// NOTE: The Metal feature set table claims this requires Apple10, but it seems to work fine on at least Apple9 as well.
-	// TODO: Test on Apple8 (M2).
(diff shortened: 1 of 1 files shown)


billhollings:
Hmmm. It doesn't seem to be that simple.

Here's what I get on my M3 Pro, which is reporting `Apple9`, macOS 26.0.1, CTS 1.4.2.1:

```
Test case 'dEQP-VK.pipeline.monolithic.sampler.view_type.1d.format.r8g8b8a8_unorm.mipmap.nearest.lod.select_bias_2_1'..
  Fail (Image mismatch)

---------

Test run totals:
  Passed:        2043/6146 (33.2%)
  Failed:        2724/6146 (44.3%)
  Not supported: 1379/6146 (22.4%)
  Warnings:      0/6146 (0.0%)
  Waived:        0/6146 (0.0%)
```

billhollings:
To make it even more confusing, according to the [Metal Feature Set Tables](https://developer.apple.com/metal/Metal-Feature-Set-Tables.pdf):

Screenshot 2025-10-06 at 2 55 50 PM
<br/><br/>
I wonder if your M4 has some undocumented Apple10 features?

squidbus:
> Here's what I get on my M3 Pro, which is reporting `Apple9`, macOS 26.0.1, CTS 1.4.2.1:

I’m using the 26.1 beta, maybe this makes a difference?

billhollings:
> > Here's what I get on my M3 Pro, which is reporting `Apple9`, macOS 26.0.1, CTS 1.4.2.1:
> 
> I’m using the 26.1 beta, maybe this makes a difference?

It's possible that they are fixing driver bugs, but given that Apple indicates `Apple10` as the minimum, my money would be on that the M4 has some, but not all, `Apple10` features, and you're just getting lucky with this one.

squidbus:
> It's possible that they are fixing driver bugs, but given that Apple indicates `Apple10` as the minimum, my money would be on that the M4 has some, but not all, `Apple10` features, and you're just getting lucky with this one.

I’ll see if I can give it a check on my test M1 later today and see if the beta makes any difference. Don’t have an M2 or M3 though.

billhollings:
> > It's possible that they are fixing driver bugs, but given that Apple indicates `Apple10` as the minimum, my money would be on that the M4 has some, but not all, `Apple10` features, and you're just getting lucky with this one.
> 
> I’ll see if I can give it a check on my test M1 later today and see if the beta makes any difference. Don’t have an M2 or M3 though.

I can try M3 on 26.1, when I upgrade. I might wait until it's out of beta though.

If we want to pull this in now though to fix this issue, perhaps we should enable LoD based on `Apple10`, per the docs. We can always adjust as more info becomes available from experimentation.

squidbus:
Sure, bumped it up for now. Will get back on updated requirements based on investigation, as preferably we can expose this to as many working devices as possible.

squidbus:
It does seem to not work at all on M1, even on the beta. Going to try and look deeper (outside of this PR) and see why it works on M4, and if there's any indication given of this.

Issue #1521, opened 2022-02-05
Disallow or implement configurable primitive restart


lexaknyazev:
In Vulkan, primitive restart may be disabled for strip topologies. In Metal, it's always active.

Consider the following setup that yields a simple square. Here, index `65535` refers to an actual vertex but Metal treats it as a primitive restart command and renders only one triangle.

```c
// Four vertices aligned to the end of uint16 range.
float vertices[65536 * 2];
const float vertices_real[] = {
    -1.0f, +1.0f,
    -1.0f, -1.0f,
    +1.0f, +1.0f,
    +1.0f, -1.0f
};
for (unsigned int i = 0; i < 4; i++) {
    vertices[(65532 + i) * 2]     = vertices_real[i * 2];
    vertices[(65532 + i) * 2 + 1] = vertices_real[i * 2 + 1];
}

// Two triangles
unsigned short indices[] = {
    65532 + 0,
    65532 + 1,
    65532 + 2,
    65532 + 3,
};

/// ...

inputAssemblyStateCreateInfo.topology = VK_PRIMITIVE_TOPOLOGY_TRIANGLE_STRIP;
inputAssemblyStateCreateInfo.primitiveRestartEnable = false;

/// ...

vkCmdDrawIndexed(commandBuffer, 4, 1, 0, 0, 0);
```

This case is not covered by the CTS, internal issue 3500.

billhollings:
PR #2632 lifts the restriction on strip topologies.

PR #2632, opened 2025-10-07
Add support for disabling primitive restart using private APIs.

squidbus:
Adds support for disabling primitive restart using the Metal private API `MTLRenderCommandEncoder::setPrimitiveRestartEnabled`. This API allows specifying both an enable flag and the restart index to use, allowing us to support not only restart states for the basic Metal `UInt32` and `UInt16` index types, but also the emulated `UInt8` index type.

This support has one quirk to it: you must set `MTLRenderPassDescriptor::openGLModeEnabled` to `true`. From my research I haven't found any noticeable side-effects of this other than some private API features functioning where they would otherwise do nothing.

I did look into only enabling this only when primitive restart is disabled, but it would have caused problems with:
* Having to split the render pass any time the primitive restart enable value changes.
* Having to rework `UInt8` index buffer support as the translation result differs depending on whether primitive restart is disabled or enabled and emulated.

Thus I opted to just enable the flag unconditionally for better performance, since it is also assumed the user knows what they are doing if they enable private API usage. Can revisit this down the line if needed.

CTS results:
```
./deqp-vk --deqp-case="dEQP-VK.pipeline.monolithic.input_assembly.primitive_restart.*"

Without private API:

Test run totals:
  Passed:        27/154 (17.5%)
  Failed:        6/154 (3.9%)
  Not supported: 121/154 (78.6%)
  Warnings:      0/154 (0.0%)
  Waived:        0/154 (0.0%)

Fails tests for disabled primitive restart.

With private API:

Test run totals:
  Passed:        33/154 (21.4%)
  Failed:        0/154 (0.0%)
  Not supported: 121/154 (78.6%)
  Warnings:      0/154 (0.0%)
  Waived:        0/154 (0.0%)

Passes all tests.
```


squidbus:
Identified and updated with an additional benefit here: the `openGLModeEnabled` flag lifts an internal Metal restriction that forces primitive restart enabled only on specific primitive topologies, instead delegating to the provided enable flag from the private API. This allows us to support `VK_EXT_primitive_topology_list_restart` for non-patches, which passes the relevant tests.

MoltenVK/MoltenVK/Commands/MVKCommandBuffer.mm
```diff
@@ -30,6 +30,13 @@
 
 using namespace std;
 
+#if MVK_USE_METAL_PRIVATE_API
+// An extension of the MTLRenderCommandEncoder protocol to declare the setLineWidth: method.
```
cdavis5e: Copy-pasto.
```suggestion
// An extension of the MTLRenderPassDescriptor class to declare the setOpenGLModeEnabled: method.
```

rcaridade145:
Just adding a related  issue #1521

MoltenVK/MoltenVK/Commands/MVKCommandBuffer.mm
billhollings:
```suggestion
// An extension of the MTLRenderCommandEncoder protocol to declare the setOpenGLModeEnabled: method.
```

squidbus: Fixed, thanks.

billhollings (approved):

cdavis5e (approved):
Just one small thing.

MoltenVK/MoltenVK/Commands/MVKCommandBuffer.mm
squidbus: I think you may be looking at an older revision, this was addressed in another comment.

Issue #2237, opened 2024-05-11
provoking vertex - is it supported?


timprepscius:
Hey there,

I think "provoking vertex' is not supported.  Because the .def file doesn't include it as an extension.

And when I try to do the following (just ignore that it's pseudo code) it doesn't work:

```
    vku::DeviceMaker dm;
    dm.defaultLayers();
    // the extension doesn't seem to exist
//    dm.extension(VK_EXT_PROVOKING_VERTEX_EXTENSION_NAME);
    
// but let's just try anyway
    vk::PhysicalDeviceProvokingVertexFeaturesEXT provoking;
    provoking.setProvokingVertexLast(true);
    dm.enableCreateInfo((vku::DeviceMaker::DeviceCreateNode *)&provoking);
```

I just want to check before I go through and write some special case code to rearrange index orderings.

Thanks in advance.

cdavis5e:
Nope. This is not supported yet.

I think there might be a Metal SPI[1] for this, since I'm pretty sure OpenGL on Apple Silicon does support it.

[1]: Service Provider Interface: Apple/IBM jargon for a private programming interface intended for use by other OS components.

voxelias:
@cdavis5e does it mean that the whole `flat` shading thing is not supported or only selection of the provoking vertex? Thank you

cdavis5e:
> [@cdavis5e](https://github.com/cdavis5e) does it mean that the whole `flat` shading thing is not supported or only selection of the provoking vertex? Thank you

Only selection of the provoking vertex. Flat shading _is_ supported.

billhollings:
PR #2634 adds support for this when `MVK_USE_METAL_PRIVATE_API` is enabled.

PR #2634, opened 2025-10-08
Add support for VK_EXT_provoking_vertex

squidbus:
Adds support for `VK_EXT_provoking_vertex` using private APIs, as well as the corresponding dynamic state from `VK_EXT_extended_dynamic_state3`.

Also pulled a some private MTL enum constants and relevant extension feature structs out of `MVK_USE_METAL_PRIVATE_API` to make compiling easier to manage. As long as we gate the actual API definitions/usage and corresponding Vulkan feature flags we are fine.

Implements https://github.com/KhronosGroup/MoltenVK/issues/2237

CTS results:
```
./deqp-vk --deqp-case="dEQP-VK.rasterization.provoking_vertex.draw.*"

Test run totals:
  Passed:        20/36 (55.6%)
  Failed:        0/36 (0.0%)
  Not supported: 16/36 (44.4%)
  Warnings:      0/36 (0.0%)
  Waived:        0/36 (0.0%)

./deqp-vk --deqp-case="dEQP-VK.pipeline.monolithic.extended_dynamic_state.*provoking_vertex*"

Test run totals:
  Passed:        23/46 (50.0%)
  Failed:        0/46 (0.0%)
  Not supported: 23/46 (50.0%)
  Warnings:      0/46 (0.0%)
  Waived:        0/46 (0.0%)
```

Draft for now since it builds on https://github.com/KhronosGroup/MoltenVK/pull/2632 which is not yet merged.


squidbus:
Rebased and ready to go.

MoltenVK/MoltenVK/API/mvk_datatypes.h
```diff
@@ -349,8 +349,37 @@ MTLBlendOperation mvkMTLBlendOperationFromVkBlendOp(VkBlendOp vkBlendOp);
...
+	MTLLogicOperationNand,
+	MTLLogicOperationOr,
+	MTLLogicOperationNor,
+	MTLLogicOperationXor,
+	MTLLogicOperationEquivalence,
+	MTLLogicOperationAndReverse,
+	MTLLogicOperationAndInverted,
+	MTLLogicOperationOrReverse,
+	MTLLogicOperationOrInverted,
+};
```
billhollings, line 372: We have to be very careful with exposing these in a build. Apple is notorious and diligent about screening for private-API use in App Store submissions, and will reject macOS and iOS apps that use private APIs and symbols.

My concern here is that their AppStore filters (particularly on iOS) may sniff out private symbols like this, and reject apps using MoltenVK (particularly those that are not built with `MVK_USE_METAL_PRIVATE_API` enabled).

We've been burnt by this in the past. Even just trying to reference something simple like `IOSurface` before it became public on iOS.

squidbus: Noted, I've backtracked a bit and hid some of these in the macro (and fixed some issues making this header invalid for plain C). Not sure which symbols make their way into the final binary but best to play it on the safe side then.

MoltenVK/MoltenVK/API/mvk_datatypes.h
```diff
@@ -471,7 +508,11 @@ static inline MTLSize mvkMTLSizeFromVkExtent3D(VkExtent3D vkExtent) {
 
 /** Returns a Vulkan VkExtent3D  constructed from a Metal MTLSize. */
 static inline VkExtent3D mvkVkExtent3DFromMTLSize(MTLSize mtlSize) {
-	return { (uint32_t)mtlSize.width, (uint32_t)mtlSize.height, (uint32_t)mtlSize.depth };
+	VkExtent3D extent3D;
+	extent3D.width = (uint32_t)mtlSize.width;
+	extent3D.height = (uint32_t)mtlSize.height;
+	extent3D.depth = (uint32_t)mtlSize.depth;
+	return extent3D;
 }
```
billhollings, line 511: It's fine, but I'm curious. Why the change in these two functions? Were these breaking? Or you just prefer being explicit to validate the right things are being assigned?

squidbus: The syntax here was not supported for C code and was failing to compile when included from a non-C++ file.

billhollings: Ah. Yes. You did comment on that above. I missed that.

MoltenVK/MoltenVK/API/mvk_datatypes.h
```diff
@@ -461,7 +494,11 @@ static inline MTLOrigin mvkMTLOriginFromVkOffset3D(VkOffset3D vkOffset) {
 
 /** Returns a Vulkan VkOffset3D constructed from a Metal MTLOrigin. */
 static inline VkOffset3D mvkVkOffset3DFromMTLOrigin(MTLOrigin mtlOrigin) {
-	return { (int32_t)mtlOrigin.x, (int32_t)mtlOrigin.y, (int32_t)mtlOrigin.z };
+	VkOffset3D offset3D;
+	offset3D.x = (int32_t)mtlOrigin.x;
+	offset3D.y = (int32_t)mtlOrigin.y;
+	offset3D.z = (int32_t)mtlOrigin.z;
```
cdavis5e: Does this work?
```suggestion
	VkOffset3D offset3D = { (int32_t)mtlOrigin.x, (int32_t)mtlOrigin.y, (int32_t)mtlOrigin.z };
```
I know it works for globals, but I'm not sure if it works for local automatic variables. I know this _is_ in C at all because in C99, you can even specify exactly the members to initialize.

MoltenVK/MoltenVK/API/mvk_datatypes.h
```diff
@@ -471,7 +508,11 @@ static inline MTLSize mvkMTLSizeFromVkExtent3D(VkExtent3D vkExtent) {
 
 /** Returns a Vulkan VkExtent3D  constructed from a Metal MTLSize. */
 static inline VkExtent3D mvkVkExtent3DFromMTLSize(MTLSize mtlSize) {
-	return { (uint32_t)mtlSize.width, (uint32_t)mtlSize.height, (uint32_t)mtlSize.depth };
+	VkExtent3D extent3D;
+	extent3D.width = (uint32_t)mtlSize.width;
+	extent3D.height = (uint32_t)mtlSize.height;
+	extent3D.depth = (uint32_t)mtlSize.depth;
```
cdavis5e: Ditto.
```suggestion
	VkExtent3D extent3D = { (uint32_t)mtlSize.width, (uint32_t)mtlSize.height, (uint32_t)mtlSize.depth };
```

MoltenVK/MoltenVK/API/mvk_datatypes.h
squidbus: Yes, that seems to work, thanks. I've updated with the suggestion here.

Issue #2637, opened 2025-10-09
Question/Request: add a variant/job mac build to Actions CI with private APIs enabled..


oscarbg:
Hi,
seeing recent updates using private APIs (SPIs) to expose new extensions like:
https://github.com/KhronosGroup/MoltenVK/pull/2634
https://github.com/KhronosGroup/MoltenVK/pull/2632
wanted a Mac build with that support enabled..
seems from:
https://github.com/KhronosGroup/MoltenVK/blob/main/.github/workflows/CI.yml
that builds don't enable private APIs..
makes sense for iOS as no app can be sent to App Store with that enabled but for Mac testing would be nice if a Mac build/job is added to CI for not having to do a local build..
the artifact could be named MoltenVK-macos-spi (for example..)
thanks..

billhollings:
PR #2639 completes this issue.

PR #2639, opened 2025-10-09
Add CI job for building for macOS with private API.

squidbus:
Resolves https://github.com/KhronosGroup/MoltenVK/issues/2637

Add a CI job, to a) verify build and b) output an artifact, for a macOS build of MoltenVK with Metal private API use enabled.

Verified in my own fork actions that both variants are being built and named as expected, and that the downloads for each have the expected capabilities.


oscarbg:
nice! thanks for fast support!

Issue #2618, opened 2025-10-01
VK_EXT_host_image_copy causes VK_ERROR_OUT_OF_DEVICE_MEMORY on small allocation


haasn:
I get this error when allocating memory for an image that happens to use `VK_IMAGE_USAGE_HOST_TRANSFER_BIT`, via a dedicated memory allocation (`VkMemoryDedicatedAllocateInfoKHR`) from memory type index 1 (host visible).

```
$ ffmpeg -f lavfi -i smptehdbars -vf libplacebo -f null -
...
[libplacebo @ 0x13c80bc00] Allocating 98304 memory of type 0xf (id 1) in heap 0: ../src/utils/upload.c:358
[libplacebo @ 0x13c80bc00] Allocation of size   96K failed: VK_ERROR_OUT_OF_DEVICE_MEMORY!
[libplacebo @ 0x13c80bc00] Memory heaps supported by device:
[libplacebo @ 0x13c80bc00]     0: flags 0x1 size   16G
[libplacebo @ 0x13c80bc00] Memory types supported by device:
[libplacebo @ 0x13c80bc00]     0: flags 0x1 heap 0
[libplacebo @ 0x13c80bc00]     1: flags 0xf heap 0
[libplacebo @ 0x13c80bc00]     2: flags 0x11 heap 0
[libplacebo @ 0x13c80bc00] Memory pool 0:
[libplacebo @ 0x13c80bc00]     Compatible types: 0x3
[libplacebo @ 0x13c80bc00]     Optimal flags: 0x3
[libplacebo @ 0x13c80bc00]     Slab  0:        0 x  100K:  266K used  400K res  400K alloc from heap 0, efficiency 66.50%  [../src/utils/upload.c:246]
[libplacebo @ 0x13c80bc00]     Slab  1:       fc x   36K:   66K used   72K res  288K alloc from heap 0, efficiency 91.67%  [../src/utils/upload.c:246]
[libplacebo @ 0x13c80bc00]     Pool summary:  332K used  472K res  688K alloc, efficiency 70.34%, utilization 68.60%
[libplacebo @ 0x13c80bc00] Memory pool 1:
[libplacebo @ 0x13c80bc00]     Compatible types: 0xffffffff
[libplacebo @ 0x13c80bc00]     Optimal flags: 0x3
[libplacebo @ 0x13c80bc00]     Buffer flags: 0x3
[...]

PR #2640, opened 2025-10-10
Allow host-coherent optimal tiling images on macOS with Apple GPU.

squidbus:
This should fix https://github.com/KhronosGroup/MoltenVK/issues/2618.

Updates the check when creating a dedicated allocation for a host-coherent, optimally-tiled image, to allow macOS with Apple GPUs.

I'll probably do a pass over all of the macOS checks later to see if there's anything else left... just this for now though, to get the issue fixed.

### DIFF
--- a/MoltenVK/MoltenVK/GPUObjects/MVKDeviceMemory.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKDeviceMemory.mm
@@ -402,7 +402,6 @@ MVKDeviceMemory::MVKDeviceMemory(MVKDevice* device,
 
 	// "Dedicated" means this memory can only be used for this image or buffer.
 	if (dedicatedImage) {
-#if MVK_MACOS
 		if (!isAppleGPU() && isMemoryHostCoherent()) {
 			if (!dedicatedImage->_isLinear) {
 				setConfigurationResult(reportError(VK_ERROR_OUT_OF_DEVICE_MEMORY, "vkAllocateMemory(): Host-coherent VkDeviceMemory objects cannot be associated with optimal-tiling images."));
@@ -411,7 +410,6 @@ MVKDeviceMemory::MVKDeviceMemory(MVKDevice* device,
 				setConfigurationResult(reportError(VK_ERROR_OUT_OF_DEVICE_MEMORY, "vkAllocateMemory(): Could not allocate a host-coherent VkDeviceMemory of size %llu bytes. The maximum memory-aligned size of a host-coherent VkDeviceMemory is %llu bytes.", _allocationSize, getMetalFeatures().maxMTLBufferSize));
 			}
 		}
-#endif
         for (auto& memoryBinding : dedicatedImage->_memoryBindings) {
             _imageMemoryBindings.push_back(memoryBinding);
         }


MoltenVK/MoltenVK/GPUObjects/MVKDeviceMemory.mm
```diff
@@ -403,7 +403,7 @@
 	// "Dedicated" means this memory can only be used for this image or buffer.
 	if (dedicatedImage) {
 #if MVK_MACOS
-		if (isMemoryHostCoherent() ) {
+		if (!isAppleGPU() && isMemoryHostCoherent()) {
 			if (!dedicatedImage->_isLinear) {
```
cdavis5e, line 406: This will cause us not to create an `MTLBuffer` to map the memory on Apple Silicon.
```suggestion
		if (isMemoryHostCoherent()) {
			if (!isAppleGPU() && !dedicatedImage->_isLinear) {
```
I'm wondering now if, maybe, the code to ensure that we have the buffer should actually be _outside_ the `#if MVK_MACOS` block...

squidbus: I would assume if this was working on non-macOS it would be working fine here.

cdavis5e: 🤷

squidbus: I'll mark this as a draft and see what testing turns up I guess, maybe there was something missing on other OSes as well. Or if @billhollings has any thoughts.

billhollings: TBH, I've lost track of the logic in this section of code. And I note that your recent conditional-legacy cleanup may have impacted this too (not necessarily in a negative way, but maybe some of this logic is now obsolete).

It seems to be saying now:

1. For macOS IM GPUs where coherent memory is required, either the dedicated image needs to be linear, or it needs to be backed by a MTLBuffer.

2. For Apple GPUs, on any OS, none of that matters. Is this because images are always linear on Apple GPUs?

squidbus: My assumption here is that on Apple GPUs you can just make the image have shared storage, and on IMR it relies on it being linear which in turn can be backed by a MTLBuffer to manage the coherency.

I’ll unmask this as draft, it seems to me like this should be fine.

squidbus: Also as you say, there’s probably some paths that are no longer necessary with the recent clean ups and minimum adjustments. For example, the code to handle 3D compressed textures is Mac only, but Mac currently has native 3D compressed support set to true always, I think. I’ll do a round of cleaning on some of those cases soon.

MoltenVK/MoltenVK/GPUObjects/MVKDeviceMemory.mm
```diff
@@ -403,7 +403,7 @@
 	// "Dedicated" means this memory can only be used for this image or buffer.
 	if (dedicatedImage) {
 #if MVK_MACOS
```
billhollings: `#if MVK_MACOS` is now redundant to `!isAppleGPU()`, and I don't see any code within this guard that can't be compiled on the other platforms (although I grant that it results in unnecessary binary code). 

IMHO, removing conditional compilation helps with understanding the code. But I understand others may feel binary code reduction is preferred.

squidbus: Sure, removed.

PR #2643, opened 2025-10-13
Clean up platform preprocessor guards.

squidbus:
Another round of legacy code cleanup, this time focusing on the platform pre-processor guards:
* Remove `MVK_IOS_OR_TVOS` and `MVK_IOS_OR_TVOS_OR_VISIONOS`.
  * Most cases have been refactored to tailor for specific platforms and then catch all others in an `else`, or use `!MVK_MACOS` where it makes sense.
* Remove some features like 3D compression emulation for macOS, since the native feature is always supported with current targets.
* Make sure that pre-processor guards that were previously intended for Apple GPU differences actually use the proper GPU type check now.
* Refactor `_metalFeatures` population to no longer use platform guards.
  * The logic is now fully based on GPU family support.
  * The Mac-specific features are under `Mac2`, and then it proceeds to all of the `Apple(n)` features for all platforms.
  * This ensures consistency between all OS/device types on what features are enabled by what GPU support level, so that certain platforms don't get accidentally left behind.
* Remove `Mac1` family flag as `supportsMac2` is always set on macOS.

### DIFF
--- a/MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
@@ -95,6 +95,9 @@ uint8_t MVKMTLDeviceCapabilities::getHighestMacGPU() const {
 MVKMTLDeviceCapabilities::MVKMTLDeviceCapabilities(id<MTLDevice> mtlDev) {
 	mvkClear(this);
 	supportsMetal3 = supportsGPUFam(Metal3, mtlDev);
+#if MVK_XCODE_26 && !MVK_OS_SIMULATOR
+	supportsMetal4 = supportsGPUFam(Metal4, mtlDev);
+#endif
 	supportsApple1 = supportsGPUFam(Apple1, mtlDev);
 	supportsApple2 = supportsGPUFam(Apple2, mtlDev);
 	supportsApple3 = supportsGPUFam(Apple3, mtlDev);
@@ -3629,8 +3632,10 @@ void MVKPhysicalDevice::logGPUInfo() {
 	logMsg += "\n\tMetal Shading Language %s";
 	logMsg += "\n\tsupports the following GPU Features:";
 
+	if (supportsMTLGPUFamily(Metal4)) { logMsg += "\n\t\tGPU Family Metal 4"; } else
 	if (supportsMTLGPUFamily(Metal3)) { logMsg += "\n\t\tGPU Family Metal 3"; }
 
+	if (supportsMTLGPUFamily(Apple10)) { logMsg += "\n\t\tGPU Family Apple 10"; } else
 	if (supportsMTLGPUFamily(Apple9)) { logMsg += "\n\t\tGPU Family Apple 9"; } else
 	if (supportsMTLGPUFamily(Apple8)) { logMsg += "\n\t\tGPU Family Apple 8"; } else
 	if (supportsMTLGPUFamily(Apple7)) { logMsg += "\n\t\tGPU Family Apple 7"; } else
--- a/MoltenVK/MoltenVK/GPUObjects/MVKDevice.h
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKDevice.h
@@ -114,6 +114,7 @@ typedef struct MVKMTLDeviceCapabilities {
 	bool supportsApple10;
 	bool supportsMac2;
 	bool supportsMetal3;
+	bool supportsMetal4;
 
 	bool isAppleGPU;
 	bool supportsBCTextureCompression;


MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
```diff
@@ -2685,93 +2625,33 @@
...
-        _metalFeatures.maxSubgroupSize = (_properties.vendorID == kAMDVendorId) ? 64 : 32;
         switch (_properties.vendorID) {
+#if MVK_MACOS
             case kIntelVendorId:
+                _metalFeatures.maxSubgroupSize = 32;
                 _metalFeatures.minSubgroupSize = 8;
                 break;
             case kAMDVendorId:
+                _metalFeatures.maxSubgroupSize = 64;
                 _metalFeatures.minSubgroupSize = isAMDRDNAGPU() ? 32 : _metalFeatures.maxSubgroupSize;
```
billhollings, line 2641: Admittedly, this is pedantic, but the `#if MVK_MACOS` guard here is only to protect `isAMDRDNAGPU()`. We could get rid of the `#if MVK_MACOS` guard here by defining `isAMDRDNAGPU() { return false; }` in the already-existing `#else` section where `initGPUInfoProperties()` is defined. 😬

MoltenVK/MoltenVK/GPUObjects/MVKPixelFormats.mm
```diff
@@ -1074,7 +1065,7 @@
 	if ( !mtlPixFmt) { return; }
 
 	MVKMTLFmtCaps fmtCaps = kMVKMTLFmtCapsNone;
-	if (mtlDevCaps.isAppleGPU && mtlDevCaps.supportsMac1) {
+	if (mtlDevCaps.isAppleGPU && mtlDevCaps.supportsMac2) {
```
billhollings, line 1068: Hmmm...this is a design change.

Testing for `supportsMac1` was a definitive test for macOS, in that it was guaranteed to be set on all macOS devices (like `isAppleGPU = supportsApple1` is for Apple GPU), whereas `supportsMac2` is a test for capabilities. While I understand in our current situation `supportsMac2` is always set, that's more of an environmental happenstance, and potentially might change unexpectedly in the future, who knows?

Maybe what is needed here is a `MVKMTLDeviceCapabilities::isMacGPU`? Although what does that really mean? Or maybe `MVKMTLDeviceCapabilities::isMacOS`? Although now we're mixing GPU & OS.

Or maybe just restore `supportsMac1` to reestablish a definitive "we're on a Mac"?

MoltenVK/MoltenVK/GPUObjects/MVKPixelFormats.mm
```diff
@@ -1384,7 +1375,7 @@
 void MVKPixelFormats::modifyMTLFormatCapabilities(const MVKMTLDeviceCapabilities& gpuCaps) {
 
 	bool noVulkanSupport =  false;		// Indicated supported in Metal but not Vulkan or SPIR-V.
-	bool notMac =  gpuCaps.isAppleGPU && !gpuCaps.supportsMac1;
+	bool notMac =  gpuCaps.isAppleGPU && !gpuCaps.supportsMac2;
```
billhollings, line 1378: Ditto.

MoltenVK/MoltenVK/GPUObjects/MVKPixelFormats.mm
squidbus: I can change this if you prefer, but technically going by the Metal Feature Set Table doc they do list these pixel format capabilities as being for `Mac2` family GPUs, which makes the check for `Mac2` here make sense to me.

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
squidbus: Makes sense, made a stub version of that function to remove the include guard (and rebased).

MoltenVK/MoltenVK/GPUObjects/MVKPixelFormats.mm
billhollings: Okay. I see what you're saying. The Metal Feature Set [docs](https://developer.apple.com/metal/Metal-Feature-Set-Tables.pdf) now _**start**_ at `Mac2`. LGTM then.

MoltenVK/MoltenVK/Commands/MVKCommandPipelineStateFactoryShaderSource.h
```diff
@@ -204,69 +204,6 @@ kernel void cmdResolveColorImage2DIntArray(texture2d_array<int, access::write> d
...
-	decompressDXTnBlock(src, pixels, blockExtent, 4 * sizeof(uint), info.format);
-	device uint* destPixel = (device uint*)dest;
-	for (uint j = 0; j < blockExtent.height; ++j) {
-		for (uint i = 0; i < blockExtent.width; ++i) {
-			destPixel[j * info.destRowStride / sizeof(uint) + i] = pixels[j * 4 + i];
-		}
-	}
-}
-
 #if __METAL_VERSION__ >= 210
```
roadrunnerto100: Not for this MR but this might be able to be removed.

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
```diff
@@ -3777,9 +3630,8 @@ static uint32_t mvkGetEntryProperty(io_registry_entry_t entry, CFStringRef prope
 	logMsg += "\n\tsupports the following GPU Features:";
 
 	if (supportsMTLGPUFamily(Metal3)) { logMsg += "\n\t\tGPU Family Metal 3"; }
-#if MVK_XCODE_15 && (MVK_IOS || MVK_MACOS)
+
```
cdavis5e, line 3637:
```suggestion

#if MVK_XCODE_26
	if (supportsMTLGPUFamily(Apple10)) { logMsg += "\n\t\tGPU Family Apple 10"; } else
#endif
```
I'm amazed this hasn't been done yet.

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
```diff
@@ -3777,9 +3630,8 @@ static uint32_t mvkGetEntryProperty(io_registry_entry_t entry, CFStringRef prope
 	logMsg += "\n\tsupports the following GPU Features:";
 
 	if (supportsMTLGPUFamily(Metal3)) { logMsg += "\n\t\tGPU Family Metal 3"; }
-#if MVK_XCODE_15 && (MVK_IOS || MVK_MACOS)
```
cdavis5e, line 3780: Do the Xcode 14 SDKs have `MTLGPUFamilyApple9` in them?

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
squidbus: The macro here does not actually use the constants, it references the ‘supports(n)’ flags we have. Hence you don’t actually need to guard these. Ditto for the other comment.

MoltenVK/MoltenVK/Commands/MVKCommandPipelineStateFactoryShaderSource.h
squidbus: Removed, thanks.

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
squidbus: Essentially, this just expands to `_gpuCapabilities.supportsApple10` which always exists.

cdavis5e: I mean, though, that the log message for `Apple10` isn't being printed yet. So, the log statement on the inside, just without the `#if` guard, as you pointed out.

squidbus: Ah right… I thought I added it but I guess that was in a previous version of changes. Let me fix that

squidbus: While I was at it I also added `Metal4` detection and logging.

billhollings:
> I'm currently running a full CTS on this, and will report back.

All good. No regressions on full CTS.

oscarbg:
just quick test on Mac 26.0.1 with M4 with the build:
https://github.com/KhronosGroup/MoltenVK/actions/runs/18517449592
still shows "GPU Family Metal 3"..
it's because builders aren't using Xcode 26, right?

```
[mvk-info] GPU device:
	model: Apple M4
	type: Integrated
	vendorID: 0x106b
	deviceID: 0x1a000209
	pipelineCacheUUID: 3E1FD32F-1A00-0209-0000-000100000000
	GPU memory available: 12124 MB
	GPU memory used: 0 MB
	Metal Shading Language 3.2
	supports the following GPU Features:
		GPU Family Metal 3
		GPU Family Apple 9
		GPU Family Mac 2
		Read-Write Texture Tier 2

```

squidbus:
> just quick test on Mac 26.0.1 with M4 with the build: https://github.com/KhronosGroup/MoltenVK/actions/runs/18517449592 still shows "GPU Family Metal 3".. it's because builders aren't using Xcode 26, right?

Yes. We can either explicitly upgrade to macOS 26 runners (which are in beta) or just wait for `macos-latest` to switch over.

oscarbg:
> Yes. We can either explicitly upgrade to macOS 26 runners (which are in beta) or just wait for `macos-latest` to switch over.

thanks..
IMHO I will wait to the switch.. should be in a month or two seems..

PR #2646, opened 2025-10-15
Remove native swizzle and depth clamp simulator limitations.

squidbus:
From the previous PR, `nativeTextureSwizzle` became true always except for simulators. I went and tested on simulator and it seems that at least nowadays both `nativeTextureSwizzle` and `depthClamp` work fine.

So:
* Deprecate `nativeTextureSwizzle` as always true and remove swizzle emulation code.
* Remove logic to set `depthClamp` feature to false on simulator.

It's possible some of the other simulator limitations no longer exist but these are the ones I was able to verify easily.

### DIFF
--- a/MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
@@ -1925,7 +1925,7 @@ id<MTLTexture> MVKImageViewPlane::newMTLTexture() {
     }
 
     id<MTLTexture> texView = nil;
-    if (_useNativeSwizzle) {
+    if (_useSwizzle) {
         texView = [mtlTex newTextureViewWithPixelFormat: _mtlPixFmt
                                             textureType: _imageView->_mtlTextureType
                                                  levels: levelRange
@@ -1966,7 +1966,7 @@ MVKImageViewPlane::MVKImageViewPlane(MVKImageView* imageView,
              ((_imageView->_mtlTextureType == MTLTextureType2D || _imageView->_mtlTextureType == MTLTextureType2DArray) && is3D)) &&
             _imageView->_subresourceRange.levelCount == _imageView->_image->_mipLevels &&
             (is3D || _imageView->_subresourceRange.layerCount == _imageView->_image->_arrayLayers) &&
-            !_useNativeSwizzle) {
+            !_useSwizzle) {
             _useMTLTextureView = false;
         }
     } else {
@@ -1976,8 +1976,7 @@ MVKImageViewPlane::MVKImageViewPlane(MVKImageView* imageView,
 
 VkResult MVKImageViewPlane::initSwizzledMTLPixelFormat(const VkImageViewCreateInfo* pCreateInfo) {
 
-	_useNativeSwizzle = false;
-	_useShaderSwizzle = false;
+	_useSwizzle = false;
 	_componentSwizzle = pCreateInfo->components;
 	VkImageAspectFlags aspectMask = pCreateInfo->subresourceRange.aspectMask;
 
@@ -2074,7 +2073,7 @@ VkResult MVKImageViewPlane::initSwizzledMTLPixelFormat(const VkImageViewCreateIn
 			}
 #endif
 		}
-		
+
 		// 2. Metal's undefined values for depth/stencil sample to vec4 conversion
 		// Due to differences in Metal and Vulkan specification for sampling depth/stencil textures into vec4, we need to
 		// provide the correct mapping from Vulkan to Metal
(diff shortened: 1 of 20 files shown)


MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
```diff
@@ -1955,29 +1947,11 @@ static void signalAndUntrack(const MVKSwapchainSignaler& signaler) {
...
-             ((_imageView->_mtlTextureType == MTLTextureType2D || _imageView->_mtlTextureType == MTLTextureType2DArray) && is3D)) &&
-            _imageView->_subresourceRange.levelCount == _imageView->_image->_mipLevels &&
-            (is3D || _imageView->_subresourceRange.layerCount == _imageView->_image->_arrayLayers) &&
-            !_useNativeSwizzle) {
-            _useMTLTextureView = false;
-        }
-    } else {
-        _useMTLTextureView = false;
-    }
+    _useMTLTextureView = _imageView->_image != nullptr;
```
billhollings: Really? **_None_** of that logic applies anymore? 😲🤨

Hmmm. 🤔

I can see that in the giant `if()`, the `!_useNativeSwizzle` is going to kill it off because we no longer track `_useNativeSwizzle` here. However, `_useNativeSwizzle` isn't always enabled when native swizzling is available (otherwise it would have been called something like `_supportsNativeSwizzle`). 

In `initSwizzledMTLPixelFormat()` below, `enableSwizzle()` is _**not**_ called if the swizzle is default `RGBA` or Identity (per `SWIZZLE_MATCHES(R, G, B, A)`).  So `_useNativeSwizzle` is actually an indicator that a non-Identity swizzle is required.

The goal is that we only want to enable `_useMTLTextureView` (and thereby create a second texture to act as a view on the base `VKImage` texture), if we have to.

So, much of the `if()` logic above should be restored, _**and**_ we need to include some logic here to only enable `_useMTLTextureView` if there is actually a non-Identity swizzle in play. 

Perhaps `_useNativeSwizzle` should be restored? Or perhaps just adding `SWIZZLE_MATCHES(R, G, B, A)` to the giant `if()` here would do the same thing?

MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
```diff
@@ -2083,187 +2057,31 @@ static void signalAndUntrack(const MVKSwapchainSignaler& signaler) {
...
-			} else if (_componentSwizzle.a == VK_COMPONENT_SWIZZLE_G || _componentSwizzle.a == VK_COMPONENT_SWIZZLE_B ) {
-				_componentSwizzle.a = VK_COMPONENT_SWIZZLE_ZERO;
-			}
-			
-			return VK_SUCCESS;
+		if (_componentSwizzle.r == VK_COMPONENT_SWIZZLE_A) {
+			_componentSwizzle.r = VK_COMPONENT_SWIZZLE_ONE;
+		} else if (_componentSwizzle.r == VK_COMPONENT_SWIZZLE_G || _componentSwizzle.r == VK_COMPONENT_SWIZZLE_B) {
+			_componentSwizzle.r = VK_COMPONENT_SWIZZLE_ZERO;
 		}
```
billhollings, line 2089: What happened to the logic setting the other three components? All components need to be defined, per the comments above, and the CTS tests mentioned in the commit that defined them.

MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
```diff
@@ -2083,187 +2057,31 @@ static void signalAndUntrack(const MVKSwapchainSignaler& signaler) {
...
-			MVKAssert(0, "Image without PixelFormatView usage couldn't enable swizzling!");
+		if (_componentSwizzle.b == VK_COMPONENT_SWIZZLE_A) {
+			_componentSwizzle.b = VK_COMPONENT_SWIZZLE_ONE;
+		} else if (_componentSwizzle.b == VK_COMPONENT_SWIZZLE_G || _componentSwizzle.b == VK_COMPONENT_SWIZZLE_B || _componentSwizzle.b == VK_COMPONENT_SWIZZLE_IDENTITY) {
+			_componentSwizzle.b = VK_COMPONENT_SWIZZLE_ZERO;
+		}
+		if (_componentSwizzle.a == VK_COMPONENT_SWIZZLE_A || _componentSwizzle.a == VK_COMPONENT_SWIZZLE_IDENTITY) {
+			_componentSwizzle.a = VK_COMPONENT_SWIZZLE_ONE;
+		} else if (_componentSwizzle.a == VK_COMPONENT_SWIZZLE_G || _componentSwizzle.a == VK_COMPONENT_SWIZZLE_B ) {
+			_componentSwizzle.a = VK_COMPONENT_SWIZZLE_ZERO;
```
billhollings, line 2103: Oh. Geez. Are these the missing depth components from my comment above? 

If so, what the hell, Git diff? 🤨

MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
```diff
@@ -2083,187 +2057,31 @@ static void signalAndUntrack(const MVKSwapchainSignaler& signaler) {
...
-
-	// No format transformation swizzles were found, so we'll need to use either native or shader swizzling, if supported.
-	if ( !enableSwizzling() ) {
-		return getVulkanAPIObject()->reportError(VK_ERROR_FEATURE_NOT_PRESENT,
-												 "The value of %s::components) (%s, %s, %s, %s), when applied to a VkImageView, requires full component swizzling to be enabled both at the"
-												 " time when the VkImageView is created and at the time any pipeline that uses that VkImageView is compiled. Full component swizzling can"
-												 " be enabled via the MVKConfiguration::fullImageViewSwizzle config parameter or MVK_CONFIG_FULL_IMAGE_VIEW_SWIZZLE environment variable.",
-												 pCreateInfo->image ? "vkCreateImageView(VkImageViewCreateInfo" : "vkGetPhysicalDeviceImageFormatProperties2KHR(VkPhysicalDeviceImageViewSupportEXTX",
-												 mvkVkComponentSwizzleName(_componentSwizzle.r), mvkVkComponentSwizzleName(_componentSwizzle.g),
-												 mvkVkComponentSwizzleName(_componentSwizzle.b), mvkVkComponentSwizzleName(_componentSwizzle.a));
```
billhollings, line 2254: Okay. So I take it from this, and the other large pink section above, that we can now abandon all the special-case magic we were doing here to substitute one pixel format for another, and just let native swizzling do its job? 🚀

Sounds good, but see my other comments about the impact on `_useMTLTextureView`.

billhollings (changes requested):
Thanks for submitting this! Another round of solid clean up! 🚀

For the most part, LGTM, but the code cleanup is too aggressive in its impact on `_useMTLTextureView`, and will result in a separate texture view being created every time. See me inline comments for how to potentially fix this.

MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
squidbus: Yeah I see the issue here now, I can add a swizzle check here and I believe it should accomplish the same thing, as you say.

MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
squidbus: Yes they are, the diff looks weird because of a combination of the removed bits after and the indentation being reduced.

MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
squidbus: That's the plan yes, instead of doing complex logic to check for alternate matching formats, etc we can just rely on the native hardware swizzle.

squidbus:
To address the comments, I've re-added a `_useSwizzle` to replace the previous `_useNativeSwizzle` and `_useShaderSwizzle` to track whether swizzling is needed at all, and restored a number of internal logic paths using that new variable instead.

squidbus:
Also, this passes `dEQP-VK.pipeline.monolithic.image_view.*`:
```
Test run totals:
  Passed:        8958/13584 (65.9%)
  Failed:        0/13584 (0.0%)
  Not supported: 4626/13584 (34.1%)
  Warnings:      0/13584 (0.0%)
  Waived:        0/13584 (0.0%)
```

SRSaunders:
I am seeing a new runtime error with this PR.  Running on an AMD 6600 / x86_64 Mac using Ventura 13.7.8:

_[MVK-BUILD-ERROR] Cereal serialization Archive for mvk::SPIRVToMSLConversionResultInfo is not completely defined. Missing 1 bytes. Struct size is 104 (including an expected 41 bytes of padding) and Cereal Archive size is 62. The Cereal Archive definition may be missing members._

The application (RBDoom3BFG) continues to run properly, but perhaps something was missed in the changes here.  I  have made sure SPIRV-Cross is up to date via a fetchDependencies build.

squidbus:
> I am seeing a new runtime error with this PR. Running on an AMD 6600 / x86_64 Mac using Ventura 13.7.8:
> 
> _[MVK-BUILD-ERROR] Cereal serialization Archive for mvk::SPIRVToMSLConversionResultInfo is not completely defined. Missing 1 bytes. Struct size is 104 (including an expected 41 bytes of padding) and Cereal Archive size is 62. The Cereal Archive definition may be missing members._
> 
> The application (RBDoom3BFG) continues to run properly, but perhaps something was missed in the changes here. I have made sure SPIRV-Cross is up to date via a fetchDependencies build.

This should be fixed now, it seems removing the flag for the swizzle buffer required an extra padding byte.

SRSaunders:
Fixed now.  Thanks.

Issue #2565, opened 2025-06-13
Residency sets causing corruption


kickblade:
Platform: MacBook Air M2 with macOS 15.5 (24F74)

I have some code that loops through a set of framebuffers and resource sets and kicks off a fullscreen draw pass as follows:
```
unsigned int frameBufferCount = 3;
for( size_t i = 1; i < frameBufferCount; ++I )
{
	encoder.Bind( pipeline );
	encoder.Bind( resourceSets[i-1] );
	encoder.Bind( framebuffers[i] );
	encoder.Bind( vertexBuffers, vertexBufferCount );
	encoder.SetViewport( 0, 0, frameBufferWidth, frameBufferHeight );
	encoder.SetScissor( 0, 0, frameBufferWidth, frameBufferHeight );
	encoder.Draw( indexBuffer, indexType, indexCount );
	encoder.Bind( nullptr );
}
```
This was all working and outputting the same image as an actual android device prior to this git commit: c3f40eea3d7991657b9b10e1e01a88dea4340a7a - "Put resources in residency sets"

However, it now causes random corruption every frame at runtime.

Android Reference - Ground Truth
![Image](https://github.com/user-attachments/assets/1ef7505c-55f8-4a33-b015-654967a09092)

MacOS - Before residency sets
<img width="797" alt="Image" src="https://github.com/user-attachments/assets/24ac8cfa-9a09-4ce1-bc2a-7036d2dad865" />

MacOS - After residency sets
<img width="797" alt="Image" src="https://github.com/user-attachments/assets/b3b374a1-7be2-40cb-93d0-c702fe8c966f" />
<img width="798" alt="Image" src="https://github.com/user-attachments/assets/d9572387-3d1e-4383-8167-60391f67bb97" />

The residency sets commit is causing random squares (corruption) to appear. I am able to fix it by hard setting `_metalFeatures.residencySets` to be false and building the libs manually. I can also fix it on latest code by doing the same thing.

Residency sets seem to be ok in this kind of loop if the framebuffer count is <= 2 but any more than that and I see corruption.

dseif0x:
[...]

PR #2649, opened 2025-10-17
Update minimum CI to macOS 14 and minimum build to Xcode 15.

squidbus:
Didn't quite expect to be updating this again already, but GitHub sent out an email that starting in November, they will be doing `macos-13` CI brown-outs, followed by full removal in December. The new minimum is `macos-14`.

Thus, we need to switch the minimum CI build over to macOS 14. Consequently, the minimum Xcode becomes 15.0.1. The minimum OS target remains macOS 11.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -30,8 +30,8 @@ jobs:
             select_xcode: false
             private_api: true
           # Legacy build. Up to 3 versions behind latest (or beta).
-          - os: "macos-13"
-            xcode: "14.3.1"
+          - os: "macos-14"
+            xcode: "15.0.1"
             platform: "macos"
             upload_artifacts: false
             select_xcode: true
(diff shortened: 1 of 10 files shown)


squidbus:
> While looking at the issue #2565 and the referenced commit [c3f40ee](https://github.com/KhronosGroup/MoltenVK/commit/c3f40eea3d7991657b9b10e1e01a88dea4340a7a) a lot of code is ifdef MKV_XCODE_16 . Can this be causing issues

I'm not sure why you would put this comment here, or why those would be connected as the release builds are done with Xcode 16 and residency sets are guarded with OS version checks.

rcaridade145:
"why those would be connected as the release builds are done with Xcode 16 " - Wasn't aware sorry. Will delete that comment.

.github/workflows/CI.yml
```diff
@@ -30,8 +30,8 @@ jobs:
             select_xcode: false
             private_api: true
           # Legacy build. Up to 3 versions behind latest (or beta).
-          - os: "macos-13"
-            xcode: "14.3.1"
+          - os: "macos-14"
+            xcode: "15.0.1"
```
billhollings, line 34: Not an issue. I'm just curious why this is 15.0.1 and not 15.0? Is there a relevant difference, or is this just what Github CI lists as supported?

squidbus: This is what the GitHub specifies as the minimum available in the image yes.

PR #2651, opened 2025-10-19
Misc CMake improvements

m154k1:
Hi, I'd like to propose 4 changes in addition to #2513:

1. 69fd6956606374ab935079aa426ae7f8c6913765 - adds `MVK_USE_METAL_PRIVATE_API` option similar to `MVK_CONFIG_LOG_LEVEL`
2. ed15e45d5eae005f4532dc8e0b7424c06ae2dcec - installs ICD json file which is used for driver discovery by [Vulkan Loader](https://github.com/KhronosGroup/Vulkan-Loader)
3. 0e3fdf09e45e95e297a8a22cc4fad52b3cf2a5dc - changes pkgconfig install path to `lib` instead of `share` which is more common
4. 9f1c7d7475f1984c48ec1ac1ec3f3a645498263b - adds an option to skip building the shader converter tool (useful if you want to build the library only)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -67,6 +67,8 @@ set_property(CACHE MVK_CONFIG_LOG_LEVEL PROPERTY STRINGS "debug" "info" "warn" "
 
 option(MVK_USE_METAL_PRIVATE_API "If enabled, MoltenVK will use private interfaces exposed by Metal to implement Vulkan features that are difficult to support otherwise." OFF)
 
+option(MVK_BUILD_SHADER_CONVERTER_TOOL "If enabled, the MoltenVKShaderConverter executable will be built." OFF)
+
 # Set default minimum C++ standard
 if(MOLTEN_VK_TOPLEVEL_PROJECT)
 	set(CMAKE_CXX_STANDARD 17)
@@ -102,7 +104,9 @@ add_subdirectory("MoltenVKShaderConverter/MoltenVKShaderConverter")
 add_subdirectory("MoltenVK")
 
 ## MoltenVKShaderConverter
-add_subdirectory("MoltenVKShaderConverter/MoltenVKShaderConverterTool")
+if(MVK_BUILD_SHADER_CONVERTER_TOOL)
+    add_subdirectory("MoltenVKShaderConverter/MoltenVKShaderConverterTool")
+endif()
 
 ################################################################################
 # Install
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -65,6 +65,10 @@ project(MoltenVK
 set(MVK_CONFIG_LOG_LEVEL "info" CACHE STRING "Set the default log level for MoltenVK. Options are: debug, info, warn, error, off. Default is 'info'.")
 set_property(CACHE MVK_CONFIG_LOG_LEVEL PROPERTY STRINGS "debug" "info" "warn" "error" "off")
 
+option(MVK_USE_METAL_PRIVATE_API "If enabled, MoltenVK will use private interfaces exposed by Metal to implement Vulkan features that are difficult to support otherwise." OFF)
+
+option(BUILD_SHADER_CONVERTER_TOOL "If enabled, the MoltenVKShaderConverter executable will be built." ON)
```
squidbus: Wondering if this should be off by default since its a less common use case.

squidbus: Also should probably be given an `MVK_` option prefix.

m154k1: Right, good point. Updated, thanks!

squidbus (approved):

jcm93:
Hi, just a small question: Is `etc/vulkan/icd.d/` defined by the spec anywhere as the appropriate location within a prefix for ICD JSON files? Mesa seems to use `share/vulkan/icd.d/` when `install`ing, and ideally I guess there would be consistency.

m154k1:
> Hi, just a small question: Is `etc/vulkan/icd.d/` defined by the spec anywhere as the appropriate location within a prefix for ICD JSON files?

Hey, yes, you can find it [here](https://github.com/KhronosGroup/Vulkan-Loader/blob/main/docs/LoaderDriverInterface.md#driver-discovery-on-linux).

> Mesa seems to use `share/vulkan/icd.d/` when `install`ing, and ideally I guess there would be consistency.

`share` is a valid prefix too, and it's more common, but the downside of it is that driver discovery won't work when `XDG_DATA_DIRS` env variable is set. You can find more details [here](https://github.com/mpv-player/mpv/issues/15642#issuecomment-2571432029).

Mesa has an option to change the installation dir (`vulkan-icd-dir`), and ideally MoltenVK should have it too.

PR #2653, opened 2025-10-20
Add CMake options for dependency exclusions.

squidbus:
Adds the `MVK_EXCLUDE_SPIRV_TOOLS` and `MVK_EXCLUDE_CEREAL` compile options to CMake.

Note that in code the cereal flag is `MVK_USE_CEREAL`, but I opted to use `MVK_EXCLUDE_CEREAL` as the CMake option for consistency, while preserving the existing code flag name for backwards compatibility.

Additionally, set `CPM_USE_NAMED_CACHE_DIRECTORIES` which will help certain IDEs to display the dependencies by name instead of a long cache hash. This option places externals inside project-named directories inside their hash-named directories, instead of just in a hash directory.


billhollings (approved):
Thanks for submitting. LGTM. Once the conflicts are fixed, I'll pull this in.

squidbus:
> Thanks for submitting. LGTM. Once the conflicts are fixed, I'll pull this in.

Fixed the conflicts.

billhollings:
@squidbus  Looks like the CMake CI builds are breaking now.

squidbus:
> Looks like the CMake CI builds are breaking now.

It seems to be having the same issue on the main branch as well now. Some error with checking out SPIRV-Cross?

squidbus:
Seems like it might be gone now looking at the latest main build, maybe some temporary issue checking out from GitHub? Very strange, I rebased this again just in case.

billhollings:
> Seems like it might be gone now looking at the latest main build, maybe some temporary issue checking out from GitHub? Very strange, I rebased this again just in case.

Uggh. In that case it might be a GitHub CI caching issue. Let's see if it persists over time, or was just transient today.

PR #2659, opened 2025-10-30
Bounds check pipeline bind ops

### LINKED ISSUE #2645: Merge pull request [#2572], which caused numerous crashes due to invalid references in samplers
Merge pull request [#2572](https://github.com/KhronosGroup/MoltenVK/issues/2572) from etang-cw/DirtyTracking
New state tracker and descriptor set implementation

which caused numerous crashes due to invalid references in samplers

etang-cw:
Fixes #2645

Some time we should implement argument buffer push descriptors, but for now, this will at least prevent us from crashing if a push descriptor goes over our bind limits.


cdavis5e:
> Some time we should implement argument buffer push descriptors

Doesn't that defeat the purpose of push descriptors?

etang-cw:
> Doesn't that defeat the purpose of push descriptors?

Do you have a better solution?  Right now an application can stay within all our stated limits and still have bindings go missing.  And yes, I realize that this isn't the only place that our actual limits are lower than what we advertise.  All those places should be fixed eventually.

PR #2667, opened 2025-11-26
Fix buffer and heap out-of-sync in initExternalMemory()

osy:
If a buffer is created before a heap in MVKDeviceMemory, the buffer would
be used in some places (for example `MVKDeviceMemory::map` and the heap
would be used in other places (for example `MVKBuffer::getMTLBuffer()`).
This leads to inconsistent memory mappings.

Specifically, there are two cases when this happens:
1. When `vkAllocateMemory` is called with an imported buffer.
2. When `vkAllocateMemory` expects `HANDLE_TYPE_MTLBUFFER` to be exported.

In either case, `ensureMTLBuffer()` gets called before `ensureMTLHeap()`
which sets `_mtlBuffer` to a buffer not allocated from the placement heap.

This change fixes this by making sure that if a placement heap is needed,
we will always create it prior to allocating the buffer. In cases where
the heap is not used at all (imported buffer), we will return error if the
user tries to specify a `HANDLE_TYPE_MTLHEAP` for export.

### DIFF
--- a/MoltenVK/MoltenVK/GPUObjects/MVKDeviceMemory.h
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKDeviceMemory.h
@@ -160,7 +160,7 @@ protected:
 	bool ensureHostMemory();
 	void freeHostMemory();
 	MVKResource* getDedicatedResource();
-	void initExternalMemory(MVKImage* dedicatedImage);
+	void initExternalMemory(MVKImage* dedicatedImage, bool wantsHeap);
 
 	MVKSmallVector<MVKBuffer*, 4> _buffers;
 	MVKSmallVector<MVKImageMemoryBinding*, 4> _imageMemoryBindings;
--- a/MoltenVK/MoltenVK/GPUObjects/MVKDeviceMemory.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKDeviceMemory.mm
@@ -196,6 +196,9 @@ bool MVKDeviceMemory::ensureMTLHeap() {
 	// Can't create a MTLHeap on imported memory
 	if (_isHostMemImported) { return true; }
 
+	// Can't create a MTLHeap if we already have a _mtlBuffer
+	if (_mtlBuffer) { return true; }
+
 	// Don't bother if we don't have placement heaps.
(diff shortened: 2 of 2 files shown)


MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
```diff
@@ -47,10 +47,10 @@
         MVKImageMemoryBinding* memoryBinding = getMemoryBinding();
 		MVKDeviceMemory* dvcMem = memoryBinding->_deviceMemory;
 
-        if (_image->_is2DViewOn3DImageCompatible && !dvcMem->ensureMTLHeap()) {
+        if (_image->_is2DViewOn3DImageCompatible && !dvcMem->getMTLHeap()) {
             MVKAssert(0, "Creating a 2D view of a 3D texture currently requires a placement heap, which is not available.");
```
billhollings, line 51: Hmmm...looks like this needs a bit more work for robustness. This update causes some regression.

This CTS test (among others) now triggers on this assertion, where it didn't before:

`dEQP-VK.pipeline.monolithic.render_to_image.dedicated_allocation.3d.small.r8g8b8a8_unorm` 

Screenshot 2025-11-30 at 5 22 50 PM

osy: I believe that is an existing bug that's now being triggered due to the original check being incorrect:

```
if (_image->_is2DViewOn3DImageCompatible && !dvcMem->ensureMTLHeap()) {
```

`ensureMTLHeap()` will always return true here regardless if a heap exists. This is counter to what the assertion requires.

So there is another underlying bug that this commit exposes. However, I am not familiar with this part of the code to say confidently what it is. Do you think it makes sense if we drop the second commit and I'll file an issue on it instead?

osy: I went ahead and dropped the commit and filed https://github.com/KhronosGroup/MoltenVK/issues/2668

billhollings: Thanks. This no longer crashes those CTS tests.

PR #2681, opened 2025-12-30
Fix desync in primitive restart enable state tracking.

### LINKED ISSUE #2680: Private API: vertex explosions with primitive restart
RPCS3, and Ryujinx emulators exhibit [vertex explosions](https://github.com/RPCS3/rpcs3/issues/17913) with private API enabled, caused by [support for disabling primitive restart using private API](https://github.com/KhronosGroup/MoltenVK/pull/2632).

If I change [this line](https://github.com/KhronosGroup/MoltenVK/blob/bf60f60d0805e9b3ab989cfc1b69f780f96fd123/MoltenVK/MoltenVK/Commands/MVKCommandEncoderState.mm#L1321) to `#if 0`(forcing primitive restart to always be enabled), the issue is fixed.

I captured [2 frames](https://drive.google.com/file/d/16J2midqvENMeZ_2VFZuRLnVtNAa1bQSr/view?usp=share_link) (1.97 GB) using `gfxreconstruct` in RPCS3 on macOS 26.2.

I have only M1 device, so I cannot test with other devices.

https://github.com/user-attachments/assets/b94fb00a-b9f1-4327-a78a-da97687325c6

https://github.com/user-attachments/assets/c326f7b7-558b-46fc-a558-0d307e5a9b1a

squidbus:
* Make `PrimitiveRestartEnable` update logic more like the other adjacent enable + value state tracking logic.
  * Fixes an issue where the tracked enable flag would be flipped even if only the restart index had changed, causing the enable flag to desync.
* Fix `PrimitiveRestartEnable` flag not being set on reset.
  * With the fixed update logic, since Metal has primitive restart by default, this could lead to restart being enabled instead of disabled for first state after reset.
* Explicitly mark `PrimitiveRestartEnable` dirty when an index buffer is bound, to make sure we always check for changed index type and update the reset index if needed.
  * I don't remember if this one could actually cause issues or if some other logic was causing it to always refresh here, but it's good to be explicit just in case.

Fixes https://github.com/KhronosGroup/MoltenVK/issues/2680

CTS:
```
./deqp-vk --deqp-case="dEQP-VK.pipeline.monolithic.input_assembly.primitive_restart.*"

Test run totals:
  Passed:        95/166 (57.2%)
  Failed:        0/166 (0.0%)
  Not supported: 71/166 (42.8%)
  Warnings:      0/166 (0.0%)
  Waived:        0/166 (0.0%)
```

### DIFF
--- a/MoltenVK/MoltenVK/Commands/MVKCommandEncoderState.mm
+++ b/MoltenVK/MoltenVK/Commands/MVKCommandEncoderState.mm
@@ -1032,8 +1032,11 @@ void MVKMetalGraphicsCommandEncoderState::reset(VkSampleCountFlags sampleCount) 
 	memset(static_cast<MVKMetalGraphicsCommandEncoderStateQuickReset*>(this), 0, offsetof(MVKMetalGraphicsCommandEncoderStateQuickReset, MEMSET_RESET_LINE));
 	_lineWidth = 1;
 	_sampleCount = getSampleCount(sampleCount);
-	_primitiveRestartIndex = 0xFFFFFFFF;
 	_depthStencil.reset();
+
+	// Enabled by default in Metal.
+	_flags.set(MVKMetalRenderEncoderStateFlag::PrimitiveRestartEnable, true);
+	_primitiveRestartIndex = 0xFFFFFFFF;
 }
 
 void MVKMetalGraphicsCommandEncoderState::bindFragmentBuffer(id<MTLRenderCommandEncoder> encoder, id<MTLBuffer> buffer, VkDeviceSize offset, NSUInteger index) {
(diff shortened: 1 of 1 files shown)


V380-Ori:
After testing in RPCS3 and Ryujinx emulators, can confirm that this PR fixes the issue!

Again, thank you!

billhollings (approved):
Thanks for submitting. LGTM.

squidbus:
@billhollings I noticed you no longer have a green checkmark for reviews, are you unable to merge this?

billhollings:
> @billhollings I noticed you no longer have a green checkmark for reviews, are you unable to merge this?

Temporarily unable. My account is currently being modified because of changes at work. Should be back to normal in a couple of days.

johnfaizzjamaludin (commented):
@Plagman

PR #2693, opened 2026-02-22
WIP: Fix semantics of VK_KHR_present_wait.

kleinerm:
The VK_KHR_present_wait spec states that the wait of vkWaitForPresent() should end as close as possible to the point in time when the presented image for a presentID becomes visible to the user ("first pixel visible").

This can be achived by calling _swapchain->notifyPresentComplete() from MVKPresentableSwapchainImage::endPresentation(), ie from the Metal drawables presentCompleted handler, after Metal signals present completion.

Right now, _swapchain->notifyPresentComplete() is called from the Metal command buffers completedHandler(), ie. when the command buffer scheduling the present completes. This typically happens multiple milliseconds before actual present, and in case of presents scheduled via VK_GOOGLE_DISPLAY_TIMING far in the future, there can be a long time between completedHandler() and actual present completion.

This commit moves the _swapchain->notifyPresentComplete() to the more appropriate place, and behind _swapchain->endPresentation(), so if a client returns from a call to vkWaitForPresent(), potentially queried VK_GOOGLE_DISPLAY_TIMING presentation timestamps are already available for the client to fetch.


squidbus:
This was explicitly moved in https://github.com/KhronosGroup/MoltenVK/pull/2543 because the place where you have it now has significant unpredictable latency and/or missed callbacks compared to actual present rate, and as a result using present wait even with a few frames acceptable delay would significantly drop FPS.

SRSaunders:
I would like to second what @squidbus is saying here.  Using the internal Optik profiler within the RBDoom3BFG game, I see the following differences when observing **VK_GOOGLE_display_timing** `presentTiming.actualPresentTime`.

Look at the horizontal green _Present trace_, and also the green _Present function chart_ in the lower left corner.  You will see that the released code shows rock-solid present timing (in this case 16.771ms or 60 Hz), while this PR causes highly variable, unpredictable recording of present times.

Current MoltenVK 1.4.1 release from the 1.4.341.0 SDK:
Screenshot 2026-02-24 at 1 10 53 PM

Head of stream MoltenVK 1.4.2 with this PR applied:
Screenshot 2026-02-24 at 1 06 28 PM

billhollings:
I've marked this as WIP until the issues around inconsistent timing that would result from this PR are resolved, or until it is firmly decided that the change in the PR will not be an improvement.

PR #2701, opened 2026-03-20
WIP: Enable depthBounds, shaderCull on all Apple GPUs on macOS/iOS 26+

### LINKED ISSUE #1909: DOOM (2016) Fails to Open
```
[mvk-error] VK_ERROR_FEATURE_NOT_PRESENT: vkCreateDevice(): Requested physical device feature specified by the 15th flag in VkPhysicalDeviceFeatures is not available on this device.
[mvk-error] VK_ERROR_FEATURE_NOT_PRESENT: vkCreateDevice(): Requested physical device feature specified by the 39th flag in VkPhysicalDeviceFeatures is not available on this device.
```

voidstarone:
depthBoundsTest was previously restricted to Apple10 (M4+), but Metal's setDepthTestMinBound:maxBound: API is available on all Apple GPU families when running macOS 26+ / iOS 26+. Expand the feature flag accordingly.

shaderCullDistance was never reported as supported. SPIRV-Cross emits CullDistance built-ins as [[user(cullN)]] varyings which pass vertex to fragment correctly, providing sufficient emulation. Report the feature as supported with a limit of 8, matching maxClipDistances.

Fixes vkCreateDevice failure (VK_ERROR_FEATURE_NOT_PRESENT) for apps that require these features, including DOOM (2016). Resolves #1909.

### DIFF
--- a/ExternalDependencies.xcodeproj/TESTING.md
+++ b/ExternalDependencies.xcodeproj/TESTING.md
@@ -1,0 +1,123 @@
+# Building and Testing MoltenVK from Fork
+
+## Prerequisites
+
+- Xcode 26 beta installed (required for macOS 26 SDK)
+- Command Line Tools matching your Xcode version
+
+If you get errors during build about missing SDKs, make sure Xcode 26 is selected:
+
+```bash
+sudo xcode-select -s /Applications/Xcode-beta.app
+```
+
+---
+
+## 1. Clone
+
+```bash
+git clone git@github.com:voidstarone/MoltenVK.git
+cd MoltenVK
+git checkout fix-depth-bounds-cull-distance
+```
+
+---
+
+## 2. Fetch external dependencies
+
+This clones SPIRV-Cross, SPIRV-Tools, Vulkan-Headers, and others. Takes a few minutes.
+
+```bash
+./fetchDependencies --macos
+```
+
+---
+
+## 3. Build
+
+```bash
+make macos
+```
+
+Output will be at:
+
+```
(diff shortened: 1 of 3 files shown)


squidbus:
Have you tested these? Preferably with the relevant Vulkan CTS tests.

* For depthBoundsTest, I tested this on M4-series when macOS 26 came out and it does not work, it will assert and crash as it is not implemented in the driver. You can check the [Metal Feature Set Tables](https://developer.apple.com/metal/Metal-Feature-Set-Tables.pdf), it's marked as Apple10 only; which is M5-series by the way, not M4.
* For shaderCullDistance, it requires more than just passing the value to fragment shaders, you also need to actually perform the cull distance testing and discard, which to my knowledge is not implemented in SPIRV-Cross. Maybe DOOM (2016) appears to work fine without it, but that depends on how they use it and cannot be used as a benchmark for whether the feature is properly supported by MoltenVK or not.

billhollings:
Marked as WIP until further results available.

K0bin:
This reports support for `depthBounds` and `shaderCullDistance` but doesn't actually implement those features.

`depthBounds` can actually be implemented on HW that supports the feature (M5) but this PR doesn't do that.

squidbus:
> `depthBounds` can actually be implemented on HW that supports the feature (M5) but this PR doesn't do that.

Already been done

squidbus:
I'm going to close this out as it appears to be invalid; `depthBounds` has been tested for sure to not work on M4 or below, and `shaderCullDistance` emulation is not currently provided in SPIRV-Cross.

PR #2704, opened 2026-03-29
Include Metal constant buffer alignment in argument buffer gpuAlign calc

SRSaunders:
Provisionally fixes #2699.

This is a ~~draft~~ fix to avoid Xcode's Metal API Validation failures for non-arm64 targets, more specifically when running on the _iOS Simulator_, or on an _x86_64_ machine.

As stated in the associated PR above,  while this fix works I am not sure if this is the correct way to fix the problem.  In particular I am not sure if taking the worst case alignment for constant buffers (256 bytes on iOS Simulator, or 32 bytes on x86_64 vs. 4 bytes for arm64) vs arg buffer alignments for other types (e.g. textures, samplers, and regular buffers which are all typically 8 bytes) will lead to inefficient memory use within Argument Buffers.

I have decided to post this ~~draft~~ fix anyways, if only to spark conversation about the issue and how to fix correctly.

### DIFF
--- a/MoltenVK/MoltenVK/GPUObjects/MVKDescriptorSet.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKDescriptorSet.mm
@@ -1900,6 +1900,7 @@ MVKDescriptorPool* MVKDescriptorPool::Create(MVKDevice* device, const VkDescript
 	const MVKPhysicalDeviceArgumentBufferSizes& sizes = device->getPhysicalDevice()->getArgumentBufferSizes();
 	uint32_t gpuAlign = std::max(std::max<uint32_t>(sizes.texture.align, sizes.sampler.align), std::max<uint32_t>(sizes.pointer.align, 1));
 	uint32_t cbufAlign = std::max<uint32_t>(gpuAlign, sizes.cbuffer.align);
+	uint32_t dataAlign = gpuAlign;
 	uint32_t cpuAlign = alignof(id);
 	uint32_t gpuSize = 0;
 	uint32_t cpuSize = 0;
@@ -1907,7 +1908,6 @@ MVKDescriptorPool* MVKDescriptorPool::Create(MVKDevice* device, const VkDescript
 	uint32_t numAuxOffset = 0;
 	uint32_t inlineUniformSize = 0;
 	bool usesAuxBuffer = false;
-	bool usesTexCombined = false;
 
 	for (const auto& pool : pools) {
 		if (pool.type == VK_DESCRIPTOR_TYPE_INLINE_UNIFORM_BLOCK) {
@@ -1918,14 +1918,7 @@ MVKDescriptorPool* MVKDescriptorPool::Create(MVKDevice* device, const VkDescript
 		MVKDescriptorCPULayout cpu = pickCPULayout(pool.type, 1, argBufMode, device);
 		numElem += descriptorGPUBindingCount(gpu) * pool.descriptorCount;
 		cpuSize += alignDescriptorOffset(descriptorCPUSize(cpu), cpuAlign) * pool.descriptorCount;
-		if (gpu == MVKDescriptorGPULayout::TexBufSoA || gpu == MVKDescriptorGPULayout::TexSampSoA ||
-			gpu == MVKDescriptorGPULayout::Tex2SampSoA || gpu == MVKDescriptorGPULayout::Tex3SampSoA) {
-			// Combined textures use Metal Constant Buffer alignment
-			gpuSize += alignDescriptorOffset(maxGPUSize(gpu, sizes), cbufAlign) * pool.descriptorCount;
-			usesTexCombined = true;
-		}
-		else
-			gpuSize += alignDescriptorOffset(maxGPUSize(gpu, sizes), gpuAlign) * pool.descriptorCount;
+		gpuSize += alignDescriptorOffset(maxGPUSize(gpu, sizes), gpuAlign) * pool.descriptorCount;
 		numAuxOffset += needsAuxOffset(gpu) ? pool.descriptorCount : 0;
 		usesAuxBuffer |= gpu == MVKDescriptorGPULayout::BufferAuxSize;
 	}
@@ -1945,10 +1938,10 @@ MVKDescriptorPool* MVKDescriptorPool::Create(MVKDevice* device, const VkDescript
 				// Add space for the pointers
 				gpuSize += alignDescriptorOffset(sizes.pointer.size, gpuAlign) * info->maxInlineUniformBlockBindings;
 				// Outlined data is 16-byte aligned
-				gpuAlign = std::max(gpuAlign, 16u);
+				dataAlign = std::max(gpuAlign, 16u);
 			}
 			if (gpuLayout != MVKDescriptorGPULayout::None)
-				gpuSize += calcGroupSizeWithPadding(inlineUniformSize, info->maxInlineUniformBlockBindings, 4, gpuAlign);
+				gpuSize += calcGroupSizeWithPadding(inlineUniformSize, info->maxInlineUniformBlockBindings, 4, dataAlign);
 		}
 	}
 
@@ -1958,14 +1951,12 @@ MVKDescriptorPool* MVKDescriptorPool::Create(MVKDevice* device, const VkDescript
 		gpuSize += alignDescriptorOffset(maxGPUSize(MVKDescriptorGPULayout::Buffer, sizes), gpuAlign) * pCreateInfo->maxSets;
 		// The aux buffer is sized relative to the number of total elements in a descriptor, rather than the number of bindings with aux buffers
 		uint32_t size = numElem * sizeof(uint32_t);
-		// Aux buffers use Metal Constant Buffer alignment
-		gpuAlign = cbufAlign;
-		gpuSize += calcGroupSizeWithPadding(size, pCreateInfo->maxSets, alignof(uint32_t), gpuAlign);
+		// Aux buffers require Metal constant buffer alignment padding between descriptor sets
+		gpuSize += calcGroupSizeWithPadding(size, pCreateInfo->maxSets, alignof(uint32_t), cbufAlign);
 	}
-	else if (usesTexCombined) {
-		// Combined textures use Metal Constant Buffer alignment
-		gpuAlign = cbufAlign;
-	}
+
+	// Set overall gpu alignment incorporating Metal constant buffer and inline uniform data alignment
+	gpuAlign = std::max(cbufAlign, dataAlign);
 
 	bool hostOnly = mvkIsAnyFlagEnabled(pCreateInfo->flags, VK_DESCRIPTOR_POOL_CREATE_HOST_ONLY_BIT_EXT) && argBufMode != MVKArgumentBufferMode::ArgEncoder;
 	void* cpuBuffer;
--- a/MoltenVK/MoltenVK/GPUObjects/MVKDescriptorSet.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKDescriptorSet.mm
@@ -1923,6 +1923,9 @@ MVKDescriptorPool* MVKDescriptorPool::Create(MVKDevice* device, const VkDescript
 		usesAuxBuffer |= gpu == MVKDescriptorGPULayout::BufferAuxSize;
 	}
 
+	// Apply Metal constant buffer alignment padding for each descriptor set
+	gpuSize = calcGroupSizeWithPadding(gpuSize, pCreateInfo->maxSets, gpuAlign, cbufAlign);
+
 	if (numAuxOffset) {
 		// Aux offsets are allocated on the CPU buffer
 		cpuSize += calcGroupSizeWithPadding(numAuxOffset * sizeof(uint32_t), pCreateInfo->maxSets, alignof(uint32_t), cpuAlign);
@@ -1951,7 +1954,7 @@ MVKDescriptorPool* MVKDescriptorPool::Create(MVKDevice* device, const VkDescript
 		gpuSize += alignDescriptorOffset(maxGPUSize(MVKDescriptorGPULayout::Buffer, sizes), gpuAlign) * pCreateInfo->maxSets;
 		// The aux buffer is sized relative to the number of total elements in a descriptor, rather than the number of bindings with aux buffers
 		uint32_t size = numElem * sizeof(uint32_t);
-		// Aux buffers require Metal constant buffer alignment padding between descriptor sets
+		// Apply Metal constant buffer alignment padding for each descriptor set
 		gpuSize += calcGroupSizeWithPadding(size, pCreateInfo->maxSets, alignof(uint32_t), cbufAlign);
 	}
 
--- a/MoltenVK/MoltenVK/GPUObjects/MVKDescriptorSet.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKDescriptorSet.mm
@@ -1898,7 +1898,7 @@ MVKDescriptorPool* MVKDescriptorPool::Create(MVKDevice* device, const VkDescript
 	MVKArrayRef pools(pCreateInfo->pPoolSizes, pCreateInfo->poolSizeCount);
 	MVKArgumentBufferMode argBufMode = pickArgumentBufferMode(device);
 	const MVKPhysicalDeviceArgumentBufferSizes& sizes = device->getPhysicalDevice()->getArgumentBufferSizes();
-	uint32_t gpuAlign = std::max(std::max<uint32_t>(sizes.texture.align, sizes.sampler.align), std::max<uint32_t>(sizes.pointer.align, 1));
+	uint32_t gpuAlign = std::max(std::max<uint32_t>(sizes.texture.align, sizes.sampler.align), std::max<uint32_t>(sizes.pointer.align, sizes.uniform.align));
 	uint32_t cpuAlign = alignof(id);
 	uint32_t gpuSize = 0;
 	uint32_t cpuSize = 0;
--- a/MoltenVK/MoltenVK/GPUObjects/MVKDescriptorSet.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKDescriptorSet.mm
@@ -1898,7 +1898,8 @@ MVKDescriptorPool* MVKDescriptorPool::Create(MVKDevice* device, const VkDescript
 	MVKArrayRef pools(pCreateInfo->pPoolSizes, pCreateInfo->poolSizeCount);
 	MVKArgumentBufferMode argBufMode = pickArgumentBufferMode(device);
 	const MVKPhysicalDeviceArgumentBufferSizes& sizes = device->getPhysicalDevice()->getArgumentBufferSizes();
-	uint32_t gpuAlign = std::max(std::max<uint32_t>(sizes.texture.align, sizes.sampler.align), std::max<uint32_t>(sizes.pointer.align, sizes.uniform.align));
+	uint32_t gpuAlign = std::max(std::max<uint32_t>(sizes.texture.align, sizes.sampler.align), std::max<uint32_t>(sizes.pointer.align, 1));
+	uint32_t cbufAlign = std::max<uint32_t>(gpuAlign, sizes.cbuffer.align);
 	uint32_t cpuAlign = alignof(id);
 	uint32_t gpuSize = 0;
 	uint32_t cpuSize = 0;
@@ -1906,6 +1907,7 @@ MVKDescriptorPool* MVKDescriptorPool::Create(MVKDevice* device, const VkDescript
 	uint32_t numAuxOffset = 0;
 	uint32_t inlineUniformSize = 0;
 	bool usesAuxBuffer = false;
+	bool usesTexCombined = false;
 
 	for (const auto& pool : pools) {
 		if (pool.type == VK_DESCRIPTOR_TYPE_INLINE_UNIFORM_BLOCK) {
@@ -1916,7 +1918,14 @@ MVKDescriptorPool* MVKDescriptorPool::Create(MVKDevice* device, const VkDescript
 		MVKDescriptorCPULayout cpu = pickCPULayout(pool.type, 1, argBufMode, device);
 		numElem += descriptorGPUBindingCount(gpu) * pool.descriptorCount;
 		cpuSize += alignDescriptorOffset(descriptorCPUSize(cpu), cpuAlign) * pool.descriptorCount;
-		gpuSize += alignDescriptorOffset(maxGPUSize(gpu, sizes), gpuAlign) * pool.descriptorCount;
+		if (gpu == MVKDescriptorGPULayout::TexBufSoA || gpu == MVKDescriptorGPULayout::TexSampSoA ||
+			gpu == MVKDescriptorGPULayout::Tex2SampSoA || gpu == MVKDescriptorGPULayout::Tex3SampSoA) {
+			// Combined textures use Metal Constant Buffer alignment
+			gpuSize += alignDescriptorOffset(maxGPUSize(gpu, sizes), cbufAlign) * pool.descriptorCount;
+			usesTexCombined = true;
+		}
+		else
+			gpuSize += alignDescriptorOffset(maxGPUSize(gpu, sizes), gpuAlign) * pool.descriptorCount;
 		numAuxOffset += needsAuxOffset(gpu) ? pool.descriptorCount : 0;
 		usesAuxBuffer |= gpu == MVKDescriptorGPULayout::BufferAuxSize;
 	}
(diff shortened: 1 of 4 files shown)


SRSaunders:
Updated to only apply Metal Constant buffer alignment to Aux Buffer and Combined Texture argument buffer types.  This seems to be more memory efficient while eliminating Metal API Validation failures for Sascha Willems' Vulkan examples and mostly for Khronos Vulkan-Samples.  There are a few remaining Metal API validation failures for Vulkan-Samples but these are not related to Metal Constant Buffer alignment issues.

Tested on iOS Simulatior and my x86_64  + AMD Mac.  Note this PR should be a no-op on Apple Silicon.

Ready to be reviewed now.

SRSaunders:
I would like to ask @etang-cw to review this PR.  Note this change attempts to solve a regression since MoltenVK 1.4.1, starting with pull request https://github.com/KhronosGroup/MoltenVK/pull/2572.  I believe you submitted the rewrite of descriptor set handling, so I imagine you would be the best person to look at this PR.  Thanks.

cdavis5e:
> I would like to ask @etang-cw to review this PR.

Unfortunately, it does not appear to be possible to assign him as a reviewer, probably because the GitHub user profile is not part of the KhronosGroup org.

SRSaunders:
Thanks @cdavis5e for looking at this.  Even if @etang-cw can't be an official reviewer, perhaps he could look at it and offer comments since he was the author of the new code.

cdavis5e (approved):
LGTM, FWIW.

etang-cw:
First of all, why only combined texture + buffer/samplers?

Second of all, each alignment can only happen once per descriptor set, not once per descriptor, so I think you're vastly overestimating the potential space lost to alignment.  You should be calculating it as a multiple of maxSets (plus the number of aux and outlined data buffers), not of the number of some type of contained descriptor.  You might want to also check the logic for allocating those outlined data and aux buffers, they might also need to be updated for the new alignment.  (Does SPIRV-Cross make aux buffers `constant` or `device const`?  I don't remember.)  At the very least there's a comment around the `OutlinedData` enum definition that should be updated.  Note that OutlinedData still needs to be 16B aligned (rather than 4B) on Apple GPUs, since structs that contain `float4`s are 16B aligned in MSL.

Third, cbuffer align is an attribute of the pointed-to memory, not of the pointer in the argument buffer, so it doesn't make sense to put it in `_argumentBufferSizes`, which stores the size and align of argument buffer struct members.  Just grab the value from `MVKPhysicalDeviceMetalFeatures`.

Finally, you're violating the comment at the top of `MVKPhysicalDeviceMetalFeatures` by adding your new entry in the middle.  Though you should also check if what you're defining is any different from `mtlBufferAlignment` after doing what I mentioned above.  Or maybe `mtlBufferAlignment` should be updated to drop to 32 based on an os version + Mac2 check.  (I haven't checked the full content of the file, and only see it being set to 256 around one of the diffs on this PR, so IDK whether it is getting dropped to 32 or not.)

SRSaunders:
Thank you @etang-cw for your detailed comments.  I have tried to incorporate all your feedback and have adjusted my code based on my understanding of your notes above.  Please forgive me if I have misunderstood something or missed an important point.

Note that I did make one adjustment to your existing code:  when reviewing the Outlined data section, I noticed that the `16u` alignment was permanently applied to `gpuAlign`, which is then used in the Aux Buffer section.  I am not sure if this is correct, so I went ahead and defined a new variable `dataAlign` which has reduced scope within Outline, but then is applied again after the Aux Buffer section for the final overall alignment prior to allocation.  I have tested this with some inline uniform block examples and it seems to work ok.  Please let me know if I am incorrect here.

Tested on macOS Sequoia x86_64 (32 byte cbuf align), iOS 26 on iPhone 16 (4 byte cbuf align), and iOS 26 on iOS Simulator (256 byte cbuf align).  I will try to test on my M1 Air tomorrow.

MoltenVK/MoltenVK/GPUObjects/MVKDescriptorSet.mm
```diff
@@ -1899,6 +1899,9 @@ static uint32_t calcGroupSizeWithPadding(uint32_t size, uint32_t groups, uint32_
 	MVKArgumentBufferMode argBufMode = pickArgumentBufferMode(device);
 	const MVKPhysicalDeviceArgumentBufferSizes& sizes = device->getPhysicalDevice()->getArgumentBufferSizes();
 	uint32_t gpuAlign = std::max(std::max<uint32_t>(sizes.texture.align, sizes.sampler.align), std::max<uint32_t>(sizes.pointer.align, 1));
+	const uint32_t mtlCbufAlign = (uint32_t)device->getPhysicalDevice()->getMetalFeatures()->mtlConstantBufferAlignment;
+	uint32_t cbufAlign = std::max(gpuAlign, argBufMode != MVKArgumentBufferMode::Off ? mtlCbufAlign : 1u);
```
etang-cw:
```suggestion
	const uint32_t mtlCbufAlign = (uint32_t)device->getPhysicalDevice()->getMetalFeatures()->mtlConstantBufferAlignment;
	uint32_t cbufAlign = std::max(gpuAlign, hostOnly ? 1u : mtlCbufAlign);
```
I think it would be nicer to handle `MVKArgumentBufferMode::Off` by only adding the padding if `gpuSize` is nonzero.

On the other hand, we don't need to align hostOnly buffers (pull the definition of hostOnly up from below).

MoltenVK/MoltenVK/GPUObjects/MVKDescriptorSet.mm
```diff
@@ -1921,6 +1924,9 @@ static uint32_t calcGroupSizeWithPadding(uint32_t size, uint32_t groups, uint32_
 		usesAuxBuffer |= gpu == MVKDescriptorGPULayout::BufferAuxSize;
 	}
 
+	// Apply Metal constant buffer alignment padding for each descriptor set
+	gpuSize = calcGroupSizeWithPadding(gpuSize, pCreateInfo->maxSets, gpuAlign, cbufAlign);
```
etang-cw:
```suggestion
	if (gpuSize)
		gpuSize = calcGroupSizeWithPadding(gpuSize, pCreateInfo->maxSets, gpuAlign, dataAlign);
```
And wait until after inline uniforms and aux buffer have been handled for this, since they add space for pointers in the main descriptor.

MoltenVK/MoltenVK/GPUObjects/MVKDescriptorSet.mm
```diff
@@ -1936,10 +1942,10 @@ static uint32_t calcGroupSizeWithPadding(uint32_t size, uint32_t groups, uint32_
 				// Add space for the pointers
 				gpuSize += alignDescriptorOffset(sizes.pointer.size, gpuAlign) * info->maxInlineUniformBlockBindings;
 				// Outlined data is 16-byte aligned
-				gpuAlign = std::max(gpuAlign, 16u);
+				dataAlign = std::max(gpuAlign, 16u);
```
etang-cw:
```suggestion
				dataAlign = std::max(dataAlign, 16u);
```
Outlined data is pointed to by constant pointers too, so it also needs to be updated for the new alignment.

This is why I mentioned updating comments earlier.

(And this affecting the subsequent aux buffer descriptor size was a mistake.)

MoltenVK/MoltenVK/GPUObjects/MVKDescriptorSet.mm
```diff
@@ -1949,9 +1955,13 @@ static uint32_t calcGroupSizeWithPadding(uint32_t size, uint32_t groups, uint32_
 		gpuSize += alignDescriptorOffset(maxGPUSize(MVKDescriptorGPULayout::Buffer, sizes), gpuAlign) * pCreateInfo->maxSets;
 		// The aux buffer is sized relative to the number of total elements in a descriptor, rather than the number of bindings with aux buffers
 		uint32_t size = numElem * sizeof(uint32_t);
-		gpuSize += calcGroupSizeWithPadding(size, pCreateInfo->maxSets, alignof(uint32_t), gpuAlign);
+		// Apply Metal constant buffer alignment padding for each descriptor set
+		gpuSize += calcGroupSizeWithPadding(size, pCreateInfo->maxSets, alignof(uint32_t), cbufAlign);
```
etang-cw:
```suggestion
		gpuSize += calcGroupSizeWithPadding(size, pCreateInfo->maxSets, alignof(uint32_t), dataAlign);
```

MoltenVK/MoltenVK/GPUObjects/MVKDescriptorSet.mm
```diff
@@ -1899,6 +1899,9 @@ static uint32_t calcGroupSizeWithPadding(uint32_t size, uint32_t groups, uint32_
 	MVKArgumentBufferMode argBufMode = pickArgumentBufferMode(device);
 	const MVKPhysicalDeviceArgumentBufferSizes& sizes = device->getPhysicalDevice()->getArgumentBufferSizes();
 	uint32_t gpuAlign = std::max(std::max<uint32_t>(sizes.texture.align, sizes.sampler.align), std::max<uint32_t>(sizes.pointer.align, 1));
+	const uint32_t mtlCbufAlign = (uint32_t)device->getPhysicalDevice()->getMetalFeatures()->mtlConstantBufferAlignment;
+	uint32_t cbufAlign = std::max(gpuAlign, argBufMode != MVKArgumentBufferMode::Off ? mtlCbufAlign : 1u);
+	uint32_t dataAlign = gpuAlign;
```
etang-cw:
```suggestion
	uint32_t dataAlign = cbufAlign;
```

MoltenVK/MoltenVK/GPUObjects/MVKDescriptorSet.mm
```diff
@@ -1949,9 +1955,13 @@ static uint32_t calcGroupSizeWithPadding(uint32_t size, uint32_t groups, uint32_
 		gpuSize += alignDescriptorOffset(maxGPUSize(MVKDescriptorGPULayout::Buffer, sizes), gpuAlign) * pCreateInfo->maxSets;
 		// The aux buffer is sized relative to the number of total elements in a descriptor, rather than the number of bindings with aux buffers
 		uint32_t size = numElem * sizeof(uint32_t);
-		gpuSize += calcGroupSizeWithPadding(size, pCreateInfo->maxSets, alignof(uint32_t), gpuAlign);
+		// Apply Metal constant buffer alignment padding for each descriptor set
+		gpuSize += calcGroupSizeWithPadding(size, pCreateInfo->maxSets, alignof(uint32_t), cbufAlign);
 	}
 
+	// Set overall gpu alignment incorporating Metal constant buffer and inline uniform data alignment
+	gpuAlign = std::max(cbufAlign, dataAlign);
```
etang-cw:
```suggestion
	gpuAlign = dataAlign;
```
(Since it was applied at the beginning)

etang-cw (commented):
It might make sense to merge `cbufAlign` and `dataAlign`, since I think they're referring to very similar ideas.

>  I have tried to incorporate all your feedback and have adjusted my code based on my understanding of your notes above. Please forgive me if I have misunderstood something or missed an important point.

If you're not sure whether you understood something, I'd rather you ask for clarification instead of implementing stuff without fully understanding what's going on.

MoltenVK/MoltenVK/GPUObjects/MVKDescriptorSet.mm
SRSaunders: This is fine as long as we apply padding using `if (gpuSize)` (i.e. non-zero) as you recommend below.

However, what happens to the `MVKArgumentBufferMode::Off` case?  The current definition of `hostOnly` does not take this into account and `cbufAlign` will be >1 when arg buffers are off.  While this should not affect the overall calculation of `gpuSize` if we pad only when non-zero, this seems inconsistent with the behaviour of `gpuAlign` when arg buffers are off (i.e. = 1). Are you okay with this?

MoltenVK/MoltenVK/GPUObjects/MVKDescriptorSet.mm
SRSaunders: Interestingly, I tested this idea before submitting and it appears to be unnecessary.  At least Metal API Validation does not flag >16u alignment for inline constant buffers as a requirement on any machine that I have available (x86_64, or arm64, or Simulator).  I am not sure whether this is a miss on the part of the validator, or a valid special case for inline. I kept `dataAlign` and `cbufAlign` separate for this reason.

Do you have insight on whether >16u alignment for inline is truly needed or not?

MoltenVK/MoltenVK/GPUObjects/MVKDescriptorSet.mm
SRSaunders: See above in code.  I kept these separate since I could not prove that >16u alignment is truly needed.

MoltenVK/MoltenVK/GPUObjects/MVKDescriptorSet.mm
SRSaunders: Same as above in code.  I kept these separate until I have more info about >16u for inline.

MoltenVK/MoltenVK/GPUObjects/MVKDescriptorSet.mm
SRSaunders: Again, same comment as above in code.  Kept separate until more info available about >16u for inline.

MoltenVK/MoltenVK/GPUObjects/MVKDescriptorSet.mm
SRSaunders: This is an interesting one.  I agree completely with the `if (gpuSize)` test.

However, I am having second thoughts about the need for constant buffer alignment using `maxSets`.  I am concerned this can overstate the padding requirement, and the situation would only be made worse by waiting until after inline and aux buffers to adjust.

Now this is where I am on shakier ground and need guidance.  I am wondering whether padding based on the sum of descriptorCounts from pools (with certain gpu layouts) that actually require constant buffer alignment may be sufficient.  I am interested in your feedback on this.

MoltenVK/MoltenVK/GPUObjects/MVKDescriptorSet.mm
etang-cw: In `MVKArgumentBufferMode::Off`, all the gpu layouts will be `MVKDescriptorGPULayout::None`, which has size 0, leaving a `gpuSize` of 0.  Current use of `gpuAlign` is with `alignDescriptorOffset`, which will properly calculate that 0 is fully aligned (and leave it as 0).  `calcGroupSizeWithPadding` doesn't currently have that property, though maybe it would make more sense to give it that by adding this to the beginning:
```cpp
	groups = std::min(groups, size / elemAlign); // Each group needs at least one element
```
Then you wouldn't have to do the `if (gpuSize)` since a 0 `gpuSize` would result in it calculating that you'd end up with a maximum of 0 groups and not add any padding.

MoltenVK/MoltenVK/GPUObjects/MVKDescriptorSet.mm
etang-cw: >  I am wondering whether padding based on the sum of descriptorCounts from pools (with certain gpu layouts) that actually require constant buffer alignment may be sufficient (and don't include descriptorCount from inlines for example).

Which GPU layouts?  My understanding is that this would be required for all GPU layouts (except `MVKDescriptorGPULayout::None`).  Do you have an example of which GPU layouts could be excluded?

MoltenVK/MoltenVK/GPUObjects/MVKDescriptorSet.mm
etang-cw: Metal API validation probably doesn't validate argument buffer contents.  Whether that's because they don't need validating (e.g. because they don't get the special constant buffer handling that gets applied to direct bindings, and are effectively treated like `device const*` buffers) or just because it's very difficult to validate argument buffers I don't know.  If we're going to keep these unaligned, then we can also keep the aux buffers unaligned, since they're referenced in the same way as the outlined buffers (exclusively from other argument buffers).

I suspect the GPU arch that's forcing 32B constant buffer alignment is Intel (Nvidia maxes out at Mac1, and AMD loads constant buffers in the shader identically to `device const*` buffers), so if you wanted to run some tests on some Intel GPUs to see how they handle this, you could.

MoltenVK/MoltenVK/GPUObjects/MVKDescriptorSet.mm
SRSaunders: Thanks for this suggestion.  It works as long as `size` contains allocations for all pool types, including inline uniforms.  Otherwise `size / elemAlign` is too small and it unfairly reduces the group count.  This is easy to overcome by locating `calcGroupSizeWithPadding()` after the inline section.  I will add this change to my next commit.

MoltenVK/MoltenVK/GPUObjects/MVKDescriptorSet.mm
SRSaunders: Well, Metal API Validation does check the alignment of the allocation when arg buffers are enabled.  An example of that is shown in the associated issue #2699 for this PR.  That 32B alignment error occurs on my x86_64 Mac with an AMD GPU.  No Intel GPU here as my machines with that setup are too old.

And if the pool size is too small, a Vulkan runtime allocation error will occur.  That's how I can tell if we have an _underallocation_ scenario.  Regarding detecting Aux Buffers underallocation, I can reproduce that quite easily using the _debugprintf_ example from Sascha Willems' Vulkan samples repo.

For the Inline case I have the _inlineuniformblocks_ example from the same repo.  That one does not fail with alignment or underallocation errors when using 16U.  While that is not a definitive test, it as least is something to consider.

And lastly, I am using the _timeline_semphore_ example from Khronos Vulkan-Samples to test the main padding via `calcGroupSizeWithPadding()`.  That example has no Inline or Aux buffers and relies solely on the main allocation loop for Combined Image Sampler and Storage Image pool types.

The current state of this PR results in no Metal API Validation alignment errors on macOS x86_64+AMD, on iOS iPhone 16, and on iOS Simulator.  This covers all Metal Constant Buffer alignment options (32, 4, and 256 respectively).  I have tested all platforms using the full suite of Sascha Willems' Vulkan examples and also with the full set of Khronos samples with Metal API Validation on.

Nonetheless, if you feel strongly about the 16u issue and merging cbufAlign and dataAlign, I will make that change.  That should be the final item to update and would resolve your outstanding comments.  

Please advise.

MoltenVK/MoltenVK/GPUObjects/MVKDescriptorSet.mm
SRSaunders: Empirically, I believe the Texture and Sampler types could be excluded since they do not involve constant buffers with associated Metal constant buffer alignment.  However, I don't know this area well enough to make that call.  I am mainly a Vulkan on Mac guy and not a Metal nor Argument Buffer expert.  Normally I don't get into the internals of MoltenVK.  So I need to defer to you and others on this aspect of things.

SRSaunders: I don't plan to pursue this possible optimization any further.  I don't want to make this a never ending project and using `maxSets` for the main padding is fine.

MoltenVK/MoltenVK/GPUObjects/MVKDescriptorSet.mm
etang-cw: > Well, Metal API Validation does check the alignment of the allocation when arg buffers are enabled. An example of that is shown in the associated issue https://github.com/KhronosGroup/MoltenVK/issues/2699 for this PR. That 32B alignment error occurs on my x86_64 Mac with an AMD GPU. No Intel GPU here as my machines with that setup are too old.

```metal
struct ArgBuf {
	constant int* pointer;
};
kernel void test(constant ArgBuf* args) { ... }
```
There's two (or more) constant pointers here, `args`, and `args[n].pointer`

API validation will definitely check the alignment of `args`, but will it check the alignment of `args[n].pointer`?

Alignment of the main descriptor is like alignment of `args`.  Alignment of both the outlined constants and the aux buffer is like alignment of `args[n].pointer`.

I don't feel strongly about whether we should align constant pointers in argument buffers.  But I do think you should do it consistently.  Either align both outlined pointers and aux buffers, or neither, but not just one.

MoltenVK/MoltenVK/GPUObjects/MVKDescriptorSet.mm
etang-cw: > Empirically, I believe the Texture and Sampler types could be excluded since they do not involve constant buffers with associated Metal constant buffer alignment.

The validation layer is not complaining about buffers in the descriptor set being unaligned.  It's complaining about the descriptor set itself being unaligned.

Unlike on Vulkan where a descriptor set is something managed by the driver, here we *are* the driver.  The descriptor set is memory managed by us, and we bind it as a Metal resource.  Which means it has to obey the rules of Metal just as much as the resources in it do.

p.s. this is why I wanted you to ask for clarification on things that didn't make sense to you.  If you had questioned why I was telling you to apply alignment to everything and not just combined resources, you probably would have gotten this explanation sooner.

SRSaunders: Thanks @etang-cw for the background info.  I appreciate it.

MoltenVK/MoltenVK/GPUObjects/MVKDescriptorSet.mm
SRSaunders: I have done the following:

1. Don't align and restore original logic for aux buffers.  As you already suspected this is probably not needed.  I have tested across multiple setups and platforms with all Vulkan examples and don't get any pool depletion errors with Argument Buffers enabled.  And you have clearly indicated why Metal API Validation does not look at this.  Note that I just found a new issue with Arg Buffers off on iOS and macOS arm64 only - but that appears to be separate and is pre-existing prior to this PR.  I will describe the issue below and ask for your input.
2. Reduce the scope of `dataAlign` (16u) to inline uniforms only.  And as you indicated above, having the original code leak through to aux buffers and onwards was a mistake.  This update fixes that.
3. Moved the descriptor set padding logic to the end, also as you also suggested earlier.  This simplifies the changes by adding only a single code block in one location.  Note that I had to add a test for Argument buffers == Off (i.e. OR with `hostOnly`) when defining `cbufAlign`.  Otherwise an incorrect value would be assigned to `gpuAlign` going into the allocations.

I think this should resolve the remaining open issues here.  @etang-cw please let me know if anything else is needed. Again, many thanks for all your assistance.

MoltenVK/MoltenVK/GPUObjects/MVKDescriptorSet.mm
SRSaunders: Should be resolved by https://github.com/KhronosGroup/MoltenVK/pull/2704/commits/e91a00e33dde8674a78da95878a4d0c1ec91f207

MoltenVK/MoltenVK/GPUObjects/MVKDescriptorSet.mm
SRSaunders: Should be resolved by https://github.com/KhronosGroup/MoltenVK/pull/2704/commits/e91a00e33dde8674a78da95878a4d0c1ec91f207

MoltenVK/MoltenVK/GPUObjects/MVKDescriptorSet.mm
SRSaunders: Should be resolved by https://github.com/KhronosGroup/MoltenVK/pull/2704/commits/e91a00e33dde8674a78da95878a4d0c1ec91f207

SRSaunders:
Hopefully with the resolutions above this PR should be ready, subject to approval from @etang-cw.

However, during testing I have identified an additional validation issue on iOS and macOS arm64 when **Arg Buffers = Off**.  This issue is pre-existing prior to this PR.  I get the following Metal API Validation error on some Vulkan examples:

_-[MTLDebugRenderCommandEncoder validateCommonDrawErrors:]:6001: failed assertion `Draw Errors Validation
Vertex Function(main0): argument ubo[0] from Buffer(0) with offset(0) and length(136) has space for 136 bytes, but argument has a length(144)._

It appears there is an off-by-one error in the allocation on iOS / macOS arm64 when Arg Buffers = Off.  This error does not appear on macOS x86_64 nor on the iOS Simulator on x86_64, nor on iOS / macOS arm64 when Arg Buffers = On.  I have ~~not yet~~ tested on my M1 (arm64) ~~but will do so soon and report back~~ and the error occurs there as well.

@etang-cw I am certain you have a view on this, and whether a resolution belongs in this PR or not.

etang-cw (approved):
> -[MTLDebugRenderCommandEncoder validateCommonDrawErrors:]:6001: failed assertion `Draw Errors Validation
Vertex Function(main0): argument ubo[0] from Buffer(0) with offset(0) and length(136) has space for 136 bytes, but argument has a length(144).

I suspect this is not an off-by-one error but another alignment issue.

If Vulkan defines a uniform buffer like
```metal
struct {
	float4 a;
	float2 b;
};
```
then in Vulkan that's a 24B struct, but in Metal, which uses C layout rules, `a` is 16B aligned so the size is 16B aligned, making the whole struct 32B.  If Vulkan then binds a 24B buffer and we do the same, Metal validation will complain that it's too small.  We try to align all buffers up to 16B to work around this, but we probably missed one somewhere.

I'd say fix that in a separate PR.

SRSaunders:
> We try to align all buffers up to 16B to work around this, but we probably missed one somewhere.
> 
> I'd say fix that in a separate PR.

Thanks @etang-cw for your analysis and suggestion.  I will open a separate issue for that.

@cdavis5e this PR is therefore complete and ready for final review. I see that @etang-cw has approved - thanks.

billhollings (approved):
@SRSaunders  Thanks for the submission, and for working through the issues.

@etang-cw  Thanks for the extensive review.

LGTM.

PR #2719, opened 2026-04-21
Vk khr portability subset metal

johnfaizzjamaludin:


billhollings:
Closing this as unworkable and unnecessary.

1. Pulling 2 year old code from internal Khronos GitLab is a non-starter.

2. The `VK_KHR_portability_subset_metal` branch is never going to get merged into `main` and be released in the wild, because Kosmic Krisp has demonstrated that the portability extensions are not needed for conformance.

PR #2722, opened 2026-04-22
Fix same-size swapchain recreation resulting in 1x1 `drawableSize`

### LINKED ISSUE #2226: Issues with the swap chain in 1.3.275.0 and later versions
Hello!

After updating Vulkan SDK to the latest version, we started seeing issues with the swap chain that did not happen before (versions 1.3.268.1 and earlier).

First, when running application from XCode with Metal API Validation enabled, the app triggers Metal assertion:

```
validateRenderPassDescriptor:991: failed assertion `RenderPass Descriptor Validation
renderTargetWidth (1600) must be <= minimum attachment width (1).
renderTargetHeight (1200) must be <= minimum attachment height (1).
```

When disabling Metal API Validation, Vulkan validation displays errors:

```
Diligent Engine: ERROR: Vulkan debug message (validation): UNASSIGNED-VkPresentInfoKHR-pImageIndices-MissingAcquireWait
                 Validation Error: [ UNASSIGNED-VkPresentInfoKHR-pImageIndices-MissingAcquireWait ] Object 0: handle = 0x808562000000003f, type = VK_OBJECT_TYPE_IMAGE; Object 1: handle = 0x9f9b41000000003c, name = Swap chain image acquired semaphore 2, type = VK_OBJECT_TYPE_SEMAPHORE; Object 2: handle = 0x5c5283000000003e, type = VK_OBJECT_TYPE_FENCE; | MessageID = 0x1b6b9ef2 | vkQueuePresentKHR(): pPresentInfo->pImageIndices[0] was acquired with a semaphore VkSemaphore 0x9f9b41000000003c[Swap chain image acquired semaphore 2] and fence VkFence 0x5c5283000000003e[] and neither of them have since been waited on
                 Object[0] (image): Handle 0x808562000000003f
                 Object[1] (semaphore): Handle 0x9f9b41000000003c, Name: 'Swap chain image acquired semaphore 2'
                 Object[2] (fence): Handle 0x5c5283000000003e

Diligent Engine: ERROR: Vulkan debug message (validation): VUID-vkDestroyFence-fence-01120
                 Validation Error: [ VUID-vkDestroyFence-fence-01120 ] Object 0: handle = 0x5c5283000000003e, type = VK_OBJECT_TYPE_FENCE; | MessageID = 0x5d296248 | vkDestroyFence(): fence (VkFence 0x5c5283000000003e[]) is in use. The Vulkan spec states: All queue submission commands that refer to fence must have completed execution (https://vulkan.lunarg.com/doc/view/1.3.275.0/mac/1.3-extensions/vkspec.html#VUID-vkDestroyFence-fence-01120)
                 Object[0] (fence): Handle 0x5c5283000000003e
```

The errors do not make a lot of sense: the semaphore is signaled few lines before it is passed to `vkQueuePresentKHR`

Lastly, when resizing the window, the app hangs.

None of these issues happened before the 1.3.275.0 SDK. They also don't happen on any other platform (Windows, Linux, Android) with the same SDK version.

The behavior is the same whether we use the MoltenVK surface or VK_EXT_Metal_Surface

[Tutorial01_HelloTriangle.zip](https://github.com/KhronosGroup/MoltenVK/files/15210675/Tutorial01_HelloTriangle.zip)

h3r2tic:
Hey! I've ran into #2226 with very similar symptoms, except without any validation errors.
In my case, changing the presentation mode from FIFO to IMMEDIATE without resizing the window would trigger the bug. The output of my game's rendering would be a constant color. Resizing the window would unbug it.

It seems that the root cause is just the order in which the `forceUnpresentedImageCompletion` workaround runs relative to the configuration of a new swapchain. If `forceUnpresentedImageCompletion` goes last, it will internally change the `CAMetalLayer` `drawableSize` to 1x1, overriding the `mtlLayer.drawableSize = ...` that was being done for the new swapchain.

I don't know how to test the Metal regression the comments discuss, but this change resolves the 1x1 drawable size issue for me. Tested on an MBP M1.

Fixes #2226

### DIFF
--- a/MoltenVK/MoltenVK/GPUObjects/MVKSwapchain.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKSwapchain.mm
@@ -486,6 +486,18 @@ void MVKSwapchain::initCAMetalLayer(const VkSwapchainCreateInfoKHR* pCreateInfo,
 	auto* mtlLayer = getCAMetalLayer();
 	if ( !mtlLayer || getIsSurfaceLost() ) { return; }
 
+	VkExtent2D prevSurfaceExtent = _surface->getExtent();
+	auto* oldSwapchain = (MVKSwapchain*)pCreateInfo->oldSwapchain;
+
+	// Because of a regression in Metal, the most recent one or two presentations may not
+	// complete and call back. Changing the CAMetalLayer drawableSize will force any incomplete
+	// presentations on the oldSwapchain to complete and call back, but if the drawableSize
+	// is not changing from the previous, we need to force those completions before resetting
+	// the layer to the same size on the replacement swapchain.
+	if (oldSwapchain && mvkVkExtent2DsAreEqual(pCreateInfo->imageExtent, prevSurfaceExtent)) {
+		oldSwapchain->forceUnpresentedImageCompletion();
+	}
+
 	auto minMagFilter = getMVKConfig().swapchainMinMagFilterUseNearest ? kCAFilterNearest : kCAFilterLinear;
 	mtlLayer.drawableSize = mvkCGSizeFromVkExtent2D(_imageExtent);
 	mtlLayer.device = getMTLDevice();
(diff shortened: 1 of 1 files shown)


cdavis5e (approved):
LGTM.

Issue #2203, opened 2024-04-08
Low maxVariableDescriptorCount being reported


MennoVink:
I'm receiving a rather low variable descriptor count which i expected to be higher.

I'm on Intel Iris 6100 and these are my limits:
```
maxPerStageDescriptorSamplers	uint32_t	16
maxPerStageDescriptorUniformBuffers	uint32_t	31
maxPerStageDescriptorStorageBuffers	uint32_t	31
maxPerStageDescriptorSampledImages	uint32_t	128
maxPerStageDescriptorStorageImages	uint32_t	8
maxPerStageDescriptorInputAttachments	uint32_t	128
maxPerStageResources	uint32_t	159
maxDescriptorSetSamplers	uint32_t	80
maxDescriptorSetUniformBuffers	uint32_t	155
maxDescriptorSetUniformBuffersDynamic	uint32_t	155
maxDescriptorSetStorageBuffers	uint32_t	155
maxDescriptorSetStorageBuffersDynamic	uint32_t	155
maxDescriptorSetSampledImages	uint32_t	640
maxDescriptorSetStorageImages	uint32_t	40
maxDescriptorSetInputAttachments	uint32_t	640
```

The values important to me are 16 samplers and 128 sampled images (both per-stage).
This is the layout for my descriptor set (glsl):

```
struct Constants
{
	mat4 world;
	int albedoIndex;
};
layout( std430, set = 0, binding = 0 ) readonly buffer buf
{
	Constants constants[];
};
layout( set = 0, binding = 1) uniform sampler albedoSampler;
layout( set = 0, binding = 2 ) uniform texture2D albedoImages[];
```

My layout bindings match this and i have bindFlags of 0, 0, 8 (variable only for the albedoImages).

Now when i call getDescriptorSetLayoutSupport attaching a DescriptorSetVariableDescriptorCountLayoutSupportEXT, i get the maximum variable count being 15.

Where does this 15 come from? How come it doesn't tell me that i can use 128 sampled images?

MennoVink:
I tried to get this to run in the cube example but it's configured to disallow debugging so not sure how to recreate it there.

Here's what i've got so far:
```
    	VkDescriptorSetLayoutBinding bindings[ 3 ] = {};
    	bindings[ 0 ].binding = 0;
    	bindings[ 0 ].descriptorType = VK_DESCRIPTOR_TYPE_STORAGE_BUFFER;
    	bindings[ 0 ].descriptorCount = 1;
[...]

PR #2723, opened 2026-04-24
Fix variable descriptor count query

MennoVink:
https://registry.khronos.org/VulkanSC/specs/1.0-extensions/man/html/VkDescriptorSetVariableDescriptorCountLayoutSupport.html

**maxVariableDescriptorCount indicates the maximum number of descriptors supported in the highest numbered binding of the layout, if that binding is variable-sized.**

MoltenVK was putting the maximum number of resources for the last binding that is **not** the variable binding in that variable count field, even if that binding is of a different type.
This fixes the two year old issue https://github.com/KhronosGroup/MoltenVK/issues/2203 and makes bindless renderers without UAB feasible.


cdavis5e (approved):
Looks to me like this was a victim of a refactoring, but I'd have to investigate the history to verify that.

Anyway, LGTM.

(@billhollings hasn't had time to maintain this project, which is why you haven't seen him around much lately.)

MennoVink:
@cdavis5e cool thanks for merging. I think that fixes #2203 so you can close that as well

PR #2731, opened 2026-05-07
MVKRenderPass: encode barriers for subpass dependencies

js6i:
@billhollings that fixes the bloom example flicker for me. I also noticed that we don't always call barrier updates when ending  render encoders, which looks like it may case other bugs. The second commit addresses that.

I'm a little uneasy about edge cases of auxiliary encoders that get created, or encoders getting shared between uses. The subpass resolves are handled, I end it immediately so it updates the right fences.

I'm low on time and don't have games or CTS set up so please handle testing to see if I didn't accidentally break something else.

@cdavis5e


SRSaunders:
I can confirm this solves the problem highlighted in #2706.  The _bloom_ example now works properly on my M1 Air.  In addition, it does not cause any visible regressions in other examples within Sascha Willems' library.

Thanks @js6i for this fix.

Perhaps @kickblade may want to look at this PR regarding #2565.

squidbus (approved):

squidbus:
Ran through some testing myself and seems to be working okay.

kickblade:
Looks very promising, I’ll take a look at whether it fixes #2565 very soon, thanks heaps @squidbus

squidbus:
> Looks very promising, I’ll take a look at whether it fixes #2565 very soon, thanks heaps @squidbus

I wasn’t the one who made the change, but hope it’s helpful.

xen2:
Thanks for this!
Confirming it fixes a nasty bug on iOS Simulator. Our scenario: vkCmdClearColorImage + image memory barrier (Transfer→ColorAttachmentOutput) + a subsequent render pass with loadOp=Load was showing uninitialized tile content instead of the cleared color (alpha-blended draws then composed over garbage).

I was hunting it down and noticed it already got fixed with latest `main`!
With this PR's fence-timing fix the tile cache loads what the clear actually wrote, and our renders come out clean.

Looking forward to v1.4.2!

kickblade:
@squidbus it does appear this also fixed #2565, thanks for letting me know and thank you @js6i for the fix!! Amazing stuff!

Issue #2727, opened 2026-05-01
immediate crashes: vkQuake 1.34.1, MoltenVK version 1.4.1; macos 12.7.6 Monterey; MacBookPro12,1 Intel i5;  Intel(R) Iris(TM) Graphics 6100


sto6:
Thank you: MoltenVK (for vkQuake) has breathed new life in my 2015 Intel MacBook Pro...

works: vkQuake 1.32.2: MoltenVK version 1.2.9, supporting Vulkan version 1.2.283 (Without & with env-var MVK_CONFIG_USE_MTLHEAP=0, no difference.)

CRASH: vkQuake 1.34.1: MoltenVK version 1.4.1, supporting Vulkan version 1.4.334 Crash is unique for without(A) or with(B) env-var MVK_CONFIG_USE_MTLHEAP=0.

Not building myself, just using releases here: [MacSourcePorts/...](https://github.com/MacSourcePorts/MSPBuildSystem/releases?q=vkquake&expanded=true). Yes, means I'm comparing releases with changes to both vkQuake source, and (less frequently) the MoltenVK version, so a cause could be in either.

Here are abbreviated stracktrace & logs for the A/B cases. Full crash-reports: [crashA-1.34.1.txt](https://github.com/user-attachments/files/27285285/crashA-1.34.1.txt) [crashB-1.34.1.txt](https://github.com/user-attachments/files/27285287/crashB-1.34.1.txt)

```
A: no env-var:
Exception Type:        EXC_CRASH (SIGABRT)
Exception Codes:       0x0000000000000000, 0x0000000000000000
Exception Note:        EXC_CORPSE_NOTIFY

Thread 0 Crashed::  Dispatch queue: com.apple.main-thread
0   libsystem_kernel.dylib                  0x7ff800cd3fce __pthread_kill + 10
1   libsystem_pthread.dylib                 0x7ff800d0a1ff pthread_kill + 263
2   libsystem_c.dylib                       0x7ff800c55d14 abort + 123
3   libsystem_c.dylib                       0x7ff800c550bb __assert_rtn + 314
4   Metal                                   0x7ff809793864 MTLReportFailure.cold.1 + 43
5   Metal                                   0x7ff80977dbc5 MTLReportFailure + 529
6   Metal                                   0x7ff809776338 _MTLMessageContextEnd + 1278
7   Metal                                   0x7ff80974be94 -[MTLHeapDescriptorInternal validateWithDevice:] + 441
[...]

PR #2733, opened 2026-05-09
Restore some checks for Mac1 GPUs.

squidbus:
Going by https://github.com/KhronosGroup/MoltenVK/issues/2727 it seems we may have some Mac1 GPUs still supported. It's not clear which GPUs since the Metal Feature Set Tables don't seem to acknowledge Mac1 anymore, but I went back to a previous version and separated out some of the feature flags which Mac1 does not support.

I'm not sure if this fully solves the linked issue but it should at least prevent one of the bugs they're seeing.

### DIFF
--- a/MoltenVK/MoltenVK/GPUObjects/MVKDevice.h
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKDevice.h
@@ -112,6 +112,7 @@ typedef struct MVKMTLDeviceCapabilities {
 	bool supportsApple8;
 	bool supportsApple9;
 	bool supportsApple10;
+	bool supportsMac1;
 	bool supportsMac2;
 	bool supportsMetal3;
 	bool supportsMetal4;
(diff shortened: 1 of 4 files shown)


cdavis5e (approved):
Needs a rebase since I pulled in #2734, but otherwise LG.

Had I pulled this one in first, it's likely #2734 would've needed a rebase.

cdavis5e:
Or, if you like, I can resolve the conflict myself from the web interface and then merge it.

squidbus:
Rebased it.

TellowKrinkle:
BTW if it's helpful to anyone, [May 2022](https://web.archive.org/web/20220506093727/https://developer.apple.com/metal/Metal-Feature-Set-Tables.pdf) is the last version of the feature set tables to have Mac1 info.

I think the final macOS version to support Mac1 GPUs is macOS 12 with the [21" iMac](https://everymac.com/systems/apple/imac/specs/imac-core-i5-1.6-21-inch-aluminum-late-2015-specs.html) and [MacBook Air](https://everymac.com/systems/apple/macbook-air/specs/macbook-air-core-i5-1.6-11-early-2015-specs.html) both using Broadwell CPUs with their associated 6000-series iGPUs.  Later if you want to support people using OCLP.  It seems the current MVK minimum is macOS 11.

squidbus:
> Later if you want to support people using OCLP.

FWIW I believe OCLP is already broken on some hardware because of features and MSL version logic based on the OS version that don't work with the ported up drivers, I don't believe it's intended to be supported.

TellowKrinkle:
Also native swizzle is missing from mac1 GPUs (#2646)

PR #2737, opened 2026-05-11
Do not convert UINT8 indices until draw time.

squidbus:
This came up in `dEQP-VK.draw.renderpass.indexed_draw.update_index_buffer_before_draw_8` (which was added after UINT8 support here).

The UINT8 index implementation attempted to optimize by converting the index buffer once at bind time, but it turns out this is not valid behavior. The test mentioned above checks that, if the index buffer is modified after bind and before draw, it will draw with the modified contents.

The test still fails now due to us not supporting disabling primitive restart, but the output is now more correct.


billhollings (approved):
LGTM.

Issue #2682, opened 2026-01-01
Occlusion queries: wrap detection uses stale baseline, crashes when visibility buffer starts near end of pass


BXYMartin:
I am running MeloNX on iOS with MoltenVK 1.4.1. Random crashes (SIGSEGV) are observed in-game consistently and reproducible across multiple devices, while calling for QueueSubmit. In the debugger, it crashed at

```assembly
AGXMetal13_3`-[AGXG13GFamilyRenderContext setVisibilityResultMode:offset:]:
    0x11a383bf8 <+0>:   pacibsp 
    0x11a383bfc <+4>:   stp    x20, x19, [sp, #-0x20]!
    0x11a383c00 <+8>:   stp    x29, x30, [sp, #0x10]
    0x11a383c04 <+12>:  add    x29, sp, #0x10
    0x11a383c08 <+16>:  adrp   x8, 1021
    0x11a383c0c <+20>:  ldrsw  x8, [x8, #0x27c]
    0x11a383c10 <+24>:  ldr    x8, [x0, x8]
    0x11a383c14 <+28>:  add    x9, x8, #0x14, lsl #12    ; =0x14000 
    0x11a383c18 <+32>:  add    x19, x9, #0x8c0
    0x11a383c1c <+36>:  mov    w9, #0x88dc               ; =35036 
    0x11a383c20 <+40>:  add    x9, x8, x9
    0x11a383c24 <+44>:  ubfx   x10, x3, #3, #29
->  0x11a383c28 <+48>:  strh   w10, [x9, #0x500]
    0x11a383c2c <+52>:  cmp    w2, #0x1
    0x11a383c30 <+56>:  cset   w11, eq
    0x11a383c34 <+60>:  cmp    w2, #0x0
    0x11a383c38 <+64>:  cset   w12, ne
    0x11a383c3c <+68>:  ldr    w13, [x9]
    0x11a383c40 <+72>:  lsl    w12, w12, #15
    0x11a383c44 <+76>:  and    w13, w13, #0xffff3fff
```

[...]

Issue #2728, opened 2026-05-02
1.4.2 Release Plan


billhollings:
Thanks to everyone who has been working on the latest updates here! 

It's about time an updated release was done. Since I haven't been in the active loop recently, let's get some input on when a good release point would be. 

For a 1.4.2 release, now that #2718 has been dealt with, from what I can see, it would be good to include:

- [x] A fix or negative-configuration to deal with #2706
- [x] #2724 once that last little cleanup is done
- [x] Add fix suggested in #2682

Please add to this list as appropriate.

@cdavis5e @squidbus

rcaridade145:
Can this be in #2682?

billhollings:
> Can this be in [#2682](https://github.com/KhronosGroup/MoltenVK/issues/2682)?

Hmmm...sounds reasonable, once a review of the fix suggested there is done.

skooch:
@billhollings That cleanup for #2724 is now completed :)

squidbus:
Looks like the main remaining issue is the linked occlusion query one, if the proposed fix can be verified by someone knowledgeable about the relevant code area.

BXYMartin:
> Looks like the main remaining issue is the linked occlusion query one, if the proposed fix can be verified by someone knowledgeable about the relevant code area.

I am happy to submit a PR to include the fix in #2682 but would definitely need an expert to sign it off/propose a better fix.

Here's the PR: #2739

squidbus:
Looks like all the identified fix items and some more have been done, I think we're good for release now @billhollings

schm1dtmac:
Sorry to bump, what's the current status of 1.4.2's release given it's been close to 2 months since the tasks listed were all completed. At a glance there are some new issue tickets, PRs and merged changes that have cropped up since then, which may also end up extending the list of 'good-to-includes' before releasing.

billhollings:
> Sorry to bump, what's the current status of 1.4.2's release given it's been close to 2 months since the tasks listed were all completed. At a glance there are some new issue tickets, PRs and merged changes that have cropped up since then, which may also end up extending the list of 'good-to-includes' before releasing.

Thanks for the tap on the shoulder. I'll try to get this done this week.

KarenGhavam-lunarG:
> Thanks for the tap on the shoulder. I'll try to get this done this week.

[...]

PR #2739, opened 2026-05-11
Occlusion queries: wrap detection uses stale baseline, crashes when visibility buffer starts near end of pass

BXYMartin:
Fix stale baseline in warp detection and may crash when visibility buffer starts near end of pass mentioned in #2682

### DIFF
--- a/MoltenVK/MoltenVK/Commands/MVKCommandBuffer.mm
+++ b/MoltenVK/MoltenVK/Commands/MVKCommandBuffer.mm
@@ -780,6 +780,9 @@ void MVKCommandEncoder::beginMetalRenderPass(MVKCommandUse cmdUse) {
 		if (!_pEncodingContext->visibilityResultBuffer.buffer()) {
 			_pEncodingContext->visibilityResultBuffer = _device->getVisibilityBuffer();
 		}
+		// Track the starting visibility offset for this Metal render pass so wrap detection compares
+		// against the correct baseline even when the buffer was already partially consumed.
+		_pEncodingContext->firstVisibilityResultOffsetInRenderPass = _pEncodingContext->visibilityResultBuffer.offset();
 		mtlRPDesc.visibilityResultBuffer = _pEncodingContext->visibilityResultBuffer.buffer();
 	}
 


squidbus:
@etang-cw I remember you did some occlusion query fixes some time back, would you be able to check if this fix is correct?

etang-cw:
The change looks like something we should have, but an unnecessary render pass restart shouldn't result in a crash.

Looking back at the handling for this, allowing `nextMetalQuery` to restart a render pass probably breaks a lot of things, and that path has probably just never been hit before since it requires either this bug or a single render pass with >16k queries.  We probably need to move the restart detection to `MVKCommandEncoder::restartMetalRenderPassIfNeeded`.

Can you swap this section with an MVKAssert and either add a check to `restartMetalRenderPassIfNeeded` or if you're not confident in your ability to do that, put a todo there and I'll deal with it later?
https://github.com/KhronosGroup/MoltenVK/blob/e9b7bd2e4d6f365b36d29f5f57b8d39ca92f157f/MoltenVK/MoltenVK/Commands/MVKCommandEncoderState.mm#L1942-L1947

billhollings:
@etang-cw 

> The change looks like something we should have, but an unnecessary render pass restart shouldn't result in a crash.
> ...
> Can you swap this section with an MVKAssert and either add a check to `restartMetalRenderPassIfNeeded` or if you're not confident in your ability to do that, put a todo there and I'll deal with it later?

Can you clarify what you're requesting here, please.

Are you saying this PR is fine, but there is also a larger issue that we need to deal with, and you're hinting at the path in your comments. Or are you saying the fix in this PR needs further work to be correct?

If the former, let's get this PR in and 1.4.2 released (see #2728), and then create a separate issue to deal with the concern you are raising above.

etang-cw:
> Are you saying this PR is fine, but there is also a larger issue that we need to deal with, and you're hinting at the path in your comments.

Correct.  The PR solves the problem "we think we're running out of space when we're actually not" but does not solve the problem "if we think we're running out of space we crash instead of making more space".  I was mostly asking to change the `if` to a `MVKAssert(offset != cmdEncoder->_pEncodingContext->firstVisibilityResultOffsetInRenderPass);` since if the contents of that `if` are pretty much guaranteed to crash MoltenVK anyways (and the full solution will involve deleting that if and adding code elsewhere), we might as well swap it to an assert now.

BXYMartin:
> > Are you saying this PR is fine, but there is also a larger issue that we need to deal with, and you're hinting at the path in your comments.
> 
> Correct. I was mostly asking to change the `if` to a `MVKAssert(offset != cmdEncoder->_pEncodingContext->firstVisibilityResultOffsetInRenderPass);` since if the contents of that `if` are pretty much guaranteed to crash MoltenVK anyways (and the full solution will involve deleting that if and adding code elsewhere), we might as well swap it to an assert now.

@etang-cw I probably don't have enough background knowledge to carry out the fix that you proposed so I will leave that to you to decide which way is better.

squidbus (approved):
Looks like this should be good to land then, I ran through occlusion related CTS too and all passing still.

PR #2740, opened 2026-05-12
Add support for VK_KHR_draw_indirect_count, resolve #168

### LINKED ISSUE #168: Request: support for VK_KHR_draw_indirect_count
As of Vulkan 1.1.76 the VK_KHR_draw_indirect_count extension was added to the spec. The extension allows for the gpu to set the instance counts. Could this be added to MoltenVK?

yohjimane:


squidbus:
Have you run any [CTS](https://github.com/KhronosGroup/VK-GL-CTS) for this?

I ran through `dEQP-VK.draw.*indirect_count*` and got the following results:

```
Test run totals:
  Passed:        32/416 (7.7%)
  Failed:        128/416 (30.8%)
  Not supported: 256/416 (61.5%)
  Warnings:      0/416 (0.0%)
  Waived:        0/416 (0.0%)
```

The failures seem to come from `vertex_attribute_divisor` tests, e.g. `dEQP-VK.draw.renderpass.vertex_attribute_divisor.ext.static_pipeline.draw_indirect_count.zero.0`.

It would be helpful to verify whether these failures are from gaps in the implementation here that need fixing or other known limitations of MoltenVK. But it seems to fail all of these `vertex_attribute_divisor` tests while the equivalent for other draw types mostly pass.

squidbus:
Also, looking at the code, it seems to completely skip tessellation and triangle fans instead of handling them.

yohjimane:
I'll get these issues fixed! Thank you for the speedy response/review.

yohjimane:
@squidbus just pushed up some fixes which should resolve the CTS and mark the unsupported features as such

as for the tessellation and tri fans, I'm working on implementing tessellation in my game engine so I can properly build that out and test it, hopefully will get that resolved later tonight or tomorrow.

yohjimane:
@squidbus I got all the tests passing, but my implementation of tessellation kinda slow so I've backed that out of this PR for now. you can see the wip changes here: https://github.com/yohjimane/MoltenVK/pull/new/wip-tessellation-icb

I just realized while testing that I get OOM errors when falling back to the existing non-count tessellation path with a large payload. Is that expected? If so, it'd probably be better to group that work in it's own PR.

squidbus:
> but my implementation of tessellation kinda slow so I've backed that out of this PR for now.

I'm not sure this can really be left out, if we claim to support both tessellation and draw indirect count they need to properly work together.

FYI... KosmicKrisp (Vulkan over Metal driver in Mesa) is able to do draw indirect count by just running a compute shader beforehand which emits to a new buffer either a copy of the original indirect draw or a draw of all zeroed parameters, based on the value in the draw count buffer, then does its normal Vulkan indirect draw logic. This approach sounds to me like it would be a lot simpler and preserve paths like tessellation, triangle fans, uint8 indexes, etc.

MoltenVK/MoltenVK/Commands/MVKCmdDraw.mm
```diff
@@ -733,7 +753,7 @@
         vertexCount = kMVKMaxDrawIndirectVertexCount;
         patchCount = mvkCeilingDivide(vertexCount, inControlPointCount);
         VkDeviceSize indirectSize = (2 * sizeof(MTLDispatchThreadgroupsIndirectArguments) + sizeof(MTLDrawPatchIndirectArguments) + sizeof(MTLStageInRegionIndirectArguments)) * _drawCount;
-		paramsIncr = std::max((size_t)dvcLimits.minUniformBufferOffsetAlignment, sizeof(uint32_t) * 2);
+		paramsIncr = 256;
```
squidbus: What's this change for, seems now we're going to use a lot more buffer space than we needed to for lower alignment sizes?

yohjimane: https://github.com/KhronosGroup/MoltenVK/blob/main/MoltenVK/MoltenVK/Commands/MVKCommandPipelineStateFactoryShaderSource.h#L347

Shader side has this:
`device auto* params = reinterpret_cast<device uint32_t*>(paramsBuff + idx * 256);`
  
but the CPU side was doing:
`paramsIncr = std::max((size_t)dvcLimits.minUniformBufferOffsetAlignment, sizeof(uint32_t) * 2);`
  
max(16, 8) = 16 on my hardware
  
I think this would work fine for single draw calls but for multi draw indirect it would break

yohjimane: Fixed with f7c439e3 !

billhollings (commented):
What's the status of this now?

And what is the status of tessellation and triangle fans? I can't tell from the above discussion or looking at the code whether this is working now (although I can see a bunch of tessellation work in there).

Also, I ran through `dEQP-VK.draw.*indirect_count*` again on CTS 1.4.5.4 and got the following results:

```
Test run totals:
  Passed:        144/416 (34.6%)
  Failed:        16/416 (3.8%)
  Not supported: 256/416 (61.5%)
  Warnings:      0/416 (0.0%)
  Waived:        0/416 (0.0%)
```

So it looks like things are much better, but not quite complete. Can this be cleaned up?

Seems to be a pattern on the first test of each pipeline type, although all other tests pass for the same pipeline types. For example:

```
Test case 'dEQP-VK.draw.renderpass.vertex_attribute_divisor.ext.static_pipeline.draw_indirect_count.non_zero.0'..
  Fail (Unexpected results in output buffers; check log for details -- at vktDrawVertexAttribDivisorTests.cpp:695)
```

billhollings:
Closing. This has not been worked on recently, and is a duplicate of #2775 which looks to be a better option.

Issue #1743, opened 2022-10-14
Macbook M1 Multi Draw indicrect: MSL conversion error: DrawIndex is not supported in MSL.


mistycube:
When using gl_DrawIDARB in vertex shader to index draw data. gl_DrawIDARB is not supported. How to index the draw data in this case even though multi draw indirect is supported in Apple M1 ?

cdavis5e:
This is a known limitation: MoltenVK does not support the `DrawIndex` built-in. At the time, we considered this reasonable, because `VertexIndex` and `InstanceIndex` a) were more widely used, and b) are directly supported by Metal. Can't say I didn't see this coming, though...

squidbus:
Completed in https://github.com/KhronosGroup/MoltenVK/pull/2742

PR #2742, opened 2026-05-13
Add gl_DrawID / DrawIndex support for MacOS

denisk-dev:
Related to these reported issues:

MoltenVK: #1743  
SPIRV-Cross: [669](https://github.com/KhronosGroup/SPIRV-Cross/issues/669)
 
Corresponding SPIRV-Cross change:
https://github.com/denisk-dev/SPIRV-Cross/tree/msl-add-gl-draw-id

Kitten Space Agency (KSA) is using Vulkan and gl_DrawID is used in the rocket parts shaders.


denisk-dev:
This is my first time contributing. Please, be gentle!

denisk-dev:
If I'm not mistaken, the checks might be failing because the corresponding change in SPIRV-Cross is not available yet

MoltenVK/MoltenVK/Commands/MVKCmdDraw.mm
```diff
@@ -768,8 +768,35 @@
 
 	MVKPiplineStages stages;
     pipeline->getStages(stages);
+    
+    // Before the for-loop over drawCount
+    if (pipeline->needsDrawIDBuffer()) {
+        uint32_t firstDrawID = 0;  // usually 0 for a new vkCmdDraw* call,
+                                     // or accumulated baseDrawID if you support multi-draw indirect with offset
```
squidbus: I'm not sure we need this constant since it's just always 0, seems like extra.

MoltenVK/MoltenVK/Commands/MVKCmdDraw.mm
```diff
@@ -768,8 +768,35 @@
...
     for (uint32_t drawIdx = 0; drawIdx < _drawCount; drawIdx++) {
+        // === gl_DrawID hack: bind current draw index as [[buffer(N)]] ===
+        if (pipeline->needsDrawIDBuffer()) {
+            
+            if (drawIdx == 0) {
+                [cmdEncoder->_mtlRenderEncoder setVertexBuffer:cmdEncoder->_drawIDBuffer
+                                                              offset:0
+                                                             atIndex:pipeline -> getMetalBufferIndexForDrawID()];   // your reserved index
+                [cmdEncoder->_mtlRenderEncoder setVertexBufferOffset:0
+                                                             atIndex:pipeline -> getMetalBufferIndexForDrawID()];   // set the correct offset for draw #0
```
squidbus: Should not need to call `setVertexBufferOffset` here since you just called `setVertexBuffer` with the offset.

MoltenVK/MoltenVK/Commands/MVKCmdDraw.mm
```diff
@@ -1081,7 +1108,32 @@
 	MVKPiplineStages stages;
     pipeline->getStages(stages);
 
+    if (pipeline->needsDrawIDBuffer()) {
+        uint32_t firstDrawID = 0;  // usually 0 for a new vkCmdDraw* call,
+                                     // or accumulated baseDrawID if you support multi-draw indirect with offset
```
squidbus: Same here regarding unneeded constant.

MoltenVK/MoltenVK/Commands/MVKCmdDraw.mm
```diff
@@ -1081,7 +1108,32 @@
...
+
     for (uint32_t drawIdx = 0; drawIdx < _drawCount; drawIdx++) {
+        // === gl_DrawID hack: bind current draw index as [[buffer(N)]] ===
+        if (pipeline->needsDrawIDBuffer()) {
+            if (drawIdx == 0) {
+                [cmdEncoder->_mtlRenderEncoder setVertexBuffer:cmdEncoder->_drawIDBuffer
+                                                              offset:0
+                                                             atIndex:pipeline -> getMetalBufferIndexForDrawID()];   // your reserved index
+                [cmdEncoder->_mtlRenderEncoder setVertexBufferOffset:0
+                                                             atIndex:pipeline -> getMetalBufferIndexForDrawID()];   // set the correct offset for draw #0
```
squidbus: Same here regarding redundant `setVertexBufferOffset` call.

MoltenVK/MoltenVK/Commands/MVKCommandBuffer.h
```diff
@@ -250,6 +250,10 @@ class MVKCommandEncoder : public MVKBaseDeviceObject {
 
 	/** Returns the Vulkan API opaque object controlling this object. */
 	MVKVulkanAPIObject* getVulkanAPIObject() override { return _cmdBuffer->getVulkanAPIObject(); };
+    
+    id<MTLBuffer> _drawIDBuffer = nil;
+    uint32_t _currentMaxDrawCount = 0;   // track allocated size to avoid frequent realloc
```
squidbus: I believe these should be with other fields in the class and not up here with the function definitions.

MoltenVK/MoltenVK/Commands/MVKCommandBuffer.mm
```diff
@@ -422,6 +422,22 @@
 	addPerformanceInterval(getPerformanceStats().queue.commandBufferEncoding, startTime);
 }
 
+void MVKCommandEncoder::ensureDrawIDBuffer(uint32_t maxDrawCount) {
+    uint32_t neededBytes = maxDrawCount * sizeof(uint32_t);
+    if (_drawIDBuffer && [_drawIDBuffer length] >= neededBytes) return;
+
+    // Release old one
+    [_drawIDBuffer release];
```
squidbus: Is this safe, what if the existing draw ID buffer is still being used by a previous draw?

FYI there is an existing `getTempMTLBuffer` function used for other temporaries in draw commands, which is pool allocated. You may be better off just removing the extra draw ID buffer stuff you've added and using that to set up your draw IDs.

denisk-dev: Something like this? (I'll remove the commented out lines before committing)

```
    if (pipeline->needsDrawIDBuffer()) {

        tempDrawIDBuff = cmdEncoder->getTempMTLBuffer(_drawCount * sizeof(uint32_t));

        // cmdEncoder->ensureDrawIDBuffer(_drawCount);   // or the indirect drawCount

        // uint32_t* drawIDs = (uint32_t*)[cmdEncoder->_drawIDBuffer contents];
        uint32_t* drawIDs = (uint32_t*)[tempDrawIDBuff->_mtlBuffer contents];
        for (uint32_t i = 0; i < _drawCount; i++) {
            drawIDs[i] = i;     // sequential IDs starting from 0
        }
    }
```

And then use it like so?

```
            if (drawIdx == 0) {
                [cmdEncoder->_mtlRenderEncoder setVertexBuffer:tempDrawIDBuff->_mtlBuffer
                                                              offset:0
                                                             atIndex:pipeline -> getMetalBufferIndexForDrawID()];   // your reserved index
            } else {
                [cmdEncoder->_mtlRenderEncoder setVertexBufferOffset:drawIdx * sizeof(uint32_t)
                                                             atIndex:pipeline -> getMetalBufferIndexForDrawID()];   // set the correct offset for draw #1+
            }
```

Did a quick test with KSA and it seems to work. Not sure if "offset" needs to use tempDrawIDBuff->_offset along with drawIdx * sizeof(uint32_t).

If that seems correct, I'll be able to drop my changes in MVKCommandBuffer files

squidbus: Yes, and you should use the offset from the temp buffer added on, yes.

denisk-dev: This seems to break it. Am I adding the offset incorrectly in setVertexBufferOffset?

```
if (drawIdx == 0) {
                [cmdEncoder->_mtlRenderEncoder setVertexBuffer:tempDrawIDBuff->_mtlBuffer
                                                              offset:tempDrawIDBuff->_offset
                                                             atIndex:pipeline -> getMetalBufferIndexForDrawID()];   // your reserved index
            } else {
                [cmdEncoder->_mtlRenderEncoder setVertexBufferOffset:tempDrawIDBuff->_offset + drawIdx * sizeof(uint32_t)
                                                             atIndex:pipeline -> getMetalBufferIndexForDrawID()];   // set the correct offset for draw #1+
            }
```

squidbus: Hmm, going by the other usages it should need the offset to have the right part of the pool allocation, for example: https://github.com/KhronosGroup/MoltenVK/blob/a14cbf433414aa95da7046694b069f0f3f9263a5/MoltenVK/MoltenVK/Commands/MVKCmdDraw.mm#L219-L223

denisk-dev: I wonder if it's breaking due to the offset being explicitly used for DrawID mapping? It seems to work if I leave it as 0 for drawIdx == 0 and drawIdx * sizeof(uint32_t) for the others.

squidbus: Ah the issue I think is that you need to offset the pointer you're writing the draw IDs to.

The buffer you get back is the whole buffer for the pooled allocations and offset is your allocated position in that buffer. So both for binding and for accessing the `contents`, you need to apply the offset.

denisk-dev: Ok, this seems to work.

```
        uint32_t* drawIDs = (uint32_t*)[tempDrawIDBuff->_mtlBuffer contents] + tempDrawIDBuff->_offset;
        for (uint32_t i = 0; i < _drawCount; i++) {
            drawIDs[i] = i;     // sequential IDs starting from 0
        }
```

I'm currently using `cmdEncoder->getTempMTLBuffer(_drawCount * sizeof(uint32_t), true, true);` (isPrivate and isDedicated). Do you think I need/should use those?

squidbus: You want to not specify either of the booleans for this, just the size. It shouldn't be private since you need to write to it with IDs from the CPU, and it doesn't need to have its own dedicated allocation.

denisk-dev: Ok, had to fix the way I applied offset, but it's working now without either of those booleans. Let me clean up the code and push the changes

MoltenVKShaderConverter/MoltenVKShaderConverter/SPIRVToMSLConverter.h
```diff
@@ -254,6 +254,8 @@ namespace mvk {
 		bool isPositionInvariant = false;
 		bool needsOutputBuffer = false;
 		bool needsPatchOutputBuffer = false;
+        bool usesDrawId = false;
+        uint32_t drawIDBuffer = 19;
```
squidbus: I think these should be added to `void serialize(Archive & archive, SPIRVToMSLConversionResultInfo& scr)` in pipelines.

squidbus (approved):
LGTM minus one thing. Passes the shader draw parameters CTS tests too.

MoltenVKShaderConverter/MoltenVKShaderConverter/SPIRVToMSLConverter.cpp
```diff
@@ -384,6 +384,10 @@ MVK_PUBLIC_SYMBOL bool SPIRVToMSLConverter::convert(SPIRVToMSLConversionConfigur
 		} else {
 			ctxSI.outIsUsedByShader = pMSLCompiler->is_msl_shader_input_used(ctxSI.shaderVar.location);
 		}
+        
+        if (pMSLCompiler->has_active_builtin(spv::BuiltInDrawIndex, spv::StorageClassInput)) {
```
squidbus: Also one more thing, this check doesn't seem to need to be in the loop that it's a part of, it doesn't use the loop parameter at all.

MoltenVKShaderConverter/MoltenVKShaderConverter/SPIRVToMSLConverter.h
denisk-dev: Just the usesDrawId or drawIDBuffer as well?

MoltenVKShaderConverter/MoltenVKShaderConverter/SPIRVToMSLConverter.cpp
denisk-dev: Does this seem correct? Pulled it out of the for-loop

```
    if (pMSLCompiler && pMSLCompiler->has_active_builtin(spv::BuiltInDrawIndex, spv::StorageClassInput)) {
        conversionResult.resultInfo.usesDrawId = true;
    }

	for (auto& ctxSI : shaderConfig.shaderInputs) {
		if (ctxSI.shaderVar.builtin != spv::BuiltInMax) {
			ctxSI.outIsUsedByShader = pMSLCompiler->has_active_builtin(ctxSI.shaderVar.builtin, spv::StorageClassInput);
		} else {
			ctxSI.outIsUsedByShader = pMSLCompiler->is_msl_shader_input_used(ctxSI.shaderVar.location);
		}
	}
```

MoltenVKShaderConverter/MoltenVKShaderConverter/SPIRVToMSLConverter.h
squidbus: Serialize `usesDrawId`.

Probably you can just remove `drawIDBuffer`, add a constant like `kMVKDrawIDBufferIndex` around [here](https://github.com/KhronosGroup/MoltenVK/blob/a14cbf433414aa95da7046694b069f0f3f9263a5/MoltenVK/MoltenVK/GPUObjects/MVKPipeline.h#L149), and make sure to set the shader conversion config draw ID buffer index to that constant.

Since we're just using a constant buffer ID anyway, no reason to have it be a runtime parameter like this.

MoltenVKShaderConverter/MoltenVKShaderConverter/SPIRVToMSLConverter.cpp
squidbus: Yes, or even just `conversionResult.resultInfo.usesDrawId = pMSLCompiler && pMSLCompiler->has_active_builtin(spv::BuiltInDrawIndex, spv::StorageClassInput);`

MoltenVKShaderConverter/MoltenVKShaderConverter/SPIRVToMSLConverter.h
squidbus: Missed removing `drawIDBuffer` here, and please set `kMVKDrawIDBufferIndex` as the `draw_id_buffer_index` in the compiler options for SPIRV-Cross. Then should be set.

denisk-dev: Something like this, under bool MVKGraphicsPipeline::addVertexShaderToPipeline?

```
	shaderConfig.options.mslOptions.disable_rasterization = !_isRasterizing;
    shaderConfig.options.mslOptions.draw_id_buffer_index = kMVKDrawIDBufferIndex;
```

squidbus: Yeah something like that.

denisk-dev: Sounds good. Thank you for all your help!

squidbus (approved):

PR #2747, opened 2026-06-01
Fix invalidation of cached metal textures

visigoth:
`MVKImage` caches the underlying Metal texture in `_mtlTexture`. There are cases where the underlying source is dynamic and freely vends updated/new texture objects. For instance, `CAMetalLayer` may vend new `CAMetalDrawable` objects each time `nextDrawable` is invoked, which may contain a new texture object. Thus, the caching of the texture object in `MVKImage` is not always correct. This can be demonstrated with my example program at https://github.com/visigoth/mvk-swapchain-texture-aliasing by running `./build-broken/mvk_swapchain_texture_aliasing --drawable-size-jitter 8`.

### DIFF
--- a/MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
@@ -29,8 +29,26 @@
 #import "MTLSamplerDescriptor+MoltenVK.h"
 #import "CAMetalLayer+MoltenVK.h"
 
+#import <IOSurface/IOSurfaceRef.h>
+
 using namespace std;
 using namespace SPIRV_CROSS_NAMESPACE;
+
+static id<MTLTexture> mvkGetRootMTLTexture(id<MTLTexture> tex) {
+    while (tex.parentTexture && tex.parentTexture != tex) {
+      tex = tex.parentTexture;
+    }
+    return tex;
+}
+
+static id<MTLTexture> mvkGetBaseMTLTexture(id<MTLTexture> tex) {
+    return tex.parentTexture ? tex.parentTexture : tex;
+}
+
+static uint32_t mvkGetMTLTextureIOSurfaceID(id<MTLTexture> tex) {
+    IOSurfaceRef ioSurface = tex.iosurface;
+    return ioSurface ? IOSurfaceGetID(ioSurface) : 0;
+}
 
 #pragma mark -
 #pragma mark MVKImagePlane
@@ -1843,17 +1861,29 @@ void MVKImageViewPlane::propagateDebugName() { _imageView->setMetalObjectLabel(_
 id<MTLTexture> MVKImageViewPlane::getMTLTexture() {
     // If we can use a Metal texture view, lazily create it, otherwise use the image texture directly.
     if (_useMTLTextureView) {
+        id<MTLTexture> baseMTLTexture = _imageView->_image->getMTLTexture(_planeIndex);
+
+        if (_mtlTexture && !matchesMTLTextureViewBase(baseMTLTexture)) {
+            lock_guard<mutex> lock(_imageView->_lock);
+            if (_mtlTexture && !matchesMTLTextureViewBase(baseMTLTexture)) {
+                releaseMTLTexture();
+            }
+        }
+
         if ( !_mtlTexture && _mtlPixFmt ) {
 
             // Lock and check again in case another thread created the texture view
             lock_guard<mutex> lock(_imageView->_lock);
-            if (_mtlTexture) { return _mtlTexture; }
+            if (_mtlTexture) {
+              if (!matchesMTLTextureViewBase(baseMTLTexture)) {
+                releaseMTLTexture();
+              } else {
+                return _mtlTexture;
+              }
+            }
 
-            id<MTLTexture> tex = newMTLTexture(); // retained
-            getDevice()->getLiveResources().add(tex);
-            _mtlTexture = tex;
-
-            propagateDebugName();
+            MVKAssert(baseMTLTexture, "Attempting to create an MTLTexture view from a nil base texture.");
+            initializeMTLTexture(newMTLTextureFromBaseMTLTexture(baseMTLTexture)); // retained
         }
         return _mtlTexture;
     } else {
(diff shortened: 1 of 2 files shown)


MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
```diff
@@ -28,10 +28,28 @@
 
 #import "MTLSamplerDescriptor+MoltenVK.h"
 #import "CAMetalLayer+MoltenVK.h"
+#import <IOSurface/IOSurface.h>
```
cdavis5e: This header is not present in the IOS SDK.
```suggestion
#import <IOSurface/IOSurfaceRef.h>
```

MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
```diff
@@ -1848,36 +1861,87 @@ static void signalAndUntrack(const MVKSwapchainSignaler& signaler) {
...
+        return cachedRootIOSurfaceID == rootIOSurfaceID;
+    }
+
+    uint32_t cachedBaseIOSurfaceID = mvkGetMTLTextureIOSurfaceID(cachedBaseMTLTexture);
+    uint32_t baseIOSurfaceID = mvkGetMTLTextureIOSurfaceID(mtlTexture);
+    if (cachedBaseIOSurfaceID || baseIOSurfaceID) {
+        return cachedBaseIOSurfaceID == baseIOSurfaceID;
+    }
+
+    return cachedBaseMTLTexture == mtlTexture && cachedRootMTLTexture == rootMTLTexture;
```
cdavis5e: This expression will always be false. Your code checked it above on line 1893; by this point, we know that the two are unequal.

```suggestion
    return false;
```

Or perhaps you meant:

```suggestion
    return cachedRootMTLTexture == rootMTLTexture;
```
? If you go with this latter one, the check should probably be ordered before the `IOSurface` check on the root `MTLTexture`.

MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
```diff
@@ -1848,36 +1861,87 @@ static void signalAndUntrack(const MVKSwapchainSignaler& signaler) {
 id<MTLTexture> MVKImageViewPlane::getMTLTexture() {
     // If we can use a Metal texture view, lazily create it, otherwise use the image texture directly.
     if (_useMTLTextureView) {
+        id<MTLTexture> baseMTLTexture = _imageView->_image->getMTLTexture(_planeIndex);
+
+        if (_mtlTexture && !matchesMTLTextureViewBase(baseMTLTexture)) {
+            lock_guard<mutex> lock(_imageView->_lock);
+            if (_mtlTexture && !matchesMTLTextureViewBase(baseMTLTexture)) {
```
cdavis5e, line 1868: I get that the apparently redundant check is needed the way this code is structured, since the `_mtlTexture` may have changed between the time of the first check and the time the lock is acquired, but I feel like there's some alternative way to restructure this so that it doesn't check twice.

MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
```diff
@@ -1848,36 +1861,87 @@ static void signalAndUntrack(const MVKSwapchainSignaler& signaler) {
...
+            }
             if (_mtlTexture) { return _mtlTexture; }
 
-            id<MTLTexture> tex = newMTLTexture(); // retained
-            getDevice()->getLiveResources().add(tex);
-            _mtlTexture = tex;
-
-            propagateDebugName();
+            MVKAssert(baseMTLTexture, "Attempting to create an MTLTexture view from a nil base texture.");
+            initializeMTLTexture(newMTLTextureFromBaseMTLTexture(baseMTLTexture));
```
cdavis5e: Bear in mind that this project uses manual retain/release. The points where new retained references are created should be marked for future reference.
```suggestion
            initializeMTLTexture(newMTLTextureFromBaseMTLTexture(baseMTLTexture));    // retained
```

MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
```diff
@@ -1848,36 +1861,87 @@ static void signalAndUntrack(const MVKSwapchainSignaler& signaler) {
...
+    if (cachedRootIOSurfaceID || rootIOSurfaceID) {
+        return cachedRootIOSurfaceID == rootIOSurfaceID;
+    }
+
+    uint32_t cachedBaseIOSurfaceID = mvkGetMTLTextureIOSurfaceID(cachedBaseMTLTexture);
+    uint32_t baseIOSurfaceID = mvkGetMTLTextureIOSurfaceID(mtlTexture);
+    if (cachedBaseIOSurfaceID || baseIOSurfaceID) {
+        return cachedBaseIOSurfaceID == baseIOSurfaceID;
+    }
+
```
cdavis5e: I presume you've made this distinction between the "base" and "root" as an optimistic optimization. In that case, checking the base texture `IOSurface` should probably happen before getting the root texture at all:

```suggestion
    uint32_t cachedBaseIOSurfaceID = mvkGetMTLTextureIOSurfaceID(cachedBaseMTLTexture);
    uint32_t baseIOSurfaceID = mvkGetMTLTextureIOSurfaceID(mtlTexture);
    if (cachedBaseIOSurfaceID || baseIOSurfaceID) {
        return cachedBaseIOSurfaceID == baseIOSurfaceID;
    }

    id<MTLTexture> cachedRootMTLTexture = mvkGetRootMTLTexture(_mtlTexture);
    id<MTLTexture> rootMTLTexture = mvkGetRootMTLTexture(mtlTexture);

    uint32_t cachedRootIOSurfaceID = mvkGetMTLTextureIOSurfaceID(cachedRootMTLTexture);
    uint32_t rootIOSurfaceID = mvkGetMTLTextureIOSurfaceID(rootMTLTexture);
    if (cachedRootIOSurfaceID || rootIOSurfaceID) {
        return cachedRootIOSurfaceID == rootIOSurfaceID;
    }

```

MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
```diff
@@ -1848,36 +1861,87 @@ static void signalAndUntrack(const MVKSwapchainSignaler& signaler) {
...
+        }
+
         if ( !_mtlTexture && _mtlPixFmt ) {
 
             // Lock and check again in case another thread created the texture view
             lock_guard<mutex> lock(_imageView->_lock);
+            if (_mtlTexture && !matchesMTLTextureViewBase(baseMTLTexture)) {
+                releaseMTLTexture();
+            }
             if (_mtlTexture) { return _mtlTexture; }
```
cdavis5e:
```suggestion
            if (_mtlTexture) {
                if (!matchesMTLTextureViewBase(baseMTLTexture)) { releaseMTLTexture(); }
                else { return _mtlTexture; }
            }
```

MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
```diff
@@ -28,10 +28,28 @@
...
 #import "MTLSamplerDescriptor+MoltenVK.h"
 #import "CAMetalLayer+MoltenVK.h"
+#import <IOSurface/IOSurface.h>
 
 using namespace std;
 using namespace SPIRV_CROSS_NAMESPACE;
 
+static id<MTLTexture> mvkGetRootMTLTexture(id<MTLTexture> tex) {
+    id<MTLTexture> rootTex = tex;
+    for (uint32_t i = 0; rootTex.parentTexture && i < 16; i++) {
```
cdavis5e: Why the limit on 16 levels of texture views?

cdavis5e (changes requested):

MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
visigoth: yes, the latter was intended. i'll change it.

MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
visigoth: bookending a conditional check that is (somewhat) likely to fail is a common performance optimization. it may be fair to say that we are unlikely to have contention here, but i don't know how projects use MoltenVK and at least in my project we have a fair amount of threading to record command buffers and such. i can perhaps make the code look a bit nicer, but the pattern should remain for correctness + performance optimization. if you just want the code to look nicest, we could let the performance optimization go and return to it if someone complains? let me know your thoughts.

MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
visigoth: got it; will review this more deeply before posting an updated PR.

MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
visigoth: agreed; thanks. yes, checking the base is worthwhile so that we avoid walking the chain. will change this around.

visigoth: there is some logic that changes in being optimistic here: `cachedBaseIOSurfaceID || baseIOSurfaceID` - so if either of the underlying APIs return non-zero, we'll end up in that eager comparison. we could perhaps be more conservative and say if either API returned zero, fall through to the root texture check. i'm going to code it up that way unless you feel it's wrong.

visigoth: i.e. change the condition to `cachedBaseIOSurfaceID && baseIOSurfaceID`

MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
visigoth: bad assumption from my test. will update to properly walk.

visigoth:
thanks for the quick feedback! i've updated my changes locally. but, before i push and update the PR, please let me know how you feel about my questions above. i'll update appropriately.

MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
visigoth: i have left this as-is in the updated commit.

visigoth:
@cdavis5e i updated the PR per all the actionable feedback. i did not make any changes to the synchronization, and i made the validation check conservative per my comments. lmk how you'd like to proceed.

visigoth:
ping?

billhollings (approved):
Good catch. Thanks for submitting this! LGTM now.

Sorry about the delay in follow up!

PR #2748, opened 2026-06-02
Do not enable blending when unsupported by format

squidbus:
This is heading into UB territory, but if a game insists on trying to blend with a format that does not support it, we can avoid the application crashing from a Metal assert by not setting blending on the pipeline and instead logging our own warning.

Fixes an assert encountered in https://github.com/KhronosGroup/MoltenVK/issues/2054

### DIFF
--- a/MoltenVK/MoltenVK/GPUObjects/MVKPipeline.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKPipeline.mm
@@ -1934,7 +1934,9 @@ void MVKGraphicsPipeline::addFragmentOutputToPipeline(MTLRenderPipelineDescripto
 			uint32_t caLoc = _colorAttachmentLocations[caIdx];
 			if (caLoc == VK_ATTACHMENT_UNUSED) { continue; }
 
-			MTLPixelFormat mtlPixFmt = getPixelFormats()->getMTLPixelFormat(pRendInfo->pColorAttachmentFormats[caIdx]);
+			VkFormat attachFmt = pRendInfo->pColorAttachmentFormats[caIdx];
+			MTLPixelFormat mtlPixFmt = getPixelFormats()->getMTLPixelFormat(attachFmt);
+			bool supportsBlend = getPixelFormats()->getCapabilities(mtlPixFmt) & kMVKMTLFmtCapsBlend;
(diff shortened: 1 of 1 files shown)


MoltenVK/MoltenVK/GPUObjects/MVKPipeline.mm
```diff
@@ -1950,7 +1952,7 @@ static void addCommonImplicitBuffersToShaderConfig(SPIRVToMSLConversionConfigura
             // Don't set the blend state if we're not using this attachment.
             // The pixel format will be MTLPixelFormatInvalid in that case, and
             // Metal asserts if we turn on blending with that pixel format.
```
cdavis5e, line 1954: This comment could probably use expansion to explain why we're excusing enabling blending with an integer format when the spec doesn't require us to do this.

cdavis5e (changes requested):
I suppose pressuring the game devs to fix their broken game isn't an option?

Technically, we're well within our rights to do this, since the behavior _is_ undefined, although I'm not a fan of having to do more work than the spec requires; on the other hand, it's not _that_ much extra code. I just hope this isn't the first of many app-specific hacks and kluges.

Honestly, though, they were lucky it never broke on any other driver. Did they not run against the Validation Layer?

Just one minor but important issue.

MoltenVK/MoltenVK/GPUObjects/MVKPipeline.mm
squidbus: Good point, updated.

squidbus:
> I suppose pressuring the game devs to fix their broken game isn't an option?

I would assume not since it's a 7 year old Windows game running through wine...

> Technically, we're well within our rights to do this, since the behavior _is_ undefined, although I'm not a fan of having to do more work than the spec requires; on the other hand, it's not _that_ much extra code. I just hope this isn't the first of many app-specific hacks and kluges.

Yeah, pretty much I went for this since it's such a small change to avoid crashing, but I wouldn't want to upstream crazy workarounds for much more than this (for example, the game also sometimes uses shaders with vertex attributes not defined in the pipeline...).

cdavis5e (approved):

PR #2752, opened 2026-06-08
Fix incorrect placement heap check in MVKImagePlane::getMTLTexture()

### LINKED ISSUE #2668: Placement check incorrect in `MVKImagePlane::getMTLTexture()`
We have the following checks:

```objc
        if (_image->_is2DViewOn3DImageCompatible && !dvcMem->ensureMTLHeap()) {
            MVKAssert(0, "Creating a 2D view of a 3D texture currently requires a placement heap, which is not available.");
        }
        if (_image->_isBlockTexelViewCompatible && !dvcMem->ensureMTLHeap()) {
            MVKAssert(0, "Creating an uncompressed view of a compressed texture currently requires a placement heap, which is not available.");
        }
```

But `dvcMem->ensureMTLHeap()` always returns true unless `[getMTLDevice() newHeapWithDescriptor: heapDesc]` returns nil. Specifically, there are many cases where a placement heap is unavailable (`_mtlHeap == nil`) and this check passes.

I attempted to fix this with https://github.com/osy/MoltenVK/commit/953f03086bd2bcaa465dc2aa7cb6a74c7b2f1624 but it seems to now fail `dEQP-VK.pipeline.monolithic.render_to_image.dedicated_allocation.3d.small.r8g8b8a8_unorm` (https://github.com/KhronosGroup/MoltenVK/pull/2667#discussion_r2575008021)

BoiledElectricity:
Fixes #2668.

The 2D-view-on-3D and block-texel-view checks in `getMTLTexture()` test `ensureMTLHeap()`, which returns true even when no placement heap exists and only returns false on an actual creation failure, so they never catch the case where the feature needs a heap but none is available. This checks `getMTLHeap()` for a real heap instead, while still calling `ensureMTLHeap()` first so the heap is created when possible (the texture-creation path below depends on that).

The checks are `MVKAssert` (debug only), so release behavior is unchanged.

Verified with VK-GL-CTS on Apple M4 Pro (macOS, Metal 4), no regressions:
- `render_to_image.dedicated_allocation.3d.small.r8g8b8a8_unorm`: pass
- `render_to_image.*.3d.*`: 103/0
- `image.2d_array_compatible.*`: 6/0
- `image.texel_view_compatible.*`: 848/0

### DIFF
--- a/Docs/Whats_New.md
+++ b/Docs/Whats_New.md
@@ -22,6 +22,7 @@ Released TBD
   - `VK_KHR_external_semaphore_fd`
 - Fix buffer and heap out-of-sync in `initExternalMemory()`.
 - Fix incorrect varable usage in `MVKImagePlane::getMTLTexture()`.
+- Fix incorrect placement heap check in `MVKImagePlane::getMTLTexture()`.
 - Fix device loss when using imported `MTLTexture` objects with Metal argument buffers.
 - Update copyright notices to year 2026.
 - Update to latest SPIRV-Cross:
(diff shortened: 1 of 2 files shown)


squidbus (approved):
Seems reasonable enough.

PR #2758, opened 2026-06-15
Fix regressions and crashes on Mac 1 (NVIDIA/Intel) GPUs

### LINKED ISSUE #2727: immediate crashes: vkQuake 1.34.1, MoltenVK version 1.4.1; macos 12.7.6 Monterey; MacBookPro12,1 Intel i5;  Intel(R) Iris(TM) Graphics 6100

vkedwardli:
This PR fixes Mac1 regressions observed on Intel Iris / Iris Pro and NVIDIA GT 750M MacBook Pro hardware

The original version of this PR used broad feature disabling. After review and additional testing, the changes have been narrowed to the specific failing paths.

### 1. Native texture swizzle is advertised but not usable on NVIDIA Mac1
On NVIDIA Mac1, `NVMTLTexture` does not implement `newTextureViewWithPixelFormat:textureType:levels:slices:swizzle:`
Using MoltenVK's native texture swizzle path therefore crashes with an unrecognized selector. This PR marks nativeTextureSwizzle unavailable for NVIDIA Mac1 and restores the shader-side image-view swizzle fallback path.

### 2. Some Mac1 argument-buffer descriptor paths are unstable
Testing showed that some descriptor types need to stay on MoltenVK's discrete-resource path on Mac1

Current fallback list:

- When shader-side swizzle fallback is active:
- `VK_DESCRIPTOR_TYPE_COMBINED_IMAGE_SAMPLER`
- `VK_DESCRIPTOR_TYPE_SAMPLED_IMAGE`
- `VK_DESCRIPTOR_TYPE_INPUT_ATTACHMENT`

These descriptor types may require shader-side image-view swizzle metadata, which is handled correctly by the discrete-resource path.

- Intel Mac1:
  - `VK_DESCRIPTOR_TYPE_STORAGE_IMAGE`

This fixes vkQuake failures around `vkCreateComputePipelines(update_lightmap_pipeline)` / Metal compiler internal errors.

- NVIDIA Mac1:
  - `VK_DESCRIPTOR_TYPE_STORAGE_BUFFER`

This fixes IOAF Code 9 / Invalid Resource device-loss failures observed in Flycast/vkQuake.

### 3. Discarded-fragment store checks fail to compile on Mac1
Flycast's OIT draw shader also triggered a Metal compiler internal error when SPIRV-Cross emitted `simd_is_helper_thread()` for discarded-fragment store checks. 
This PR disables check_discarded_frag_stores on NVIDIA Mac1 to avoid generating that failing shader path.

Attached are some logs for reference, v1.2.9 is used for comparing the recent crashes since v1.3.0
[1.2.9.txt](https://github.com/user-attachments/files/28957211/1.2.9.txt)
[1.3.0.txt](https://github.com/user-attachments/files/28957213/1.3.0.txt)
[1.4.2_92935d4c.txt](https://github.com/user-attachments/files/28957214/1.4.2_92935d4c.txt)

Closes #2727

### DIFF
--- a/MoltenVK/MoltenVK/Commands/MVKCommandEncoderState.mm
+++ b/MoltenVK/MoltenVK/Commands/MVKCommandEncoderState.mm
@@ -272,6 +272,7 @@ static uint32_t getCPUMetaOffset(MVKDescriptorCPULayout layout) {
 
 enum class ImplicitBufferData {
 	BufferSize,
+	TextureSwizzle,
 };
 
 static bool isTexelBuffer(VkDescriptorType type) {
@@ -312,12 +313,15 @@ static void bindImplicitBufferData(uint32_t* target, MVKDescriptorSetLayout* lay
 		uint32_t perDescCount;
 		switch (DataType) {
 			case ImplicitBufferData::BufferSize:     perDescCount = binding.perDescriptorResourceCount.buffer;  break;
+			case ImplicitBufferData::TextureSwizzle: perDescCount = binding.perDescriptorResourceCount.texture; break;
 		}
 		if (perDescCount == 0)
 			continue;
 		size_t stride = descriptorCPUSize(binding.cpuLayout);
 		size_t metaOff = getCPUMetaOffset(binding.cpuLayout);
-		if (metaOff == 0) {
+		if (DataType == ImplicitBufferData::TextureSwizzle && isTexelBuffer(binding.descriptorType)) {
+			mvkClear(target, count);
+		} else if (metaOff == 0) {
 			assert(DataType != ImplicitBufferData::BufferSize && "All buffers should have metadata");
 			mvkClear(target, count);
 		} else {
@@ -332,6 +336,9 @@ static void bindImplicitBufferData(uint32_t* target, MVKDescriptorSetLayout* lay
 							target[i] = reinterpret_cast<const MVKDescriptorMetaImage*>(base)->size;
 						else
 							target[i] = reinterpret_cast<const MVKDescriptorMetaBuffer*>(base)->size;
+						break;
+					case ImplicitBufferData::TextureSwizzle:
+						target[i] = reinterpret_cast<const MVKDescriptorMetaImage*>(base)->swizzle;
 						break;
 				}
 			}
@@ -374,6 +381,10 @@ static void bindDescriptorSets(MVKImplicitBufferData& target,
 		if (setLayout->argBufMode() == MVKArgumentBufferMode::Off) {
 			// If we grab these now, we can be guaranteed the sets are valid
 			// If we wait until draw time, we can only be guaranteed statically used sets are valid
+			if (stride.textureIndex) {
+				mvkEnsureSize(target.textureSwizzles, offsets.textureIndex + stride.textureIndex);
+				bindImplicitBufferData<ImplicitBufferData::TextureSwizzle>(&target.textureSwizzles[offsets.textureIndex], setLayout, set->cpuBuffer, vkStage, varCount);
+			}
 			if (stride.bufferIndex) {
 				mvkEnsureSize(target.bufferSizes, offsets.bufferIndex + stride.bufferIndex);
 				bindImplicitBufferData<ImplicitBufferData::BufferSize>(&target.bufferSizes[offsets.bufferIndex], setLayout, set->cpuBuffer, vkStage, varCount);
@@ -667,6 +678,9 @@ static void bindMetalResources(id<MTLCommandEncoder> encoder,
 		switch (nvbuffer) {
 			case MVKNonVolatileImplicitBuffer::PushConstant:
 				bindImmediateData(encoder, mvkEncoder, pushConstants, common._layout->getPushConstantsLength(), idx, binder);
+				break;
+			case MVKNonVolatileImplicitBuffer::Swizzle:
+				bindImmediateData(encoder, mvkEncoder, getImplicitBindingData(implicitBufferData.textureSwizzles, resourceCounts.textureIndex), idx, binder);
 				break;
 			case MVKNonVolatileImplicitBuffer::BufferSize:
 				bindImmediateData(encoder, mvkEncoder, getImplicitBindingData(implicitBufferData.bufferSizes, resourceCounts.bufferIndex), idx, binder);
--- a/MoltenVK/MoltenVK/Commands/MVKCommandEncoderState.mm
+++ b/MoltenVK/MoltenVK/Commands/MVKCommandEncoderState.mm
@@ -381,7 +381,7 @@ static void bindDescriptorSets(MVKImplicitBufferData& target,
 		if (setLayout->argBufMode() == MVKArgumentBufferMode::Off) {
 			// If we grab these now, we can be guaranteed the sets are valid
 			// If we wait until draw time, we can only be guaranteed statically used sets are valid
-			if (stride.textureIndex) {
+			if (!layout->getMetalFeatures().nativeTextureSwizzle && stride.textureIndex) {
 				mvkEnsureSize(target.textureSwizzles, offsets.textureIndex + stride.textureIndex);
 				bindImplicitBufferData<ImplicitBufferData::TextureSwizzle>(&target.textureSwizzles[offsets.textureIndex], setLayout, set->cpuBuffer, vkStage, varCount);
 			}
@@ -841,6 +841,7 @@ template <typename MTLState>
 static void invalidateDescriptorSetImplicitBuffers(MTLState& state) {
 	invalidateImplicitBuffer(state, MVKNonVolatileImplicitBuffer::BufferSize);
 	invalidateImplicitBuffer(state, MVKNonVolatileImplicitBuffer::DynamicOffset);
+	invalidateImplicitBuffer(state, MVKNonVolatileImplicitBuffer::Swizzle);
 }
 
 static bool isGraphicsStage(MVKShaderStage stage) {
--- a/MoltenVK/MoltenVK/Commands/MVKCommandResourceFactory.mm
+++ b/MoltenVK/MoltenVK/Commands/MVKCommandResourceFactory.mm
@@ -168,6 +168,32 @@ id<MTLRenderPipelineState> MVKCommandResourceFactory::newCmdClearMTLRenderPipeli
 	return rps;
 }
 
+static NSString* getSwizzleComponentExpr(VkComponentSwizzle swizzle, const char* identityComponent, bool isFloat) {
+	switch (swizzle) {
+		case VK_COMPONENT_SWIZZLE_ZERO:     return isFloat ? @"0.0" : @"0";
+		case VK_COMPONENT_SWIZZLE_ONE:      return isFloat ? @"1.0" : @"1";
+		case VK_COMPONENT_SWIZZLE_R:        return @"sample.x";
+		case VK_COMPONENT_SWIZZLE_G:        return @"sample.y";
+		case VK_COMPONENT_SWIZZLE_B:        return @"sample.z";
+		case VK_COMPONENT_SWIZZLE_A:        return @"sample.w";
+		case VK_COMPONENT_SWIZZLE_IDENTITY:
+		default:                            return [NSString stringWithUTF8String: identityComponent];
+	}
+}
+
+static NSString* getSwizzledSampleExpr(uint32_t packedSwizzle, NSString* typeStr) {
+	if (!packedSwizzle) { return @"sample"; }
+
+	VkComponentMapping swizzle = mvkUnpackSwizzle(packedSwizzle);
+	bool isFloat = [typeStr isEqualToString: @"float"];
+	return [NSString stringWithFormat: @"%@4(%@, %@, %@, %@)",
+			typeStr,
+			getSwizzleComponentExpr(swizzle.r, "sample.x", isFloat),
+			getSwizzleComponentExpr(swizzle.g, "sample.y", isFloat),
+			getSwizzleComponentExpr(swizzle.b, "sample.z", isFloat),
+			getSwizzleComponentExpr(swizzle.a, "sample.w", isFloat)];
+}
+
 id<MTLFunction> MVKCommandResourceFactory::newBlitFragFunction(MVKRPSKeyBlitImg& blitKey) {
 	@autoreleasepool {
 		bool isLayeredBlit = blitKey.dstSampleCount > 1 ? getMetalFeatures().multisampleLayeredRendering : getMetalFeatures().layeredRendering;
@@ -272,7 +298,9 @@ id<MTLFunction> MVKCommandResourceFactory::newBlitFragFunction(MVKRPSKeyBlitImg&
 			[msl appendLineMVK];
 		}
 		if (!mvkIsAnyFlagEnabled(blitKey.srcAspect, (VK_IMAGE_ASPECT_DEPTH_BIT | VK_IMAGE_ASPECT_STENCIL_BIT))) {
-			[msl appendFormat: @"    out.color = tex.sample(ce_sampler, varyings.v_texCoord%@%@, level(subRez.lod));", coordArg, sliceArg];
+			[msl appendFormat: @"    auto sample = tex.sample(ce_sampler, varyings.v_texCoord%@%@, level(subRez.lod));", coordArg, sliceArg];
+			[msl appendLineMVK];
+			[msl appendFormat: @"    out.color = %@;", getSwizzledSampleExpr(blitKey.srcSwizzle, typeStr)];
 			[msl appendLineMVK];
 		}
 		[msl appendLineMVK: @"    return out;"];
--- a/MoltenVK/MoltenVK/GPUObjects/MVKDescriptorSet.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKDescriptorSet.mm
@@ -192,14 +192,16 @@ static MVKArgumentBufferMode pickArgumentBufferMode(MVKDevice* dev, const VkDesc
 	if (mvkIsAnyFlagEnabled(pCreateInfo->flags, VK_DESCRIPTOR_SET_LAYOUT_CREATE_PUSH_DESCRIPTOR_BIT))
 		return MVKArgumentBufferMode::Off;
 	auto gpuCaps = dev->getPhysicalDevice()->getMTLDeviceCapabilities();
+	auto* metalFeatures = dev->getPhysicalDevice()->getMetalFeatures();
 	if (dev->getPhysicalDevice()->isNVIDIAGPU() &&
 		gpuCaps.supportsMac1 && !gpuCaps.supportsMac2 &&
-		dev->getPhysicalDevice()->getMetalFeatures()->needsArgumentBufferEncoders) {
+		metalFeatures->needsArgumentBufferEncoders) {
 		for (uint32_t i = 0; i < pCreateInfo->bindingCount; i++) {
 			const VkDescriptorSetLayoutBinding& bind = pCreateInfo->pBindings[i];
 			if (bind.descriptorCount == 0)
 				continue;
-			if (bind.descriptorType == VK_DESCRIPTOR_TYPE_COMBINED_IMAGE_SAMPLER)
+			if (bind.descriptorType == VK_DESCRIPTOR_TYPE_COMBINED_IMAGE_SAMPLER ||
+				(!metalFeatures->nativeTextureSwizzle && bind.descriptorType == VK_DESCRIPTOR_TYPE_SAMPLED_IMAGE))
 				return MVKArgumentBufferMode::Off;
 		}
 	}
@@ -348,17 +350,20 @@ static MVKDescriptorCPULayout pickCPULayout(
 	if (count == 0)
 		return MVKDescriptorCPULayout::None;
 	bool nativeTAtomic = dev->getPhysicalDevice()->getMetalFeatures()->nativeTextureAtomics;
+	bool nativeTextureSwizzle = dev->getPhysicalDevice()->getMetalFeatures()->nativeTextureSwizzle;
 	switch (type) {
 		case VK_DESCRIPTOR_TYPE_COMBINED_IMAGE_SAMPLER:
 			// Multiplanar images are for ycbcr, which requires a swizzle of IDENTITY, so we don't need to store it.
 			// Immutable samplers are accessible from the layout so they also don't need to be stored.
+			if (!nativeTextureSwizzle && !planes.hasYCBCR())
+				return planes.hasImmutableSamplers() ? MVKDescriptorCPULayout::OneIDMeta : MVKDescriptorCPULayout::TwoIDMeta;
 			if (planes.planeCount() > 2)
 				return MVKDescriptorCPULayout::TwoIDMeta;
 			if (planes.planeCount() == 1 || planes.hasNonYCBCR())
 				return MVKDescriptorCPULayout::OneID;
 			return MVKDescriptorCPULayout::OneIDMeta;
 		case VK_DESCRIPTOR_TYPE_SAMPLER:                return planes.hasImmutableSamplers() ? MVKDescriptorCPULayout::None : MVKDescriptorCPULayout::OneID;
-		case VK_DESCRIPTOR_TYPE_SAMPLED_IMAGE:          return MVKDescriptorCPULayout::OneID;
+		case VK_DESCRIPTOR_TYPE_SAMPLED_IMAGE:          return nativeTextureSwizzle ? MVKDescriptorCPULayout::OneID : MVKDescriptorCPULayout::OneIDMeta;
 		case VK_DESCRIPTOR_TYPE_STORAGE_IMAGE:          return nativeTAtomic ? MVKDescriptorCPULayout::OneID : MVKDescriptorCPULayout::TwoID2Meta;
 		case VK_DESCRIPTOR_TYPE_UNIFORM_TEXEL_BUFFER:   return MVKDescriptorCPULayout::OneID;
 		case VK_DESCRIPTOR_TYPE_STORAGE_TEXEL_BUFFER:   return nativeTAtomic ? MVKDescriptorCPULayout::OneID : MVKDescriptorCPULayout::TwoID2Meta;
@@ -1206,7 +1211,7 @@ static void writeDescriptorSetCPUBuffer(
 								if (immutableSamplers[i]->isYCBCR())
 									desc->b = img->getMTLTexture(1);
 								else
-									desc->meta.img = { static_cast<uint32_t>([tex height] * [tex bufferBytesPerRow]) };
+									desc->meta.img = { static_cast<uint32_t>([tex height] * [tex bufferBytesPerRow]), img->getPackedSwizzle() };
 							} else {
 								*desc = {};
 							}
@@ -1222,7 +1227,7 @@ static void writeDescriptorSetCPUBuffer(
 						if (auto* img = reinterpret_cast<MVKImageView*>(static_cast<const VkDescriptorImageInfo*>(src)->imageView)) {
 							id<MTLTexture> tex = img->getMTLTexture();
 							desc->a = tex;
-							desc->meta.img = { static_cast<uint32_t>([tex height] * [tex bufferBytesPerRow]) };
+							desc->meta.img = { static_cast<uint32_t>([tex height] * [tex bufferBytesPerRow]), img->getPackedSwizzle() };
 						} else {
 							*desc = {};
 						}
@@ -1248,7 +1253,7 @@ static void writeDescriptorSetCPUBuffer(
 							desc->c = img->getMTLTexture(2);
 						} else {
 							desc->b = nullptr;
-							desc->meta.img = { static_cast<uint32_t>([tex height] * [tex bufferBytesPerRow]) };
+							desc->meta.img = { static_cast<uint32_t>([tex height] * [tex bufferBytesPerRow]), img->getPackedSwizzle() };
 						}
 					} else {
 						*desc = {};
@@ -1259,7 +1264,7 @@ static void writeDescriptorSetCPUBuffer(
 					if (img) {
 						id<MTLTexture> tex = img->getMTLTexture();
 						desc->a = tex;
-						desc->meta.img = { static_cast<uint32_t>([tex height] * [tex bufferBytesPerRow]) };
+						desc->meta.img = { static_cast<uint32_t>([tex height] * [tex bufferBytesPerRow]), img->getPackedSwizzle() };
 					} else {
 						desc->a = nil;
 						desc->meta = {};
@@ -1293,7 +1298,7 @@ static void writeDescriptorSetCPUBuffer(
 							desc->a = tex;
 							desc->b = [tex buffer];
 							desc->offset = [tex bufferOffset];
-							desc->meta.img = { static_cast<uint32_t>([tex height] * [tex bufferBytesPerRow]) };
+							desc->meta.img = { static_cast<uint32_t>([tex height] * [tex bufferBytesPerRow]), img->getPackedSwizzle() };
 						} else {
 							*desc = {};
 						}
--- a/MoltenVK/MoltenVK/GPUObjects/MVKDescriptorSet.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKDescriptorSet.mm
@@ -193,6 +193,17 @@ static MVKArgumentBufferMode pickArgumentBufferMode(MVKDevice* dev, const VkDesc
 		return MVKArgumentBufferMode::Off;
 	auto gpuCaps = dev->getPhysicalDevice()->getMTLDeviceCapabilities();
 	auto* metalFeatures = dev->getPhysicalDevice()->getMetalFeatures();
+	if (!metalFeatures->nativeTextureSwizzle) {
+		for (uint32_t i = 0; i < pCreateInfo->bindingCount; i++) {
+			const VkDescriptorSetLayoutBinding& bind = pCreateInfo->pBindings[i];
+			if (bind.descriptorCount == 0)
+				continue;
+			if (bind.descriptorType == VK_DESCRIPTOR_TYPE_COMBINED_IMAGE_SAMPLER ||
+				bind.descriptorType == VK_DESCRIPTOR_TYPE_SAMPLED_IMAGE ||
+				bind.descriptorType == VK_DESCRIPTOR_TYPE_INPUT_ATTACHMENT)
+				return MVKArgumentBufferMode::Off;
+		}
+	}
 	if (dev->getPhysicalDevice()->isNVIDIAGPU() &&
 		gpuCaps.supportsMac1 && !gpuCaps.supportsMac2 &&
 		metalFeatures->needsArgumentBufferEncoders) {
@@ -201,8 +212,7 @@ static MVKArgumentBufferMode pickArgumentBufferMode(MVKDevice* dev, const VkDesc
 			if (bind.descriptorCount == 0)
 				continue;
 			if (bind.descriptorType == VK_DESCRIPTOR_TYPE_COMBINED_IMAGE_SAMPLER ||
-				bind.descriptorType == VK_DESCRIPTOR_TYPE_INPUT_ATTACHMENT ||
-				(!metalFeatures->nativeTextureSwizzle && bind.descriptorType == VK_DESCRIPTOR_TYPE_SAMPLED_IMAGE))
+				bind.descriptorType == VK_DESCRIPTOR_TYPE_INPUT_ATTACHMENT)
 				return MVKArgumentBufferMode::Off;
 		}
 	}
--- a/MoltenVK/MoltenVK/GPUObjects/MVKDescriptorSet.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKDescriptorSet.mm
@@ -191,7 +191,6 @@ static MVKArgumentBufferMode pickArgumentBufferMode(MVKDevice* dev, const VkDesc
 	// Push descriptors are always binding-based
 	if (mvkIsAnyFlagEnabled(pCreateInfo->flags, VK_DESCRIPTOR_SET_LAYOUT_CREATE_PUSH_DESCRIPTOR_BIT))
 		return MVKArgumentBufferMode::Off;
-	auto gpuCaps = dev->getPhysicalDevice()->getMTLDeviceCapabilities();
 	auto* metalFeatures = dev->getPhysicalDevice()->getMetalFeatures();
 	if (!metalFeatures->nativeTextureSwizzle) {
 		for (uint32_t i = 0; i < pCreateInfo->bindingCount; i++) {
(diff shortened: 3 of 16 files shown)


MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
```diff
@@ -2491,15 +2491,15 @@
...
-		_metalFeatures.memoryBarriers = true;
 	}
 
     if (supportsMTLGPUFamily(Mac2)) {
 		_metalFeatures.multisampleLayeredRendering = true;
 		_metalFeatures.stencilFeedback = true;
 		_metalFeatures.depthResolve = true;
 		_metalFeatures.stencilResolve = true;
+		_metalFeatures.simdPermute = true;
+		_metalFeatures.memoryBarriers = true;
```
squidbus: These are both listed as supported for Mac1 on the last feature set table to include it, I don't think these should be moved unconditionally. If you really need to, do a check for not NVIDIA vendor instead.

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
```diff
@@ -2677,7 +2677,7 @@
 
 	// Metal argument buffer support for descriptor sets is supported on macOS 11.0 or later,
 	// or on older versions of macOS using an Intel GPU, or on iOS & tvOS 16.0 or later (Metal 3).
-	_metalFeatures.descriptorSetArgumentBuffers = (mvkOSVersionIsAtLeast(11.0, 16.0, 1.0) ||
+	_metalFeatures.descriptorSetArgumentBuffers = ((mvkOSVersionIsAtLeast(11.0, 16.0, 1.0) && supportsMTLGPUFamily(Mac2)) ||
```
squidbus: `The redesigned Argument Buffer architecture (writing descriptors directly to unified memory) violates Tier 1 Argument Buffer limitations on NVIDIA hardware.`

You'll have to elaborate on this, we already check the argument buffers tier (`argumentBuffersTier` set right above...) and act appropriately (`needsArgumentBufferEncoders` below...).

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
```diff
@@ -2733,7 +2733,15 @@
 	_metalFeatures.events = true;
 	_metalFeatures.ioSurfaces = true;
 	_metalFeatures.renderWithoutAttachments = true;
-	_metalFeatures.nativeTextureSwizzle = true;
+	_metalFeatures.nativeTextureSwizzle = mvkOSVersionIsAtLeast(10.15, 13.0, 1.0);
+	if (_metalFeatures.nativeTextureSwizzle) {
+		MTLTextureDescriptor* mtlTexDesc = [MTLTextureDescriptor texture2DDescriptorWithPixelFormat: MTLPixelFormatRGBA8Unorm width: 1 height: 1 mipmapped: NO];
+		id<MTLTexture> mtlTex = [_mtlDevice newTextureWithDescriptor: mtlTexDesc];
+		if ( ! [mtlTex respondsToSelector: @selector(newTextureViewWithPixelFormat:textureType:levels:slices:swizzle:)] ) {
```
squidbus: The hardware feature flag check makes sense but there is no reason to do a `respondsToSelector` check here, the API was added in macOS 10.15, and our minimum is 11.

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
```diff
@@ -2733,7 +2733,15 @@
 	_metalFeatures.events = true;
 	_metalFeatures.ioSurfaces = true;
 	_metalFeatures.renderWithoutAttachments = true;
-	_metalFeatures.nativeTextureSwizzle = true;
+	_metalFeatures.nativeTextureSwizzle = mvkOSVersionIsAtLeast(10.15, 13.0, 1.0);
```
squidbus: Our minimum OS is 11, so this does nothing. You should do a check for the appropriate hardware families. Ideally remove this line and put the `_metalFeatures.nativeTextureSwizzle = true;` under the correct hardware families.

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
vkedwardli: This check is added because I was facing `[NVMTLTexture newTextureViewWithPixelFormat:textureType:levels:slices:swizzle:]: unrecognized selector sent to instance`

Similar to https://github.com/google/filament/issues/3754 or https://github.com/Vita3K/Vita3K/issues/3502

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
vkedwardli: In commit 30c5b926, MoltenVK's Argument Buffer implementation was fundamentally redesigned to write descriptors directly into unified memory pools. While highly optimized for modern GPUs, this new architecture violates the strict limitations of Tier 1 Argument Buffers on older Mac hardware, causing the Metal driver to instantly kill the render pass.

I'm getting `Invalid Resource (IOAF code 9)` error as shown in the attached 1.4.2_92935d4c.txt

Maintaining a parallel, legacy descriptor architecture just to support Tier 1 Argument Buffers on older hardware seems impractical?

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
squidbus: Right, but the value you set to `nativeTextureSwizzle` is wrong. Fix that and you don't need this.

(also the second you posted is unsupported case because it's using OCLP)

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
vkedwardli: Thanks for the info, this is fixing the hard crash on my nvidia, I will see if I can condition it without hard coding vendors

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
squidbus: Please point to the specific code and documentation language that's in violation here, we are using argument encoders for tier 1 support as documented.

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
squidbus: Also side note, if this is being used the deprecation language should be removed [here](https://github.com/KhronosGroup/MoltenVK/blob/5843e5da2e1f561261cb06a2f859ad39663d054f/MoltenVK/MoltenVK/API/mvk_deprecated_api.h#L121).

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
squidbus: It's okay to disable for vendors if needed, for example we disable heaps by default for AMD: https://github.com/vkedwardli/MoltenVK/blob/d3941ee1c260f0a2a72fe6375d448929a94a4af7/MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm#L2460

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
vkedwardli: My earlier statement was too broad, after looking at the specific code around argument buffers, this appears to be a NVIDIA Mac1 runtime issue with combined image sampler descriptor set layouts encoded through Metal argument buffers

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
vkedwardli: You’re right, the Mid-2012/OCLP case can be ignored.

The supported target here is GT 750M on the Late-2013 Retina MacBook Pro, which officially supports macOS 11. Since native texture swizzle crashes on this NVIDIA Mac1 device, I think we need to restore the non-native shader-side swizzle path when `nativeTextureSwizzle` is false

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
vkedwardli: Thank you, wasn't aware of this

vkedwardli:
I force-pushed a cleaned version of this PR.

The original broad changes around `simdPermute`, `memoryBarriers`, and global argument-buffer disabling have been removed.

1. `nativeTextureSwizzle = false` for NVIDIA Mac1, plus shader-side swizzle fallback restore
- Fixes the unrecognized selector crash: `[NVMTLTexture newTextureViewWithPixelFormat:textureType:levels:slices:swizzle:]: unrecognized selector`

2. Avoid Metal argument buffers for NVIDIA Mac1 descriptor set layouts containing `VK_DESCRIPTOR_TYPE_COMBINED_IMAGE_SAMPLER`
- Fixes the Flycast startup/device-loss path where combined image sampler descriptors through Tier 1 argument buffers caused `Invalid Resource (IOAF code 9)`

3. Avoid Metal argument buffers for NVIDIA Mac1 descriptor set layouts containing `VK_DESCRIPTOR_TYPE_INPUT_ATTACHMENT`
- Fixes the Flycast Vulkan OIT resolve crash. The reduced shader showed `tex.read(...)` from the input attachment crashed with `Invalid Resource (IOAF code 9)` through argument buffers, and stopped crashing with discrete bindings

4. Avoid argument buffers for `VK_DESCRIPTOR_TYPE_SAMPLED_IMAGE` on NVIDIA Mac1 when shader-side swizzle fallback is active
- Needed because native texture swizzle is disabled on this hardware, and MoltenVK must bind swizzle metadata alongside the image view

5. Disable discarded-fragment store checks on NVIDIA Mac1
- Fixes the Flycast Vulkan OIT draw-shader compile failure:
`Render pipeline compile failed (Error code 2): Compiler encountered an internal error`
- The generated MSL failed when SPIRV-Cross emitted:
`gl_HelperInvocation = simd_is_helper_thread();`

Verified with a clean build on GT 750M / GPU Family Mac 1: Flycast Vulkan starts, Vulkan OIT renders correctly, and the swizzle repro passes descriptor sampling and blit-path checks.

vkedwardli:
Attached is a small test for checking the shader-side image-view swizzle fallback

It covers:
  - mutable combined image sampler
  - mutable separate sampled image + sampler
  - immutable combined image sampler
  - immutable separate sampled image + sampler
  - stencil descriptor sampling
  - color blit path
  - stencil blit format support query
 
Result on NVIDIA GT 750M / Mac1
```
S8_UINT optimal features: sampled=1 transferDst=1
S8_UINT stencil blit support: src=0 dst=0
D24_UNORM_S8_UINT stencil blit support: src=0 dst=0
D32_SFLOAT_S8_UINT stencil blit support: src=0 dst=0
descriptor source RGBA: 17 34 51 68
mutable combined BGRA: 51 34 17 68
mutable separate BGRA: 51 34 17 68
immutable combined BGRA: 51 34 17 68
immutable separate BGRA: 51 34 17 68
stencil source S8: 123
mutable stencil combined ZRZ1: 0 123 0 1
mutable stencil separate ZRZ1: 0 123 0 1
immutable stencil combined ZRZ1: 0 123 0 1
immutable stencil separate ZRZ1: 0 123 0 1
stencil expected ZRZ1: 0 123 0 1
blit src raw BGRA: 51 34 17 68
blit dst raw RGBA: 17 34 51 68
descriptor expected BGRA: 51 34 17 68
PASS: swizzled color/stencil descriptor sampling and color blit paths produced expected results
```

I also tested the fallback path on Apple Silicon by temporarily forcing `nativeTextureSwizzle = false`. That exposed an issue where Metal argument buffers did not apply the shader-side swizzle metadata correctly; the latest patch keeps sampled image descriptors on the discrete binding path when native texture swizzle is unavailable.

[swizzle-repro.zip](https://github.com/user-attachments/files/29029448/swizzle-repro.zip)

p.s. I understand unsupported macOS configurations are out of scope. My main concern here is preserving compatibility for officially supported Intel Mac1 systems where possible, since MoltenVK remains especially important for older Intel Macs while newer Apple Silicon-focused paths continue to evolve separately

MoltenVK/MoltenVK/Commands/MVKCommandEncoderState.mm
```diff
@@ -374,6 +381,10 @@ static void bindDescriptorSets(MVKImplicitBufferData& target,
 		if (setLayout->argBufMode() == MVKArgumentBufferMode::Off) {
 			// If we grab these now, we can be guaranteed the sets are valid
 			// If we wait until draw time, we can only be guaranteed statically used sets are valid
+			if (stride.textureIndex) {
```
squidbus:
```suggestion
			if (!layout->getMetalFeatures().nativeTextureSwizzle && stride.textureIndex) {
```
Seems to be missing feature check from the original code.

MoltenVK/MoltenVK/Commands/MVKCommandEncoderState.mm
squidbus, line 845:
```suggestion
	invalidateImplicitBuffer(state, MVKNonVolatileImplicitBuffer::Swizzle);
}
```

Seems to be missing from the original code.

MoltenVK/MoltenVK/Commands/MVKCommandResourceFactory.mm
squidbus, line 297: The original swizzle emulation code would swizzle stencil as well, is this needed?

MoltenVK/MoltenVK/GPUObjects/MVKDescriptorSet.mm
```diff
@@ -191,6 +191,21 @@ static MVKArgumentBufferMode pickArgumentBufferMode(MVKDevice* dev, const VkDesc
 	// Push descriptors are always binding-based
 	if (mvkIsAnyFlagEnabled(pCreateInfo->flags, VK_DESCRIPTOR_SET_LAYOUT_CREATE_PUSH_DESCRIPTOR_BIT))
 		return MVKArgumentBufferMode::Off;
+	auto gpuCaps = dev->getPhysicalDevice()->getMTLDeviceCapabilities();
+	auto* metalFeatures = dev->getPhysicalDevice()->getMetalFeatures();
+	if (dev->getPhysicalDevice()->isNVIDIAGPU() &&
+		gpuCaps.supportsMac1 && !gpuCaps.supportsMac2 &&
```
squidbus: `supportsMac1` is always true here.

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
```diff
@@ -2734,6 +2734,11 @@
 	_metalFeatures.ioSurfaces = true;
 	_metalFeatures.renderWithoutAttachments = true;
 	_metalFeatures.nativeTextureSwizzle = true;
+
+	// NVIDIA Mac1 textures do not support native texture swizzle.
+	if (isNVIDIAGPU() && supportsMTLGPUFamily(Mac1) && !supportsMTLGPUFamily(Mac2)) {
```
squidbus: `supportsMTLGPUFamily(Mac1)` always true.

MoltenVK/MoltenVK/GPUObjects/MVKPipeline.mm
```diff
@@ -1649,6 +1652,10 @@ static void addCommonImplicitBuffersToShaderConfig(SPIRVToMSLConversionConfigura
 		shaderConfig.options.mslOptions.capture_output_to_buffer = false;
 		shaderConfig.options.mslOptions.fixed_subgroup_size = mvkIsAnyFlagEnabled(pFragmentSS->flags, VK_PIPELINE_SHADER_STAGE_CREATE_ALLOW_VARYING_SUBGROUP_SIZE_BIT) ? 0 : mtlFeats.maxSubgroupSize;
 		shaderConfig.options.mslOptions.check_discarded_frag_stores = true;
+		auto gpuCaps = getPhysicalDevice()->getMTLDeviceCapabilities();
+		if (getPhysicalDevice()->isNVIDIAGPU() && gpuCaps.supportsMac1 && !gpuCaps.supportsMac2) {
```
squidbus: `gpuCaps.supportsMac1` always true.

MoltenVK/MoltenVK/GPUObjects/MVKPipeline.mm
```diff
@@ -3174,7 +3186,7 @@ void mvkValidateCeralArchiveDefinitions() {
 	missingBytes += mvkValidateCerealArchiveSize<mvk::MSLResourceBinding>(2);
 	missingBytes += mvkValidateCerealArchiveSize<mvk::DescriptorBinding>();
 	missingBytes += mvkValidateCerealArchiveSize<mvk::SPIRVToMSLConversionConfiguration>(106);	// Contains collection
-	missingBytes += mvkValidateCerealArchiveSize<mvk::SPIRVToMSLConversionResultInfo>(41);		// Contains collection
+	missingBytes += mvkValidateCerealArchiveSize<mvk::SPIRVToMSLConversionResultInfo>(40);		// Contains collection
```
squidbus, line 3189: Just to confirm, you've verified that MVK isn't printing out any warnings about the cereal sizes on init? The serialization stuff can be a little tricky.

vkedwardli: Yes, with `MVK_CONFIG_DEBUG=1 MVK_CONFIG_LOG_LEVEL=3`, MoltenVK initializes cleanly and does not print any `[MVK-BUILD-ERROR] Cereal serialization Archive ...` warning.

 I also tested this explicitly by reverting only the validator count change. MoltenVK then prints:
```
[MVK-BUILD-ERROR] Cereal serialization Archive for mvk::SPIRVToMSLConversionResultInfo is not completely defined. Missing -1 bytes. Struct size is 104 (including an expected 41 bytes of padding) and Cereal Archive size is 64. The Cereal Archive definition may be missing members.
Assertion failed: (missingBytes == 0 && "Cereal Archive definitions incomplete. See previous logged errors."), function mvkValidateCeralArchiveDefinitions, file MVKPipeline.mm, line 3193.
```

MoltenVK/MoltenVK/GPUObjects/MVKDescriptorSet.mm
vkedwardli: I used `isNVIDIAGPU() && supportsMac1 && !supportsMac2` mainly to document the intended scope as "NVIDIA Mac1 only"

would you prefer `isNVIDIAGPU() && !supportsMac2` or just `isNVIDIAGPU()`? since all Metal-capable NVIDIA is Mac1 only

squidbus: Would prefer the simplest version of the check. If all supported NVIDIA GPUs are Mac1 anyway we can just leave it as an NVIDIA GPU check only.

MoltenVK/MoltenVK/Commands/MVKCommandResourceFactory.mm
vkedwardli: I checked this with the updated swizzle-repro

For descriptor sampling, stencil does need the shader-side image-view swizzle fallback when native texture swizzle is unavailable. The repro now covers `S8_UINT` with a `ZERO/R/ZERO/ONE` component mapping.

For `vkCmdBlitImage`, the tested stencil formats do not advertise stencil blit support:
```
S8_UINT stencil blit support: src=0 dst=0
D24_UNORM_S8_UINT stencil blit support: src=0 dst=0
D32_SFLOAT_S8_UINT stencil blit support: src=0 dst=0
```

So I kept shader-side stencil swizzle support for descriptor sampling, but did not add stencil swizzling to the blit shader path

MoltenVK/MoltenVK/Commands/MVKCommandEncoderState.mm
vkedwardli: included, thanks a lot

MoltenVK/MoltenVK/Commands/MVKCommandEncoderState.mm
vkedwardli: included, thanks a lot

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
squidbus: Actually I missed something with this one. `nativeTextureSwizzle` is not supported at all for `Mac1`, would ask that you just move the `_metalFeatures.nativeTextureSwizzle = true;` under the existing blocks of features for `Mac2` and `Apple1`.

MoltenVK/MoltenVK/GPUObjects/MVKPipeline.mm
squidbus: Also ideally some kind of comment on why this flag needs to be false.

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
vkedwardli: Do I need to check and verify this on some `Mac1` Intel iGPU? I'm not familiar with the feature sets

squidbus: The last version of the feature set table with `Mac1` included is here: https://web.archive.org/web/20220506093727/https://developer.apple.com/metal/Metal-Feature-Set-Tables.pdf

You can see "Texture swizzle" has no check mark for `Mac1`, and is checked for `Apple1` and up and `Mac2`.

Previously `Mac2` was being set unconditionally for macOS, hence it was assumed `Mac1` wasn't a concern anymore. However as we can see this has proven untrue so things like emulated texture swizzle are actually needed for any `Mac1` GPU we still support.

vkedwardli: wow this table is very useful (for my other emulator related features), thanks a lot.

Updated.

MoltenVK/MoltenVK/GPUObjects/MVKDescriptorSet.mm
vkedwardli: Done

MoltenVK/MoltenVK/GPUObjects/MVKPipeline.mm
vkedwardli: Added, thanks

squidbus (approved):
Seems okay, I'd like if someone could take a second look though.

squidbus:
There's actually one additional thing... based on https://github.com/KhronosGroup/MoltenVK/issues/2727 it may just be simpler to disable argument buffers on all Mac1 GPUs rather than doing the NVIDIA workaround for argument buffers, it seems like there's some broken stuff going on in general. So only enable argument buffers use if `Mac2` or `Apple1` when its determined in the `MVKDevice`.

vkedwardli:
Oh, may be I can check Intel Mac1 also and see what's going on for the argument buffers and see if my fix is specific to NVIDIA or should actually target Mac1

vkedwardli:
## Now this PR is compatible with vkQuake, Flycast Vulkan, Flycast Vulkan OIT for Intel Mac1 and NVIDIA Mac1
Here's the latest findings:
for vkQuake, MacPorts build:
```
[mvk-info] MoltenVK version 1.4.1, supporting Vulkan version 1.4.334.
[mvk-info] GPU device:
	model: Intel Iris Pro Graphics
	type: Integrated
	vendorID: 0x8086
	deviceID: 0x0d26
	pipelineCacheUUID: DB445FF2-0B07-0200-0000-000100000000
	GPU memory available: 1536 MB
	GPU memory used: 0 MB
	Metal Shading Language 2.3
	supports the following GPU Features:
		GPU Family Mac 2
		Read-Write Texture Tier 1

-[MTLHeapDescriptorInternal validateWithDevice:]:327: failed assertion `Heap Descriptor Validation Placement heap type is not supported.'
```
Which is expected, since it does not contain the GPUFamilyMac1 fix.

With latest lib on the main branch:
```
[mvk-info] MoltenVK version 1.4.2, supporting Vulkan version 1.4.350.
[mvk-info] GPU device:
	model: Intel Iris Pro Graphics
	type: Integrated
	vendorID: 0x8086
	deviceID: 0x0d26
	pipelineCacheUUID: 5843E5DA-0B07-0000-0000-000100000000
	GPU memory available: 1536 MB
	GPU memory used: 0 MB
	Metal Shading Language 2.3
	supports the following GPU Features:
		GPU Family Mac 1
		Read-Write Texture Tier 1


[mvk-info] Compiling Metal shader with FastMath enabled and PreserveInvariance disabled.
[mvk-error] VK_ERROR_INITIALIZATION_FAILED: Compute pipeline compile failed (Error code 2):
Compiler encountered an internal error.

ERROR-OUT BEGIN


QUAKE ERROR: vkCreateComputePipelines failed (update_lightmap_pipeline) with code -3
STACK TRACE:
0x10bec12b3 0x10bec103a 0x10bdd5978 0x10bdf043b 0x10bdeaa58 0x10bdd9b9a 0x10bdaeac9 0x10bead327 0x10bdf96ea 0x10be24c7c 0x7fff2061df3d 
0 : 0   vkquake                             0x000000010bec12b3 Sys_StackTrace + 51
1 : 1   vkquake                             0x000000010bec103a Sys_Error + 362
2 : 2   vkquake                             0x000000010bdd5978 R_CreatePipelines + 16776
3 : 3   vkquake                             0x000000010bdf043b GL_CreateRenderResources + 21995
4 : 4   vkquake                             0x000000010bdeaa58 GL_BeginRendering + 104
5 : 5   vkquake                             0x000000010bdd9b9a SCR_UpdateScreen + 1610
6 : 6   vkquake                             0x000000010bdaeac9 Con_Printf + 985
7 : 7   vkquake                             0x000000010bead327 S_Init + 279
8 : 8   vkquake                             0x000000010bdf96ea Host_Init + 282
9 : 9   vkquake                             0x000000010be24c7c main + 236
10: 10  libdyld.dylib                       0x00007fff2061df3d start + 1
```
Still crashing.


With my NVIDIA only patch:
```
========= Quake Initialized =========

execing quake.rc
execing default.cfg
execing config.cfg
couldn't exec autoexec.cfg
3 demo(s) in loop
Playing demo from demo1.dem.



the Necropolis
Using protocol 15
zsh: segmentation fault
```
Still crashing, then I identified `VK_DESCRIPTOR_TYPE_STORAGE_IMAGE` as the remaining Intel Mac1 argument-buffer case needed by vkQuake


After adding the `VK_DESCRIPTOR_TYPE_STORAGE_IMAGE` skip:
Screenshot 2026-06-19 at 7 01 11 AM


For Flycast:
With latest lib on the main branch:
Screenshot 2026-06-19 at 5 07 55 AM

With my NVIDIA only patch:
Screenshot 2026-06-19 at 5 06 57 AM

But when testing with the Vulkan OIT renderer:
```
[mvk-error] VK_ERROR_INITIALIZATION_FAILED: Render pipeline compile failed (Error code 2):
Compiler encountered an internal error.
2026-06-19 05:16:23.048 Flycast[24731:7337502] rend/vulkan/oit/oit_renderer.cpp:114 E[RENDERER]: Vulkan system error vk::Device::createGraphicsPipelineUnique: ErrorInitializationFailed
2026-06-19 05:16:23.051 Flycast[24731:7337502] rend/vulkan/oit/oit_renderer.cpp:32 N[RENDERER]: OITVulkanRenderer::Init
```
So it requires `shaderConfig.options.mslOptions.check_discarded_frag_stores = false` on Intel also.

After disabling, Flycast OIT runs perfectly.


Current argument-buffer fallback list from testing:

- All Mac1 devices when `nativeTextureSwizzle == false`:
  - `VK_DESCRIPTOR_TYPE_COMBINED_IMAGE_SAMPLER`
  - `VK_DESCRIPTOR_TYPE_SAMPLED_IMAGE`
  - `VK_DESCRIPTOR_TYPE_INPUT_ATTACHMENT`

Reason: these descriptor types can require shader-side image-view swizzle metadata. That metadata is handled correctly by the discrete-resource path, but not reliably through argument buffers on Mac1.

- Intel Mac1 with argument-buffer encoders:
  - `VK_DESCRIPTOR_TYPE_STORAGE_IMAGE`

Reason: vkQuake still failed after the generic image-descriptor fallback. Disabling argument buffers for storage images fixed the `vkCreateComputePipelines(update_lightmap_pipeline)`/internal compiler failure path and allowed rendering.

- NVIDIA Mac1 with argument-buffer encoders:
  - `VK_DESCRIPTOR_TYPE_STORAGE_BUFFER`

Reason: Flycast/vkQuake IOAF `Invalid Resource` failures persisted until storage-buffer descriptors were also forced to the discrete-resource path on NVIDIA Mac1.

p.s. Intel fix was verified on Iris Pro `0x0d26` and Iris `0x0a2e`

PR #2761, opened 2026-06-22
Do not rely on extension enable for vkGetPhysicalDeviceToolPropertiesEXT proc addr

squidbus:
`vkGetPhysicalDeviceToolProperties` was previously changed to be marked as an instance entrypoint in https://github.com/KhronosGroup/MoltenVK/pull/2598, as it is a physical device entrypoint.

However, labeling it this way requires the corresponding instance extension to be enabled pre-1.3. Since `VK_EXT_tooling_info` is a device extension, this is impossible.

Add a labeling for instance entrypoints which belong to device extensions, and check that along with `isDevice` when checking if it should be returned from the instance. I've checked with other drivers that this entrypoint is returned unconditionally at the instance level.


billhollings (approved):
LGTM. Thanks.

PR #2762, opened 2026-06-29
WIP: Defer residency set removal until referencing command buffers complete

cyyever:
## Problem

With `MTLResidencySet` enabled (Metal 3+), MoltenVK removed an allocation from the residency set the moment the app destroyed the resource — but committed command buffers may still reference it through argument-buffer descriptors, and per-draw `useResource:` tracking is off while a residency set exists. Reading non-resident memory on Apple GPUs loses the device or returns undefined data.

Two independent defects, one per commit:

**1 — Defer removals until referencing command buffers complete.** Removals are queued, tagged with the newest command-buffer sequence committed before the destroy, and executed one drain after that sequence completes (also on present completions; `waitIdle` and teardown flush the rest). Fixes `VK_ERROR_DEVICE_LOST`.

**2 — Reference-count membership by underlying allocation.** Metal dedups set membership by underlying allocation, which live resources can share (heap sub-allocations, buffer texture views); removing one destroyed resource evicted the shared allocation for its still-live siblings. Membership is now reference-counted per base allocation. Fixes full-screen white-out.

## Validation

macOS, Apple M2 (x86_64 under Rosetta), a DXVK D3D9 game via Wine with the Steam overlay injected — its continuous resource churn triggered both defects within seconds of launch. After: no device loss, no Metal validation failures, and the white-out is gone across repeated sessions. Standalone smoke tests (instance/device/buffer lifecycle, offscreen render + readback) also pass.

A separate color-fidelity issue on this stack (a subtle tint sampled as grayscale) is **not** residency-related — it reproduces with removals fully disabled — and is out of scope here.

Investigation, fix, and in-game validation done with the assistance of Claude Code.

### DIFF
--- a/MoltenVK/MoltenVK/GPUObjects/MVKDevice.h
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKDevice.h
@@ -31,6 +31,7 @@
 #include <string>
 #include <mutex>
 #include <set>
+#include <unordered_map>
 #include <os/lock.h>
 
 #import <Metal/Metal.h>
@@ -986,10 +987,10 @@ public:
 	void drainResidencyRemovals() {}
 	void flushResidencyRemovalsIfIdle() {}
 #else
-	/** Adds the allocation to the residency set, cancelling any pending removal of it. */
+	/** Adds the resource's underlying allocation to the residency set (reference-counted). */
 	void makeResident(id<MTLAllocation> allocation);
 
-	/** Queues removal of the allocation from the residency set once no command buffer can reference it. */
+	/** Queues removal of the resource's underlying allocation once no command buffer can reference it. */
 	void removeResidency(id<MTLAllocation> allocation);
 
 	/** Registers a command buffer being committed and returns its tracking sequence. */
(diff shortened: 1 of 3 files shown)


squidbus:
This seems like a workaround rather than a fix, we should be handling the lifetime ourselves in this case, which is why retain is disabled.

squidbus:
Also, experimentally, calling `addAllocation` on a residency set with a resource does retain a reference to it (`retainCount` increases by 1), so I'm not sure what you're referring to there. Most likely this is completely unrelated to residency set semantics and is simply because freeing the Vulkan resource removes it from residency.

billhollings:
Marking as WIP until questions above are resolved.

etang-cw:
> MoltenVK removed an allocation from the residency set the moment the app destroyed the resource

Can you show a way to use Vulkan that would be broken by doing this?

All the object destruction commands I know of have things like this: 
[VUID-vkDestroyImage-image-01000
All submitted commands that refer to `image`, either directly or via a `VkImageView`, must have completed execution](https://docs.vulkan.org/refpages/latest/refpages/source/vkDestroyImage.html#VUID-vkDestroyImage-image-01000)

PR #2769, opened 2026-07-08
Disable non-working quad control flow

squidbus:
Originally in https://github.com/KhronosGroup/MoltenVK/pull/2533 it was assumed that some of the fragment terminate related quad control tests were failing due to other problems with demote vs terminate semantics in MoltenVK, but it has since been determined that they just do not work. So it makes sense to me to disable quad control for now until it gets fixed, if possible.

Disabling the flag in `_metalFeatures` will in turn disable the extension elsewhere.


billhollings (approved):
Thanks for the catch. LGTM.

PR #2770, opened 2026-07-11
Disable 64-bit float control properties

squidbus:
These were enabled due to problems with CTS falsely requiring them, but said problems are being fixed. We do not support 64-bit floats.

See https://github.com/KhronosGroup/VK-GL-CTS/issues/610


billhollings (approved):
Thanks! LGTM.

Issue #2663, opened 2025-11-13
ErrorIncompatibleDriver when loading libMoltenVK.dylib using VK_LUNARG_direct_driver_loading extension


QuisPic:
When calling VkCreateInstance with the VkDirectDriverLoadingListLUNARG structure attached to the pNext chain of the VkInstanceCreateInfo, which contains the address of the vk_icdGetInstanceProcAddr function in libMoltenVK.dylib, I receive the following error:

> Context::createInstance: ErrorIncompatibleDriver

When using the `VK_LOADER_DEBUG=error` environment variable I get the following message from the Vulkan Loader:

> [Vulkan Loader] ERROR | DRIVER: loader_add_direct_driver: VkDirectDriverLoadingInfoLUNARG structure at index 0 supports interface version 5, which is incompatible with the Loader Driver Interface version that supports the VK_LUNARG_direct_driver_loading extension, skipping.
[Vulkan Loader] ERROR | DRIVER: vkCreateInstance: Found no drivers!

I was able to pinpoint the issue to the vk_icdNegotiateLoaderICDInterfaceVersion function. The function receives the value 7 in the pSupportedVersion variable and then changes it to 5, but it seems that Vulkan Loader rejects this value.

Changing one line in vk_icdNegotiateLoaderICDInterfaceVersion from:
`*pSupportedVersion = 5;`
to
`*pSupportedVersion = 7;` 
fixes the issue for me, but I doubt that's the solution.

The same libMoltenVK.dylib and libvulkan.dylib work properly and without any issues when loaded normally from /usr/local/lib using MoltenVK_icd.json.

Vulkan SDK version is 1.4.328

cdavis5e:
> Changing one line in vk_icdNegotiateLoaderICDInterfaceVersion from:
> `*pSupportedVersion = 5;`
> to
> `*pSupportedVersion = 7;`
> fixes the issue for me, but I doubt that's the solution.

Maybe not, but it's certainly part of the solution. We also need to investigate what changed in the ICD interface between versions 5 and 7, and make any appropriate changes.

PR #2642, opened 2025-10-11
WIP: Add support for mesh shaders.

squidbus:
Made an attempt at wiring up mesh shader support in MoltenVK. The [Vulkan-Samples mesh_shading example](https://github.com/KhronosGroup/Vulkan-Samples/tree/main/samples/extensions/mesh_shading) works, however trying to run CTS, well...

```
./deqp-vk --deqp-case="dEQP-VK.mesh_shader.ext.smoke.monolithic.mesh_task_shader_triangle"

...

[mvk-error] VK_ERROR_INITIALIZATION_FAILED: Render pipeline compile failed (Error code 3):
Internal compiler error..
  Fail (retcode: VK_ERROR_INITIALIZATION_FAILED at vkPipelineConstructionUtil.cpp:176)
```

If anyone has any ideas on debugging that, would be appreciated.

TO-DO list:
* [x] Successfully run at least one relevant CTS test.
* [ ] Fix internal compile error for most shaders.
* [ ] Deal with missing `DrawIndex` SPIR-V built-in which is used by many test cases.
* [ ] Finish filling out properties structure. Took care of the ones I could figure out, if anyone can help figure out the remaining ones would be appreciated as well.

And lastly, multiview is not supported in this PR. It seems like a lot would need to be done to wire up the way multiview works currently to mesh shaders, and the extension has a feature flag for this, so it's disabled for now.

Issue #2375, opened 2024-10-13
VK_EXT_mesh_shader: Further plans 


zmarlon:
Hello, I have a suggestion regarding the implementation of the extension. At the moment it is the case that emitMeshTasksEXT terminates the execution of the task shader. However, this is not the case in Metal. The possibilities that exist now are the following:

1) Not to support Task Shader at all
2) Support a subset of the task shaders, which would still cover a huge number of use cases.

I don't like this solution, as mesh shaders offer an enormous benefit for writing a high-performance modern game engine. Therefore, I would suggest the second solution.

@billhollings What is your opinion on this? It would be good if we could come to an agreement on this, because if the SPIRV Cross PR was merged we could finally close the PR on the MoltenVK side. Future changes regarding the terminating could still be made on the Spirv Cross side, but it would be a good start.

zmarlon:
Tagging a few people that might be Interested @BeastLe9enD @spnda @Try @squidbus

BeastLe9enD:
I think it makes sense to add initial support for task shaders even if only a subset is supported, because only supporting mesh shaders would be against the spec too and so it makes more sense to add the existing task shader subset. The only question is how to expose this to the user to get a feedback that the extension is only half-supported.

We could add a VkPhysicalDeviceMeshShaderPortabilityFeaturesMVK like structure that needs to be passed along the VkPhysicalDeviceFeatures2 when VkPhysicalDeviceMeshShaderFeaturesEXT is inside pNext in order to succeed device creation.

When task shaders are fully supported, the only thing we need to do is remove this check that ensures device creation only succeeds when our portability struct is passed a long.

Im open for further discussion, but this is the way I would take.

Try:
Mesh shader for spirv-cross:
https://github.com/KhronosGroup/SPIRV-Cross/pull/2400, and https://github.com/KhronosGroup/SPIRV-Cross/pull/2401 are merged.

Task shader on review:
https://github.com/KhronosGroup/SPIRV-Cross/pull/2402

billhollings:
> Mesh shader for spirv-cross: [KhronosGroup/SPIRV-Cross#2400](https://github.com/KhronosGroup/SPIRV-Cross/pull/2400), and [KhronosGroup/SPIRV-Cross#2401](https://github.com/KhronosGroup/SPIRV-Cross/pull/2401) are merged.
> 
[...]

PR #2771, opened 2026-07-12
WIP: Add Metal ray query and ray tracing support

dttdrv:
This PR adds opt-in Vulkan ray tracing support through Metal. It implements VK_KHR_acceleration_structure, VK_KHR_ray_query, VK_KHR_ray_tracing_pipeline, VK_KHR_pipeline_library, and VK_KHR_ray_tracing_maintenance1. The implementation remains disabled unless MVK_CONFIG_ENABLE_EXPERIMENTAL_RAY_TRACING=1 is set.

It covers acceleration-structure build, update, copy and compaction, ray queries, shader binding tables, recursive traces up to depth 4, callable shaders, pipeline libraries, descriptors, push constants, and cross-queue synchronization. rayTracingPipelineTraceRaysIndirect2 remains false.

Supported any-hit and procedural intersection shaders use Metal intersection function buffers on the Metal 4 path. Visible function tables handle closest-hit and callable dispatch, linked pipeline libraries, and cases that cannot use the native path safely. MoltenVK derives the ray dispatch layout from the pipeline instead of duplicating a fixed descriptor-set layout.

The compiler work is in KhronosGroup/SPIRV-Cross#2663. MoltenVK temporarily pins that revision until the compiler change is merged.

The Metal 4 and compatibility configurations build successfully. The focused ray-query, ray-pipeline, depth-4, any-hit, procedural-hit, acceleration-structure update, and cross-queue synchronization paths pass with Metal validation on my M5 Pro.


dttdrv:
Hi @billhollings and @cdavis5e, would you be willing to review this ray query and ray tracing implementation when you have time? CI and the CLA are green. Thanks.

squidbus:
It's a lot to look at, but the first thing that stands out to me is we generally prefer to upstream changes to SPIRV-Cross rather than maintaining patch sets. Is there a particular reason not to get it upstreamed?

dttdrv:
You’re right. There isn’t a reason to keep this as a MoltenVK patch. I put it here to test the SPIRV-Cross and MoltenVK changes together. I’ll move the compiler changes and tests to SPIRV-Cross, then update this PR to use that revision. I’ll check the approach against KhronosGroup/SPIRV-Cross#2065 as I do it.

dttdrv:
The compiler changes and tests are now in KhronosGroup/SPIRV-Cross#2649. This PR pins that exact commit and no longer carries a SPIRV-Cross patch. Clean MoltenVK Debug and Release CMake builds pass against the pinned revision.

denisk-dev:
Hi @dttdrv, I ran into this while trying your branch against Kitten Space Agency with the ray query (finally!) enabled:
[mvk-error] SPIR-V to MSL conversion error: Ray tracing instance metadata could not be resolved.

Do you know why this error might be happening?

dttdrv:
@denisk-dev Thanks, I found it. Metadata provenance was being lost when an acceleration structure was returned from a helper function. The fix is now in the SPIRV-Cross PR, and this branch has been updated to use it. The regression passes the optimized and unoptimized MSL suites, Metal 3.0 compilation, and MoltenVK Debug and Release builds. Could you retry with the updated branch?

denisk-dev:
@dttdrv, got a bunch of warnings and errors. I'll attach as a file. Seems like the majority of errors are related to undeclared identifiers.
[ksa_ray_errors.txt](https://github.com/user-attachments/files/30022616/ksa_ray_errors.txt)

dttdrv:
@denisk-dev I'm on it. Problem is the only RT game I want to play is Doom The Dark Ages, which is very heavy for the slow internet I have, thus essentially prohibiting me from testing. This will take some time, but any help is appreciated.

oscarbg:
> @denisk-dev I'm on it. Problem is the only RT game I want to play is Doom The Dark Ages, which is very heavy for the slow internet I have, thus essentially prohibiting me from testing. This will take some time, but any help is appreciated.
> 

have you tested q2rtx? has both ray query and raytracing pipeline modes..

oscarbg:
also a question supports recursion depth greater than 1? for samples like:
https://github.com/DerRM/VulkanRaytracingProceduralGeometry
at least the original d3d12 needs it..

dttdrv:
@oscarbg Great idea, haven't though of that. I even have it in my Steam Library... I checked the current sample and it sets maxPipelineRayRecursionDepth to 1, even though its closest-hit shader traces additional rays. I’ve now run the Vulkan CTS recursive-trace test at depth 2 against this branch and it passes with Metal validation enabled. I’m testing higher depths and will also try the sample with its recursion depth raised.

oscarbg:
Hi,
tested a q2rtx native MacOS build..
had to modify source as it requires pipelineStatisticsQuery and samplerFilterMinmax..

log:
```
Vulkan error: VK_ERROR_FEATURE_NOT_PRESENT: vkCreateDevice(): Requested physical device feature specified by the 25th flag in VkPhysicalDeviceFeatures2 is not available on this device.

Vulkan error: VK_ERROR_FEATURE_NOT_PRESENT: vkCreateDevice(): Requested physical device feature specified by the 31st flag in VkPhysicalDeviceVulkan12Features is not available on this device.
```

1)25th flag is pipelineStatisticsQuery which is unsupported by MoltenVK (and KK)
luckily seems not used in code (A pipeline statistics query is begun and ended by calling vkCmdBeginQuery and vkCmdEndQuery) and found no references of vkCmdBeginQuery and vkCmdEndQuery..

2)31th flag is samplerFilterMinmax.. there is some branch by @squidbus exposing it but it would require M5 or higher (Apple10 family)
sadly it's used in code godrays.c uses min mode..

upon changes:
then fails with:
```
][mvk-error] VK_ERROR_INITIALIZATION_FAILED: Shader library compile failed (Error code 3):
program_source:1565:63: error: use of undeclared identifier 'topLevelAS_mvkMetadata_1'
    spvRayQueryMetadataAddress_3254 = reinterpret_cast<ulong>(topLevelAS_mvkMetadata_1[0]);
                                                              ^
program_source:1566:20: error: type 'const Ray' does not provide a call operator
    rayQuery.reset(ray(ray.origin, ray.direction, ray.t_min, ray.t_max), topLevelAS[0], (uint(instance_mask) & 255u), spvMakeIntersectionParams(rayFlags));
                   ^~~
program_source:1583:19: warning: unused variable 'hitT' [-Wunused-variable]
            float hitT = _3291;
                  ^
program_source:1587:18: warning: unused variable 'isProcedural' [-Wunused-variable]
            bool isProcedural = _3295 == 1u;
                 ^
program_source:2214:63: error: use of undeclared identifier 'topLevelAS_mvkMetadata_1'
validation layer 2 4096: VK_ERROR_INITIALIZATION_FAILED: Shader library compile failed (Error code 3):
program_source:1565:63: error: use of undeclared identifier 'topLevelAS_mvkMetadata_1'
    spvRayQueryMetadataAddress_3254 = reinterpret_cast<ulong>(topLevelAS_mvkMetadata_1[0]);
                                                              ^
program_source:1566:20: error: type 'const Ray' does not provide a call operator
    rayQuery.reset(ray(ray.origin, ray.direction, ray.t_min, ray.t_max), topLevelAS[0], (uint(instance_mask) & 255u), spvMakeIntersectionParams(rayFlags));
                   ^~~
program_source:1583:19: warning: unused variable 'hitT' [-Wunused-variable]
            float hitT = _3291;
                  ^
program_source:1587:18: warning: unused variable 'isProcedural' [-Wunused-variable]
            bool isProcedural = _3295 == 1u;
                 ^
program_source:2214:63: error: use of undeclared identifier 'topLevelAS_mvkMetadata_1'
    spvRayQueryMetadataAddress_3546 = reinterpret_cast<ulong>(topLevelAS_mvkMetadata_1[1]);
                                                              ^
program_source:2215:22: error: type 'const Ray' does not provide a call operator
    rayQuery_1.reset(ray(ray.origin, ray.direction, ray.t_min, ray.t_max), topLevelAS[1], (instance_mask & 255u), spvMakeIntersectionParams(rayFlags));
                     ^~~
program_source:2565:18: warning: unused variable '_4987' [-Wunused-variable]
            bool _4987 = panini_forward(param_4, param_5, param_6, param_7, _1804);
                 ^
program_source:2576:18: warning: unused variable '_4997' [-Wunused-variable]
            bool _4997 = stereographic_forward(param_8, param_9, param_10, param_11, _1804);
                 ^
program_source:2587:18: warning: unused variable '_5007' [-Wunused-variable]
            bool _5007 = cylindrical_forward(param_12, param_13, param_14, param_15, _1804);
                 ^
program_source:2598:18: warning: unused variable '_5017' [-Wunused-variable]
            bool _5017 = equirectangular_forward(param_16, param_17, param_18, param_19, _1804);
                 ^
program_source:2609:18: warning: unused variable '_5027' [-Wunused-variable]
            bool _5027 = mercator_forward(param_20, param_21, param_22, param_23, _1804);
                 ^
program_source:2620:18: warning: unused variable '_4977' [-Wunused-variable]
            bool _4977 = rectilinear_forward(param, param_1, param_2, param_3, _1804);
                 ^
program_source:3242:5: error: no matching function for call to 'trace_geometry_ray'
    trace_geometry_ray(param_5, param_6, param_7, spvDescriptorSet2.global_texture_descriptors, spvDescriptorSet2.global_texture_descriptorsSmplr, spvDescriptorSet3.primitive_buffers, (*spvDescriptorSet1.m_1329), (*spvDescriptorSet3.m_1605), (*spvDescriptorSet1.m_1804), ray_payload_geometry, rayQuery, spvRayQueryMetadataAddress_3254, spvDescriptorSet0.topLevelAS, reinterpret_cast<ulong>(reinterpret_cast<device const uint2*>(spvRayMetadata_6527)));
    ^~~~~~~~~~~~~~~~~~
program_source:1553:6: note: candidate function not viable: no known conversion from 'const constant raytracing::acceleration_structure<raytracing::instancing>[2]' (aka 'const constant _acceleration_structure<metal::raytracing::instancing>[2]') to 'const constant spvUnsafeArray<raytracing::acceleration_structure<raytracing::instancing>, 2> &' (aka 'const constant spvUnsafeArray<_acceleration_structure<metal::raytracing::instancing>, 2> &') for 13th argument
void trace_geometry_ray(thread const Ray& ray, thread const bool& cull_back_faces, thread const int& instance_mask, constant array<texture2d<float>, 8192>& global_texture_descriptors, constant array<sampler, 8192>& global_texture_descriptorsSmplr, device PRIMITIVE_BUFFER* constant (&primitive_buffers)[8194], device InstanceSSBO& _1329, device LIGHT_BUFFER& _1605, constant UBO& _1804, thread RayPayloadGeometry& ray_payload_geometry, thread raytracing::intersection_query<raytracing::instancing, raytracing::triangle_data>& rayQuery, thread ulong& spvRayQueryMetadataAdVulkan error: VK_ERROR_INITIALIZATION_FAILED: Shader library compile failed (Error code 3):
program_source:1565:63: error: use of undeclared identifier 'topLevelAS_mvkMetadata_1'
    spvRayQueryMetadataAddress_3254 = reinterpret_cast<ulong>(topLevelAS_mvkMetadata_1[0]);
                                                              ^
program_source:1566:20: error: type 'const Ray' does not provide a call operator
    rayQuery.reset(ray(ray.origin, ray.direction, ray.t_min, ray.t_max), topLevelAS[0], (uint(instance_mask) & 255u), spvMakeIntersectionParams(rayFlags));
                   ^~~
program_source:1583:19: warning: unused variable 'hitT' [-Wunused-variable]
            float hitT = _3291;
                  ^
program_source:1587:18: warning: unused variable 'isProcedural' [-Wunused-variable]
            bool isProcedural = _3295 == 1u;
                 ^
program_source:2214:63: error: use of undeclared identifier 'topLevelAS_mvkMetadata_1'
    spvRayQueryMetadataAddress_3546 = reinterpret_cast<ulong>(topLevelAS_mvkMetadata_1[1]);
                                                              ^
program_source:2215:22: error: type 'const Ray' does not provide a call operator
    rayQuery_1.reset(ray(ray.origin, ray.direction, ray.t_min, ray.t_max), topLevelAS[1], (instance_mask & 255u), spvMakeIntersectionParams(rayFlags));
                     ^~~
program_source:2565:18: warning: unused variable '_4987' [-Wunused-variable]
            bool _4987 = panini_forward(param_4, param_5, param--- QVK_MOD_PRIMARY_RAYS_RGEN 15

validation layer 2 4096: VK_ERROR_INITIALIZATION_FAILED: Compute shader function could not be compiled into pipeline. See previous logged error.
]Vulkan error: VK_ERROR_INITIALIZATION_FAILED: Compute shader function could not be compiled into pipeline. See previous logged error.
--- (null) 19

Failed to create ray tracing compute pipeline #0, vkCreateComputePipelines error code is VK_ERROR_INITIALIZATION_FAILED
Closing console log.
```

dttdrv:
@oscarbg Thanks, the log helped. It exposed three SPIRV-Cross issues: acceleration-structure array metadata, native Metal array arguments, and a shader variable named ray shadowing the Metal ray type. They are fixed now and both PR branches are updated. I also built Q2RTX locally; with the two feature workarounds you mentioned, both ray-query and ray-tracing-pipeline modes render. Pipeline mode is very slow for now.

denisk-dev:
@dttdrv, can confirm the latest version seems to work with Kitten Space Agency. Doing the interior camera (that's ray query heavy) is incredibly slow though, but I haven't played with graphics settings yet.

dttdrv:
@denisk-dev Yeah, further optimization is inbound. Quake 2 RTX is also not flawless, but it runs at least.

oscarbg:
@dttdrv  incredible job! are you interested on working on mesh shader support eventually?.. seems the other modern GPU extension still not supported (both HW support for raytracing and mesh shading introduced in M3 era..)
2 issue already opened:
https://github.com/KhronosGroup/MoltenVK/issues/1618
https://github.com/KhronosGroup/MoltenVK/issues/2375
latest one shows MSL support merged already seems and 
latest MVK integration https://github.com/ProjectKML/MoltenVK/commit/88e5f00677e8e5487e2777f6e47bfde836817bea

denisk-dev:
@oscarbg, I think that's more of a question for @dttdrv. I can definitely assist with testing it against KSA (which is fully built on Vulkan). Not sure if it's using mesh shaders though. I'll have to check

oscarbg:
@denisk-dev sorry fixed..

dttdrv:
@oscarbg Thanks. Yes, I’m interested. I found the newer WIP in #2642, so I’d continue from that instead of restarting from the older ProjectKML patch. Just glad it's working properly as of now. One step closer to playin Doom: The Dark Ages. I'll continue work tomorrow.

denisk-dev:
@dttdrv, for #2642, that todo list should technically be updated with "DrawIndex" checked now. Hopefully, that helps.

johnfaizzjamaludin:
# jfaizjamaludinOn 15 Jul 2026, at 08:19, Deyan Todorov ***@***.***> wrote:﻿dttdrv left a comment (KhronosGroup/MoltenVK#2771)
@oscarbg Great idea, haven't though of that. I even have it in my Steam Library... I checked the current sample and it sets maxPipelineRayRecursionDepth to 1, even though its closest-hit shader traces additional rays. I’ve now run the Vulkan CTS recursive-trace test at depth 2 against this branch and it passes with Metal validation enabled. I’m testing higher depths and will also try the sample with its recursion depth raised.

oscarbg:
> @dttdrv, can confirm the latest version seems to work with Kitten Space Agency. Doing the interior camera (that's ray query heavy) is incredibly slow though, but I haven't played with graphics settings yet.

@dttdrv @denisk-dev curious what Apple SOC you test with ? M3 and higher required for (fast) HW raytracing..

denisk-dev:
@oscarbg M3 Max with 16/40 cores and 128GB

dttdrv:
> > @dttdrv, can confirm the latest version seems to work with Kitten Space Agency. Doing the interior camera (that's ray query heavy) is incredibly slow though, but I haven't played with graphics settings yet.
> 
> @dttdrv @denisk-dev curious what Apple SOC you test with ? M3 and higher required for (fast) HW raytracing..

I use a base M5 Pro with 24GB of RAM

oscarbg:
> > > @dttdrv, can confirm the latest version seems to work with Kitten Space Agency. Doing the interior camera (that's ray query heavy) is incredibly slow though, but I haven't played with graphics settings yet.
> > 
> > 
> > @dttdrv @denisk-dev curious what Apple SOC you test with ? M3 and higher required for (fast) HW raytracing..
> 
> I use a base M5 Pro with 24GB of RAM


@dttdrv thanks for sharing.. I checked q2rtx and it works finally!! nice to see.. if you use an M5 then you should be able to use this patch: https://github.com/squidbus/MoltenVK/commit/6f75b5b95a68e88891f701811a731f97281d569f (from issue: https://github.com/KhronosGroup/MoltenVK/issues/1559) and don't disable samplerFilterMinmax on your q2rtx custom build.. then in theory maybe you get correct rendering in q2rtx.. would be nice to know!!

dttdrv:
@billhollings @cdavis5e @squidbus @denisk-dev @oscarbg Quick update on the mesh shader work. I continued from #2642 and kept it separate from the ray tracing branches for now.

The taskless path is working end to end now. Direct and indirect draws, DrawIndex, the required 4,194,304 logical workgroup limit, triangle/line/point output, output spilling and replay, clip/cull distances, specialization constants, provoking vertex, and the pipeline cache are covered. The current full `dEQP-VK.mesh_shader.ext.*` run with `meshShader` enabled and `taskShader` disabled is 353 passed, 26,567 unsupported, 0 failed, and 0 warnings. For comparison, #2642 reported 46 passed, 569 failed, and 26,152 unsupported. The focused taskless benchmark improved from 39.58 to 78.79 FPS over five runs, about 1.99x, at 4,194,240 one-invocation mesh workgroups. That's a throughput microbenchmark, not expected game FPS.

I have also added SPIRV-Cross handling for `OpEmitMeshTasksEXT` termination through nested calls, but I have not exposed `taskShader` yet because the full task/payload/dynamic-grid CTS coverage is not done. `vkCmdDrawMeshTasksIndirectCountEXT`, legal memoryless attachment handling, full limit/device validation, multiview, mesh queries, and fragment shading rate still need work, so `VK_EXT_mesh_shader` remains hidden in the real branch.

Where would you prefer me to submit this? I can open separate draft PRs for SPIRV-Cross and MoltenVK, or coordinate the MoltenVK side as a successor/update to #2642. I didn't want to mix it into the ray tracing PRs without asking.

dttdrv:
@oscarbg Thanks, I missed that patch. I'll try it on my M5 Pro, re-enable `samplerFilterMinmax` in Q2RTX, and let you know how it goes.

squidbus:
> Where would you prefer me to submit this? I can open separate draft PRs for SPIRV-Cross and MoltenVK, or coordinate the MoltenVK side as a successor/update to #2642. I didn't want to mix it into the ray tracing PRs without asking.

Feel free to open a successor PR, and of course any changes needed to SPIRV-Cross can be PR'd there. I never got to looking into any of the remaining shader translation side issues. Should be kept separate from ray tracing though.

dttdrv:
@squidbus Sounds good. I’ll open a successor PR for MoltenVK and a separate SPIRV-Cross PR, and keep both out of the ray tracing work.

oscarbg:
> @oscarbg Thanks, I missed that patch. I'll try it on my M5 Pro, re-enable `samplerFilterMinmax` in Q2RTX, and let you know how it goes.

thanks.. waiting to see if fixes completely q2rtx rendering..

don’t want to put more work on your side but if you could run someday the specific sampler filter minmax CTS tests on your M5 and they run OK without regressions, maybe it could be merged.. think reason is not merged it’s because was not testable by @squidbus at the time on Apple10 GPU..

dttdrv:
Latest branch is at `e2eed8ef`, pinned to SPIRV-Cross `e2fe4af5`.

On M5 Pro:

- Q2RTX loads a playable level and renders correctly with ray queries at 1920x1200. Both the shader fallback and native sampler min/max path work; native sampler support is split into #2776.
- Position fetch passed indexed offsets, refit, compact copy, serialization, and deserialization coverage.
- Acceleration-structure lifetime stress passed 500 iterations in both destruction orders under Metal API Validation and NSZombie.

A live fullscreen mode switch can still leave stale attachments. Launching directly at 1920x1200 works.

oscarbg:
@dttdrv Impressive work, jointly with mesh shader stuff!!.. (+adding requested rt position fetch).. nice to hear q2rtx render issues were due to sampler filter minmax..

I see you mention “fragment shading rate” in your mesh shader work.. are you working on it also?
There is an open issue 4 years ago and it seems either Metal API or Apple GPU were not enough to implement it at the time:
https://github.com/KhronosGroup/MoltenVK/issues/1710

I ask because for fragment shading rate ext because Turing GPUs bringed 4 big changes together almost 8 years ago:
Raytracing
Mesh shaders
Variable rate shading ( fragmemt shade rate vk ext)
And tensor cores ( cooperative matrix vk ext)

3 first formed dx12 ultimate..

So seems MVK is getting PRs for 3 of these 4 features(exts) almost simultaneously..

@billhollings @cdavis5e @squidbus
Also as it’s all unrelated to this PR but seeing that it seems we’re getting mvk support for:
Raytracing (+ position fetch ext)
Mesh shaders
Cooperative matrix
Draw indirect count
Sampler filter minmax

It might warrant a 1.5.0 or even 2.x.x release ( or maybe 2.x.x is reserved for some future Metal4 enhancements/path to the codebase?)..

dttdrv:
@oscarbg Q2RTX now renders correctly with ray queries and native sampler min/max enabled, so your suggestion was right. I am not working on fragment shading rate. I only mentioned it because `PrimitiveShadingRateKHR` is part of the mesh shader interface and remains unsupported. I checked the current Metal 4 API and replied in #1710. It still needs separate work from #2777.

billhollings:
Amazing work here! Thanks very much for the submission!

Marking as WIP until the SPIRV-Cross update is available, and further testing and review.

johnfaizzjamaludin (approved):

denisk-dev:
@dttdrv, this is a bit off-topic, but what's the current status with Metal IndirectCommandBuffers? Could ICB's be used for DrawIndexedIndirect? KSA has some places where it's very heavy on indirect draw calls and I think the current implementation might not be keeping up.

dttdrv:
@denisk-dev MoltenVK doesn't use ICBs today. `vkCmdDrawIndexedIndirect` encodes as a per-draw loop of Metal indirect draws and re-finalizes state for each one, so CPU cost scales with the draw count. ICBs could batch that down. Not today, but I'll look into it once I get some free time to continue tomorrow.

dttdrv:
@denisk-dev I tested it. ICBs were slower than direct Metal encoding through 4,096 draws and only 1.15x faster at 65,536. The existing loop was mostly paying for repeated draw-state finalization. Skipping that after the first draw gave 2.29x at 4,096 and 3.12x at 16,384 and 65,536 in the submit-and-wait benchmark. I opened #2788 with 572/572 indexed-indirect CTS cases passing.

denisk-dev:
@dttdrv, thanks for looking into it! Unfortunate that ICB's were slower, but glad that you found an optimization. I did a quick test with KSA and it does seem to improve frame rate. Are you doing any kind of frame captures with Xcode? I used to have MTLHeap disabled, in order to do frame captures, but I need to have it enabled now with your accelerated structures addition. My frame captures seem to fail with MTLHeap enabled.

dttdrv:
This is now rebased onto current main and pins SPIRV-Cross #2649 at `62db8c83`. Debug and Release builds pass, along with the position fetch lifecycle coverage and variable descriptor count CTS

dttdrv:
@denisk-dev Yeah, good to hear it helps KSA. #801 reports nearly the same failure with `MTLHeap`. The RT branch should still work with `MVK_CONFIG_USE_MTLHEAP=0`, using standalone acceleration-structure storage. Could you send what fails with heaps disabled, along with your Xcode/macOS versions and any capture error? That should tell us whether this is the existing Xcode capture issue or a fallback bug

denisk-dev:
@dttdrv doesn't this line prevent RT from working if MTLHeap is not used?
`_metalFeatures.accelerationStructures &=
		_isUsingMetalArgumentBuffers && _metalFeatures.placementHeaps &&
		getMVKConfig().enableExperimentalRayTracing;`

The captures worked for me when I had MTLHeap disabled, so it might be hitting the 801 issue. I'm using XCode 26.6 and MacOS 26.5.2. I've got a screenshot of where it crashed with MTLHeap enabled, but I also misread your request. I expect the capture to still work with it disabled.

MoltenVK/MoltenVK/GPUObjects/MVKAccelerationStructure.mm
```diff
@@ -0,0 +1,922 @@
...
+	generation->release();
+	bool published = false;
+	{
+		std::lock_guard<std::mutex> lock(_lock);
+		if (!_isDestroyed && !_isMaterialized.load(std::memory_order_relaxed) && _backingBuffer == backingBuffer) {
+			_storage = storage;
+			_storageMemory = memory;
+			_nativeCapacity = nativeCapacity;
+			_metadataCapacity = metadataCapacity;
+			_address = _requestedDeviceAddress ? _requestedDeviceAddress : bufferAddress + _bufferOffset;
```
0xbeefd1ed: Consider returning the Metal gpuResourceID as the AS device address

_address is derived from the backing buffer's real GPU address. Because a Metal acceleration structure has no GPU address of its own, this then forces the whole address-resolution apparatus that follows: the device-wide `_gpuAddressableAccelerationStructures` list, the per-encoder open-addressed hash table in `MVKDevice::getAccelerationStructureAddressTable` (line 4818), the matching `mvkAccelerationStructureReferenceAddress` probe loop in the instance-conversion kernel, the per-encoder descriptor-set snapshots, and the semaphore encoding-token timeline in `MVKSync`/`MVKQueue`.

Vulkan only ever uses this value opaquely: TLAS instance descriptors store it in `accelerationStructureReference`, and `OpConvertUToAccelerationStructureKHR` turns it back into an AS in shaders. It is never dereferenced as memory. Returning `accelerationStructure.gpuResourceID` instead makes both of those a direct mapping with no table, no probe, and no CPU-side resolution — which in turn removes the need for the address table to be rebuilt and re-uploaded per encoder, and removes the ordering problem that the encoding tokens exist to solve.

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
```diff
@@ -4600,13 +4749,185 @@ static uint32_t mvkGetEntryProperty(io_registry_entry_t entry, CFStringRef prope
...
 
+static uint64_t mvkAccelerationStructureAddressHash(uint64_t value) {
+	value ^= value >> 30;
+	value *= 0xbf58476d1ce4e5b9ull;
+	value ^= value >> 27;
+	value *= 0x94d049bb133111ebull;
+	return value ^ (value >> 31);
+}
+
+void MVKDevice::getAccelerationStructureAddressTable(MVKCommandEncoder* commandEncoder,
```
0xbeefd1ed: Once you make Vulkan's deviceAddress gpuResourceId, you can delete this table entirely.

The only other thing using this table is the per-TLAS instance metadata (uint2(customIndex, sbtRecordOffset)) for an acceleration structure obtained in-shader from a raw u64 rather than from a descriptor. Descriptor-bound structures need nothing: your AccelerationStructureAux descriptor already carries the metadata pointer next to the structure (MVKDescriptorSet.mm:1955), which is the common path.

That residual is small enough to cut. customIndex is redundant — Metal's get_candidate_user_instance_id() returns it, and you already use exactly that for the ray-query version of the getter (spirv_msl.cpp:11855), so the convert kernel's dst.userID write already covers it. That leaves only `instanceShaderBindingTableRecordOffset`, which Metal accepts on the instance descriptor but exposes no MSL getter to read back. Dropping the metadata buffer and the table with it costs you two things:

- rayQueryGetIntersectionInstanceShaderBindingTableRecordOffsetEXT returns 0. In practice almost no application calls it.
- While traversal is done in software, a trace against a raw-address structure would ignore non-zero per-instance SBT offsets when picking the hit record. This one disappears entirely if you move traversal to Metal's intersection function buffer — Metal folds the instance's intersectionFunctionTableOffset into [[function_id]] itself, so the driver never needs the value for dispatch.

Given this extension is opt-in behind MVK_CONFIG_ENABLE_EXPERIMENTAL_RAY_TRACING and explicitly not chasing conformance yet, you should document these gaps explicitly and file feedback to Apple instead of complicating your implementation and sacrificing performance.

dttdrv:
@denisk-dev Thanks for catching this. The gate and fallback were added together, so I assumed the placement heap check was intentional and did not trace the no-heap path far enough. It was blocking the fallback. I removed it and the full RT probe now passes with `MVK_CONFIG_USE_MTLHEAP=0`. The heap-enabled capture crash still looks like #801

dttdrv:
@denisk-dev I should be clearer: I added the gate and the standalone fallback together and missed that they contradicted each other. The fix is now pushed in eefe7375, and the full RT probe passes with heaps disabled. Thanks again for catching it

denisk-dev:
@dttdrv, hitting this with use MTLHeap=1: `[mvk-error] VK_ERROR_OUT_OF_DEVICE_MEMORY: vkCreateAccelerationStructureKHR(): Metal could not allocate acceleration-structure resources.`. I had about 80GB of RAM available, and the most I've seen this game run was at ~25GB

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
```diff
@@ -4600,13 +4749,185 @@ static uint32_t mvkGetEntryProperty(io_registry_entry_t entry, CFStringRef prope
...
+	lock_guard<mutex> lock(_rezLock);
+	_gpuAddressableAccelerationStructures.push_back(accelerationStructure);
+}
+
+void MVKDevice::removeGPUAddressableAccelerationStructure(MVKAccelerationStructure* accelerationStructure) {
+	lock_guard<mutex> lock(_rezLock);
+	mvkRemoveFirstOccurance(_gpuAddressableAccelerationStructures, accelerationStructure);
+}
+
+MVKBuffer* MVKDevice::getBufferAtAddress(VkDeviceAddress address, VkDeviceSize& offset, VkDeviceSize requiredSize) {
```
0xbeefd1ed, line 5006: Every geometry of every BLAS build resolves its vertex, index and transform addresses through this scan, as does every SBT region in vkCmdTraceRays. With VK_KHR_buffer_device_address in use the list can be large, and the scan is done while holding _rezLock, which the encoder threads also contend for.

dttdrv:
> @dttdrv, hitting this with use MTLHeap=1: `[mvk-error] VK_ERROR_OUT_OF_DEVICE_MEMORY: vkCreateAccelerationStructureKHR(): Metal could not allocate acceleration-structure resources.`. I had about 80GB of RAM available, and the most I've seen this game run was at ~25GB

I'll look into it later today, won't have access to internet for a couple of hours. We'll fix it. Also, FYI, managed to get a boot window of Doom TDA, albeit with errors. Shows a black screen. Really nice.

MoltenVK/MoltenVK/GPUObjects/MVKPipeline.mm
```diff
@@ -2563,6 +2750,697 @@ static void setEmulatedReversedDepthViewportConfig(SPIRVToMSLConversionConfigura
...
+	if (!_mtlPipelineState) {
+		_hasValidMTLPipelineStates = false;
+		finishFeedback();
+		return;
+	}
+
+	auto makeTable = [&](NSUInteger functionCount) {
+		MTLVisibleFunctionTableDescriptor* tableDescriptor = [MTLVisibleFunctionTableDescriptor new];
+		tableDescriptor.functionCount = functionCount;
+		id<MTLVisibleFunctionTable> table = [_mtlPipelineState newVisibleFunctionTableWithDescriptor:tableDescriptor];
```
0xbeefd1ed, line 3318: All six tables here are MTLVisibleFunctionTable, and the generated traversal in SPIRV-Cross #2649 is a while (spvRayQuery.next()) loop that calls any-hit and intersection shaders as ordinary visible functions. Metal offers intersection function buffer (which is different from intersection function table) and that's what you want to use for anyhit and intersection shader; you can continue to use visible function table for closest hit shader and callable shader.

MoltenVK/MoltenVK/Commands/MVKCmdAccelerationStructure.mm
```diff
@@ -0,0 +1,662 @@
...
+		if (generationResult < 0 || !dstGeneration) {
+			if (dstGeneration) { dstGeneration->release(); }
+			cmdEncoder->reportError(VK_ERROR_OUT_OF_DEVICE_MEMORY,
+				"vkCmdBuildAccelerationStructuresKHR(): The destination acceleration structure has insufficient storage capacity.");
+			[descriptor release];
+			continue;
+		}
+		id<MTLAccelerationStructure> dstAccStruct = dstGeneration->getMTLAccelerationStructure();
+		MVKAccelerationStructureCanonicalBuild canonicalBuild;
+		VkResult canonicalResult = canonicalBuild.prepareAndEncode(cmdEncoder, mvkDstAccStruct,
```
0xbeefd1ed, line 261: canonicalBuild.prepareAndEncode() runs on every build, not only when the application intends to serialize. For a large static BLAS this doubles the memory footprint of the geometry and adds a full copy of vertex/index/transform/AABB data to every build's bandwidth, permanently, so that vkCmdCopyAccelerationStructureToMemory can rebuild later.

You should drop the serialization / deserialization feature and file a feature request.

MoltenVK/MoltenVK/Commands/MVKCmdAccelerationStructure.mm
```diff
@@ -0,0 +1,662 @@
...
+void MVKCmdWriteAccelerationStructuresProperties::encode(MVKCommandEncoder* cmdEncoder) {
+	if (_accelerationStructures.empty()) { return; }
+	if (_queryType == VK_QUERY_TYPE_ACCELERATION_STRUCTURE_COMPACTED_SIZE_KHR) {
+		id<MTLAccelerationStructureCommandEncoder> encoder =
+			cmdEncoder->getMTLAccelerationStructureEncoder(kMVKCommandUseWriteAccelerationStructuresProperties);
+		for (uint32_t index = 0; index < _accelerationStructures.size(); index++) {
+			auto* generation = _accelerationStructures[index]->retainCurrentGeneration();
+			uint32_t query = _firstQuery + index;
+			if (generation) {
+				[encoder writeCompactedAccelerationStructureSize:generation->getMTLAccelerationStructure()
```
0xbeefd1ed, line 607: getBuildSizes inflates accelerationStructureSize by placementAlignment - 256 so materialize() has room to round the placement offset up (line 655). writeCompactedAccelerationStructureSize writes Metal's raw compacted size straight into the query buffer, without that reservation.

Failure scenario: app builds with ALLOW_COMPACTION, queries the compacted size, creates a buffer and AS of exactly that size, and calls vkCmdCopyAccelerationStructureKHR with COMPACT. In materialize(), displacement can consume up to placementAlignment - 1 bytes of the new allocation, so the binary search in getMaximumMetalSize either returns a Metal size smaller than the compacted structure or fails outright with VK_ERROR_OUT_OF_DEVICE_MEMORY. The compacted size needs the same + alignment, round-up-to-alignment treatment as getBuildSizes, applied GPU-side to the value Metal writes.

0xbeefd1ed (changes requested):
Review was done with help of AI since the PR was obviously done with help of AI. However, please note that the PR as it currently stands has a huge penalty on performance and memory. You should adopt intersection function buffer (not intersection function table, and definitively not visible function table), drop serialization and the table for rayQueryGetIntersectionInstanceShaderBindingTableRecordOffsetEXT at a minimum before merging this PR. Missing Metal features should be reported by filling a feedback request instead of slowing down every application.

oscarbg:
hi,
just kosmickrisp patches with raytracing support have appeared:
https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/43303

seems they support 2 additional extensions vs here (but here we have position fetch support missing in kk MR)..

so sorry, but asking this 2 more extensions in this commit if possible(?):
1) VK_KHR_pipeline_library
2) VK_KHR_ray_tracing_maintenance1

VK_KHR_ray_tracing_maintenance1 seems can be supported with Metal seeing KK support.. and also 
VK_KHR_pipeline_library which seems was part of official Vulkan KHR raytracing extensons according to NV VK dev drv:

```
November 23rd, 2020 - Windows 457.44, Linux 455.46.02

    New:
        The official release of [Khronos Vulkan Ray Tracing](https://www.khronos.org/blog/vulkan-ray-tracing-final-specification-release) for Pascal 6GB GPUs and above:
            [VK_KHR_acceleration_structure](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/VK_KHR_acceleration_structure.html)
            [VK_KHR_ray_tracing_pipeline](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/VK_KHR_ray_tracing_pipeline.html)
            [VK_KHR_ray_query](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/VK_KHR_ray_query.html) [RTX GPUs]
            [VK_KHR_pipeline_library](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/VK_KHR_pipeline_library.html)
            [VK_KHR_deferred_host_operations](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/VK_KHR_deferred_host_operations.html)
            [SPV_KHR_ray_tracing](http://htmlpreview.github.io/?https://github.com/KhronosGroup/SPIRV-Registry/blob/master/extensions/KHR/SPV_KHR_ray_tracing.html)
            [SPV_KHR_ray_query](http://htmlpreview.github.io/?https://github.com/KhronosGroup/SPIRV-Registry/blob/master/extensions/KHR/SPV_KHR_ray_query.html) [RTX GPUs]
            [GLSL_EXT_ray_tracing](https://github.com/KhronosGroup/GLSL/blob/master/extensions/ext/GLSL_EXT_ray_tracing.txt)
            [GLSL_EXT_ray_query](https://github.com/KhronosGroup/GLSL/blob/master/extensions/ext/GLSL_EXT_ray_query.txt) [RTX GPUs]
            [GLSL_EXT_ray_flags_primitive_culling](https://github.com/KhronosGroup/GLSL/blob/master/extensions/ext/GLSL_EXT_ray_flags_primitive_culling.txt)
            For Vulkan Ray Tracing samples and tutorials see https://github.com/nvpro-samples/vk_raytracing_tutorial_KHR
```

and also AMD post:
https://gpuopen.com/news/vulkan-ray-tracing-extensions/

johnfaizzjamaludin:
# jfaizjamaludinOn 29 Jul 2026, at 05:05, Deyan Todorov ***@***.***> wrote:﻿dttdrv left a comment (KhronosGroup/MoltenVK#2771)
@denisk-dev Yeah, good to hear it helps KSA. #801 reports nearly the same failure with MTLHeap. The RT branch should still work with MVK_CONFIG_USE_MTLHEAP=0, using standalone acceleration-structure storage. Could you send what fails with heaps disabled, along with your Xcode/macOS versions and any capture error? That should tell us whether this is the existing Xcode capture issue or a fallback bug

dttdrv:
I went back through this after the review and reworked a large part of the implementation. Vulkan acceleration-structure addresses now remain stable across Metal generations, compact placement is limited to the declared Vulkan range, and the semaphore-token graph and broad acceleration-structure retention are gone. I also replaced the linear buffer-address scan, removed duplicate serialization and metadata state, dropped position fetch from this PR, and added `VK_KHR_ray_tracing_maintenance1`. `rayTracingPipelineTraceRaysIndirect2` remains false.

I did not remove everything mentioned in the review. VFT remains the compatibility path, while a native intersection-function-buffer path for newer hardware still needs to be implemented and measured. Serialization and the address translation needed for raw acceleration-structure addresses and SBT offsets also remain because removing them would change observable Vulkan behavior. The remaining serialization cost still needs work.

The SPIRV-Cross side needs to be redone as well. I plan to split the generic ray-query fixes from the larger pipeline lowering before opening another compiler PR. Until that is settled, this remains WIP.

Debug and Release builds pass. The focused ray-query, ray-pipeline, acceleration-structure, maintenance1 and cross-queue synchronization checks pass on my M5 Pro with Metal validation. This does not establish support across the other Apple GPU generations yet.

MoltenVK/MoltenVK/GPUObjects/MVKAccelerationStructure.mm
dttdrv: I tried this direction, but rebuilding the same Vulkan acceleration structure can replace its Metal object and change the `gpuResourceID` after the application has already received an address. I kept the Vulkan address stable and now translate it only when Metal consumes the reference. This also allowed me to remove the semaphore-token timeline and broad per-encoder retention.

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
dttdrv: I narrowed this rather than deleting it. Descriptor-bound acceleration structures already use their reference directly. Translation remains for TLAS instance addresses, `OpConvertUToAccelerationStructureKHR` and SBT record offsets. Returning zero for the SBT offset would make an advertised query return the wrong result, so I did not take that shortcut.

MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
dttdrv: Agreed, this scan was on the wrong path. Buffer-address lookup now uses a lazily rebuilt interval index instead of walking every buffer for every geometry and SBT lookup. I kept the existing overlap and tie behavior and compared it with the old scan in a focused lookup benchmark.

MoltenVK/MoltenVK/GPUObjects/MVKPipeline.mm
dttdrv: I have not fixed this one yet. VFT still handles the compatibility path, including any-hit and intersection shaders. A native intersection-function-buffer path looks like the better direction for newer hardware, but I need to implement and measure it first, and it would not replace the older fallback. This thread should stay open.

MoltenVK/MoltenVK/Commands/MVKCmdAccelerationStructure.mm
dttdrv: I confirmed this cost. I removed the duplicate storage and schema, but the canonical capture still runs for every build and is not free. I did not drop serialization because `VK_KHR_acceleration_structure` does not expose it as a separate optional feature. This part is still unresolved.

MoltenVK/MoltenVK/Commands/MVKCmdAccelerationStructure.mm
dttdrv: Thanks, this was a real bug. The code was treating backing-buffer tail space outside the declared Vulkan acceleration-structure range as usable. Placement now happens only when the complete Metal allocation fits within `_size - displacement`; otherwise it uses standalone storage. I did not inflate Metal’s compacted-size result because I could not justify that formula from the API contract.

dttdrv:
@oscarbg Thanks. `VK_KHR_pipeline_library` was already included, and I added `VK_KHR_ray_tracing_maintenance1`. CullMask, the new stage and access bits, and the acceleration-structure size and pointer-count queries are supported. `rayTracingPipelineTraceRaysIndirect2` remains false.

MoltenVK/MoltenVK/GPUObjects/MVKPipeline.mm
0xbeefd1ed: Intersection Function Buffers are supported on all devices with hardware ray tracing (M3+). You gain nothing by maintaining the "older fallback".

dttdrv: Not really maintaining, just keeping it there just in case a newer one breaks something. Give me a bit of time.

johnfaizzjamaludin:
# jfaizjamaludinOn 29 Jul 2026, at 13:38, Deyan Todorov ***@***.***> wrote:﻿dttdrv left a comment (KhronosGroup/MoltenVK#2771)
@denisk-dev Thanks for catching this. The gate and fallback were added together, so I assumed the placement heap check was intentional and did not trace the no-heap path far enough. It was blocking the fallback. I removed it and the full RT probe now passes with MVK_CONFIG_USE_MTLHEAP=0. The heap-enabled capture crash still looks like #801

dttdrv:
Quick update, I managed to get Doom: The Dark Ages running with some gaming-specific tweaks on top: https://www.reddit.com/r/macgaming/comments/1vdlg24/experimental_doom_the_dark_ages_on_a_macbook_m5/

oscarbg:
@dttdrv nice! any moltenvk branch with this fixes to test?

dttdrv:
@oscarbg soo, because some of these tweaks would bloat this PR and are gaming-specific, which MoltenVK is not only made for, I made github.com/dttdrv/magevk and github.com/dttdrv/mage. they feature sparse-buffer emulation, GPU identity spoofing, a native traceRay fast path, fixes for acceleration-structure rejections and empty-geometry vertex-buffer lifetimes, a reported-VRAM heap cap and a zero-group-count dispatch guard. I made this because it helps me test different things on games with this PR. I made many fixes locally, which help, but would be a problem for this PR, so I keep them in MageVK. In that regard, it's kind of bloated right now. Another thing is that I'm trying to reduce SPRIV-C dependency, so this simplification and optimization of the code has been hitting hard, I'll commit sometime later today or tomorrow.

denisk-dev:
@dttdrv, hello! I ran into a new issue with KSA, after a new update they pushed, stemming from them using layout set >7 (9th set), which goes over the kMaxArgumentBuffers = 8 / kMVKMaxDescriptorSetCount (read from SPIRV-Cross under MoltenVK). I noticed in your code that you have `static_assert(kMVKMaxDescriptorSetCount == 8);`. Before I fork your SPIRV-Cross and make modifications to kMaxArgumentBuffers on it, is there a necessity to do that assert in your MoltenVK code (under MVKCMDDispatch)?

dttdrv:
@denisk-dev Yep, that assert is needed. It ties two copies of the same GPU-visible layout to each other. It's not a Metal hardware limit                                
                                                                                                                                                                  
   The RT dispatch path works like this:                                                                                                                          
                                                                                                                                                                  
   1. MVKCmdDispatch.mm packs a fixed-size struct on the CPU side. It contains descriptorSetAddresses[kMVKMaxDescriptorSetCount].                                 
   2. The MSL that SPIRV-Cross emits reads that struct back as spvRayTracingDispatch, with descriptorSet0 through descriptorSet7 written out as individual fields (emitted in spirv_msl.cpp in the pinned fork).                                                                                                              
   3. The two layouts are maintained independently. If only one side changes size, the GPU reads past the struct. The assert turns that into a build error instead of silent memory corruption.                                                                                                                                
                                                                                                                                                                  
   To raise the limit for KSA's 9th set, change both sides:                                                                                                       
                                                                                                                                                                  
   1. In MoltenVK: bump kMVKMaxDescriptorSetCount and update the assert to the new value.                                                                         
   2. In the SPIRV-Cross fork: bump kMaxArgumentBuffers in spirv_msl.hpp and extend the emitted spvRayTracingDispatch struct in spirv_msl.cpp. The discrete and device-storage masks are 32-bit, so 9 sets encode without changes there.                                                                                    
                                                                                                                                                                  
   Two path-specific notes:                                                                                                                                                                                         On the RT dispatch path, descriptor sets are passed by GPU address inside the dispatch buffer, not through encoder buffer slots. There is no Metal-side slot pressure there.                                                                                                                                              
On the classic argument buffer path, each descriptor set costs buffer indices in Metal's argument table, which caps at 31 entries per stage minus MoltenVK's implicit buffers.

johnfaizzjamaludin:
> # jfaizjamaludinOn 15 Jul 2026, at 08:19, Deyan Todorov ***@***.***> wrote:﻿dttdrv left a comment ([KhronosGroup/MoltenVK#2771](https://github.com/KhronosGroup/MoltenVK/pull/2771)) @oscarbg Great idea, haven't though of that. I even have it in my Steam Library... I checked the current sample and it sets maxPipelineRayRecursionDepth to 1, even though its closest-hit shader traces additional rays. I’ve now run the Vulkan CTS recursive-trace test at depth 2 against this branch and it passes with Metal validation enabled. I’m testing higher depths and will also try the sample with its recursion depth raised. —Reply to this email directly, view it on GitHub, or unsubscribe.You are receiving this because you are subscribed to this thread.Message ID: ***@***.***>

dttdrv:
I rebased the branch onto current `main` and reworked the unpublished ray-tracing changes. The new version uses Metal's intersection-function-buffer path for any-hit shaders where available, keeps the runtime in MoltenVK instead of SPIRV-Cross, supports descriptor-set index 8 without duplicating the dispatch layout, and fixes GPU-address residency, semaphore propagation, acceleration-structure storage, and serialization. It also includes `VK_KHR_pipeline_library` and `VK_KHR_ray_tracing_maintenance1`.

The paired compiler change is in https://github.com/KhronosGroup/SPIRV-Cross/pull/2663. The rebased branch builds in Release and passes Metal validation for the IFB and no-residency paths, as well as the four synchronization modes.

Only my M5 Pro has been tested directly. Metal 4 may bring significant ray-tracing performance improvements on M3 and M4 as well. Would it be okay for me to add an optional Metal 4 fast path while keeping the existing fallback?

denisk-dev:
@dttdrv, thanks for including the descriptor set index 8! One thing I noticed while testing with KSA is if I have "in-cockpit" ray query enabled, it seems to crash with pCreateInfo->maxPipelineRayRecursionDepth == 4 (did some debug printing), which triggers `setConfigurationResult(reportError(VK_ERROR_FEATURE_NOT_PRESENT, "vkCreateRayTracingPipelinesKHR(): The pipeline shader groups, library interface, or recursion depth are unsupported."));`

dttdrv:
@denisk-dev Thanks for catching the depth-4 failure. It came from the old recursion limit. This update now advertises and accepts maxPipelineRayRecursionDepth == 4, and fixes Metal stack sizing for pipelines that combine nested traceRay calls with callable shaders. I cannot run KSA here, so I would appreciate another test of the cockpit path.

I also went back through the rest of the implementation instead of adding another isolated fix. On Metal 4, supported triangle any-hit and procedural intersection shaders now use intersection function buffers. Visible function tables remain for closest-hit and callable dispatch, linked pipeline libraries, and cases that cannot use the native path safely.

The fixed dispatch layout tied to eight descriptor sets is gone. Descriptor-set addresses are now uploaded separately using the pipeline’s real layout, so set 8 works without changing SPIRV-Cross’s normal default. The old semaphore-token graph, repeated linear buffer scan, and broad per-encoder retention are also gone. Acceleration-structure addresses stay stable when their Metal storage changes, cross-queue dependencies use MoltenVK’s existing queue boundary, and the generated dispatcher libraries are cached per device instead of being compiled for every pipeline.

Both the Metal 4 and compatibility builds pass. The depth-4, triangle any-hit, procedural-hit, acceleration-structure update, and cross-queue paths pass with Metal validation on my M5 Pro. Please test and share, because your result would be useful!

denisk-dev:
@dttdrv, hello, please see the errors generated in the attached file. Thanks!
[RT_mvk_errors_8_11.log](https://github.com/user-attachments/files/30963407/RT_mvk_errors_8_11.log)

dttdrv:
@denisk-dev Thanks for sending the log. I found the mismatch. This is separate from the depth-4 issue.

SPIRV-Cross emitted `tlasArray` as a native array of two acceleration-structure pointers at the entry point, but helper functions still expected the old wrapped array type with a declared size of one. The first return-type error in the log caused the later `traceLighting`, `traceReflections`, and `calculateDirectLighting` errors.

I updated [SPIRV-Cross #2663](https://github.com/KhronosGroup/SPIRV-Cross/pull/2663) so acceleration-structure arrays passed through helper functions preserve the caller's Metal type, descriptor count, and address space. The regression case uses descriptor set 3 with a nonzero binding, a shader-declared size of one, a bound descriptor count of two, and helper-to-helper forwarding.

The generated device, constant, and discrete forms compile with Metal 3 and Metal 4. This PR now pins the corrected compiler revision.

Please test the updated build and share the result.

denisk-dev:
@dttdrv, looks like that fixed it! Also, not sure if it's the game's improvement (since it's constantly evolving at this pre-alpha stage) or your code, but the cockpit camera went from something like 1-2fps with the previous version, to a proper 20+fps! By the way, KSA is free (at least for now), in case you ever want to check it out.

dttdrv:
@denisk-dev This is awesome news, thanks for testing it. The last compiler fix only corrected the acceleration-structure array type, so it should not explain that frame-rate jump by itself. The wider update did remove expensive lookup and synchronization work and added the native Metal 4 intersection-function-buffer path, so part of the improvement may be ours. I can't verify if any updates in the game introduced a performance improvement, but maybe I should try, could lead to new optimizations. Still, this jump is great, that's what I was aiming for with these large changes. I think I'll download KSA to check it out, I thought it was paid like KSP. Again, thank you very much!

denisk-dev:
@dttdrv, I haven't been able to run your repo once they added that 8th set, and I believe you made more changes since then. This was the first time I could run it.

dttdrv:
Thanks @Z3r0Glitch22 for their findings on CPU overhead in Metal4RT, particularly repeated descriptor and residency bookkeeping. Those findings helped guide the investigation here: RT dispatch now reuses the existing resource tracker, and indexed serialization capture reuses already-validated buffer ranges.

This update also fixes normal-build serialization roundtrips, zero vertex stride, and compute binding state. All seven focused checks pass with Metal validation enabled.

dttdrv:
So, fixed the stale compute bindings after ray-tracing dispatches. the same compute shader could produce a different result after tracing because the binding cache no longer matched Metal's state. the fix uses the existing tracker and passes the six affected binding-slot checks.
also added native traversal for eligible opaque closest-hit rays and word-sized copies for aligned AS inputs. those reduced GPU time by 2.56% and 1.94% in the respective tests on my M5 Pro. the shadow control was 0.87% slower. i'm working on a larger performance improvement separately, but it seems like it’ll take some time.

PR #2772, opened 2026-07-13
Raise minimum target to macOS 12.0 / iOS 15.0 / tvOS 15.0

mrpippy:
Xcode 27 does not support targeting macOS 11 or iOS/tvOS 14, and building currently fails with errors like `error: The macOS deployment target 'MACOSX_DEPLOYMENT_TARGET' is set to 11.0, but the range of supported deployment target versions is 12.0 to 27.0.x. (in target 'SPIRV-Cross-macOS' from project 'ExternalDependencies')`.


billhollings (approved):
Thanks for taking care of this. LGTM.

PR #2775, opened 2026-07-17
Support draw indirect count

antisaling:


squidbus:
This looks to me like a simpler implementation than the other one, which is nice. Do you have any test results from CTS for this?

oscarbg:
also advertise the VK_KHR_draw_indirect_count ext, the other one has it..

adds in MoltenVK/MoltenVK/Layers/MVKExtensions.def:
>MVK_EXTENSION(KHR_draw_indirect_count,                  KHR_DRAW_INDIRECT_COUNT,                  DEVICE,   10.11,  8.0,  1.0)

antisaling:
I was just playing around trying to make Godot more GPU driven, will try to get results for the CTS, this should've been a draft I guess but I was so happy with how simple it is.

antisaling:
CTS run got the same results as the other PR:
```
Test run totals:
  Passed:        144/416 (34.6%)
  Failed:        16/416 (3.8%)
  Not supported: 256/416 (61.5%)
  Warnings:      0/416 (0.0%)
  Waived:        0/416 (0.0%)
```

I will try to address, then mark as ready

squidbus:
If you are able to fix those up, and expose `VK_KHR_draw_indirect_count` as mentioned, I think this looks good to me.

antisaling:
btw it seems the correct filter for this would be `--deqp-case='dEQP-VK.draw.*indirect*'`, there are some tests that use `*indirect_draw*`

dttdrv:
@antisaling I fixed the 16 divisor-zero failures and added the extension advertisement. I also stopped the count path from changing recorded command state during encoding. Debug and Release both pass all 160 supported `dEQP-VK.draw.*indirect_count*` cases. The broader indirect filter still reports 106 failures, none in the count cases. The patch is here: https://github.com/dttdrv/MoltenVK/commit/fa97c4b2737d1912cbff37372bc01bc370bd1edd

antisaling:
I have a wip change here that I'm still testing, will compare it against yours and cherry-pick if yours look better, thanks a lot @dttdrv !

antisaling:
So on a quick look it seems that you are re-running full count conversion once per zero-divisor binding and preserving my initial bad command-state mutation, both of which I've addressed on my wip change. so I'm going to finish the current CTS pass and upload soon. however the vertex binding translation is something I'd forgotten so I'm taking the liberty of copying that bit, ok with you @dttdrv ?

dttdrv:
of course feel free to use it and thanks for catching that. dumb mistake I made with the re-runs, I gotta pay more attention...

antisaling:
Reran CTS using `--deqp-case='dEQP-VK.draw.*indirect*'` and got

```
Test run totals:
  Passed:        2266/3874 (58.5%)
  Failed:        0/3874 (0.0%)
  Not supported: 1608/3874 (41.5%)
  Warnings:      0/3874 (0.0%)
  Waived:        0/3874 (0.0%)
```

against upstream/main
```
Test run totals:
  Passed:        1156/3874 (29.8%)
  Failed:        90/3874 (2.3%)
  Not supported: 2628/3874 (67.8%)
  Warnings:      0/3874 (0.0%)
  Waived:        0/3874 (0.0%)
```

MoltenVK/MoltenVK/Commands/MVKCmdDraw.mm
```diff
@@ -1235,9 +1442,8 @@
 					[mtlTessCtlEncoder setStageInRegionWithIndirectBuffer: mtlIndBuff
 						                             indirectBufferOffset: mtlTempIndBuffOfst];
 					mtlTempIndBuffOfst += sizeof(MTLStageInRegionIndirectArguments);
-					// If this is a synthetic command that originated in a direct call, and there are vertex bindings with a zero vertex
-					// divisor, I need to offset them by _firstInstance * stride, since that is the expected behaviour for a divisor of 0.
-					cmdEncoder->getState().offsetZeroDivisorVertexBuffers(*cmdEncoder, stage, pipeline, _directCmdFirstInstance);
+					// Bind the copied zero-divisor vertex data for this indirect command.
```
antisaling, line 1445: IMO this comment got a bit redundant, this is the only thing I can think of.

squidbus (approved):
Seems alright to me, not sure if we need to wait to pull this in since a release is in progress.

billhollings:
> Seems alright to me, not sure if we need to wait to pull this in since a release is in progress.

Yeah. If we can wait until 1.4.2 is out, that would be simplest. Should just be a few more days. Waiting for @etang-cw to have a look at #2781 so I can get a clean CTS run if possible.

I can branch 1.4.2 if this is needed before then.

PR #2776, opened 2026-07-18
Add native sampler min/max filtering

dttdrv:
Adds `VK_EXT_sampler_filter_minmax` on Apple10 GPUs where Metal exposes native sampler reduction.

- Reports the Vulkan feature and format capability only on supported devices and OS versions.
- Maps `VkSamplerReductionModeCreateInfo` to Metal minimum, maximum, and weighted-average sampling.
- Leaves older devices unchanged.

Validation:

- MoltenVK Debug and Release builds.
- `dEQP-VK.*sampler*minmax*`: 1,118 passed, 0 failed, 2,832 unsupported.
- M5 Pro with Metal API Validation.


MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
```diff
@@ -109,6 +109,9 @@
 	supportsApple9 = supportsGPUFam(Apple9, mtlDev);
 #if MVK_XCODE_26
 	supportsApple10 = supportsGPUFam(Apple10, mtlDev);
+#endif
+#if MVK_XCODE_26 && !MVK_TVOS && !MVK_VISIONOS
+	supportsSamplerReduction = supportsApple10 && mvkOSVersionIsAtLeast(26.0);
```
squidbus: Usually for this kind of thing, it's added to `_metalFeatures` instead of the base caps here. For example, how sampler lod bias support is flagged.

squidbus: Well, unless this is to make it accessible to the format flag stuff. In which case I think it would be good to at least move this down around `supportsRenderLinearTextures`.

dttdrv: Right, it is used by `MVKPixelFormats` for `VK_FORMAT_FEATURE_2_SAMPLED_IMAGE_FILTER_MINMAX_BIT`, and `_pixelFormats` is constructed before `initMetalFeatures()`. I moved it beside `supportsRenderLinearTextures`.

MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
```diff
@@ -2442,6 +2442,31 @@ static MSLSamplerYCbCrRange getSpvSamplerYcbcrRangeFromVkSamplerYcbcrRange(VkSam
 		mtlSampDesc.compareFunction = mvkMTLCompareFunctionFromVkCompareOp(pCreateInfo->compareOp);
 	}
 
+#if MVK_XCODE_26 && !MVK_TVOS && !MVK_VISIONOS
+	if (@available(macOS 26.0, iOS 26.0, *)) {
```
squidbus, line 2446: Small nit, but I don't think this is needed given `supportsSamplerReduction` is already OS gated.

MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
```diff
@@ -2442,6 +2442,31 @@ static MSLSamplerYCbCrRange getSpvSamplerYcbcrRangeFromVkSamplerYcbcrRange(VkSam
...
+						case VK_SAMPLER_REDUCTION_MODE_MIN:
+							mtlSampDesc.reductionMode = MTLSamplerReductionModeMinimum;
+							break;
+						case VK_SAMPLER_REDUCTION_MODE_MAX:
+							mtlSampDesc.reductionMode = MTLSamplerReductionModeMaximum;
+							break;
+						default:
+							mtlSampDesc.reductionMode = MTLSamplerReductionModeWeightedAverage;
+							break;
+					}
```
squidbus: This should probably be a function like `mvkMTLSamplerReductionModeFromVkSamplerReductionMode` in `mvk_datatypes`, similar to other conversions.

squidbus (commented):
Other than small comments, looks okay to me.

dttdrv:
@squidbus I moved the conversion to `mvkMTLSamplerReductionModeFromVkSamplerReductionMode()` in `mvk_datatypes`. I kept the `@available` check because Clang cannot infer the OS version from `supportsSamplerReduction` and otherwise emits `-Wunguarded-availability-new`.

squidbus (approved):

billhollings (approved):
Thanks for submitting this. LGTM.

Issue #2783, opened 2026-07-22
MoltenVK version 1.4.2-rc1: immediate abort at vkCreateComputePipelines: vkQuake 1.35.0, macos 12.7.6 Monterey; MacBookPro12,1 Intel i5; Intel(R) Iris(TM) Graphics 6100


sto6:
Since [issue#2727](2727) resolution (by pull-requests), these events have occurred:

1. vkQuake 1.35.0 release; and build of same available on MacSourcePorts. That build predates the MoltenVk 1.4.2rc1. FAILS as-is (as expected) due to issues from: [issue#2727](2727). But WORKS with that issue's libMoltenVk 1.4.2 built by vkedwardli on Jun 18.
2. Subsequently [issue#2728](2728): MoltenVK 1.4.2rc1 pre-release (with macos binaries) became available on Jul 19.

So I tried to put these together, but it fails with a new (to me) error, on same machine (Intel Mac & Intel GPU):
```
Model Name:     MacBook Pro
Model Identifier:       MacBookPro12,1
Processor Name: Dual-Core Intel Core i5
Processor Speed:        2.9 GHz
Number of Processors:   1
Total Number of Cores:  2
L2 Cache (per Core):    256 KB
L3 Cache:       3 MB
Hyper-Threading Technology:     Enabled
```
As before, I didn't yet compile anything: just swapping out the libMoltenVk.dylib used by a vkQuake.app application bundle.
Effectively comparing these two variants of vkQuake-1.35.0:

A: fails(1.4.2rc1): [mvk-info] MoltenVK version 1.4.2, supporting Vulkan version 1.4.357.
B: works(vkedwardli): [mvk-info] MoltenVK version 1.4.2, supporting Vulkan version 1.4.350.

Below are relevant diffs in the logs. Note:

The Vulkan version differs (1.4.357 vs 1.4.350): crude indication of how much newer the rc1 is vs vkedwardli.
The working version has three extra repeats of the paired: Create 3 swapchain/VK_ERROR_FEATURE_NOT_PRESENT lines,
where the rc1 has no repeats: likely not relevant?
Next two errors are new, "SPIR-V to MSL conversion" and "Compute shader function could not be compiled into pipeline",
and yield the final "QUAKE ERROR: vkCreateComputePipelines failed".

```
A: fails(1.4.2.rc1):
[mvk-info] MoltenVK version 1.4.2, supporting Vulkan version 1.4.357.
...
[...]

PR #2755, opened 2026-06-13
Fix AMD subgroup size (64→32) and disable Intel UHD 630 simdReduction

csfercoci:
Two GPU correctness fixes for MoltenVK:

1. **AMD**: Reduce maxSubgroupSize from 64 to 32 — Metal simdgroup operations produce incorrect results at 64-wide on AMD
2. **Intel**: Disable simdReduction on UHD 630 (Gen9) — Mac2 family reporting is inaccurate for Gen9, simd_sum produces wrong results
3. **Pipeline**: Update emulate_subgroups to !simdPermute || !simdReduction

Fixes #2756, #2757

PR #2785, opened 2026-07-22
Do not enable subgroup emulation for reduction

squidbus:
Fixes regression https://github.com/KhronosGroup/MoltenVK/issues/2783 in 1.4.2-rc1

SPIRV-Cross basically only supports subgroup op emulation for `OpGroupNonUniformElect`. This is in line with how, with no Metal SIMD capabilities, we only advertise `VK_SUBGROUP_FEATURE_BASIC_BIT`.

The `emulate_subgroups` flag cannot be enabled when just `simdReduction` is disabled, as in https://github.com/KhronosGroup/MoltenVK/pull/2755, because:
* It violates the SPIRV-Cross assumption that we only support `VK_SUBGROUP_FEATURE_BASIC_BIT` (i.e. `OpGroupNonUniformElect`) when emulating subgroups, causing subgroup ops that are natively available to take the emulation path and fail.
* It does not support emulating these ops anyway, so it makes no sense.

When `simdReduction` is false, we will already correctly not advertise the associated subgroup op bits. If emulation is needed to support GPUs without working reduction ops, it would need to be implemented in SPIRV-Cross first, along with changes to ensure other native ops can still be used.


billhollings (approved):
Thanks! LGTM.

PR #2788, opened 2026-07-28
Avoid repeated state finalization for indexed indirect draws

dttdrv:
`vkCmdDrawIndexedIndirect` currently finalizes unchanged draw state before every draw in an ordinary indirect batch. This finalizes it once for that path while preserving per-draw behavior for tessellation, multiview and triangle-fan conversion. Pipeline validity is still checked for every draw.

In a submit-and-wait microbenchmark using one `vkCmdDrawIndexedIndirect` call and zero instances, 4,096 draws improved from 1.115 ms to 0.487 ms, 16,384 from 4.163 ms to 1.335 ms and 65,536 from 15.569 ms to 4.993 ms on a 15/16 M5 Pro. The code path is hardware-neutral; the M5 Pro is only the measured machine.

Debug and Release builds pass. The u16 and u32 correctness matrix passes, along with all 572 indexed render-pass and dynamic-rendering CTS cases.

PR #2790, opened 2026-08-04
Avoid redundant non-indexed indirect draw finalization

dttdrv:
## Summary

Avoid calling `finalizeDrawState()` for every ordinary non-indexed indirect draw after the first draw has already submitted unchanged state.

Tessellation and instance-adjustment paths continue to finalize each draw because they update per-draw state. Pipeline validation remains unconditional.

This mirrors the state-finalization reduction made for indexed indirect draws in #2788.

## Validation

- Release MoltenVK build
- Metal-validation correctness probes for ordinary rendering and DrawID-only positioning at 1, 64, 256, 1,024, and 4,096 draws
- Dynamic viewport/scissor and active occlusion-query witnesses
- 105 targeted Vulkan CTS cases passed with zero failures
- Controlled M5 Pro A/B with command-buffer prefill disabled showed reductions of 13.3% at 4,096 draws, 17.1% at 16,384 draws, and 30.1% at 65,536 draws

Performance figures are local M5 Pro measurements only.


squidbus (approved):
Tests look okay

PR #2794, opened 2026-08-07
Fix areQueriesDeviceAvailable

kisasexypantera94:
`areQueriesDeviceAvailable` expects its second argument to be endQuery, while callers actually pass queryCount.

This causes vkCmdCopyQueryPoolResults() to incorrectly consider queries with nonzero firstQuery. With Venus query feedback this breaks occlusion queries, causing Portal and other Source Engine games to get stuck polling them indefinitely.


kisasexypantera94:
While testing I found another offset bug in the availability buffer. The second commit fixes that and resolves the Portal hang as well.

squidbus (approved):

PR #2797, opened 2026-08-12
Add support for VK_EXT_nested_command_buffer.

utzcoz:
Secondary command buffers are encoded by replaying their commands onto the command encoder, which is already recursive, so a secondary command buffer that executes another one needs no additional handling. Replaying does not modify the secondary command buffer, so it may also be executing elsewhere, and nesting is bounded only by host stack, so no nesting limit is reported.

dEQP-VK.api.command_buffers, the five cases covering nested command buffers:

  before   0 passed, 0 failed, 5 unsupported
  after    5 passed, 0 failed, 0 unsupported

Those cases cover each of the three features reported: nested_execute and nested_execute_multiple_levels for nestedCommandBuffer, nested_render_pass_continue for nestedCommandBufferRendering, and record_simul_use_nested and record_simul_use_twice_nested for nestedCommandBufferSimultaneousUse.

maxCommandBufferNestingLevel is reported as the maximum value, as the implementation imposes no limit of its own. Submitting chains of nested secondary command buffers succeeded at every depth tested, up to 65536.

### DIFF
--- a/Docs/MoltenVK_Runtime_UserGuide.md
+++ b/Docs/MoltenVK_Runtime_UserGuide.md
@@ -370,6 +370,7 @@ In addition to core *Vulkan* functionality, **MoltenVK**  also supports the foll
   - *Requires Metal 2.0.*
 - `VK_EXT_metal_objects`
 - `VK_EXT_metal_surface`
+- `VK_EXT_nested_command_buffer`
 - `VK_EXT_non_seamless_cube_map`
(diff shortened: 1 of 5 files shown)


squidbus (approved):
Makes sense. I ran through some additional tests using nested command buffers in the draw, dynamic_state, multiview, and renderpasses sets, and looks good.

PR #2798, opened 2026-08-13
Add support for VK_EXT_multi_draw.

utzcoz:
vkCmdDrawMultiEXT() and vkCmdDrawMultiIndexedEXT() encode each element of the supplied array as a separate draw command, honouring the array stride and, for indexed draws, the optional vertex offset override. Since no batching limit is imposed, maxMultiDrawCount is reported as the maximum value.

Each draw of a multi-draw call gets its own index, which the shader reads as DrawIndex, so pass the index of the draw to the command, to be written into the draw ID buffer. Previously that buffer was always filled with zero, since only indirect multi-draw calls supplied an index. The buffer is a suballocation, so write through getContents(), which accounts for the offset the buffer is bound at, rather than to the start of the underlying MTLBuffer.

dEQP-VK.draw.renderpass.multi_draw, 9024 cases:

  before   0 passed, 0 failed, 9024 unsupported
  after    3384 passed, 0 failed, 5640 unsupported

The remainder need features MoltenVK does not have. No case regresses.


squidbus (approved):
LGTM

PR #2800, opened 2026-08-16
Fix image memoryTypeBits consistency

utzcoz:
Fix inconsistent image memoryTypeBits for input attachments

Vulkan requires identical memoryTypeBits for all images sharing a tiling, sparse binding flag, external handle types and transient attachment usage. MoltenVK also withheld the lazily allocated memory type from images using VK_IMAGE_USAGE_INPUT_ATTACHMENT_BIT, which is not permitted to affect it.

That exclusion existed because an input attachment is read through a texture binding, which memoryless storage does not support. It is dropped, and the memory is instead committed when binding such an image.

https://registry.khronos.org/vulkan/specs/latest/html/vkspec.html#VkMemoryRequirements
https://registry.khronos.org/vulkan/specs/latest/man/html/VkMemoryPropertyFlagBits.html
https://developer.apple.com/documentation/metal/mtlstoragemode/memoryless

### DIFF
--- a/Docs/Whats_New.md
+++ b/Docs/Whats_New.md
@@ -22,6 +22,7 @@ Released TBD
   - `VK_EXT_multi_draw`
   - `VK_EXT_nested_command_buffer`
   - `VK_EXT_ycbcr_2plane_444_formats`
+- Fix inconsistent image `memoryTypeBits` when `VK_IMAGE_USAGE_INPUT_ATTACHMENT_BIT` is used.
 
 
 
--- a/MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
@@ -954,13 +954,12 @@ VkResult MVKImage::getMemoryRequirements(VkMemoryRequirements* pMemoryRequiremen
 		mvkDisableFlags(pMemoryRequirements->memoryTypeBits, mvkPD->getPrivateMemoryTypes());
 	}
 
-    // Only transient attachments may use memoryless storage.
-	// Using memoryless as an input attachment requires shader framebuffer fetch, which MoltenVK does not support yet.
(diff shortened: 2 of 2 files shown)


MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
```diff
@@ -1146,6 +1145,14 @@ static MTLRegion getMTLRegion(const ImgRgn& imgRgn) {
 
     if (_ioSurface && stgMode == MTLStorageModePrivate) { stgMode = MTLStorageModeShared; }
 
+	// An input attachment is read through a texture binding, which memoryless storage does not
+	// support, so commit the memory instead. Lazily allocated memory only promises that an
+	// implementation may defer the allocation, and never that it must.
+	if (stgMode == MTLStorageModeMemoryless &&
```
squidbus, line 1153: Should carry forward the TODO about being able to improve this with framebuffer fetch.

Docs/Whats_New.md
```diff
@@ -22,6 +22,9 @@ Released TBD
   - `VK_EXT_multi_draw`
   - `VK_EXT_nested_command_buffer`
   - `VK_EXT_ycbcr_2plane_444_formats`
+- Support lazily allocated memory for input attachments.
```
squidbus: I don't think it's correct to say "support" here when it's really just falling back to private.

squidbus (commented):
Not sure on the stage interface stuff, would prefer if that was separate.

utzcoz:
> Not sure on the stage interface stuff, would prefer if that was separate.

No problem. I will split two commits to two PRs.

MoltenVK/MoltenVK/GPUObjects/MVKImage.mm
utzcoz: Added.

Docs/Whats_New.md
utzcoz: Removed.

squidbus (approved):

utzcoz:
I failed to modify PR information yesterday because of GitHub's incident. Thanks for your modifications of title.

PR #2712, opened 2026-04-13
feat: implement VK_EXT_depth_clip_enable

anshulnoori:
Adds support for the `VK_EXT_depth_clip_enable` extension in MoltenVK.

* Added `EXT_depth_clip_enable` to the device extension definitions in `MVKExtensions.def`.
* Defined `DepthClipEnable` in `MVKDeviceFeatureStructs.def`.
* Added case for `VkPhysicalDeviceDepthClipEnableFeaturesEXT` in `MVKDevice.mm`.
* Updated pipeline creation in `MVKPipeline.mm` to respect `VkPipelineRasterizationDepthClipStateCreateInfoEXT`, setting the depth clamp state based on the `depthClipEnable` flag.

PR #2802, opened 2026-08-18
Add support for VK_EXT_depth_clip_enable

darrinperry83:
`VK_EXT_depth_clip_enable` decouples depth clipping from `depthClampEnable`, which in core Vulkan implicitly disables clipping.

Metal exposes depth clipping directly through `MTLDepthClipMode`, which MoltenVK already drives for `depthClampEnable`, so this maps onto existing machinery without emulation.

### Mapping

The clip mode follows `depthClipEnable` alone:

* clipping **enabled** → `MTLDepthClipModeClip`
* clipping **disabled** → `MTLDepthClipModeClamp`

`depthClampEnable` needs no special case when `VkPipelineRasterizationDepthClipStateCreateInfoEXT` is present. With clipping enabled nothing survives outside the near and far planes, so every remaining fragment's depth already lies within the viewport's `[minDepth, maxDepth]`, and clamping to that range is a no-op. When the struct is absent the core rule applies unchanged: clipping is enabled exactly when `depthClampEnable` is `VK_FALSE`.

### Verification

Tested on Apple M2 Max, macOS 26.5.1. A triangle is rendered offscreen to a colour + depth target, depth test `VK_COMPARE_OP_LESS_OR_EQUAL`, depth cleared to `1.0`, and the result is checked for whether the primitive survived rasterisation:

| gl_Position.z | depthClipEnable | depthClampEnable | result |
|---|---|---|---|
| 0.5 | enabled | disabled | drawn |
| 0.5 | disabled | disabled | drawn |
| 0.5 | enabled | enabled | drawn |
| 1.5 | enabled | disabled | clipped |
| 1.5 | disabled | disabled | drawn |
| 1.5 | enabled | enabled | clipped |
| 1.5 | disabled | enabled | drawn |

The `z = 0.5` rows are controls. They confirm the pipeline rasterises at all, so that a pipeline drawing nothing could not make the clipped rows pass by accident. The `z = 1.5` rows are the behaviour under test: identical geometry beyond the far plane is removed with clipping enabled and retained with it disabled, and the two `depthClampEnable = enabled` rows confirm that clipping takes precedence as described above.

Without this change the extension is not advertised, and the same test reports it as unavailable.

### Notes

* No new files, so no new copyright headers.
* No Vulkan validation is added; the mapping covers every combination without a deviation to report.
* No `autorelease` use.
* Documentation updated in `Docs/MoltenVK_Runtime_UserGuide.md` and `Docs/Whats_New.md`.

### DIFF
--- a/Docs/MoltenVK_Runtime_UserGuide.md
+++ b/Docs/MoltenVK_Runtime_UserGuide.md
@@ -335,6 +335,7 @@ In addition to core *Vulkan* functionality, **MoltenVK**  also supports the foll
 - `VK_EXT_debug_report`
 - `VK_EXT_debug_utils`
 - `VK_EXT_depth_clip_control`
+- `VK_EXT_depth_clip_enable`
 - `VK_EXT_descriptor_indexing`
   - *Initial release limited to Metal Tier 1: 96/128 textures,
     16 samplers, except macOS 11.0 (Big Sur) or later, or on older versions of macOS using
--- a/Docs/Whats_New.md
+++ b/Docs/Whats_New.md
@@ -19,6 +19,7 @@ MoltenVK 1.4.3
 Released TBD
 
 - Add support for the following extensions:
+  - `VK_EXT_depth_clip_enable`
   - `VK_EXT_multi_draw`
   - `VK_EXT_nested_command_buffer`
   - `VK_EXT_ycbcr_2plane_444_formats`
--- a/MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKDevice.mm
@@ -682,6 +682,11 @@ void MVKPhysicalDevice::getFeatures(VkPhysicalDeviceFeatures2* features) {
 				depthFeatures->depthClipControl = true;
 				break;
 			}
+			case VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_DEPTH_CLIP_ENABLE_FEATURES_EXT: {
+				auto* depthClipFeatures = (VkPhysicalDeviceDepthClipEnableFeaturesEXT*)next;
+				depthClipFeatures->depthClipEnable = true;
+				break;
+			}
 			case VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_EXTENDED_DYNAMIC_STATE_FEATURES_EXT: {
 				auto* extDynState = (VkPhysicalDeviceExtendedDynamicStateFeaturesEXT*)next;
 				extDynState->extendedDynamicState = true;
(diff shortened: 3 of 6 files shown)


squidbus:
This is not correct, you need to be able to actually handle when both are set to off and both are set to on.

> With clipping enabled nothing survives outside the near and far planes, so every remaining fragment's depth already lies within the viewport's [minDepth, maxDepth], and clamping to that range is a no-op.

For this case where both are on: depth can be written by the fragment shader, which must also be clamped.

And when both are off, you want to have no clamp or clip other than the usual restriction to [0, 1] depth range (since we don't support unrestricted depth).

As an example of how this can be done, using how KosmicKrisp does it:
* Both off: Clamp mode, viewport depth range set to [0, 1], manual viewport Z transform inserted into vertex shader.
* Both on: Clip mode, manual depth clamp inserted into fragment shader if it writes to depth.

The extension's entire purpose is to allow these cases, for example in Direct3D translation where clip is toggle-able and clamp is always expected to be on. If it were equivalent to just toggling between clamp and clip there would be no point.

darrinperry83:
Thanks, you're right on both counts.

I missed that a fragment shader can write `gl_FragDepth` after clipping, so clamping is not a no-op when both are enabled — my justification for collapsing the two cases was simply wrong. And the broader point stands: if this reduces to a clamp/clip toggle it defeats the purpose of the extension.

I've read the KosmicKrisp implementation you pointed at (the mapping in `kk_cmd_draw.c` and the lowering in `msl_nir_lower_common.c`) and intend to follow it:

* clamp off / clip off — Metal clamp mode, viewport depth range set to `[0, 1]`, viewport Z transform emulated in the vertex shader (`z * (zmax - zmin) + w * zmin`, which gives the intended `(z/w) * zscale + zmin` after the perspective divide)
* clamp off / clip on — Metal clip mode
* clamp on / clip off — Metal clamp mode
* clamp on / clip on — Metal clip mode, output depth clamped in the fragment shader when it writes depth

On the MoltenVK side that means two shader-conversion flags plus an implicit buffer carrying the viewport Z range, following the existing `emulate_reversed_depth_viewport` precedent, with the matching emission added to SPIRV-Cross (which already tracks `writes_to_depth` for the fragment-side guard). I also noted the detail that the clamp path wants the min/max of the viewport range while the viewport-Z path needs the raw `minDepth`/`maxDepth` so inverted depth still works.

Marking this as a draft while I do that. Is anyone already working on it? Happy to stand down if so rather than duplicate the effort.

anshulnoori:
@darrinperry83 @squidbus implemented the advised approach in #2712

darrinperry83:
closing as the advised approach has been implemented.

PR #2803, opened 2026-08-18
Fix stage interface matching of 16-bit floating point variables.

utzcoz:
When matching the interface between two shader stages, MoltenVK took each variable's width from a VkFormat mapped from its base type. A format describes how many bits a component occupies rather than how it is interpreted, so a 16-bit float was described as 32-bit, and no format is ever described as a half.

A variable the adjoining stage does not access is laid out from that width alone, so a tessellation control shader writing a half wrote two bytes where the evaluation shader reserved four, shifting every later variable in the interface.

Take the width from the base type directly, in both directions of the match. Only one base type changes: a half is now described as 16-bit.

CTS over dEQP-VK.memory, query_pool, dynamic_state, clipping, fragment_operations, multiview, descriptor_indexing, info, geometry, tessellation and conditional_rendering, 29692 tests:

  Pass 7072 -> 7092, Fail 44 -> 24, no test regressed.

The 20 newly passing tests are dEQP-VK.tessellation.tess_io.max_in_out.with_f16.*. The 64-bit integer equivalents, with_i64.*.tcs_vert_writes{,_reads}_tes_na, still fail, because an interface variable can currently be described only as 16-bit or 32-bit. The 64-bit float ones do not run, as shaderFloat64 is not supported.

https://registry.khronos.org/vulkan/specs/latest/html/vkspec.html#interfaces-iointerfaces-matching

### DIFF
--- a/Docs/Whats_New.md
+++ b/Docs/Whats_New.md
@@ -23,6 +23,7 @@ Released TBD
   - `VK_EXT_nested_command_buffer`
   - `VK_EXT_ycbcr_2plane_444_formats`
 - Fix inconsistent image `memoryTypeBits` when `VK_IMAGE_USAGE_INPUT_ATTACHMENT_BIT` is used.
+- Fix shader stage interface matching of 16-bit floating point variables.
 
 
 
--- a/MoltenVK/MoltenVK/GPUObjects/MVKPipeline.mm
+++ b/MoltenVK/MoltenVK/GPUObjects/MVKPipeline.mm
@@ -1248,76 +1248,22 @@ MTLComputePipelineDescriptor* MVKGraphicsPipeline::newMTLTessVertexStageDescript
 	return plDesc;
 }
 
-static VkFormat mvkFormatFromOutput(const SPIRVShaderOutput& output) {
-	switch (output.baseType) {
-		case SPIRType::SByte:
-			switch (output.vecWidth) {
-				case 1: return VK_FORMAT_R8_SINT;
-				case 2: return VK_FORMAT_R8G8_SINT;
-				case 3: return VK_FORMAT_R8G8B8_SINT;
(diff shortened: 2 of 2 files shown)


squidbus (commented):
I feel like not having any of the formats described as half might be an oversight... or maybe we just don't need the half format specifier. Either way this seems simpler and equivalent to intended so it looks fine to me.

> The equivalent 64-bit tests still fail, because an interface variable can currently be described only as 16-bit or 32-bit.

I don't think the 64-bit tests should be running, ~~we don't support `shaderFloat64`. At least on current main branch CTS it seems to properly check for this.~~ Unless this is referring to 64-bit integer I guess, I'm kind of surprised there isn't an equivalent to `storageInputOutput16` for this though?

squidbus (approved):

utzcoz:
> half might be an oversight

It's unreachable — I found all four 16-bit SFLOAT formats are registered as `ColorFloat`, so
`getFormatType()` never returns it. Five of the six switches handling it share an arm with
`kMVKFormatColorFloat` anyway. Can send a follow-up removing it.

> I'm kind of surprised there isn't an equivalent to storageInputOutput16 for this though?

Right, they're the integer ones. `shaderFloat64` is never set and `with_f64` is 80/80
NotSupported; `shaderInt64` is set, so `with_i64` runs and 20 of 80 fail, all `*_tes_na` — same
shape as the f16 bug. Reworded to say 64-bit integer.

squidbus:
> I found all four 16-bit SFLOAT formats are registered as ColorFloat, so
getFormatType() never returns it.

Yeah that’s what I meant.

utzcoz:
If you want, I think I can send a new PR to remove half.

utzcoz:
https://github.com/KhronosGroup/MoltenVK/pull/2804 here.

PR #2804, opened 2026-08-19
Remove the unused kMVKFormatColorHalf format type.

utzcoz:
No format descriptor ever produced it. Every 16-bit floating point format is registered as kMVKFormatColorFloat, so getFormatType() could not return it, and the six switches naming it were unreachable.

Five of those switches shared an arm with kMVKFormatColorFloat and are unaffected. The sixth, getMTLFormatTypeString(), was the only one to give it a distinct result, a half MSL type for the attachment clear shader; a 16-bit float attachment already reaches the float arm there.

This removes an enumerator from a public header, so a value of MVKFormatType compiled against an earlier release no longer matches.


utzcoz:
Hi @squidbus , could you help to review this PR? Thanks.

squidbus:
To be honest, I'm unsure whether it makes more sense to just remove this or set the 16-bit float color formats to use it, which is why I haven't left a review.

utzcoz:
Understood. I can close this one.

PR #2809, opened 2026-08-28
Fix memory accumulation of pooled command objects that retain their content

### LINKED ISSUE #2808: Command object pools leak under multithreaded recording, USE_COMMAND_POOLING=0 avoids it
## Description

Command object pools leak when command buffers are recorded from more than one thread. Memory grows by roughly 90MB per minute while rendering, mostly 64KB push descriptor allocations, and the process aborts after about an hour. Setting `MVK_CONFIG_USE_COMMAND_POOLING=0` removes the leak entirely: 447MB growth over a 5 minute replay with pooling on, 83MB with it off, no measurable fps difference between the two.

## Environment

- MoltenVK 1.4.2
- macOS 26, Apple Silicon (M3 Max, also reported on M3 Ultra)
- mesa zink (26.2-devel) on top of MoltenVK, threaded context
- App: Beyond All Reason (OpenGL through zink)

## Steps to Reproduce

1. Run a zink based app with a threaded context on default MoltenVK config
2. Watch process memory while it renders: steady growth, ~90MB per minute here
3. Repeat with `MVK_CONFIG_USE_COMMAND_POOLING=0`: memory stays flat

## Notes

`MVKObjectPool` documents `acquireObject()`/`returnObject()` as single thread only and has `Safely` variants next to them, but `MVKCommandPool` uses the unsafe ones. So the question: is the app expected to serialize everything that touches these pools? Then this is a zink bug and I'll take it to mesa instead. Or should the command pools be using the safe variants? If you tell me which way you would want this fixed, I can put together a PR. The leak reproduces here in minutes, so testing is easy.

Vandomas:
Pooled commands keep their retained content while parked in the pool. `MVKCmdPushDescriptorSetWithTemplate` holds about 60 KB of template data per object, and under zink that accumulates around 90 MB per minute until the process aborts.

The fix releases the content when the command buffer is reset and out of the pending state, via a `releaseRetainedContent()` virtual implemented by the two push descriptor commands. On the same workload retained payload after 5 minutes drops from 516 MB to 106 MB of commands actually in flight, fps unchanged.

Fixes #2808.


Vandomas:
Converted to draft: found a use after free in game testing. The immediate encoding path releases commands mid-recording while the encoder still holds a pointer to their content, which ends in a lost device. The release needs to happen only on the reset path. Reworking and retesting.

PR #2814, opened 2026-09-03
Add MVK_CONFIG_MAXIMUM_PRESENT_FRAME_RATE to cap the presentation rate

berksozgen:
Metal can pace presentation through `-[MTLDrawable presentAfterMinimumDuration:]`,
which lets the OS reduce GPU clocks, which is usually the reason an app chooses a
lower frame rate in the first place. MoltenVK never calls it, so an app that wants
to run below the rate of the display can only throttle its own render loop, which
leaves the OS to infer the cadence rather than being told it.

This adds a configuration parameter that declares the maximum presentation rate.
When it is zero, the default, presentation is unchanged.

### Measurements

Frame interval sampled over one second windows, on an iPhone 17 and an iPad Pro M4,
both with 120Hz displays:

| | default | 80 | 45 | 30 |
|---|---|---|---|---|
| iPhone 17 | 60 | 60 | 30 | 30 |
| iPad Pro M4 | 120 | 60 | 40 | 30 |

The resulting rate is quantized to the refresh rate of the display, since a
presentation is shown on a refresh cycle: on a display being refreshed at `R`,
requesting `F` presents at `R / ceil(R / F)`. The rows above are exactly that,
with `R` being 60 on the iPhone and 120 on the iPad. This is documented, because
an app asking for 45 and being shown 40 is otherwise surprising.

Verified against an unpatched build of the same source on both devices: with the
default, the frame interval is identical.

### Notes

- The duration is resolved once, at swapchain creation, rather than at each
  presentation, because presentation occurs from a `MTLCommandBuffer` callback.
- Deriving a default from the refresh rate of the display was tried and rejected:
  a minimum duration equal to the refresh cycle duration is a knife edge, and on
  the iPad it lowered the measured rate from 120 to 80, as jitter pushed alternate
  frames to the following refresh cycle. The parameter is therefore purely opt-in.
- Built and tested on `main`; the measurements above were taken with the same
  change applied to `v1.4.2`, which is the version shipped in Vulkan SDK 1.4.357.1.


berksozgen:
A follow-up measurement on this, which exposes a rough edge in what I proposed.
I do not think it should change the patch, but it belongs in the documentation.

The patch passes the duration through raw: `_minPresentDuration = 1.0 / maximumPresentFrameRate`.
When the requested rate divides the refresh rate exactly, that duration lands
precisely on a cycle boundary — and those are the values people actually ask for
(30 on a 60Hz panel, 60 or 30 on a 120Hz panel).

Measured on iPhone 17, panel at 60Hz, one-second windows:

| requested cap | duration | position     | achieved | frame interval  |
|---------------|----------|--------------|----------|-----------------|
| 30            | 33.33 ms | 2.00 periods | 30       | **20.1 - 35.0 ms** |
| 31            | 32.26 ms | 1.94 periods | 30       | **31.6 - 34.5 ms** |

Same achieved rate, but pacing is roughly five times tighter one step off the
boundary. Average frame rate is not the thing that degrades; frame-to-frame
consistency is.

The principled fix would be to aim for the middle of the intended cycle rather
than its edge: with `k = ceil(R / F)`, use `(k - 0.5) / R` instead of `1 / F`.

**I am not proposing that here, for two reasons.**

First, MoltenVK does not reliably know `R`. `MVKSwapchain::getRefreshCycleDuration`
reads `screen.maximumFramesPerSecond` on iOS, which is the panel's *maximum*, not
its current rate. Those differ as soon as anything in the process declares a
`CAFrameRateRange` — in my case the panel sits at 80Hz while `maximumFramesPerSecond`
still reports 120.

Second, `presentAfterMinimumDuration:` is a *minimum*, so shortening it can move
a frame to an earlier cycle and exceed the maximum the caller asked for. For a
setting named MAXIMUM_PRESENT_FRAME_RATE that seems worse than the pacing jitter.

A fixed offset such as `1 / (F + 1)` does not work either: it fixes 30 and breaks
29, 19 and 14 on a 60Hz panel, because it relocates the boundary rather than
avoiding it. I measured 29 as a clean 20fps today; with the offset it would sit
on the boundary instead.

So my suggestion is a documentation note rather than a behaviour change — roughly:

> The requested rate is quantized to the display's refresh cycle. Requesting a
> rate that divides the refresh rate exactly places the minimum duration on a
> cycle boundary, where small variations in frame time can move a presentation to
> the next cycle. The average rate is unaffected, but frame pacing is more
> consistent if the request is placed slightly above such a value.

I am happy to write that into `Docs/MoltenVK_Configuration_Parameters.md` in this
PR if you would like it there.

Issue #2805, opened 2026-08-21
Commands implemented using compute may have incorrect barrier fences when combined with other compute


squidbus:
Consider the following example (received from @aitor-lunarg):
```
         vkCmdDispatch(cmd, 1, 1, 1);

         /* arm "gap": the same command, now the encoder's last use */
         if (arm == 2)
            vkCmdFillBuffer(cmd, scratch, 0, 1024, 0xabcdabcdu);

         /* the dependency under test */
         VkMemoryBarrier2 mb = {
            .sType = VK_STRUCTURE_TYPE_MEMORY_BARRIER_2,
            .srcStageMask = VK_PIPELINE_STAGE_2_COMPUTE_SHADER_BIT,
            .srcAccessMask = VK_ACCESS_2_SHADER_STORAGE_WRITE_BIT,
            .dstStageMask = VK_PIPELINE_STAGE_2_COMPUTE_SHADER_BIT,
            .dstAccessMask = VK_ACCESS_2_SHADER_STORAGE_READ_BIT |
                             VK_ACCESS_2_SHADER_STORAGE_WRITE_BIT };
         VkDependencyInfo di = { .sType = VK_STRUCTURE_TYPE_DEPENDENCY_INFO,
                                 .memoryBarrierCount = 1,
                                 .pMemoryBarriers = &mb };
         vkCmdPipelineBarrier2(cmd, &di);

         /* consumer: reads the payload */
         struct push p_cons = { budget, expect, 1u };
         vkCmdPushConstants(cmd, layout, VK_SHADER_STAGE_COMPUTE_BIT, 0,
                            sizeof(p_cons), &p_cons);
         vkCmdDispatch(cmd, 1, 1, 1);
```

This results in two compute encoders:
* Compute encoder 1: The first dispatch and the fill buffer compute
* Compute encoder 2: The second dispatch

The combined first encoder is a problem as we determine the barrier stage fence to update at the end of the encoder only from the last command use, like so:
```
void MVKCommandEncoder::encodeBarrierWaits(MVKCommandUse use) {
	...
	if (_mtlComputeEncoder) {
		auto stage = commandUseToBarrierStage(use);
		if (stage != kMVKBarrierStageNone) {
			barrierWait(stage, _mtlComputeEncoder);
		}
	}
	...
}
```

[...]

PR #2819, opened 2026-09-05
Preserve execution dependencies across Metal encoders

dttdrv:
Track every barrier stage used by a shared compute encoder, so closing it updates all required fences rather than only the last command's stage. Preserve the source and destination stages of legacy execution-only pipeline barriers when no memory barrier records are supplied. Both changes use the existing fence machinery and preserve encoder sharing.

Related to #2805.

### DIFF
--- a/MoltenVK/MoltenVK/Commands/MVKCmdPipeline.mm
+++ b/MoltenVK/MoltenVK/Commands/MVKCmdPipeline.mm
@@ -104,6 +104,10 @@ VkResult MVKCmdPipelineBarrier<N>::setContent(MVKCommandBuffer* cmdBuff,
 	for (uint32_t i = 0; i < imageMemoryBarrierCount; i++) {
 		_barriers.emplace_back(pImageMemoryBarriers[i], srcStageMask, dstStageMask);
 	}
+	if (_barriers.empty()) {
+		VkMemoryBarrier barrier = {VK_STRUCTURE_TYPE_MEMORY_BARRIER};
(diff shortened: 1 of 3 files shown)


squidbus:
Has this been tested with something along the lines of the sample in https://github.com/KhronosGroup/MoltenVK/issues/2805?

dttdrv:
I tested the equivalent sequence on this PR’s Release build. Both fill and no-fill cases passed 1,000 runs with Metal validation. Fence tracing confirms the required updates without an extra encoder split.

MoltenVK/MoltenVK/Commands/MVKCmdPipeline.mm
```diff
@@ -104,6 +104,10 @@
 	for (uint32_t i = 0; i < imageMemoryBarrierCount; i++) {
 		_barriers.emplace_back(pImageMemoryBarriers[i], srcStageMask, dstStageMask);
 	}
+	if (_barriers.empty()) {
+		VkMemoryBarrier barrier = {VK_STRUCTURE_TYPE_MEMORY_BARRIER};
+		_barriers.emplace_back(barrier, srcStageMask, dstStageMask);
```
squidbus, line 109: I think you should probably add a new constructor overload to `MVKPipelineBarrier` that just takes the stage masks and applies a new type `Execution`, so it can skip the parts specific to memory/resource barriers when encoding.

dttdrv: done, i added an 'Execution' barrier type with a stage-mask-only constructor

PR #2817, opened 2026-09-03
Minor fix for tessellation

denisk-dev:
Increasing the indirect max vertex count and fixing what seems to be an incorrect buffer offset.

This fixes terrain tessellation with Kitten Space Agency.

Please, let me know if this change makes sense, or if it's a one-off that fixes KSA.

Issue #2376, opened 2024-10-13
Vertex indices seem invalid when using indexed draw with tessellation


tomaka:
Here's my situation:
I have a draw call that uses a tessellation shaders. Everything is working.
If I write the `VkDrawIndirectCommand` corresponding to my call to a buffer, and replace `vkCmdDrawIndexed` with `vkCmdDrawIndexedIndirect` pointing to this buffer, the object no longer renders.

Debugging with XCode shows that my indirect buffer indeed contains the correct data. The problem comes from the fact that all the elements in the geometry of the object (the tessellation evaluation shader output) are identical.

Digging a bit into what MoltenVK is doing, it seems that it's because the MVK-injected compute shader that executes [`cmdDrawIndexedCopyIndex32Buffer`](https://github.com/KhronosGroup/MoltenVK/blob/f79ab2a29ec2da87069f389604cc02f32510f69d/MoltenVK/MoltenVK/Commands/MVKCommandPipelineStateFactoryShaderSource.h#L492) has a workgroup size of 0, and thus all the indices in the index buffer remain at their initial value of 0, and thus the MVK-injected compute shader that later generates the vertex data (it's not clear to me whether it's the vertex shader input or the tessellation shader input, but this doesn't matter) repeatedly copies vertex element 0 over the entire vertex data.

I have the impression that this workgroup size is supposed to be written [here](https://github.com/KhronosGroup/MoltenVK/blob/f79ab2a29ec2da87069f389604cc02f32510f69d/MoltenVK/MoltenVK/Commands/MVKCommandPipelineStateFactoryShaderSource.h#L467-L469). However, `destVTX` points to offset 12 of `destBuf`, while the shader that executes `cmdDrawIndexedCopyIndex32Buffer` later reads its indirect calls from offset 0 of that same buffer. Offset 0 is where [`destSI`](https://github.com/KhronosGroup/MoltenVK/blob/f79ab2a29ec2da87069f389604cc02f32510f69d/MoltenVK/MoltenVK/Commands/MVKCommandPipelineStateFactoryShaderSource.h#L476-L478) is located. 
Maybe I'm misunderstanding what MoltenVK is doing here, but I suspect that there's a bug somewhere with these offsets and that the workgroup size is supposed to read from `destVTX`?

MoltenVK version: 1.2.11

PR #2820, opened 2026-09-05
Correct tessellation patch counts and indexed-indirect command layout

dttdrv:
Discard incomplete patches per instance and pack the captured vertices accordingly, so direct and indirect draws produce the same complete patches. Also reserve both dispatch records written by indexed-indirect conversion and advance source commands by their supplied stride.

The index-copy dispatch offset is already fixed by #2817 and is excluded from this proposal.

Related to #2376.


dttdrv:
I fixed the index copy bounds and the buffer size after uint8 conversion. the copy now uses the valid index count and copies one instance row.
the indirect conversion kernels write records at a 256-byte stride, but the host was advancing by the uniform-buffer alignment. that could make later draws read padding, making them both use 256 bytes now

PR #2821, opened 2026-09-05
Add support for VK_EXT_shader_object.

utzcoz:
Metal bakes into a MTLRenderPipelineState most of the state a shader object leaves dynamic, and MoltenVK generates MSL that depends on the adjoining stages and the vertex layout, so a VkShaderEXT cannot be translated when it is created. It holds its SPIR-V, and a draw builds or reuses the pipeline that the bound shaders and the current state amount to, keyed on everything Metal cannot change afterwards. Cheap state such as the viewport stays dynamic, so it does not force a rebuild. A shader binary is the SPIR-V behind a header naming the driver that produced it, sharing the pipeline cache UUID because the same changes invalidate both.

Both vkCmdSetVertexInputEXT and vkCmdBindVertexBuffers2 set the vertex stride and the later call wins, so which of them set it last selects the pipeline, and decides whether the stride is baked in or read from the buffer bindings.

Eleven dynamic state setters that were empty stubs now record their state, since enabling shader objects makes all of them usable. An entry point may now name a second extension that also provides it, because VK_EXT_shader_object supplies every setter the extended dynamic state extensions do. The NV dynamic state setters that VK_EXT_extended_dynamic_state3 names are provided as no-ops, the way vkCmdSetColorBlendAdvancedEXT already is, with their feature bits reporting false.

A draw whose shaders cannot be made into a pipeline is skipped rather than run against nothing, which Metal would fault on. Changing any of the state a pipeline was built for drops it, so that the next draw resolves again instead of silently rendering with the old values. Writing a value that is already there is not a change, because applications commonly re-set every dynamic state before each draw and that must not cost a pipeline resolve every time.

Tessellation shader objects work except in point mode, which MoltenVK reports as unsupported through VK_KHR_portability_subset. Linked shaders cannot be exercised at all, because every CTS case for them also needs geometry shaders.

CTS over dEQP-VK.pipeline.monolithic.vertex_input, pipeline.monolithic.dynamic and dynamic_state, 13967 tests:

  Pass 9203 -> 9383, Fail 5 -> 5. The only change is 180 tests that
  were skipped and now pass; no test changed in any other way.

CTS over dEQP-VK.shader_object.create, binary and misc, 1203 tests:

  Pass 0 -> 522, Fail 0 -> 0, and no test crashes the driver.

CTS over the rest of dEQP-VK.shader_object: every case of binding, link, tessellation, performance, pipeline_interaction and api, plus one in sixty of rendering, 4535 tests:

  Pass 0 -> 1039, Fail 0 -> 27, and no test crashes the driver.
  Nothing in the rendering sample fails. The failures are 23
  performance tests, which time a single command recording call and
  compare maxima over a hundred iterations, and whose outcome varies
  from run to run on the same build, and four that need tessellation
  point mode, which MoltenVK reports as unsupported through
  VK_KHR_portability_subset.

### DIFF
--- a/MoltenVK/MoltenVK/Commands/MVKCmdTransfer.h
+++ b/MoltenVK/MoltenVK/Commands/MVKCmdTransfer.h
@@ -133,6 +133,7 @@ typedef MVKCmdBlitImage<4> MVKCmdBlitImageMulti;
 typedef struct {
     VkImageSubresourceLayers srcSubresource;
     VkImageSubresourceLayers dstSubresource;
+	uint32_t dstDepthPlane;		/**< The depth plane resolved into, for a 3D destination image. */
 } MVKMetalResolveSlice;
 
 /**
--- a/MoltenVK/MoltenVK/Commands/MVKCmdTransfer.mm
+++ b/MoltenVK/MoltenVK/Commands/MVKCmdTransfer.mm
@@ -829,6 +829,7 @@ void MVKCmdResolveImage<N>::encode(MVKCommandEncoder* cmdEncoder) {
 	uint32_t expCnt = 0;
 	uint32_t copyCnt = 0;
 	uint32_t sliceCnt = 0;
+	bool isDst3D = _dstImage->getImageType() == VK_IMAGE_TYPE_3D;
 
 	for (const auto& vkIR : _vkImageResolves) {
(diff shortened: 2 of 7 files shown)


alyssarosenzweig:
This is fundamentally against the entire spirit of the EXT and indeed Vulkan itself, which is to eliminate draw-time recompiles.

utzcoz:
@alyssarosenzweig Thanks for your feedback. I will check whether I can avoid it.

